PR #7, opened 2018-12-24
Modifications to ESS determination

alongd:
Added an optional `ess_settings` user argument to override the `determine_remote` method.
Added a diagnostics option to `determine_remote` along with an iPy notebook to run just diagnostics.
Other general improvements to the method.


alongd:
An example `ess_settings` argument is:
`ess_settings = {'gaussian': 'pharos', 'molpro': 'rmg', 'qchem': 'pharos'}`

PR #69, opened 2019-02-23
Thermo Calculations for Hydrogen Bonded Conformers

mjohnson541:
This PR enables ARC to calculate thermo for RMG molecules with hydrogen bonds.  In order to do this it finds the lowest energy conformer and the lowest energy hydrogen bonded conformer and calculates them both with all rotors.  The hydrogen bonded conformer is then calculated without any rotors that break a hydrogen bond and the rotors that break the hydrogen bond are used to estimate the barrier to breaking the hydrogen bond which is used to calculate a characteristic temperature for the hydrogen bonding.  The hydrogen bonded conformer thermo and the lowest energy conformer with all rotor's thermo is then fused using thermodynamic properties of the hydrogen bonded conformer below Tchar and the other above Tchar.  

This should be functional right now, but not final (tests will be added after this is finalized), but I thought I'd put it up for programmatic and technical discussion now.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -24,7 +24,7 @@ import arc.rmgdb as rmgdb
 from arc.settings import arc_path, default_levels_of_theory, check_status_command, servers, valid_chars
 from arc.scheduler import Scheduler, time_lapse
 from arc.exceptions import InputError, SettingsError, SpeciesError
-from arc.species import ARCSpecies
+from arc.species import ARCSpecies, get_Hbonds
 from arc.reaction import ARCReaction
 from arc.processor import Processor
 from arc.job.ssh import SSH_Client
@@ -272,6 +272,14 @@ class ARC(object):
                 if i in indices_to_pop:
                     self.arc_species_list.pop(i)
             self.arc_species_list.extend(converted_species_list)
+            for arc_spc in self.arc_species_list: #if it has hydrogen bonds we need to calculate the unbonded confomer too
+                if get_Hbonds(arc_spc.mol):
+                    arc_mol = arc_spc.mol.copy(deep=True)
+                    arc_mol.remove_H_bonds()
+                    spc = Species(molecule=[arc_mol])
+                    spc.label = str(arc_spc.label+'noHbonds')
+                    self.arc_species_list.append(ARCSpecies(is_ts=False,rmg_species=spc))
+                    logging.info("ARC Species {} is H-bonded so calculating un-H-bonded conformer {} as well".format(arc_spc.label,spc.label))
             for arc_spc in self.arc_species_list:
                 if arc_spc.label not in self.unique_species_labels:
                     self.unique_species_labels.append(arc_spc.label)
--- a/arc/processor.py
+++ b/arc/processor.py
@@ -120,6 +120,9 @@ class Processor(object):
             rotors = '\n\nrotors = ['
             rotors_description = '1D rotors:\n'
             for i in range(species.number_of_rotors):
+                if 'HBonded' in species.rotors_dict[i].keys() and species.rotors_dict[i]['HBonded']: #skip H-bonded rotors
+                    logging.info("skipping H-bonded rotor of species {0} between pivots {1}".format(species.label,species.rotors_dict[i]['pivots']))
+                    continue
                 pivots = str(species.rotors_dict[i]['pivots'])
                 scan = str(species.rotors_dict[i]['scan'])
                 if species.rotors_dict[i]['success']:
--- a/arc/processor.py
+++ b/arc/processor.py
@@ -182,6 +182,8 @@ class Processor(object):
         species_for_thermo_lib = list()
         for species in self.species_dict.values():
             if species.generate_thermo and not species.is_ts and 'ALL converged' in self.output[species.label]['status']:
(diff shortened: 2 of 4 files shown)


alongd:
Can you take a look at the tests? Seems something trivial in species rotors

mjohnson541:
Tests should be fixed

alongd:
Can you take a look at the Codacy report? Specifically, initializing an argument with an empty list `def get_min_energy_conformer(xyzs, energies, hbonds=[])`
BTW, we have an unwritten rule in the ARC community that each PR has to increase the coverage, even by just a little bit... Could you add some tests?

mjohnson541:
I'll change that initialization  As I said this is just for your initial perusal and comments on the programming and approach.  Don't worry, I'm going to add all of the tests after I'm certain this is how this should be done and once I get the approach approved by Bill.

arc/main.py
```diff
@@ -272,6 +272,14 @@ def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn
                 if i in indices_to_pop:
                     self.arc_species_list.pop(i)
             self.arc_species_list.extend(converted_species_list)
+            for arc_spc in self.arc_species_list: #if it has hydrogen bonds we need to calculate the unbonded confomer too
```
alongd: Can you place the comment underneath the `for` so the line isn't too long?

arc/main.py
```diff
@@ -272,6 +272,14 @@ def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn
...
+                if get_Hbonds(arc_spc.mol):
```
alongd: I think you need to first make sure arc_spc.mol exists, it could be `None` if only xyz is given as the input for a species.

arc/main.py
```diff
@@ -272,6 +272,14 @@ def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn
...
+                    arc_mol = arc_spc.mol.copy(deep=True)
+                    arc_mol.remove_H_bonds()
+                    spc = Species(molecule=[arc_mol])
+                    spc.label = str(arc_spc.label+'noHbonds')
+                    self.arc_species_list.append(ARCSpecies(is_ts=False,rmg_species=spc))
```
alongd: Can we have H bonds in TSs? Perhaps do `is_ts=arc_spc.is_ts)`

arc/main.py
alongd: I think that this section deals with species inputted via the API (e.g., Jupyter notebooks), but we should have a similar check for species inputted via a YAML input file in `from_dict` (we should also check that the noHbonded species wasn't already generated since the input file could be a restart file already containing that duplicate species)

arc/species.py
```diff
@@ -508,11 +508,19 @@ def find_conformers(self, mol, method='all'):
             rd_xyzs, rd_energies = _get_possible_conformers_rdkit(mol)
             if rd_xyzs:
                 rd_xyz = get_min_energy_conformer(xyzs=rd_xyzs, energies=rd_energies,hbonds=hbonds)
-                self.xyzs.append(get_xyz_string(xyz=rd_xyz, mol=mol))
+                if rd_xyz:
+                    self.xyzs.append(get_xyz_string(xyz=rd_xyz, mol=mol))
+                else:
+                    logging.warn("did not find a hydrogen bonded conformer for {} from RDKit search".format(self.label))
```
alongd: Could it be that a conformer wasn't found for a different reason?

arc/species.py
```diff
@@ -538,9 +538,16 @@ def determine_rotors(self):
                 mol_list = self.mol_list
             for mol in mol_list:
                 if get_Hbonds(mol): #We need the rotor scans of the H-bonded molecule later for determining when the H-bond breaks
-                    mol = mol.copy(deep=True)
```
alongd: This part deletes the content of commit `make sure we get scans for the rotors that break the H-bond`
Can you remove the commit instead?

arc/processor.py
```diff
@@ -316,6 +317,32 @@ def clean_output_directory(self):
...
+        if Tchar > Tlist[i]:
+            Cplist[i] += thermo_HB.getHeatCapacity(Tlist[i])
+        else:
+            Cplist[i] += thermo_noHB.getHeatCapacity(Tlist[i])
+    H298 += thermo_HB.getEnthalpy(298.0)
+    S298 += thermo_HB.getEntropy(298.0)
+    #H-bonded molecule assumed to be non-linear, polyatomic
+    Nrotors = len(spc.rotors_dict) #all possible rotors (CpInf is high T)
+    Nfreq = 3*spc.number_of_atoms-6-Nrotors
+    Cp0 = 4.0*8.314
```
alongd: Please add comments for unit conversions

arc/species.py
```diff
@@ -1100,7 +1100,9 @@ def _get_possible_conformers_openbabel(mol):
     return xyzs, energies
 
 
-def get_min_energy_conformer(xyzs, energies, hbonds=[]):
```
alongd: Should this be a fixup commit?

alongd:
@mjohnson541, unless you're tied-up, let's get this one in

mjohnson541:
Desired functionality for hydrogen bonded species has changed so I'm closing this PR.

PR #70, opened 2019-02-23
Parse xyz for a species from a file

alongd:
Currently allowing .xyz and .gjf


alongd:
Add Molpro and QChem input files, add an output file parser through Arkane

PR #85, opened 2019-03-08
Troubleshoot rotor scans

alongd:
By freezing all other dihedrals, also increasing the scan resolution


alongd:
@mjohnson541, can you try running your molecule on this branch and see if the scan troubleshooting is any good?

mjohnson541:
Yep

PR #86, opened 2019-03-08
TS rotors

mjohnson541:
This PR makes it possible for ARC to do Rotors for TS's.  The user can specify on the reaction object the `atom_length_constraint` that is a list of length 2 lists of 1-indexed atom indices.  Each pair of indices correspond to atoms that need to stay close together in the TS.  

After rotors are calculated normal checks as in thermo are done with two exceptions:  
1) Rotor scan files are parsed and the distances between constrained atom pairs both at the optimized geometry and in the rotor scan are computed.  If the change in distance as a fraction of the distance at the first point in the rotor scan is more than `ts_length_change` or as a fraction of distance in the optimized geometry is more than `ts_opt_length_change` the rotor is considered to break the TS and is invalidated.  
2) If the initial scan point is determined to not be a well instead of launching a new optimization job the rotor is simply invalidated under the assumption that the rotor is moving along the TS.  

Lots of small bug fixes.  

Open to better naming suggestions for user facing parameters.

### DIFF
--- a/arc/job/inputs.py
+++ b/arc/job/inputs.py
@@ -42,7 +42,7 @@ name
 
 {charge} {multiplicity}
 {xyz}
-{scan}(scan_trsh}
+{scan}{scan_trsh}
 
 
 """,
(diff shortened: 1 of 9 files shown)


alongd:
closed by #320

PR #96, opened 2019-03-18
Fix to xyz/2D loading and atom reordering in ARCSpecies

mliu49:
### Before
In `__init__`:
- If `self.mol` was `None`, then a molecule would be generated from the xyz coordinates but not saved to self, and `self.mol_list` would also not be set.

In `mol_from_xyz`:
- It seemed like the intention was to maintain consistent atom IDs between `self.mol` and `self.mol_list`, which was not the case, since the atom orders changed due to resonance structure generation.

### After
- Simplified `__init__` code to call `mol_from_xyz` for all non-TS species
- Handling for whether or not `self.mol` exists moved to `mol_from_xyz`
- Fixed atom id copying when generating resonance structures
- Added unit test for atom ids from mol_to_xyz

### Reviewer Notes
Please double check that my interpretation of `mol_from_xyz` is correct. If you had tests for that method, it might be good to run them/add them to unit tests.

### DIFF
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -212,13 +212,12 @@ class ARCSpecies(object):
                 self.multiplicity = multiplicity
                 self.charge = charge
 
-            if self.mol is None and not self.is_ts:
-                xyz = self.final_xyz if self.final_xyz else self.initial_xyz
-                _, mol = molecules_from_xyz(xyz)
-                self.number_of_atoms = len(mol.atoms)
-            elif not self.is_ts:
-                if self.initial_xyz is not None:
+            if not self.is_ts:
+                # Perceive molecule from xyz coordinates
+                # This also populates mol_list
+                if self.final_xyz or self.initial_xyz:
                     self.mol_from_xyz()
+                # Generate bond list for applying bond corrections
                 if not self.bond_corrections:
                     self.bond_corrections = self.mol.enumerate_bonds()
                     if self.bond_corrections:
@@ -232,6 +231,7 @@ class ARCSpecies(object):
             elif not self.bond_corrections and self.generate_thermo:
                 logging.warn('Cannot determine bond additivity corrections (BAC) for species {0} based on xyz'
                              ' coordinates only. For better thermoproperties, provide bond corrections.')
+
             if self.initial_xyz is not None:
                 # consider the initial guess as one of the conformers if generating others.
                 # otherwise, just consider it as the conformer.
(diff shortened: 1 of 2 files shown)


arc/species/speciesTest.py
```diff
@@ -454,6 +454,23 @@ def test_get_min_energy_conformer(self):
...
+        mol = self.spc6.mol
+        mol_list = self.spc6.mol_list
+
+        self.assertEqual(len(mol_list), 1)
+        res = mol_list[0]
+
+        self.assertTrue(mol.atomIDValid())
+        self.assertTrue(res.atomIDValid())
+
+        print(mol.toAdjacencyList())
```
alongd: Were the print statements left here on purpose?

arc/species/species.py
```diff
@@ -800,24 +800,33 @@ def mol_from_xyz(self, xyz=None):
...
         if self.mol_list is None:
+            # Assign atom ids first, so they carry through to the resonance structures
             self.mol.assignAtomIDs()
-            self.mol_list = self.mol.generate_resonance_structures(keep_isomorphic=False, filter_structures=True)
-            # The generate_resonance_structures methods changes atom order, reorder atoms in self.mol:
-            id_mol = self.mol.copy(deep=True)
-            _, self.mol = molecules_from_xyz(xyz)
-            for i, atom in enumerate(self.mol.atoms):
-                atom.id = id_mol.atoms[i].id
+            # The generate_resonance_structures methods changes atom order
```
alongd: This is my typo originally. Can you change `methods` to `method` in this comment?

arc/species/speciesTest.py
```diff
@@ -454,6 +454,23 @@ def test_get_min_energy_conformer(self):
...
+        print(res.toAdjacencyList())
+
+        self.assertTrue(mol.isIsomorphic(res))
+        self.assertTrue(mol.isIdentical(res))
+
```
alongd: Can you add another test for a species with more than one entry in `mol_list`?
For example, HSO3 has two resonance structures, you can use it's xyz:
```
S      -0.12383700    0.10918200   -0.21334200
O       0.97332200   -0.98800100    0.31790100
O      -1.41608500   -0.43976300    0.14487300
O       0.32370100    1.42850400    0.21585900
H       1.84477700   -0.57224200    0.35517700
```

Could we check that `S` and `H` have the same ID and are in the same order in `self.mol` and in all entries of `self.mol_list`?

arc/species/species.py
```diff
@@ -800,24 +800,33 @@ def mol_from_xyz(self, xyz=None):
...
-            raise InputError('XYZ and the 2D graph representation of the Molecule are not isomorphic.\n'
-                             'Got xyz:\n{0}\n\nwhich corresponds to {1}\n{2}\n\nand: {3}\n{4}'.format(
-                               xyz, self.mol.toSMILES(), self.mol.toAdjacencyList(),
-                               original_mol.toSMILES(), original_mol.toAdjacencyList()))
+            xyz = self.final_xyz or self.initial_xyz
+
+        if self.mol is not None:
+            # self.mol should have come from another source, e.g. SMILES or yml
+            original_mol = self.mol
+            self.mol = molecules_from_xyz(xyz)[1]
```
alongd: We have a problem here (in the original code, not introduced by this PR) that when we overrid self.mol we might change the multiplicity. For example, `molecules_from_xyz()` will return multiplicity 3 for birads, but self.mol could originally be singlet. Then the ESS calculation will be affected. I'm trying to solve these issues on the master branch, it also affects isomorphism checks after optimization. If you get this in first, you won't have to rebase.

arc/species/speciesTest.py
mliu49: Sorry, they're left over from debugging. I'll remove them.

arc/species/species.py
mliu49: I've had that issue, but it will usually fail the isomorphism comparison with the 2D structure. Have you found a way to fix the molecule generated from the XYZ to match the expected multiplicity?

alongd: Yes, I'll fix that after merging this branch

arc/species/species.py
mliu49: Done

arc/species/speciesTest.py
mliu49: Done

PR #97, opened 2019-03-18
Save all log files

oscarwumit:
ARC now saves all log files regardless of the number of restarts
The old log files are saved in
<project_folder>/log/arc.old.[local_time].log

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -6,6 +6,7 @@ import logging
 import sys
 import os
 import time
+import datetime
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


oscarwumit:
@alongd I modified the code as follows. ARC now backups all versions of log and restart files. If a user restarted an ARC job more than once, ARC creates a folder named "backups" in the project folder. The backup files are saved in two subfolders of "backups" named "log" and "restart", respectively.

Please let me know if more changes are needed.

alongd:
I like the implementation with the time stamp in the file name, although the time we save is the time when the file was copied, not its creation time.

I suggest renaming `backups` to `log_and_restart_archive`. It's a long name, but very clear. Also, it could be OK to have all files there without a subfolder division (the files are easily differentiable), but whatever you decide is fine by me.

Also, now that we have such a fancy archiving method, there's no need in the `.old` files. Could you replace the code that saved the `.old` with your code?

oscarwumit:
@alongd Thanks for your comment. Now all backup log and restart files are saved under log_and_restart_archive. I also replaced the code that was used to make .old files. Please review the code and let me know if more changes are needed.

arc/main.py
```diff
@@ -680,11 +685,11 @@ def initialize_log(self, verbose=logging.INFO, log_file=None):
         # Create file handler
         if log_file:
             if os.path.isfile(log_file):
-                old_file = os.path.join(os.path.dirname(log_file), 'arc.old.log')
-                if os.path.isfile(old_file):
-                    os.remove(old_file)
-                shutil.copy(log_file, old_file)
-                os.remove(log_file)
```
alongd: We still need to keep this line (of removing the previous log file), otherwise we'll get one log file that contains all runs

alongd:
Looking good! I added a small comment, you can also squash the commits

arc/main.py
oscarwumit: Thanks. I just added that line back.

alongd:
@oscarwumit, looking great!
Did you try it out to see that it behaves as expected?

oscarwumit:
@alongd I tested the archive feature and it works as expected.

PR #100, opened 2019-03-21
Use the Gaussian check file

alongd:
The Gaussian check file `check.chk` will now be used from previous Gaussian jobs to speed up subsequent jobs and maintain WF consistency.
The most recent path of a check file from a composite / opt / conformer job is saved as a Species attribute, and passed to the next job executed for this species. If the next job uses Gaussian, the check file will be uploaded along with the input file (and will be used).

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
 # encoding: utf-8
+"""
+The ARC Job module
+"""
+
 
 from __future__ import (absolute_import, division, print_function, unicode_literals)
 import os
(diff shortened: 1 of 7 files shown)


alongd:
@dranasinghe, try using this branch, please let me know if it behaves as expected. Thanks!

alongd:
I think I haven't added `guess=read` to the input files if the check file is available...

arc/job/job.py
```diff
@@ -150,21 +150,21 @@ def __init__(self, project, settings, species_name, xyz, job_type, level_of_theo
                             self.method, self.basis_set))
                     self.software = 'gaussian'
                 elif 'b97' in self.method or 'm06-2x' in self.method or 'def2' in self.basis_set:
-                    if not self.settings['qchem']:
-                        raise JobError('Could not find the QChem software to run {0}/{1}'.format(
+                    if not self.settings['gaussian']:
```
alongd: As discussed, let's make QChem the fallback software if Gaussian isn't found.
Also, QChem should deal with `m06-2x`, and Gaussian should deal with `m062x` (is this correct?)
Same comment for scan jobs below

arc/job/job.py
```diff
@@ -410,12 +410,12 @@ def write_input_file(self):
...
                 if self.is_ts:
-                    job_type_1 = 'opt=(ts, calcfc, noeigentest)'
+                    job_type_1 = 'opt=(ts, calcfc, noeigentest) guess=mix'
                 else:
-                    job_type_1 = 'opt=(calcfc, noeigentest)'
+                    job_type_1 = 'opt=(calcfc, noeigentest) guess=mix'
                 if self.fine:
                     # Note that the Acc2E argument is not available in Gaussian03
-                    fine = 'scf=(tight, direct) integral=(grid=ultrafine, Acc2E=12)'
+                    fine = 'scf=(tight, direct) integral=(grid=ultrafine, Acc2E=12) guess=read'
```
alongd: `fine` is appended to `job_type_1` when writing the input file. This will result in having both `guess=mix` and `guess=read` in the same Gaussian directive. Is this the desirable behaviour?

arc/job/job.py
```diff
@@ -491,7 +491,7 @@ def write_input_file(self):
 
         if self.job_type == 'sp':
             if self.software == 'gaussian':
-                job_type_1 = 'scf=(tight, direct) integral=(grid=ultrafine, Acc2E=12)'
+                job_type_1 = 'scf=(tight, direct) integral=(grid=ultrafine, Acc2E=12) guess=read'
```
alongd: What if the optimization was done in a different software, and then we run sp in Gaussian? Will it complain if it has `guess=read` when the checkfile isn't available?

arc/job/job.py
```diff
@@ -565,8 +565,9 @@ def write_input_file(self):
             f.write(self.input)
         if self.settings['ssh']:
             self._upload_input_file()
-            if os.path.isfile(self.checkfile):
-                self._upload_check_file(local_check_file_path=self.checkfile)
+            if self.checkfile is not None:
+                if os.path.isfile(self.checkfile):
```
alongd: Please change to `if self.checkfile is not None and  os.path.isfile(self.checkfile):`

arc/job/job.py
```diff
@@ -588,6 +589,7 @@ def _upload_check_file(self, local_check_file_path=None):
             else local_check_file_path
         if os.path.isfile(local_check_file_path) and self.software.lower() == 'gaussian':
             ssh.upload_file(remote_file_path=remote_check_file_path, local_file_path=local_check_file_path)
+            logging.info('uploading checkpoint file\n')
```
alongd: Was this logging statement added on purpose, or was it left from debugging and testing this brunch?
If we choose to keep it, maybe change `logging.info` to `logging.debug` to only log this when debugging?

arc/main.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn
                  ts_guess_level='', fine=True, generate_conformers=True, scan_rotors=True, use_bac=True,
                  model_chemistry='', ess_settings=None, initial_trsh=None, t_min=None, t_max=None, t_count=None,
                  verbose=logging.INFO, project_directory=None, max_job_time=120, allow_nonisomorphic_2d=False,
-                 job_memory=1500):
+                 job_memory=15000):
```
alongd: Please also change the default memory in Scheduler's `init` function.
Then, In Scheduler under `troubleshoot_ess()` where we increase the memory allocation, we should use a multiplication factor instead of the current `memory = 3000` assignment.

We should also check that this work OK for molpro, where memory is specified in MW, and I think ARC just divides the value here by 2 for a rough estimation. 7500 MW is feasible, but might consume too much resources on the RMG server.
Currently we do in job.py:
```python
        if self.software == 'molpro':
            # molpro's memory is in MW, 750 should be enough
            memory /= 2
        self.memory = memory
```

PR #101, opened 2019-03-21
bug fix: if statement in model chemistry

oscarwumit:
The if statement in model chemistry is problematic.

If self.model_chemistry is defined, then stat_mech_job.modelChemistry = self.model_chemistry
Else, when self.model_chemistry is NOT defined, we should ban the use of AtomEnergyCorrections


alongd:
@oscarwumit , this branch still appears to me as not rebased. Can you take a look?

oscarwumit:
@alongd I rebased it again and it should work now. Please let me know if any issue occurs. Thanks.

alongd:
thanks!

PR #104, opened 2019-03-25
Allow ARCSpecies to be initialized with a conformer list file

alongd:
Users can now initialize an ARCSpecies object with either of the `conformers_before_optimization.txt` or `conformers_after_optimization.txt` files. These files are automatically generated by ARC and saved in the Species 'geometry' folder. The latter also reports the conformer energies.

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -234,7 +234,10 @@ class Job(object):
         self.submit = ''
         self.input = ''
         self.server_nodes = list()
-        self._write_initiated_job_to_csv_file()
+        if job_num is None:
+            # this checks jon_num and not self.job_num on purpose
+            # if job_num was given, then don't save as initiated jobs, this is a restarted job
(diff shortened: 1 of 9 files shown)


alongd:
Thanks!
Regarding point A, the way to define an ARCSpecies with an RMG Species might be:
```python
rmg_spc = Species(molecule=[Molecule().fromSMILES("C=C[O]")])
arc_spc = ARCSpecies(rmg_species=rmg_spc, conformers_path='path/to/conformers.txt')
```
I wouldn't want to add `conformers_path` to the RMG species, since this info isn't being used in the RMG object.

Will https://github.com/ReactionMechanismGenerator/ARC/pull/106 solve point B?

oscarwumit:
@alongd  I just tested that PR #106 will NOT resolve point B. This problem is mentioned in issue #107. Otherwise, this PR works as expected. I think we should merge this PR, and address #107 in a separate PR.

PR #106, opened 2019-03-26
BugFix: Don't overwrite self.mol if self.mol_from_xyz is None

oscarwumit:
Temporary fix. Need to troubleshoot why self.mol_from_xyz is None
i.e. why it does not parse xyz as expected

### DIFF
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -74,7 +74,8 @@ class ARCSpecies(object):
     `ts_guesses`            ``list``     A list of TSGuess objects for each of the specified methods
     `successful_methods`    ``list``     Methods used to generate a TS guess that successfully generated an XYZ guess
     `unsuccessful_methods`  ``list``     Methods used to generate a TS guess that were unsuccessfully
-    `chosen_ts`             ``int``      The TSGuess index corresponding to the chosen TS conformer used for optimization
+    `chosen_ts`             ``int``      The TSGuess index corresponding to the chosen TS conformer used for
+                                            optimization
     `chosen_ts_method`      ``str``      The TS method that was actually used for optimization
     `ts_conf_spawned`       ``bool``     Whether conformers were already spawned for the Species (representing a TS)
                                            based on its TSGuess objects
@@ -210,9 +211,9 @@ class ARCSpecies(object):
                 elif smiles:
                     self.mol = Molecule(SMILES=smiles)
             if not self.is_ts and self.mol is None and self.generate_thermo:
-                logging.warn('No structure (SMILES, adjList, RMG:Species, or RMG:Molecule) was given for species'
-                             ' {0}, NOT using bond additivity corrections (BAC) for thermo computation.'.format(
-                    self.label))
+                logging.warning('No structure (SMILES, adjList, RMG:Species, or RMG:Molecule) was given for species'
+                                ' {0}, NOT using bond additivity corrections (BAC) for thermo computation.'
+                                .format(self.label))
 
             if not self.is_ts:
                 # Perceive molecule from xyz coordinates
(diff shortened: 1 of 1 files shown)


arc/species/species.py
```diff
@@ -827,6 +830,8 @@ def mol_from_xyz(self, xyz=None):
                                  'Got xyz:\n{0}\n\nwhich corresponds to {1}\n{2}\n\nand: {3}\n{4}'.format(
                                    xyz, self.mol.toSMILES(), self.mol.toAdjacencyList(),
                                    original_mol.toSMILES(), original_mol.toAdjacencyList()))
+            if self.mol is None:
+                self.mol = original_mol  # todo: Atom order will not be correct, need fix
         else:
             self.mol = molecules_from_xyz(xyz, multiplicity=self.multiplicity, charge=self.charge)[1]
```
oscarwumit: @alongd What if after the else statement, molecules_from_xyz gives None. In this case, it seems that we will still have a problem of self.mol = None. How should we resolve that?

arc/species/species.py
```diff
@@ -195,22 +195,25 @@ def __init__(self, is_ts=False, rmg_species=None, mol=None, label=None, xyz=None
                                     self.mol.toSMILES(), self.label))
                 self.multiplicity = self.rmg_species.molecule[0].multiplicity
                 self.charge = self.rmg_species.molecule[0].getNetCharge()
-            else:
-                # parameters were entered directly, not via an RMG:Species object
+
+            # parameters were entered directly, not via an RMG:Species object
```
alongd: This comment should be removed, since the code here refers to all cases now

arc/species/species.py
```diff
@@ -195,22 +195,25 @@ def __init__(self, is_ts=False, rmg_species=None, mol=None, label=None, xyz=None
...
-                if not self.is_ts and self.mol is None and self.generate_thermo:
-                    logging.warn('No structure (SMILES, adjList, RMG:Species, or RMG:Molecule) was given for species'
-                                 ' {0}, NOT using bond additivity corrections (BAC) for thermo computation.'.format(
-                        self.label))
+            if multiplicity is not None:
                 self.multiplicity = multiplicity
+            if charge is not None:
                 self.charge = charge
+            if self.mol is None:
+                if adjlist:
```
alongd: Let's add `mol` before adjList, inchi and SMILES (it can also be directly given)

arc/species/species.py
alongd: If this condition is true, then this means that self.mol is None to begin with, and `molecules_from_xyz` is our last resort. If it doesn't work, the user is expected to give some 2D representation. When we do it in an automated fashion, we have an RMG Species object, so it shouldn't be a problem

alongd:
Thanks for your contribution!
Regarding the commit message, I prefer to have a `BugFix` in the beginning, this capitalization is easier to spot when briefly looking at a bunch of commits. Also, in the commit message, please replace `= None` with `is None`.

arc/species/species.py
oscarwumit: Don't we already have self.mol = mol in the init? See line 118, species.py

arc/species/species.py
oscarwumit: I see.

arc/species/species.py
oscarwumit: Ok.

oscarwumit:
@alongd Many thanks for your review. I have changed the commit message. Please let me what do you think of the self.mol = mol in the init.

arc/species/species.py
alongd: good point

alongd:
Could you change the commit message (not the PR title)?

oscarwumit:
@alongd I updated the commit message and also rebased.

arc/species/species.py
```diff
@@ -195,22 +195,24 @@ def __init__(self, is_ts=False, rmg_species=None, mol=None, label=None, xyz=None
...
+                    self.mol = Molecule().fromAdjacencyList(adjlist=adjlist)
+                elif inchi:
+                    self.mol = rmg_mol_from_inchi(inchi)
+                elif smiles:
+                    self.mol = Molecule(SMILES=smiles)
+            if not self.is_ts and self.mol is None and self.generate_thermo:
+                logging.warn('No structure (SMILES, adjList, RMG:Species, or RMG:Molecule) was given for species'
```
alongd: Please change `logging.warn` (which will be deprecated soon) to `logging.warning`

alongd:
I added another small comment.
Also, the commit massage says "temp, need to see why we don't get the correct xyz...", so please open an issue and give the details of the species we saw the bug for, so we'll remember to deal with it.

arc/species/species.py
oscarwumit: I just made another commit to replace all `logging.warn` to `logging.warning` in species.py. I also changed the code to conform to PEP 8.

oscarwumit:
Many thanks for the comment. I opened an issue addressing our problem here.

PR #108, opened 2019-03-28
Set default max stepsize for optimization in Gaussian

oscarwumit:
As suggested by Duminda, setting MaxStep = 5 in Gaussian gives better TS
result. The default value in G16 is MaxStep = 30 (less rigorious)


alongd:
Thanks! Then let's also add it to `optfreq` and `composite` job types

oscarwumit:
I added maxstep=5 to all related gaussian TS job types. @alongd

alongd:
Thanks!!

PR #113, opened 2019-04-09
YAML file related changes

oscarwumit:
1. BugFix: Create geometry directory before copying files 
    ARC will create geometry directory in the local os.path if the directory isn't exist. This resolves a copy error occured when running yml input file

2. Feature: Define fake model_chemistry to force generate YAML species files
    Now user can specify model_chemistry.lower() == 'fake' to force ARC generate YAML species files.
    This feature is useful for rate calculation with model_chemistry not defined in Arkane.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -121,8 +121,12 @@ class ARC(object):
             self.scan_rotors = scan_rotors
             self.run_orbitals = run_orbitals
             self.use_bac = use_bac
-            self.model_chemistry = model_chemistry
-            if self.model_chemistry:
+            if model_chemistry.lower() == 'fake':
+                self.model_chemistry = 'cbs-qb3'
+                logging.warning('user specified model_chemistry is fake, ARC is using cbs-qb3 model_chemistry to '
+                                'generate YAML files for species. Warning: Do NOT use YAML for thermo calculation')
+            else:
+                self.model_chemistry = model_chemistry
                 logging.info('Using {0} as model chemistry for energy corrections in Arkane'.format(self.model_chemistry))
             if not self.fine:
                 logging.info('\n')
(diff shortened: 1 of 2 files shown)


arc/processor.py
```diff
@@ -378,5 +378,8 @@ def copy_freq_output_for_ts(self, label):
         Copy the frequency job output file into the TS geometry folder
         """
         calc_path = os.path.join(self.output[label]['freq'])
-        output_path = os.path.join(self.project_directory, 'output', 'rxns', label, 'geometry', 'frequency.out')
+        out_path = os.path.join(self.project_directory, 'output', 'rxns', label, 'geometry')
+        if not os.path.exists(out_path):
```
alongd: `output_path` and `out_path` are confusing names, not informative of their content (that one is a directory path).
Could you instead keep the original `output_path`, and change the condition to be as suggested below?
`if not os.path.exists(os.path.dirname(output_path)):`

arc/main.py
```diff
@@ -121,8 +121,12 @@ def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn
             self.scan_rotors = scan_rotors
             self.run_orbitals = run_orbitals
             self.use_bac = use_bac
-            self.model_chemistry = model_chemistry
```
alongd: in the commit message, could you replace
`model_chemistry.lower() == 'fake'`
with
`model_chemistry = 'fake'`
?

arc/main.py
```diff
@@ -121,8 +121,12 @@ def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn
...
-            if self.model_chemistry:
+            if model_chemistry.lower() == 'fake':
```
alongd: How about renaming 'fake' as 'kinetics'?

alongd:
@oscarwumit, what is the status of this PR? do we still need it?

PR #110, opened 2019-03-31
Automatic node test and swtich for errored jobs on OGE

oscarwumit:
ARC now performs automatic node test and switch for errored jobs on OGE server (currently implemented for Pharos). 

Features highlight:

1. ARC now provides detailed job status for OGE cluster. For detailed queue status and state codes explanation, see:
"https://softpanorama.org/HPC/Grid_engine/Queues/queue_states.shtml"

2. Added scripts.py to better organize ARC scripts

PR #118, opened 2019-04-24
 Node availability test and switch for error-ed jobs on OGE clusters

oscarwumit:
This PR enables node availability test and automatic switch for error-ed jobs on OGE clusters. 
This PR is improved from PR #110 (closed and not merged)

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -84,6 +84,9 @@ class Job(object):
     `occ`              ``int``           The number of occupied orbitals (core + val) from a molpro CCSD sp calc
     `project_directory` ``str``          The path to the project directory
     `max_job_time`     ``int``           The maximal allowed job time on the server in hours
+    `available_nodes_dict`  ``dict``     Dictionary of list of available nodes on severs.
+                                         {server1: [list of working nodes], server2: [list of working nodes]}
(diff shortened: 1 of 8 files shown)


oscarwumit:
@alongd I closed PR 110 and opened this one with updated code. Please let me know if you have any questions. Thanks.

arc/job/job.py
```diff
@@ -1006,34 +1007,26 @@ def troubleshoot_server(self):
...
-                else:
-                    content.insert(7, '#$ -l h=node{0}.cluster'.format(node))
-                content = ''.join(content)  # convert list into a single string, not to upset paramiko
-                # resubmit
-                ssh.upload_file(remote_file_path=os.path.join(self.remote_path,
-                                submit_filename[servers[self.server]['cluster_soft']]), file_string=content)
-                self.run()
+                if self.server_test == 3:
+                    logging.error('Server error. No nodes available. Tested 3 times.')
+                if 'pharos_8core' not in self.available_nodes_dict and 'pharos_48core' not in self.available_nodes_dict:
```
alongd: Again, we're hard-coding here for Pharos without checking which server that is first. So for non-Pharos servers this condition will always be true. We wouldn't want users to have to change code all around, just in the scripts.py

Issue #129, opened 2019-05-03
adaptive distance for auto1dmin


goldmanm:
When using auto1dmin, ARC gives a minimum distance between the bath gas and the molecule of interest of 2 A and maximum distance of 5 A. When running for larger molecules, it might make sense that the distance between the two molecules increase. To make auto1dmin more reliable, it might make sense to use adapt the range of distance depending on the size of the molecule and bath gas. For example, when running a C4H10O3, the original settings failed, but the adjusting to 3 A and 6 A gave lennard jones parameters

It also seems like the wider range given, the longer it takes to run.

alongd:
Excellent observation, thanks!

PR #131, opened 2019-05-09
Use adaptive OneDMin radii

alongd:
**Assuming** the species is sphere-like, determine the r_min and r_max radii to use for a OneDMin L-J calculation.
Addresses #129

### DIFF
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -30,7 +30,7 @@ from rmgpy.molecule.resonance import generate_kekule_structure
 from rmgpy.transport import TransportData
 from rmgpy.exceptions import InvalidAdjacencyListError
 
-from arc.common import get_logger
+from arc.common import get_logger, get_atom_radius, get_center_of_mass
 from arc.arc_exceptions import SpeciesError, RotorError, InputError, TSError
 from arc.settings import arc_path, default_ts_methods, valid_chars, minimum_barrier
 from arc.parser import parse_xyz_from_file, parse_dipole_moment, parse_polarizability
@@ -106,6 +106,8 @@ class ARCSpecies(object):
     `optical_isomers`       ``int``      Whether (=2) or not (=1) the species has chiral center/s
     `transport_data`        ``TransportData``  A placeholder for updating transport properties after Lennard-Jones
                                                  calculation (using OneDMin)
+    `r_min`            ``int``           The minimum radius for OneDMin in Angstrom
+    `r_max`            ``int``           The maximum radius for OneDMin in Angstrom
     ====================== ============= ===============================================================================
 
     Dictionary structure:
@@ -125,7 +127,7 @@ class ARCSpecies(object):
     def __init__(self, is_ts=False, rmg_species=None, mol=None, label=None, xyz=None, multiplicity=None, charge=None,
                  smiles='', adjlist='', inchi='', bond_corrections=None, generate_thermo=True, species_dict=None,
                  yml_path=None, ts_methods=None, ts_number=None, rxn_label=None, external_symmetry=None,
-                 optical_isomers=None, run_time=None, checkfile=None, number_of_radicals=None):
+                 optical_isomers=None, run_time=None, checkfile=None, number_of_radicals=None, r_min=None, r_max=None):
         self.t1 = None
         self.ts_number = ts_number
         self.conformers = list()
@@ -149,6 +151,8 @@ class ARCSpecies(object):
         self.conformer_checkfiles = dict()
         self.most_stable_conformer = None
         self.transport_data = TransportData()
+        self.r_min = r_min
+        self.r_max = r_max
 
         if species_dict is not None:
             # Reading from a dictionary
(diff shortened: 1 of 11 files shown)


arc/species/species.py
```diff
@@ -1041,6 +1042,42 @@ def set_transport_data(self, lj_path, opt_path, bath_gas, opt_level, freq_path='
             comment=str(comment)
         )
 
+    def determine_radius(self):
```
goldmanm: Does the xyz used in ARC have a center of mass at 0,0,0? If not, this method can easily overestimate the radius.

arc/species/species.py
```diff
@@ -1041,6 +1042,42 @@ def set_transport_data(self, lj_path, opt_path, bath_gas, opt_level, freq_path='
...
+        Assumes that the species is close to a sphere
+        """
+        if bath_gas not in ['He', 'Ne', 'Ar', 'Kr', 'H2', 'N2', 'O2']:
+            raise SpeciesError('Unknown bath gas. Should be either He, Ne, Ar, Kr, H2, N2, or O2, '
+                               'got {0}'.format(bath_gas))
+        bath_gas_radii_dict = {'He': 1.40, 'Ne': 1.54, 'Ar': 1.88, 'Kr': 2.0,
+                               'H2': 1.70, 'N2': 1.95, 'O2': 1.88}  # in Angstrom
+        bath_gas_r = int(math.ceil(bath_gas_radii_dict[bath_gas]))
+        species_radius = int(math.ceil(self.determine_radius()))
+        r_min = species_radius + bath_gas_r
```
goldmanm: I think this might be an overestimate, since the minimum radius could fall below this value.

goldmanm: Taking the max distance is likely to be an overestimate. Since we are taking the max distance in the previous one, maybe this would work better as r_max instead of r_min

arc/species/speciesTest.py
```diff
@@ -752,6 +753,23 @@ def test_xyz_from_dict(self):
...
+            self.assertAlmostEqual(spc.determine_radius(), expected_rs[i], 5)
+        with self.assertRaises(SpeciesError):
+            self.spc3.determine_radius()
+
+    def test_determine_onedmin_radii(self):
+        """Test determining the OneDMin radii"""
+        bath_gases = ['H2', 'N2', 'Ar', 'H2']
+        expected_rs = [(3, 6), (4, 7), (4, 7), (3, 6)]
+        for i, spc in enumerate([self.spc2, self.spc6, self.spc8, self.spc9]):
+            self.assertEqual(spc.determine_onedmin_radii(bath_gas=bath_gases[i]), expected_rs[i])
```
goldmanm: A more robust test might ensure that the actual radius that 1dmin would predict falls within the range arkane guesses.

Though even if it is within the range but near the edge, 1dmin can fail. That's what happened with my ibutanol peroxy, which eventually obtained a radius of 4.6 but failed when the range was 2-5.

arc/species/species.py
alongd: I think this should be the center of geometry, not of mass, right? I.e., not weighting atom numbers in the average

alongd:
Thanks for the comments!
I added fixups, let me know what you think. I plan to test it out with OneDMin before merging

arc/species/species.py
goldmanm: I am not entirely sure what the best way to do this should be. Jasper 2014 describes $r$ as center-of-mass distance between the bath gas and molecule, so I was thinking centering around the mass would be a better metric. Though I am not convinced that the center of mass is a better metric to determine the guess for the radius. Maybe the center of atoms/geometry would be better.

arc/species/species.py
```diff
@@ -1072,11 +1072,11 @@ def determine_onedmin_radii(self, bath_gas):
                                'got {0}'.format(bath_gas))
         bath_gas_radii_dict = {'He': 1.40, 'Ne': 1.54, 'Ar': 1.88, 'Kr': 2.0,
                                'H2': 1.70, 'N2': 1.95, 'O2': 1.88}  # in Angstrom
-        bath_gas_r = int(math.ceil(bath_gas_radii_dict[bath_gas]))
-        species_radius = int(math.ceil(self.determine_radius()))
-        r_min = species_radius + bath_gas_r
+        bath_gas_r = bath_gas_radii_dict[bath_gas]
+        species_radius = self.determine_radius()
+        r_min = species_radius + bath_gas_r + 0.2
```
goldmanm: One question I have is why add 0.2 A to the minimum radius.

goldmanm: At least with the older algorithm using ibutanol peroxy as a test case, the radius given by the method appeared to be an over estimate, such that I think 

```
r_max = max((species_radius + bath_gas_r, 4.5))
r_min = species_radius - 3
```

would lead to more successful calculations than the other algorithm.

alongd: I think we don't have the same definition to `r_max`. I think r_max is the farthest point the bath gas is at from the axis origin, so it cannot be `species_radius + bath_gas_r`, because for a spherical species this will put the bath gas right at the edge of the sphere, which is `r_min`

alongd:
I removed the xyz centring function, following the e-mail conversation with Ahren. I assume that the xyz is centered around the center of mass, which is the case for ESS output.
Let me know if you have any comments, perhaps we could meet next week to discuss it.

PR #133, opened 2019-06-03
Minor changes

xiaoruiDong:
I was trying to install the ARC based on the instruction from Wiki and encountered errors when running ESS diagnostics:

* Change the isomorphic_species_list to same_species_list to keep up with the latest update of RMG
* Avoid error when running into local address error (double checking, but treat local as SSH) in ESS check
* No ARC.determine_remote exists. I guess that it updates to ARC.determine_ess_settings, so I changed it. I also removed the output of iPython notebook, but it can be kept if you want.


alongd:
Thanks for the contribution!

I think the iPython output should stay, as an example of what to expect from this tool, unless it doesn't show well. If we do decide to remove it, then it should have its own commit.

I understand that the ESS check is buggy when running on a server. How does the addition of
```python
            if server == 'local':
                continue
```
treat local as SSH? I think the expected behaviour is that we will be able to determine ESS when running locally as well.

xiaoruiDong:
Besides, I just noticed that even if I don't have orca on my local machine, the ess will still tells me that I have. Probably, there might be another bug about it.

alongd:
Thanks, I think the code looks good, just some comments regarding the commits:
1. The iPython notebook change should be in its own separate commit, not joint with the local server check
2. Please rename the `Update the rmg_function name: same_species_list. ` commit message as `Updated isomorphic species check method name in rmgdb` (since `rmg_function` with an underscore seems like a real function name)
3. Generally, commit messages should be at a maximal length of one line, then further description can follow after a line break, to avoid getting:
```
 Update the function name in notebook and avoid local server check err…

…or in ESS diagnostics```

xiaoruiDong:
@alongd Do you think these changes in commits are slightly better? Thanks?

alongd:
Looking great! Let's wait for the tests and get the fixes in.
Thanks!

PR #136, opened 2019-06-14
Added a ZPE/freq scaling factor script to utils

alongd:
Based on 10.1016/j.cpc.2016.09.004, adapted by Duminda
Also added tests, will soon add an iPy notebook


alongd:
Todo: run this automatically if ARC identifies that Arkane lacks the freq. scaling factor for the requested level of theory

arc/utils/scale.py
```diff
@@ -0,0 +1,410 @@
...
+
+    Args:
+        zpe_dict (dict): The calculated vibrational zero-point energies at the requested level of theory.
+                         Keys are species labels, values are floats representing the ZPE in J/mol.
+        level_of_theory (str, unicode): The frequencies level of theory.
+
+    Returns:
+        float: The scale factor for the vibrational zero-point energy (lambda ZPE) as defined in reference [2].
+    """
+    if len(zpe_dict) < 15:
```
alongd: Should use `unconverged = [key for key, val in zpe_dict.items() if val is None]` here as well

alongd:
@dranasinghe, I think this branch is ready now, feel free to only focus on this commit: https://github.com/ReactionMechanismGenerator/ARC/pull/136/commits/7b4d666584844de3487423b4aa1c5f6c223c3888
Thanks!

alongd:
@dranasinghe, I'm about to merge this in soon, let me know if you have any comments re [scale.py](https://github.com/ReactionMechanismGenerator/ARC/blob/scaling_factors/arc/utils/scale.py)

dranasinghe:
Dear Alon,
everything is looking good in the scale.py. I made changes to job.py. line
570 and 581
"                    job_type_1 = 'opt'      "
and 581
 "                   job_type_1 = 'opt=(tight)' "
I removed calcfc, noeigentest from the ground state opt. Gaussian will
throw an error if we request a method that Gaussian does not have 2nd
derivative eg. CCSD. BTW this will make optimization faster.
Duminda

PR #137, opened 2019-06-15
Organized job memory handeling

alongd:
- Organized the memory used for each ESS
- Added an optional memory attribute to the servers dictionary to limit the requested memory if needed


alongd:
Update `using memory: {mem}` units in Scheduler

PR #141, opened 2019-06-28
xyz to smiles

alongd:
Added an iPy notebook with an xyz to SMILES script

Also specify numpy and RDKit version

Also, don't load the RMG database if not necessary


alongd:
Added pydot-ng to environment.yml, see https://github.com/ReactionMechanismGenerator/RMG-Py/pull/1644

PR #143, opened 2019-07-03
A new way of thinking about conformers

alongd:
- Generate non-TS conformers bottom-up instead of top-down
- Use Gromacs with a fitted force field if desired (if species.force_field is set to 'fit')


alongd:
Tasks:
- [ ] Minimize acpype as much as possible and add tests

PR #145, opened 2019-07-17
File compatibility with MAC OS (skip .DS_Store files)

xiaoruiDong:
Description: 
Bugfix for:
(Minor) RMG Library loading. 
Problem: 
Mac will generate .DS_Store for each folder. The current loading process will read these files. 
Solution:
Avoid loading these files by adding ifs.

### DIFF
--- a/arc/rmgdb.py
+++ b/arc/rmgdb.py
@@ -74,8 +74,10 @@ def load_rmg_database(rmgdb, thermo_libraries=None, reaction_libraries=None, kin
     if not thermo_libraries:
         thermo_path = os.path.join(db_path, 'thermo', 'libraries')
         for thermo_library_path in os.listdir(thermo_path):
-            thermo_library, _ = os.path.splitext(os.path.basename(thermo_library_path))
-            thermo_libraries.append(thermo_library)
+            # Avoid reading .DS_store files for compatibility with Mac OS
+            if not thermo_library_path.startswith('.'):
+                thermo_library, _ = os.path.splitext(os.path.basename(thermo_library_path))
+                thermo_libraries.append(thermo_library)
         # prioritize libraries
         thermo_priority = ['BurkeH2O2', 'thermo_DFT_CCSDTF12_BAC', 'DFT_QCI_thermo', 'Klippenstein_Glarborg2016',
                            'primaryThermoLibrary', 'primaryNS', 'NitrogenCurran', 'NOx2018', 'FFCM1(-)',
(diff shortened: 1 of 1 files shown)


arc/job/job.py
```diff
@@ -379,7 +379,7 @@ def write_input_file(self):
 {job_type_2}
 ---;"""
 
-        if self.job_type in ['conformer', 'opt']:
+        if self.job_type in ['conformer', 'opt', 'composite']:
```
alongd: We consider composite jobs in [line 497](https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/job/job.py#L497)
If it doesn't function as expected, please consider fixing it with the desired keywords instead of adding `composite` here

xiaoruiDong: Thanks Alon. I suddenly find that it is not the case, and it won't fix the problem. The actual problem happens in scheduler.py line 706 where `self.run_job(label=label, xyz=xyz, level_of_theory=self.composite_method, job_type='composite', fine=False)` . So, composite method will never run with fine_grid. May I know why is that?

alongd: Good question. Note that `run_opt_job()` has `fine=False` as well. We resubmit a second opt job with a fine grid if the user requested fine grids, but this is done in `parse_opt_geo()`.
The fix to your issue should be to the `parse_composite_geo()` method, so it behaves like `parse_opt_geo()` for everything related to resubmitting a fine grid job.
We might also consider merging these methods.

alongd:
@xiaoruiDong, let me know what the status of this PR is, or whether you'd like me to assist

xiaoruiDong:
@alongd Sorry for my delay. I will have time later today or tomorrow. Then I will go through `parse_opt_geo()` and decide whether update it or add `parse_composite_geo()`. Thanks. I will let you know if I have any questions.

xiaoruiDong:
Alon, I read through the code and find out that the fine grid job can be only submitted after a normal -gird job, am I right? However, I think it would not solve my problem. 
Let me describe the situation I ran into:
I am calculating HCO based on the default setting of ARC on c3ddb using g09. the keywords are `opt=(calcfc, noeigentest, tight) guess=read cbs-qb3 scf=(qc,nosymm) iop(2/9=2000)`. Due to the use of tight and HCO being a really small molecule, the convergence thresholds are too small for normal-grid job (O(1e-5) for both force and displacement), thus it cannot converge.

alongd:
We should solve the original problem you described, I'd happily do it.
Seems like for your case you need a troubleshooting method (we have plenty of these, let's add what you described)
Could you come up with heuristics for situations where a fine grid is needed to begin with? E.g., molecule size? Also, please upload the Gaussian log file so I'll see the error messages.

alongd:
I addressed the [original issue](https://github.com/ReactionMechanismGenerator/ARC/commit/e2e3abdd845e2a6bff3a69fc55a262a618bb6712) (already merged), perhaps that'll help. Let's talk tomorrow about additional changes.

alongd:
@xiaoruiDong, would you like to change this PR to include the MAC fix only?

xiaoruiDong:
@alongd I updated the branch. Thanks for the reminder!

arc/rmgdb.py
```diff
@@ -74,8 +74,9 @@ def load_rmg_database(rmgdb, thermo_libraries=None, reaction_libraries=None, kin
     if not thermo_libraries:
         thermo_path = os.path.join(db_path, 'thermo', 'libraries')
         for thermo_library_path in os.listdir(thermo_path):
-            thermo_library, _ = os.path.splitext(os.path.basename(thermo_library_path))
-            thermo_libraries.append(thermo_library)
+            if not thermo_library_path.startswith('.'):
```
alongd: Great! Could you add a comment, something like
`# Avoid reading ".DS_store" files for compatibility with Mac OS`
for code mainteinability?

xiaoruiDong: Yes, of course

PR #159, opened 2019-08-02
Remove rotor info related to re-run after conformer correction

xiaoruiDong:
Fix1:
Fix the logging message when detecting inconsistencies between two consecutive points.

Fix2:
Remove all the information (keep pivots, scan, and time_set_dihedrals) when a conformer correction needs to be made.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -1522,7 +1522,7 @@ class Scheduler(object):
                                     message += error_message + '; '
                                     invalidate = True
                                     invalidation_reason = 'two consecutive points are inconsistent by more than ' \
-                                                          '{0:.2f} kJ/mol'.format(inconsistency_ab)
+                                                          '{0:.2f} kJ/mol'.format(inconsistency_ab * max(v_list))
                                     if not job.scan_trsh:
                                         logger.info('Trying to troubleshoot rotor {0} of {1}...'.format(
                                             job.pivots, label))
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -1563,6 +1563,12 @@ class Scheduler(object):
                                     scan=self.species_dict[label].rotors_dict[i]['scan'],
                                     deg_increment=min_index*rotor_scan_resolution)
                                 self.delete_all_species_jobs(label)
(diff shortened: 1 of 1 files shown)


arc/scheduler.py
```diff
@@ -1489,7 +1489,7 @@ def check_scan_job(self, label, job):
                                     message += error_message + '; '
                                     invalidate = True
                                     invalidation_reason = 'two consecutive points are inconsistent by more than ' \
-                                                          '{0:.2f} kJ/mol'.format(inconsistency_ab)
+                                                          '{0:.2f} kJ/mol'.format(inconsistency_ab*max(v_list))
```
alongd: Thanks!
Please add spaces between the arguments, i.e.:
`inconsistency_ab * max(v_list)`

arc/scheduler.py
```diff
@@ -1530,6 +1530,12 @@ def check_scan_job(self, label, job):
                                     scan=self.species_dict[label].rotors_dict[i]['scan'],
                                     deg_increment=min_index*rotor_scan_resolution)
                                 self.delete_all_species_jobs(label)
+                                # Remove all completed rotor calculation information
+                                for rotor in self.species_dict[label].rotors_dict.values():
```
alongd: Do you think it'll also work to just do
```python
for rotor in self.species_dict[label].rotors_dict.values():
    rotor = dict()
```
?
Or is there a special dictionary initialization we need?

arc/scheduler.py
xiaoruiDong: Done

arc/scheduler.py
xiaoruiDong: We need to remove these attributes and still keep pivots, scan, and time_set_dihedrals, since they are only generated at the beginning of the job (seems like before conformer jobs?)

alongd: Good point!
Could you please rename this commit? Instead of "BugFix: Remove all rotor info..." have "BugFix: Remove rotor info related to re-run after conformer correction"?
Then please rebase, and looks like we're all set

PR #167, opened 2019-08-06
Better err msg for command error on server

oscarwumit:
More discriptive err msg to deal with unexpected command.

### DIFF
--- a/arc/job/local.py
+++ b/arc/job/local.py
@@ -16,6 +16,7 @@ import re
 
 from arc.job.ssh import check_job_status_in_stdout
 from arc.settings import servers, check_status_command, submit_command, submit_filename, delete_command, output_filename
+from arc.arc_exceptions import SettingsError
 
 ##################################################################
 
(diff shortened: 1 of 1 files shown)


arc/job/local.py
```diff
@@ -29,7 +30,13 @@ def execute_command(command, shell=True):
...
     if not isinstance(command, list) and not shell:
         command = [command]
-    stdout = subprocess.check_output(command, shell=shell)
+    try:
+        stdout = subprocess.check_output(command, shell=shell)
+    except subprocess.CalledProcessError as e:
+        logger.error('The following command is erroneous:\n{0}'.format(e.message))
+        raise SettingsError('The following command is erroneous: \n{0}'
+                            '\n To correct the command, modify settings.py'
+                            '\n Tips: use "which" command to locate cluster software on server.'.format(e.message))
```
alongd: We don't need to add a blank space after a line break (do `\nTo correct...` instead of `\n To correct...`)
In the "Tip", could you add an example, like `e.g., type "which sbatch" on a server running Slurm to identify the correct sbatch path to past in the submit_command dictionary."`
Instead of "to locate cluster software on server.", maybe "to locate cluster software commands on the server."?

oscarwumit: Thanks. I've made the changes you suggested. Please take a look and let me know if additional modifications are needed.

PR #169, opened 2019-08-09
Improved ARC conformer troubleshoot

oscarwumit:
Troubleshoot conformers for a species if all conformers for this
species failed 2D isomorphism test.

Also log the conformer index and xyz ARC used for geometry optimization.
This makes it easier to locate the QM files of the conformer that ARC
uses in the project folder.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -1072,6 +1072,7 @@ class Scheduler(object):
                                      xyzs=self.species_dict[label].conformers, level_of_theory=self.conformer_level,
                                      multiplicity=self.species_dict[label].multiplicity,
                                      charge=self.species_dict[label].charge, is_ts=False)  # before optimization
+        self.species_dict[label].conformers_before_opt = tuple(self.species_dict[label].conformers)
         if self.species_dict[label].initial_xyz is None and self.species_dict[label].final_xyz is None \
                 and not self.testing:
             if len(self.species_dict[label].conformers) > 1:
@@ -1243,18 +1244,22 @@ class Scheduler(object):
                                                  molecules_from_xyz(xyzs[0],
                                                                     multiplicity=self.species_dict[label].multiplicity,
                                                                     charge=self.species_dict[label].charge)[1]))
+                                    self.output[label]['status'] += 'Conformer {0} was found to be the lowest energy ' \
+                                                                    'isomorphic conformer; '.format(i)
                                 conformer_xyz = xyz
+                            self.output[label]['status'] += 'Conformers optimized and compared at {0}; '\
+                                                            .format(self.conformer_level)
                             break
                         else:
                             if i == 0:
-                                self.output[label]['status'] += 'most stable conformer ({0}) did not pass isomorphism ' \
-                                                                'check; '.format(i)
+                                self.output[label]['status'] += 'most stable conformer ({0}) did not pass ' \
+                                                                'isomorphism check; '.format(i)
                                 self.species_dict[label].conf_is_isomorphic = False
                                 logger.warning('Most stable conformer for species {0} with structure {1} was found to '
                                                'be NON-isomorphic with the 2D graph representation {2}. Searching for '
                                                'a different conformer that is isomorphic...'.format(
                                                 label, b_mol.toSMILES(), self.species_dict[label].mol.toSMILES()))
-                else:
+                else:  # all conformers for a species failed isomorphism test
                     smiles_list = list()
                     for xyz in xyzs:
                         try:
@@ -1274,11 +1279,12 @@ class Scheduler(object):
                             logger.warning('Isomorphism check cannot be done for charged species {0}'.format(label))
                         conformer_xyz = xyzs[0]
                     else:
-                        logger.error('No conformer for {0} was found to be isomorphic with the 2D graph representation'
-                                     ' {1} (got: {2}). NOT optimizing this species.'.format(
-                                      label, self.species_dict[label].mol.toSMILES(), smiles_list))
-                        self.output[label]['status'] += 'Error: No conformer was found to be isomorphic with the 2D' \
-                                                        ' graph representation!; '
+                        # troubleshoot when all conformers of a species failed isomorphic test
+                        logger.warning('Isomorphism check for species {0} failed at {1}. '
+                                       'Attempt to troubleshoot at other levels.'.format(label, self.conformer_level))
+                        self.output[label]['status'] += 'No conformer was found to be isomorphic with the 2D' \
+                                                        ' graph representation at {0}; '.format(self.conformer_level)
+                        self.troubleshoot_conformer_isomorphism(label=label)
             else:
                 logger.warning('Could not run isomorphism check for species {0} due to missing 2D graph '
                                'representation. Using the most stable conformer for further geometry'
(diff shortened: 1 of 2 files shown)


arc/scheduler.py
```diff
@@ -1274,11 +1279,13 @@ def determine_most_stable_conformer(self, label):
                             logger.warning('Isomorphism check cannot be done for charged species {0}'.format(label))
                         conformer_xyz = xyzs[0]
                     else:
-                        logger.error('No conformer for {0} was found to be isomorphic with the 2D graph representation'
-                                     ' {1} (got: {2}). NOT optimizing this species.'.format(
-                                      label, self.species_dict[label].mol.toSMILES(), smiles_list))
-                        self.output[label]['status'] += 'Error: No conformer was found to be isomorphic with the 2D' \
-                                                        ' graph representation!; '
+                        # trouble shoot when all conformers of a species failed isomorphic test
```
alongd: make troubleshoot one word

arc/scheduler.py
```diff
@@ -1274,11 +1279,13 @@ def determine_most_stable_conformer(self, label):
...
+                        logger.warning('Isomorphism check for species {0} failed at {1}. '
+                                       'Attempt to troubleshoot at other levels.'.format(label, self.conformer_level))
+                        self.output[label]['status'] += 'No conformer was found to be isomorphic with the 2D' \
+                                                        ' graph representation at {0}; '.format(self.conformer_level)
+                        for conformer in range(0, i):   # rerun conformer job at higher level for all conformers
```
alongd: I suggest using `len(self.species_dict[label].conformers)` instead of `i` for clarity.
Also, let's loop inside the troubleshooting method instead of calling it multiple times (I have no solid reason for it, I just feel it's better, feel free to disagree). That way you only need to pass `label`.

arc/scheduler.py
```diff
@@ -1287,6 +1294,55 @@ def determine_most_stable_conformer(self, label):
             if conformer_xyz is not None:
                 self.species_dict[label].initial_xyz = conformer_xyz
                 self.species_dict[label].most_stable_conformer = xyzs_in_original_order.index(conformer_xyz)
+                logger.info('Conformer {0} is used for geometry optimization, with xyz:\n{1}'
+                            .format(xyzs_in_original_order.index(conformer_xyz), conformer_xyz))
+
+    def troubleshoot_conformer_isomorphism(self, label, conformer):
+        """
+        Trouble shoot conformer optimization for a species that failed isomorphic test in
```
alongd: one word

arc/scheduler.py
```diff
@@ -1287,6 +1294,55 @@ def determine_most_stable_conformer(self, label):
             if conformer_xyz is not None:
                 self.species_dict[label].initial_xyz = conformer_xyz
                 self.species_dict[label].most_stable_conformer = xyzs_in_original_order.index(conformer_xyz)
+                logger.info('Conformer {0} is used for geometry optimization, with xyz:\n{1}'
```
alongd: I think our new conformers method already prints the xyz and displays it in 3D. Perhaps only print the index?

arc/scheduler.py
```diff
@@ -1287,6 +1294,55 @@ def determine_most_stable_conformer(self, label):
...
+                                 ' state geometries.')
+
+        job = self.job_dict[label]['conformers'][conformer]
+
+        if job.software == 'gaussian':
+            conformer_trsh_methods = ['conformer wb97xd/def2TZVP', 'conformer apfd/def2TZVP']
+        elif job.software == 'qchem':
+            conformer_trsh_methods = ['conformer wb97x-d3/def2-TZVP']
+        else:
+            raise SchedulerError('The troubleshoot_conformer_isomorphism() method has not implemented for {0}.'
```
alongd: "is" or "was" instead of "has"

arc/scheduler.py
```diff
@@ -1287,6 +1294,55 @@ def determine_most_stable_conformer(self, label):
...
+                                 .format(job.software))
+
+        for method in conformer_trsh_methods:
+            if method in job.ess_trsh_methods:
+                continue
+            else:
```
alongd: We don't need the `else` if the above conditions `continues`

arc/scheduler.py
```diff
@@ -1287,6 +1294,55 @@ def determine_most_stable_conformer(self, label):
...
+        """
+
+        if self.species_dict[label].is_ts:
+            raise SchedulerError('The troubleshoot_conformer_isomorphism() method does not deal with transition'
+                                 ' state geometries.')
+
+        job = self.job_dict[label]['conformers'][conformer]
+
+        if job.software == 'gaussian':
+            conformer_trsh_methods = ['conformer wb97xd/def2TZVP', 'conformer apfd/def2TZVP']
```
alongd: A minor suggestion: Le't define only levels of theory in these lists. below we can check:
`if 'conformer ' + method in job.ess_trsh_methods:`
and append to the ess_trsh_methods accordingly.
So treating the `ess_trsh_methods` argument is less "smooth", but it'll be more straight forward to modify these lists. Again, feel free to keep the current structure if you prefer.

oscarwumit:
> Thanks for this important contribution!
> Please document the species you used for testing in the PR's comments if in the future we'd like to make tests for this method.

I tested this PR for the following species:
OO[C]1CCCCC1
Using b3lyp/cbsb7 as the initial conformer level will lead to all conformers fail 2D isomorphism test. The troubleshooting method in this PR will then be triggered.

oscarwumit:
Thanks for the review. I've changed the code accordingly. Please take a look again and let me know if further modifications are needed.

arc/species/species.py
```diff
@@ -175,6 +175,7 @@ class ARCSpecies(object):
         conf_is_isomorphic (bool): Whether the lowest conformer is isomorphic with the 2D graph representation
                                    of the species. `True` if it is. Defaults to `None`. If `True`, an isomorphism check
                                    will be strictly enforced for the final optimized coordinates.
+        conformers_before_opt (tuple): Conformers XYZs of a species before optimization.
```
alongd: Out of curiosity: What's the benefit of it being a tuple and not a list?

arc/scheduler.py
```diff
@@ -1287,6 +1293,55 @@ def determine_most_stable_conformer(self, label):
...
+            else:
+                raise SchedulerError('The troubleshoot_conformer_isomorphism() method is not implemented for {0}.'
+                                     .format(job.software))
+
+            for method in conformer_trsh_methods:
+                if 'conformer ' + method in job.ess_trsh_methods:
+                    continue
+                job.ess_trsh_methods.append('conformer ' + method)
+                level_of_theory = method
+                logger.info('Troubleshooting conformer job in {software} for {label} using {level}'
```
alongd: Looks like this message will be printed `i` times (the number of conformers). Would you like to relocate it so it's only printed once per call of this method?

oscarwumit: I have modified the code to log the message only once for each species and each method it tries.

arc/species/species.py
oscarwumit: Short answer: a tuple is not mutable, this ensures conformers_before_opt has the correct xyzs and will never change.

Extended answer: a list, however, is mutable. species.conformers is a list, and its entries change as ARC executes. 

I created the tuple in line 1075 of scheduler.py:        self.species_dict[label].conformers_before_opt = tuple(self.species_dict[label].conformers)
The argument inside the tuple parenthesis is a list that is mutable. If I simply use:
self.species_dict[label].conformers_before_opt = self.species_dict[label].conformers
Then conformers_before_opt will change overtime, giving wrong xyzs.
Alternatively, we can do a deep copy of the list on the right-hand side, but I think this is less elegant than simply creating a tuple.

oscarwumit:
I have made the changes you suggested. Please take a look again. Thanks. @alongd

arc/scheduler.py
```diff
@@ -1287,6 +1293,59 @@ def determine_most_stable_conformer(self, label):
...
+        if self.species_dict[label].is_ts:
+            raise SchedulerError('The troubleshoot_conformer_isomorphism() method does not deal with transition'
+                                 ' state geometries.')
+
+        msg_logged = False
+        num_of_conformers = len(self.species_dict[label].conformers)
+        for conformer in range(0, num_of_conformers):  # rerun conformer job at higher level for all conformers
+            job = self.job_dict[label]['conformers'][conformer]
+
+            if job.software == 'gaussian':
```
alongd: I think these conditions should be relocated to before the `for` loop

arc/scheduler.py
```diff
@@ -1287,6 +1293,59 @@ def determine_most_stable_conformer(self, label):
...
+                if 'conformer ' + method in job.ess_trsh_methods:
+                    continue
+                job.ess_trsh_methods.append('conformer ' + method)
+                level_of_theory = method
+
+                # initial xyz before trouble shooting
+                xyz = self.species_dict[label].conformers_before_opt[conformer]
+
+                if not msg_logged:
+                    logger.info('Troubleshooting conformer job in {software} using {level}'
```
alongd: the logging message could be relocated to before this loop as well

oscarwumit: I have changed the code accordingly. Please take a look. @alongd

alongd:
@oscarwumit, could you please rebase?

oscarwumit:
I just rebased. Thanks. @alongd

PR #170, opened 2019-08-09
BugFix: Create geometry directory before copying files

oscarwumit:
ARC will create geometry directory in the local os.path if the directory
isn't exist. This resolves a copy error occured when running yml input
file

### DIFF
--- a/arc/processor.py
+++ b/arc/processor.py
@@ -415,4 +415,6 @@ class Processor(object):
         """
         calc_path = os.path.join(self.output[label]['freq'])
         output_path = os.path.join(self.project_directory, 'output', 'rxns', label, 'geometry', 'frequency.out')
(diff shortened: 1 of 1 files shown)


arc/processor.py
```diff
@@ -415,4 +415,6 @@ def copy_freq_output_for_ts(self, label):
         """
         calc_path = os.path.join(self.output[label]['freq'])
         output_path = os.path.join(self.project_directory, 'output', 'rxns', label, 'geometry', 'frequency.out')
+        if not os.path.exists(os.path.dirname(output_path)):
+            os.makedirs(output_path)
```
alongd: Shouldn't this be
`os.makedirs(os.path.dirname(output_path))`?

oscarwumit: Good catch. I fixed that. Thanks.

PR #178, opened 2019-08-29
Modify the check_scan_job module in the scheduler.py

xiaoruiDong:
Major changes:
(1) Since the `if job.job_status[1] != 'done'` is checked at the beginning, the `invalidation_reason = crash` will never  triggered. So I depreciate the relative codes
(2) The invalidation of a rotor with a barrier larger than 40 kJ should be triggered after the 'lowest-conformer' check.
(3) The previous code/logic structure is complicated. The workflow is simplified:
```
loop: find corresponding rotor
         if found:
             sequential checks (`break` right after any check is failed)
         else not found:
             raise error
assign invalidation_reason, scan_path if not trsh
plot scan curve if available
```
(4) Scan curve plot also for invalid rotors (distinguished by name)
(5) Scan curve is now dot-line curve instead of solely dots
(6) When the conformer with lower energy is found, directly rotate to the lowest conformer instead of the nearest minima.

-----------------------------------------------------------------------------------
Other thoughts:
(1) Tighten the initial-final difference threshold. As long as, there is no change in other dihedrals. The difference ~ 2 kJ (though not large) is usually caused by a change in dihedral R-O-H between initial and final conformers. 
(2) Tighten the lowest conformer threshold. I find some cases failing to find the lowest, especially when the barrier is not high. The current threshold is to avoid false positives in the check, I guess. However, it is usually not that big (from some of my methyl group's example, ~2% difference). To avoid false positives, it can be checked with symmetries.
(3) I am developing a debugging tool now. The problem is how to implement it. Let me give an example, here are three cases (1) ESS failed because of (bond breaking) (2) ESS converged but has a detectable problem (large jump) (3) ESS converged and the curve looks correct by current checking. But the debug tool finds there is an obvious dihedral change, and it is confirmed by GaussView visualization. The workflow will become less clear then. It will be desirable if it can completely replace the whole check_scan_job module, but then the choice of the outlier threshold should be more tailored. I am implementing 1.5 X IQR as the threshold. Generally looks good given some example cases, but need to test its performance on a larger scale.

### DIFF
--- a/arc/plotter.py
+++ b/arc/plotter.py
@@ -9,7 +9,6 @@ from __future__ import (absolute_import, division, print_function, unicode_liter
 import numpy as np
 import os
 import shutil
-import math
 import matplotlib.pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
 from matplotlib.backends.backend_pdf import PdfPages
@@ -174,25 +173,25 @@ def plot_rotor_scan(angle, v_list, path=None, pivots=None, comment=''):
     """
     plots a 1D rotor PES for v_list vs. angle.
     """
-    angle = angle * 180 / math.pi  # convert radians to degree
+    angle = angle * 180 / np.pi  # convert radians to degree
     v_list = np.array(v_list, np.float64)  # in kJ/mol
+    marker_color, line_color = plt.cm.viridis([0.1, 0.9])
     plt.figure(figsize=(4, 3), dpi=120)
     plt.subplot(1, 1, 1)
-    plt.plot(angle, v_list, 'g.')
+    plt.plot(angle, v_list, '.-', markerfacecolor=marker_color,
+             markeredgecolor=marker_color, color=line_color)
     plt.xlabel('dihedral (deg)')
     plt.xlim = (0, 360)
     plt.xticks(np.arange(0, 361, step=60))
     plt.ylabel('V (kJ/mol)')
     plt.tight_layout()
     plt.show()
+
     if path is not None and pivots is not None:
         if not os.path.exists(path):
             os.makedirs(path)
-        fig_path = os.path.join(path, '{0}.png'.format(pivots))
-        plt.savefig(fig_path, dpi=120, facecolor='w', edgecolor='w', orientation='portrait', papertype=None,
-                    format=str('png'), transparent=False, bbox_inches=None, pad_inches=0.1, frameon=False,
-                    metadata=None)
         if comment:
+            fig_name = '{0}-invalid.png'.format(pivots)
             txt_path = os.path.join(path, 'rotor comments.txt')
             if os.path.isfile(txt_path):
                 with open(txt_path, 'a') as f:
@@ -200,6 +199,12 @@ def plot_rotor_scan(angle, v_list, path=None, pivots=None, comment=''):
             else:
                 with open(txt_path, 'w') as f:
                     f.write(str('Pivots: {0}\nComment: {1}'.format(pivots, comment)))
+        else:
+            fig_name = '{0}.png'.format(pivots)
+        fig_path = os.path.join(path, fig_name)
+        plt.savefig(fig_path, dpi=120, facecolor='w', edgecolor='w', orientation='portrait', papertype=None,
+                    format=str('png'), transparent=False, bbox_inches=None, pad_inches=0.1, frameon=False,
+                    metadata=None)
 
 
 def log_thermo(label, path):
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -34,8 +34,7 @@ from arc.species.species import ARCSpecies, TSGuess, determine_rotor_symmetry
 from arc.species.converter import get_xyz_string, molecules_from_xyz, check_isomorphism, standardize_xyz_string
 import arc.species.conformers as conformers
 from arc.ts.atst import autotst
-from arc.settings import rotor_scan_resolution, inconsistency_ab, inconsistency_az, maximum_barrier, default_job_types,\
-    servers
+from arc.settings import inconsistency_ab, inconsistency_az, maximum_barrier, default_job_types, servers
 
 ##################################################################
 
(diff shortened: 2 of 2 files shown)


arc/plotter.py
```diff
@@ -178,17 +178,21 @@ def plot_rotor_scan(angle, v_list, path=None, pivots=None, comment=''):
     v_list = np.array(v_list, np.float64)  # in kJ/mol
     plt.figure(figsize=(4, 3), dpi=120)
     plt.subplot(1, 1, 1)
-    plt.plot(angle, v_list, 'g.')
+    plt.plot(angle, v_list, 'g.-')
```
alongd: Nice. Can you paste in the PR comments a sample image created using this modification?

arc/plotter.py
```diff
@@ -178,17 +178,21 @@ def plot_rotor_scan(angle, v_list, path=None, pivots=None, comment=''):
...
     plt.xlabel('dihedral (deg)')
     plt.xlim = (0, 360)
     plt.xticks(np.arange(0, 361, step=60))
     plt.ylabel('V (kJ/mol)')
     plt.tight_layout()
     plt.show()
+    if commment:
```
alongd: We can remove this block under the `if path is not None and pivots is not None:` since it's only being used if that statement is True

arc/scheduler.py
```diff
@@ -1746,144 +1745,138 @@ def check_scan_job(self, label, job):
...
+                    message = 'Species {label} is not oriented correctly around pivots {pivots},' \
+                              ' searching for a better conformation...'.format(label=label,
+                                                                               pivots=job.pivots)
+                    logger.info(message)
+                    # Find the rotation dihedral in degrees to the closest minimum:
+                    min_index = np.argmin(v_list)
+                    self.species_dict[label].set_dihedral(
+                        pivots=self.species_dict[label].rotors_dict[i]['pivots'],
+                        scan=self.species_dict[label].rotors_dict[i]['scan'],
+                        deg_increment=min_index*job.scan_res)
```
alongd: good idea!
(please add a space on both sides of the `*` operator)

alongd:
Thanks! @dranasinghe is also working on better rotor troubleshooting methods, perhaps the three of us can meet to decide how best to continue?

xiaoruiDong:
Yes, Duminda and I are working on that. Based on the idea from Duminda, I simplified and modified the workflow. I am trying to do a brief data analysis based on the jobs I have and try to diagnosis the behavior of the new debugger. I am free today for discussions

arc/plotter.py
xiaoruiDong: It looks like this. I personally think that adding lines can help me better observe the curve. It is helpful especially when I am debugging the rotor scan scripts.
image

arc/plotter.py
xiaoruiDong: Yes, you are right! I reframed it.

arc/scheduler.py
xiaoruiDong: Thanks! space added.

xiaoruiDong:
@alongd I update the branch. Once you think it is good, I can squash the fixup commits. Thanks!

arc/plotter.py
alongd: Great! Perhaps make the dots in a different color

arc/scheduler.py
```diff
@@ -1746,144 +1745,138 @@ def check_scan_job(self, label, job):
...
+
+                # 2. Check conformation:
+                v_diff = v_list[0] - np.min(v_list)
+                if v_diff >= 2 or v_diff > 0.5 * (max(v_list) - min(v_list)):
+                    message = 'Species {label} is not oriented correctly around pivots {pivots},' \
+                              ' searching for a better conformation...'.format(label=label,
+                                                                               pivots=job.pivots)
+                    logger.info(message)
+                    # Find the rotation dihedral in degrees to the closest minimum:
+                    min_index = np.argmin(v_list)
```
alongd: 👍

arc/scheduler.py
```diff
@@ -1746,144 +1745,138 @@ def check_scan_job(self, label, job):
...
+                    for rotor in self.species_dict[label].rotors_dict.values():
+                        rotor["scan_path"] = ''
+                        rotor["invalidation_reason"] = ''
+                        rotor["success"] = None
+                        rotor.pop('symmetry', None)
+                    self.run_opt_job(label)  # run opt on new initial_xyz with the desired dihedral
+                    break
+
+                # 3. Check the barrier height
+                if (np.max(v_list) - v_list[0]) > maximum_barrier:
```
alongd: Let's change ` v_list[0]` to `np.min(v_list)`. I know that by this point we assume there was no lower conformer, but I think it's cleaner and more comprehensive (we could have found a conformer lower by just 1 kJ/mol for which we don't re-run opt)

arc/scheduler.py
```diff
@@ -1746,144 +1745,138 @@ def check_scan_job(self, label, job):
...
+                    break
+
+                # 4. Calculate the symmetry if the scan succeed
+                if not invalidate:
+                    self.species_dict[label].rotors_dict[i]['success'] = True
+                    self.species_dict[label].rotors_dict[i]['symmetry'] = determine_rotor_symmetry(
+                        rotor_path=job.local_path_to_output_file, label=label,
+                        pivots=self.species_dict[label].rotors_dict[i]['pivots'])[0]
+                    symmetry = ' has symmetry {0}'.format(self.species_dict[label].rotors_dict[i]['symmetry'])
+                    logger.info('Rotor scan {scan} between pivots {pivots}'
```
alongd: let's combine `' has symmetry {0}'` into the `logger.info`

arc/scheduler.py
```diff
@@ -1746,144 +1745,138 @@ def check_scan_job(self, label, job):
...
+                                ' for {label}{symmetry}'.format(
+                                 scan=self.species_dict[label].rotors_dict[i]['scan'],
+                                 pivots=self.species_dict[label].rotors_dict[i]['pivots'],
+                                 label=label, symmetry=symmetry))
+                    break
         else:
             raise SchedulerError('Could not match rotor with pivots {0} in species {1}'.format(job.pivots, label))
+
+        if not trsh: # Not append info to troubleshooting scan
```
alongd: Please change to `Only continue if not troubleshooting the scan`

arc/scheduler.py
```diff
@@ -1746,144 +1745,138 @@ def check_scan_job(self, label, job):
...
+            if invalidate: # if success or reset conformer, invalidate = false
```
alongd: I didn't understand the comment

arc/scheduler.py
alongd: Double-space before inline `#` comment (here and below)

arc/scheduler.py
xiaoruiDong: Thanks! I agree that is a more comprehensive approach. BTW, in the future, what is a desirable behavior of ARC when finding a conformer with dE < 2 kJ/mol? We will switch to that conformer, right? The current status is that ARC does not tell, there is a lower conformer if the difference is within 2 kJ/mol, and I personally rerun the job.

arc/scheduler.py
xiaoruiDong: That works like a reminder. At this stage, if the rotor looks good, `invalidate = False`; if a lower conformer is found, then `invalidate = False`. Otherwise, `invalidate = True`.   It looks trivial, so I removed it in the new PR.

xiaoruiDong:
Thanks @alongd for review the PR! I made changes accordingly. Since most comments are about rephrasing comments, I squashed all the fixup commits together.

alongd:
I'm sure you have many exotic rotor scans by now, let's be sure to save theme somewhere to make unit tests out of them when we transition all troubleshooting methods into a separate module.

PR #184, opened 2019-09-06
BugFix: Scale job types fix

alongd:
Initialize job_types in scale.py

Previously scale.py used a hard-coded job_types dictionary, which isn't future compatible once new job types are added.
This was now resolved by relocating `initialize_job_types` from main to common (to avoid circular imports, and initializing job_types in scale. A test was also added.

### DIFF
--- a/arc/commonTest.py
+++ b/arc/commonTest.py
@@ -119,6 +119,31 @@ class TestARC(unittest.TestCase):
         min_lst = common.min_list(lst)
         self.assertEqual(min_lst, -79)
 
+
+    def test_initialize_job_types(self):
+        """Test the initialize_job_types() function"""
+        job_types_0 = {'conformers': False, 'opt': True, 'fine': True, 'freq': True, 'sp': False, '1d_rotors': False}
+        job_types_0_expected = {'1d_rotors': False, 'bde': False, 'conformers': False, 'fine': True,
+                                'freq': True, 'onedmin': False, 'opt': True, 'orbitals': False, 'sp': False}
+        job_types_0_initialized = common.initialize_job_types(job_types_0)
+        self.assertEqual(job_types_0_expected, job_types_0_initialized)
+
+        job_types_1 = {}
+        job_types_1_expected = {'1d_rotors': True, 'bde': False, 'conformers': True, 'fine': True, 'freq': True,
+                                'onedmin': False, 'opt': True, 'orbitals': False, 'sp': True}
+        job_types_1_initialized = common.initialize_job_types(job_types_1)
+        self.assertEqual(job_types_1_expected, job_types_1_initialized)
+
+        job_types_2 = {'bde': True}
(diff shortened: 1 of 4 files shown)


arc/utils/scale.py
```diff
@@ -72,8 +72,9 @@ def determine_scaling_factors(levels_of_theory, ess_settings=None, init_log=True
     logger.info(HEADER)
     logger.info('\n\nstarting ARC...\n')
 
-    job_types = {'conformers': False, 'opt': True, 'fine': True, 'freq': True, 'sp': False, '1d_rotors': False,
-                 'orbitals': False, 'onedmin': False}  # only run opt (fine) and freq
+    # only run opt (fine) and freq
+    job_types = {'conformers': False, 'opt': True, 'fine': True, 'freq': True, 'sp': False, '1d_rotors': False}
```
oscarwumit: Minor comment: (1) since 'opt', 'fine', and 'freq' defaults to be `True`, maybe we do not need to explicitly specify them. (2) However, since we only need to run `opt`, `fine` and `freq` here. If we assume the user might change the ARC default settings on job types (e.g., some user set 'bde' = `True` by default), then we should probably explicitly specify all job types to ensure correct calculation being spawned here.

i.e., job_types = {'conformers': False, 'opt': True, 'fine': True, 'freq': True, 'sp': False, '1d_rotors': False, 'onedmin': False, 'orbitals': False, 'bde': False}

alongd: You are right

arc/commonTest.py
```diff
@@ -119,6 +119,31 @@ def test_min_list(self):
...
 
+
+    def test_initialize_job_types(self):
+        """Test the initialize_job_types() function"""
+        job_types_0 = {'conformers': False, 'opt': True, 'fine': True, 'freq': True, 'sp': False, '1d_rotors': False}
+        job_types_0_expected = {'1d_rotors': False, 'bde': False, 'conformers': False, 'fine': True,
+                                'freq': True, 'onedmin': False, 'opt': True, 'orbitals': False, 'sp': False}
+        job_types_0_initialized = common.initialize_job_types(job_types_0)
+        self.assertEqual(job_types_0_expected, job_types_0_initialized)
+
```
oscarwumit: Minor comment: this test assumes the user does not modify default job types in ARC. Should we document somewhere that if the user would like to modify default job types, be sure to change the test as well? Or, we can define our default job types in this test, and check first if the default jobs in settings.py are in agreement with what we think.

oscarwumit:
Thanks for the bug fix. The code looks good. I leave some minor comments. Please take a look and let me know if you have any questions.

arc/commonTest.py
alongd: You're right again, but currently many other tests also fail once users make changes to the settings (e.g., servers), and we recommend in the documentation to only run tests after stashing your changes (developers should of course recognize tests that faile due to trivial changes, and modify them accordingly). So I think it's OK to leave this as is.

alongd:
Thanks for the quick review! I added a fixup, let me know whether to squash

PR #220, opened 2019-10-08
BugFix: read model_chemistry from restart

oscarwumit:
Correct a bug that over-write model chemistry when reading from the restart file.


alongd:
To document this change: Previously `model_chemistry` would only be attempted to be read from an input/restart file if `use_bde` was set to True.
This is problematic for several reasons:
A. No model chemistry is used when the user only cares about BDE's (and has no BACs), and technically need AECs for some level Arkane supports.
B. If `use_bac` is set to true but `model_chemistry` isn't given in the YAML file, ARC will attempt to read it and crush.
C. This causes a different behaviour between running ARC through the API and through an input/restart file.

PR #223, opened 2019-10-14
Minor: Change YAML `FullLoader` to `Loader`

oscarwumit:
In PyYaml version 5.x above, the `FullLoader` is not backward compatible with previous PyYaml versions. This results in ARC crash in some restart jobs. The `Loader` function, instead, is backward compatible.

-- legacy comments
PyYaml version 3.12 works well with the current ARC codebase. New versions of PyYaml leads to restart issues.

When I restart an ARC job, it crashed because PyYaml complains that 
"yaml.constructor.ConstructorError: while constructing a Python instance
expected a class, but found <class 'builtin_function_or_method'>"

After traceback the code, I find PyYaml crashed because of the following code in ARC 
yaml.load(stream=f, Loader=yaml.FullLoader)

This seems to be a common PyYaml issue, especially for new versions: 
https://github.com/facebookresearch/Detectron/issues/840

Take advice online, I changed the PyYaml version to 3.12. Because in this version, FullLoader is not available, I changed it to Loader. Now we have in ARC 
pyyaml = 3.12
yaml.load(stream=f, Loader=yaml.Loader)

After these changes, my restart job works fine :-)

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -264,7 +264,7 @@ def read_yaml_file(path):
     if not os.path.isfile(path):
         raise InputError('Could not find the YAML file {0}'.format(path))
     with open(path, 'r') as f:
-        content = yaml.load(stream=f, Loader=yaml.FullLoader)
+        content = yaml.load(stream=f, Loader=yaml.Loader)
     return content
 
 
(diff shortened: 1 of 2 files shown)


alongd:
@oscarwumit, could you upload a minimal YAML file that crashes on master? We should make a test out of it on this branch

alongd:
I prefer to keep the version 5.x, if possible, see https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation
@oscarwumit, do you still have a copy of the restart file that crushed?

PR #224, opened 2019-10-16
BugFix: Re-run job if FileNotFoundError

oscarwumit:
Avoid ARC crash and re-run the job one more time if QM output files do not exist in the path.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -721,15 +721,25 @@ class Scheduler(object):
             if job.job_type not in ['orbitals']:
                 logger.warning('Tried to determine status of job {0}, but it seems like the job never ran.'
                                ' Re-running job.'.format(job.job_name))
-                self.run_job(label=label, xyz=job.xyz, level_of_theory=job.level_of_theory, job_type=job.job_type,
-                             fine=job.fine, software=job.software, shift=job.shift, trsh=job.trsh, memory=job.memory_gb,
-                             conformer=job.conformer, ess_trsh_methods=job.ess_trsh_methods, scan=job.scan,
-                             pivots=job.pivots, occ=job.occ, scan_trsh=job.scan_trsh, scan_res=job.scan_res,
-                             max_job_time=job.max_job_time, confs=job.conformers, radius=job.radius,
-                             directed_scan_type=job.directed_scan_type, directed_scans=job.directed_scans,
-                             directed_dihedrals=job.directed_dihedrals, rotor_index=job.rotor_index)
+                self._run_a_job(job=job, label=label)
             if job_name in self.running_jobs[label]:
                 self.running_jobs[label].pop(self.running_jobs[label].index(job_name))
+
+        if not os.path.exists(job.local_path_to_output_file):
+            if 'restart_due_to_file_not_found' in job.ess_trsh_methods:
+                job.job_status[0] = 'errored'
+                job.job_status[1] = 'errored'
+                logger.warning('Job {0} errored because for the second time ARC did not find the output file path {1}.'
+                               .format(job.job_name, job.local_path_to_output_file))
+            elif job.job_type not in ['orbitals']:
+                job.ess_trsh_methods.append('restart_due_to_file_not_found')
+                logger.warning('Did not find the output file of job {0} with path {1}. Maybe the job never ran.'
+                               ' Re-running job.'.format(job.job_name, job.local_path_to_output_file))
+                self._run_a_job(job=job, label=label)
+            if job_name in self.running_jobs[label]:
+                self.running_jobs[label].pop(self.running_jobs[label].index(job_name))
+            return False
+
         if job.job_status[0] != 'running' and job.job_status[1]['status'] != 'running':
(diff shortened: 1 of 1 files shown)


arc/scheduler.py
```diff
@@ -730,6 +730,27 @@ def end_job(self, job, label, job_name):
                              directed_dihedrals=job.directed_dihedrals, rotor_index=job.rotor_index)
             if job_name in self.running_jobs[label]:
                 self.running_jobs[label].pop(self.running_jobs[label].index(job_name))
+
```
oscarwumit: Do we need a function return at line 733 after except block? Without a return statement here, what will the function return after except block if the condition on line 754 is false? I think the function might return None? @alongd

arc/scheduler.py
```diff
@@ -730,6 +730,27 @@ def end_job(self, job, label, job_name):
...
+                self.run_job(label=label, xyz=job.xyz, level_of_theory=job.level_of_theory, job_type=job.job_type,
+                             fine=job.fine, software=job.software, shift=job.shift, trsh=job.trsh, memory=job.memory_gb,
+                             conformer=job.conformer, ess_trsh_methods=job.ess_trsh_methods, scan=job.scan,
+                             pivots=job.pivots, occ=job.occ, scan_trsh=job.scan_trsh, scan_res=job.scan_res,
+                             max_job_time=job.max_job_time, confs=job.conformers, radius=job.radius,
+                             directed_scan_type=job.directed_scan_type, directed_scans=job.directed_scans,
+                             directed_dihedrals=job.directed_dihedrals, rotor_index=job.rotor_index)
+            if job_name in self.running_jobs[label]:
+                self.running_jobs[label].pop(self.running_jobs[label].index(job_name))
+            return False
```
oscarwumit: Code will crash if I do not return here. @alongd

oscarwumit:
I tested the code by restarting a previously crashed job. The restarted job finished without error.

arc/scheduler.py
```diff
@@ -730,6 +730,27 @@ def end_job(self, job, label, job_name):
                              directed_dihedrals=job.directed_dihedrals, rotor_index=job.rotor_index)
             if job_name in self.running_jobs[label]:
                 self.running_jobs[label].pop(self.running_jobs[label].index(job_name))
+
+        if not os.path.exists(job.local_path_to_output_file):
+            if 'restart_due_to_file_not_found' in job.ess_trsh_methods:
+                job.job_status[0] = 'errored'
+                job.job_status[1] = 'errored'
+
+            if job.job_type not in ['orbitals']:
```
alongd: I think this should be an `elif`

arc/scheduler.py
```diff
@@ -730,6 +730,27 @@ def end_job(self, job, label, job_name):
...
+                job.ess_trsh_methods.append('restart_due_to_file_not_found')
+                logger.warning('Did not find the output file of job {0} with path {1}. Maybe the job never ran.'
+                               ' Re-running job.'.format(job.job_name, job.local_path_to_output_file))
+                self.run_job(label=label, xyz=job.xyz, level_of_theory=job.level_of_theory, job_type=job.job_type,
```
alongd: Since this is a repetition of the above block, I suggest having this block appear just once (easier to maintain). Maybe add a flag to re-run the job, set it to true in these two instances, and execute this block depending on its value?

arc/scheduler.py
alongd: OK

oscarwumit:
Thanks for the feedback, I made the changes as you suggested. Please take a look. @alongd

arc/scheduler.py
```diff
@@ -721,15 +721,25 @@ def end_job(self, job, label, job_name):
...
             if job_name in self.running_jobs[label]:
                 self.running_jobs[label].pop(self.running_jobs[label].index(job_name))
+
+        if not os.path.exists(job.local_path_to_output_file):
+            if 'restart_due_to_file_not_found' in job.ess_trsh_methods:
+                job.job_status[0] = 'errored'
+                job.job_status[1] = 'errored'
+                logger.warning('Job {0} errored because for the second time ARC did not find the output file path {1}.'
+                               .format(job.job_name, job.local_path_to_output_file))
+            elif job.job_type not in ['orbitals']:
```
alongd: why do we exclude orbitals?

oscarwumit: I excluded orbitals here because I saw In line 721 we excluded orbitals. Do you remember what was the reason we excluded orbitals in line 721?

arc/scheduler.py
```diff
@@ -721,15 +721,25 @@ def end_job(self, job, label, job_name):
             if job.job_type not in ['orbitals']:
```
oscarwumit: Why we exclude orbitals here? Do we need to do the same for line 734? @alongd

arc/scheduler.py
alongd: I see. I guess I anticipated issues with orbital jobs (they should be improved). OK, let's keep it like this.

PR #226, opened 2019-10-19
BugFix: Read specific_job_type from yaml file

oscarwumit:
Fixed a bug that ARC did not read specific_job_type attribute from input file. 

Added a test to maintest.py to check if specific_job_type is given in the yaml file, whether ARC process this attribute to initialize job correctly.

Improved logging to indicate if user specified job_type, specific_job_type, or did not specify any.

Lesson learned: when adding new attribute to arc object, do not forget to update from_dict and as
_dict.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -415,6 +415,7 @@ def initialize_job_types(job_types, specific_job_type=''):
     """
 
     if specific_job_type:
+        logger.info(f'Specific_job_type {specific_job_type} is given by user.')
         if job_types:
             logger.warning('Both job_types and specific_job_type are given, ARC will only use specific_job_type to '
                            'populate the job_types dictionary.')
@@ -422,7 +423,7 @@ def initialize_job_types(job_types, specific_job_type=''):
         try:
             job_types[specific_job_type] = True
         except KeyError:
-            raise InputError('Specified job type "{0}" is not supported'.format(specific_job_type))
+            raise InputError(f'Specified job type {specific_job_type} is not supported.')
 
     if specific_job_type == 'bde':
         bde_default = {'opt': True, 'fine_grid': True, 'freq': True, 'sp': True}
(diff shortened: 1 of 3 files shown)


alongd:
Thanks! take a look at the tests?

arc/common.py
```diff
@@ -415,6 +415,7 @@ def initialize_job_types(job_types, specific_job_type=''):
     """
 
     if specific_job_type:
+        logger.info('Specific_job_type is given by user.')
```
alongd: Perhaps also report which specific job type was asked for, so it's logged?

arc/common.py
```diff
@@ -432,6 +433,9 @@ def initialize_job_types(job_types, specific_job_type=''):
     defaults_to_false = ['onedmin', 'orbitals', 'bde']
     if job_types is None:
         job_types = default_job_types
+        logger.info('It seems that user did NOT specify job_types, ARC using default_job_types as in settings.py')
```
alongd: Please rephrase to:
"job types were not specified, using default values from ARC's settings"

arc/common.py
```diff
@@ -432,6 +433,9 @@ def initialize_job_types(job_types, specific_job_type=''):
...
+    else:
+        logger.info('Job_type is given by user.')
```
alongd: please rephrase to:
`logger.info(f'the following job types were specified: {job_types}.')`

oscarwumit:
Thanks for the comment. I've made the changes you suggested. Please take a look. @alongd

arc/common.py
```diff
@@ -432,6 +433,9 @@ def initialize_job_types(job_types, specific_job_type=''):
     defaults_to_false = ['onedmin', 'orbitals', 'bde']
     if job_types is None:
         job_types = default_job_types
+        logger.warning('Job types were not specified, using default values from ARC settings')
```
alongd: This shouldn't be `warning`, I think `info` is more suited

oscarwumit: fixed

oscarwumit:
I changed the logging msg to info as you suggested and rebased. @alongd

PR #229, opened 2019-10-23
Feature: Try server command again before crash.

oscarwumit:
Previously ARC crashes if a command returns false (error-ed) on the server. Sometimes the error might come from a temporary server problem. ARC now rerun the command with several attempts before it raises an exception. This PR will make ARC more robust in handling command error on servers.

### DIFF
--- a/arc/job/local.py
+++ b/arc/job/local.py
@@ -12,6 +12,7 @@ import os
 import re
 import shutil
 import subprocess
+import time
 
 from arc.common import get_logger
 from arc.exceptions import SettingsError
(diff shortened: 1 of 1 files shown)


arc/job/local.py
```diff
@@ -31,27 +32,42 @@ def execute_command(command, shell=True):
...
+            logger.error('And got the following status (cmd, message, output, returncode, cms:')
+            logger.error(e.cmd)
+            logger.info('\n')
+            logger.error(e)
+            logger.info('\n')
+            logger.error(e.output)
+            logger.info('\n')
+            logger.error(e.returncode)
+            logger.info('\n')
+            logger.error(e.cmd)
```
alongd: remove this line, it's redundant

arc/job/local.py
```diff
@@ -31,27 +32,42 @@ def execute_command(command, shell=True):
...
+    i, max_times_to_try = 1, 30
+    success = False
+    sleep_time = 60  # seconds
+    while i < max_times_to_try:
+        try:
+            stdout = subprocess.check_output(command, shell=shell)
+        except subprocess.CalledProcessError as e:
+            logger.error('The server command is erroneous.')
+            logger.error(f'Tried to submit the following command:\n{command}')
+            logger.error('And got the following status (cmd, message, output, returncode, cms:')
```
alongd: change to `'And got the following status (cmd, message, output, return code):'`

arc/job/local.py
```diff
@@ -31,27 +32,42 @@ def execute_command(command, shell=True):
...
+                         ' please check if this is a server issue by executing the command manually on server.')
+            logger.info('ZZZZZ..... ZZZZZ.....')
+            time.sleep(sleep_time * i)  # in seconds
+            i += 1
+        else:
+            success = True
+            break
+    if not success:
+        raise SettingsError(f'The command "{command}" is erroneous, got: \n{e}'
+                            '\nThis maybe either a server issue or the command is wrong.'
```
alongd: when I do `print f` I try to add `f` before each line. Here it'll work because we only have arguments in the first line, but it's good practice. Could you add it?

oscarwumit:
Thanks for the comment, I've updated the PR with changes you suggested. Please take a look. @alongd

arc/job/local.py
```diff
@@ -31,27 +32,40 @@ def execute_command(command, shell=True):
...
+    i, max_times_to_try = 1, 30
+    success = False
+    sleep_time = 60  # seconds
+    while i < max_times_to_try:
+        try:
+            stdout = subprocess.check_output(command, shell=shell)
+        except subprocess.CalledProcessError as e:
+            logger.error('The server command is erroneous.')
+            logger.error(f'Tried to submit the following command:\n{command}')
+            logger.error('And got the following status (cmd, message, output, return code')
```
alongd: Could you add a closing parenthesis in the text: `(cmd, message, output, return code)`?
(it was also missing in the original version)
thanks!

arc/job/local.py
```diff
@@ -31,27 +32,40 @@ def execute_command(command, shell=True):
...
+            logger.error(e.cmd)
+            logger.info('\n')
+            logger.error(e)
+            logger.info('\n')
+            logger.error(e.output)
+            logger.info('\n')
+            logger.error(e.returncode)
+            logger.info('\n')
+            logger.error(f'ARC is sleeping for {sleep_time * i} seconds before re-trying,'
+                         ' please check if this is a server issue by executing the command manually on server.')
```
alongd: please add `f` here as well

oscarwumit:
changes made. thanks

Issue #171, opened 2019-08-09
Rotor detection fails in some special cases


xiaoruiDong:
Rotor detector fails in some special cases
(1) When one end of the dihedral has triple bonds, e.g., H-C#C-O-H. It will take C#C-O-H as a rotor.
(2) From the original structure, it seems existing rotors. But when looking at other resonance structures, then some of the rotors are fake, e.g., O=[C]C=O. It looks like O=[C]-C=O is a rotor, but it has a resonance structure: O=C=C-[O], which indicates that it is not really a rotor (The actual angle O-C-C is 180 degree in the optimized structure).

Suggestion:
These two cases fail when they have bond angle = 180. We can check triple bond or resonance structure, but there may be other cases impossible to identify beforehand. So we can check `if any bond angle == 180` in the dihedral about to be scanned, if so, ignore the rotor.

alongd:
See `check_special_non_rotor_cases` [here](https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/species/conformers.py#L1290)
Perhaps you can come up with suggestions for improvement.
Note that we don't really check the angle, since the unoptimized 3D structure has wrong angles

xiaoruiDong:
Thanks for letting me know the function. What I mean is actually check the angle before submitting the scan job. As we only submit scan job after optimization, at that time we should have an accurate bond angle:
Start ARC -> find rotors -> conformers -> opt -> check the angle -> scan

alongd:
Rotors are determined before running opt. We could reconsider that if we want. More importantly, see why check_special_non_rotor_cases doesn't address the original issue and whether it could be updated?

Regarding the resonance case, I think it's OK to declare that bond as a potential rotor (the resonance might be weak), worst case we'll find a hight barrier and invalidate the rotor.

alongd:
@xiaoruiDong, what's the status of this issue?

xiaoruiDong:
This issue relates to a special kind of molecules: species have a triple bond but the bond angle is not 180 degrees. All the cases I found relates to the open-shell species (something like [C]#CC). ARC will calculate the rotor scan. And some of them do have a slightly large barrier (larger than the free rotor threshold we use). The thing is this kind of rotor scans have a high chance of failure, and the failures happen in a similar pattern: the bond angle of triple bonds change to 180 degrees and the gaussian job crashed. I haven't figured out a solution to this yet.

whgreen:
[...]

PR #232, opened 2019-10-25
Don't run torsion scans if the dihedral has a 180 degrees angle at one of its sides

alongd:
Improved (generalized) the `check_special_non_rotor_cases` function for screening torsions.
Also, if after optimization the dihedral is (very close to) 180 degrees, don't run this scan (as suggested by @xiaoruiDong). From my experience, non-rotors like this have dihedrals of 180.0 degrees in DFT opt, here we're using a tolerance of 0.1 degrees.

fixes https://github.com/ReactionMechanismGenerator/ARC/issues/171

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -26,12 +26,13 @@ from arc.job.local import check_running_jobs_ids
 from arc.job.ssh import SSHClient
 from arc.job.trsh import trsh_negative_freq, trsh_scan_job, trsh_ess_job, trsh_conformer_isomorphism, scan_quality_check
 from arc.species.species import ARCSpecies, TSGuess, determine_rotor_symmetry
-from arc.species.converter import molecules_from_xyz, check_isomorphism, standardize_xyz_string, xyz_from_data, \
-    str_to_xyz, xyz_to_str
+from arc.species.converter import molecules_from_xyz, check_isomorphism, standardize_xyz_string, \
+    str_to_xyz, xyz_to_str, xyz_to_coords_list
 from arc.ts.atst import autotst
 from arc.settings import default_job_types, rotor_scan_resolution
 import arc.rmgdb as rmgdb
 import arc.species.conformers as conformers  # import after importing plotter to avoid circular import
+from arc.species.vectors import get_angle
 
 
 logger = get_logger()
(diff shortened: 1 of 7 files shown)


alongd:
@xiaoruiDong could you take a look at this PR?

xiaoruiDong:
@alongd, I will review it by today. Sorry for the delay.

arc/species/conformersTest.py
```diff
@@ -751,7 +762,15 @@ def test_check_special_non_rotor_cases(self):
         self.assertTrue(len(torsions1) == 1)  # expecting only the CH3 rotor
 
         torsions2 = [rotor_dict['scan'] for rotor_dict in spc2.rotors_dict.values()]
-        self.assertTrue(len(torsions2) == 1)  # expecting only the CH3 rotor
+        self.assertFalse(len(torsions2))
```
xiaoruiDong: Do you think it is better to use `assertEqual` instead of `assertFalse`?  The same for line 762, 769 and 773

arc/scheduler.py
```diff
@@ -967,8 +967,16 @@ def run_scan_jobs(self, label):
         """
         if self.job_types['rotors']:
             for i in range(self.species_dict[label].number_of_rotors):
-                scans = self.species_dict[label].rotors_dict[i]['scan']
+                scan = self.species_dict[label].rotors_dict[i]['scan']
                 pivots = self.species_dict[label].rotors_dict[i]['pivots']
+                dihedral = calculate_dihedral_angle(coords=self.species_dict[label].get_xyz(generate=False)['coords'],
+                                                    torsion=scan)
+                if abs(dihedral - 180.0) < 0.1:
```
xiaoruiDong: Do we actually want to check if a bond angle is near 180?

alongd: good catch! I'll fix it

arc/species/conformers.py
```diff
@@ -1299,28 +1299,22 @@ def check_atom_collisions(xyz):
...
-        if len(top) == 2 and mol.atoms[top[0] - 1].is_carbon() and mol.atoms[top[1] - 1].is_nitrogen() \
-                and mol.atoms[top[1] - 1].atomtype.label == 'N3t':
-            return True
-    for tp1, tp2 in [(top1, top2), (top2, top1)]:
-        # check azide group
-        if len(tp1) == 2 and mol.atoms[tp1[0] - 1].atomtype.label == 'N5tc' \
-                and mol.atoms[tp1[1] - 1].atomtype.label == 'N3t' and mol.atoms[tp2[0] - 1].atomtype.label == 'N1sc':
+        if mol.atoms[top[0] - 1].atomtype.label in ['Ct', 'N3t', 'N5tc'] \
+                and mol.atoms[top[1] - 1].atomtype.label in ['Ct', 'N3t'] and \
+                (len(top) == 2 or (len(top) == 3 and mol.atoms[top[2] - 1].is_hydrogen())):
```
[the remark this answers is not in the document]
alongd: We do `for top in [top1, top2]`, so `top` is either `top1` or `top2`

arc/species/conformersTest.py
alongd: I'll change all `assertTrue(x == 1)` to `assertEqual(x, 1)`
I think `assertFalse` is acceptable to check that something equals `0`, I'll keep that

PR #237, opened 2019-11-01
BugFix: Don't generate resonance structures for charges species via RMG

alongd:
Also consider the charge when determining the multiplicity.
Added tests, and updated the conformers iPy notebook


mbprend:
Thank you for working on this when I was there to observe. It looks like the net code coverage will be improved if this pull request is accepted.

PR #250, opened 2019-11-15
Allow max job time in hours as a float to include minutes and seconds

amarkpayne:
I was trying to submit a job to a debug queue on a server that allowed for job times of no more than 30 minutes. Since we specify the maximum job time as the number of hours as an integer, there was no way to set this 30 minute time.

I have refactored the code responsible for this so that it can handle hours as a float. 0.5 will convert to 30 minutes, 4.5 will convert to 4 hours and 30 minutes, and so on and so forth.

I have moved the code for this to a new function to reduce code complexity and allow this to be properly unit tested.

Reviewer note: I included type hints in my code, but because this doesn't match the style around it I am happy to convert this so that the type only appears in the doc string, or even get rid of it altogether.

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -66,7 +66,7 @@ class Job(object):
         server (str, optional): Server's name.
         initial_time (datetime, optional): The date-time this job was initiated.
         occ (int, optional): The number of occupied orbitals (core + val) from a molpro CCSD sp calc.
-        max_job_time (int, optional): The maximal allowed job time on the server in hours.
+        max_job_time (float, optional): The maximal allowed job time on the server in hours (can be fractional).
         scan_res (int, optional): The rotor scan resolution in degrees.
         checkfile (str, optional): The path to a previous Gaussian checkfile to be used in the current job.
         number_of_radicals (int, optional): The number of radicals (inputted by the user, ARC won't attempt to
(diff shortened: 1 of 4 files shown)


alongd:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/ARC/ARC/pullRequest?prid=4510023&bid=15197700)

arc/job/job.py
```diff
@@ -483,6 +483,31 @@ def write_completed_job_to_csv_file(self):
                    self.job_status[0], self.job_status[1]['status'], self.ess_trsh_methods, self.comments]
             writer.writerow(row)
 
+    def format_max_job_time(self, time_format: str) -> str:
```
alongd: Thanks for this contribution!
We can keep the type hints, I wonder how this will be displayed differently in the docs/API.

(minor: could you add a new line before `Args:`, and remove the new line after the `Returns:... str:...`?)

amarkpayne:
Is it just that Codacy does not like the type hints?

arc/job/job.py
amarkpayne: I think I am going to remove the type hints to play it safe. I'll also make the white space changes

alongd:
Do you mind changing some additional related docstring types?
In main.py lines 77 and 134 we need to say now `float` instead of `int`.
Same for scheduler.py lines 112, 145, and job.py lines 69 and 157.
Thanks!

amarkpayne:
I think the Travis tests have passed, but this site hasn't updated

PR #256, opened 2019-11-22
Minor: do not crash if running_jobs is empty in restart file

oscarwumit:
Fixed an issue that crashes ARC during restart.

Previously, ARC crashes if the 'running_jobs' dictionary is empty in the restart file. This crash was found by Jim. Normally 'running_jobs' should not be empty in the restart file. However, if the user would like to restart ARC by ignoring previously running jobs, they might manually delete related job labels in the restart file. The user may forget to completely delete the 'running_jobs' key in the restart file, and left it with empty value.

This PR fixed the crash by assuming if 'running_jobs' is empty or None, then all jobs should have finished.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -2545,47 +2545,54 @@ class Scheduler(object):
         Important for the restart feature so long jobs won't be ran twice.
         """
         jobs = self.restart_dict['running_jobs']
-        for spc_label in jobs.keys():
-            if spc_label not in self.running_jobs:
-                self.running_jobs[spc_label] = list()
-            for job_description in jobs[spc_label]:
-                if 'conformer' not in job_description or job_description['conformer'] < 0:
-                    self.running_jobs[spc_label].append(job_description['job_name'])
-                else:
-                    self.running_jobs[spc_label].append('conformer{0}'.format(job_description['conformer']))
-                for species in self.species_list:
-                    if species.label == spc_label:
-                        break
-                else:
-                    raise SchedulerError('Could not find species {0} in the restart file'.format(spc_label))
-                job = Job(job_dict=job_description)
-                if spc_label not in self.job_dict:
-                    self.job_dict[spc_label] = dict()
-                if job_description['job_type'] not in self.job_dict[spc_label]:
+        if not jobs:
+            del self.restart_dict['running_jobs']
+            self.running_jobs = dict()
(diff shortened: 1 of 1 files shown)


arc/scheduler.py
```diff
@@ -2535,47 +2535,54 @@ def restore_running_jobs(self):
...
-                else:
-                    raise SchedulerError('Could not find species {0} in the restart file'.format(spc_label))
-                job = Job(job_dict=job_description)
-                if spc_label not in self.job_dict:
-                    self.job_dict[spc_label] = dict()
-                if job_description['job_type'] not in self.job_dict[spc_label]:
+        if not jobs:
+            del self.restart_dict['running_jobs']
+            self.running_jobs = dict()
+            logger.info('It seems that there are no running jobs specified in the ARC restart file. Assume all jobs'
```
alongd: Add a blank space after `Assume all jobs` (so it won't print `Assume all jobshave finished.`)
Also please change `Assume` to `Assuming`.

arc/scheduler.py
```diff
@@ -2535,47 +2535,54 @@ def restore_running_jobs(self):
...
+                        'have finished.')
+        else:
+            logger.info(f'ARC restart files indicate the following jobs are still running. {jobs.keys()}')
```
alongd: Change to `f"ARC's restart file indicates that the  following"`...

oscarwumit:
Thanks for the comment. Please take a look again. @alongd

arc/scheduler.py
```diff
@@ -2545,47 +2545,54 @@ def restore_running_jobs(self):
...
-                if spc_label not in self.job_dict:
-                    self.job_dict[spc_label] = dict()
-                if job_description['job_type'] not in self.job_dict[spc_label]:
+        if not jobs:
+            del self.restart_dict['running_jobs']
+            self.running_jobs = dict()
+            logger.info('It seems that there are no running jobs specified in the ARC restart file. Assuming all jobs '
+                        'have finished.')
+        else:
+            logger.info(f"ARC's restart files indicate the following jobs are still running. {jobs.keys()}")
```
alongd: I think `job.keys()` returns an iterator, maybe we'd like to print `list(job.keys())` instead?

oscarwumit: Thanks @alongd

PR #281, opened 2020-01-15
BugFix: use copy() instead of copy(deep=True)

oscarwumit:
This PR fixed a bug related to deepcopy. The bug was found during the execution of an ARC reaction job. 

The traceback was:
"""
  File "/home/wuh104/workspace/ARC/ARC.py", line 73, in <module>
    main()
  File "/home/wuh104/workspace/ARC/ARC.py", line 67, in main
    arc_object.execute()
  File "/gpfs/workspace/users/wuh104/ARC/arc/main.py", line 531, in execute
    dont_gen_confs=self.dont_gen_confs)
  File "/gpfs/workspace/users/wuh104/ARC/arc/scheduler.py", line 246, in __init__
    display(rxn.rmg_reaction.copy(deep=True))
TypeError: copy() takes no keyword arguments
"""

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -243,7 +243,7 @@ class Scheduler(object):
                 if family_text:
                     logger.info(f'({family_text})')
                 if rxn.rmg_reaction is not None:
-                    display(rxn.rmg_reaction.copy(deep=True))
+                    display(rxn.rmg_reaction.copy())
                 rxn.determine_rxn_charge()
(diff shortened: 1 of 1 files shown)


alongd:
I think we don't need a deep copy for `display()`, perhaps delete the `deep=True` in the existing `.copy()` instead of using `deepcopy`?

oscarwumit:
@alongd that works as well. please take a look of the change.

PR #302, opened 2020-02-06
Use dispersion if given for Gaussian jobs

### LINKED ISSUE #300: Emperical dispersion does not make it to the QM input file
**Describe the bug**
Following the documentation [here](https://reactionmechanismgenerator.github.io/ARC/advanced.html#specify-job-level-of-theory) to add in the empirical dispersion model, I am finding that the dispersion keyword is not actually being added to the QM (in my case Gaussian) input file.

Looking at job.py, it appears that `self.dispersion` is being set properly, but this variable doesn't appear to be used elsewhere at first glance.

**How to reproduce**
Here is some of the code you need to reproduce this bug:

```
project = 'dispersion_test'
ess_settings = {'gaussian': ['local']}
job_types = {'rotors': False,
             'conformers': False,
             'fine': True,
             'freq': True,
             'opt': True,
             'sp': True,
             }
max_job_time = 100
model_chemistry = 'user_defined'

# Level of Theory
lot = {'method': 'b3lyp', 'basis': 'def2tzvp', 'dispersion': 'empiricaldispersion=gd3bj'}

opt_level = lot
freq_level = lot
scan_level = lot
sp_level = lot
conformer_level = lot
ts_guess_level = lot

# arc_species_list is a list of one species. smiles="O=CCl"

arc = ARC(project=project,
          arc_species_list = arc_species_list,
          ess_settings=ess_settings,
          job_types=job_types,
          max_job_time=max_job_time,
          model_chemistry=model_chemistry,
          opt_level=opt_level,
          freq_level=freq_level,
          scan_level=scan_level,
          sp_level=sp_level,
          conformer_level=conformer_level,
          ts_guess_level=ts_guess_level,
          compare_to_rmg=False,
          #calc_freq_factor=False,
          )

arc.execute()

```

**Expected behavior**
Specifying the dispersion should mean that the dispersion model is used in QM calculations.

amarkpayne:
fixes #300 

I believe this is all that is needed to fix this issue. I can add unit tests as appropriate but I first want to submit it for review so that I can be sure that this is the preferred way of fixing this issue. Using this commit I am seeing the dispersion being set in the Gaussian input file as expected.

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -638,20 +638,19 @@ class Job(object):
             else:  # gaussian, molpro
                 restricted = ''
 
-        dispersion = ''
         if self.software == 'terachem':
             # TeraChem does not accept "wb97xd3", it expects to get "wb97x" as the method and "d3" as the dispersion
             if self.method[-2:] == 'd2':
-                dispersion = 'd2'
+                self.dispersion = 'd2'
                 self.method = self.method[:-2]
             elif self.method[-2:] == 'd3':
-                dispersion = 'd3'
+                self.dispersion = 'd3'
                 self.method = self.method[:-2]
             elif self.method[-1:] == 'd':
-                dispersion = 'yes'
+                self.dispersion = 'yes'
                 self.method = self.method[:-2]
             else:
-                dispersion = 'no'
+                self.dispersion = 'no'
 
         job_type_1, job_type_2, fine = '', '', ''
 
@@ -1170,7 +1169,7 @@ end
                                                job_options_blocks=job_options_blocks,
                                                job_options_keywords=job_options_keywords,
                                                method_class=method_class, auxiliary_basis=self.auxiliary_basis_set,
-                                               shortcut_keywords=shortcut_keywords, dispersion=dispersion,
+                                               shortcut_keywords=shortcut_keywords, dispersion=self.dispersion,
                                                ) \
                     if self.input is not None else None
             except KeyError:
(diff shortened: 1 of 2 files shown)


arc/job/job.py
```diff
@@ -591,7 +591,7 @@ def write_input_file(self):
         """
         # Initialize variables
         orca_options_keywords_dict, orca_options_blocks_dict, restricted, method_class = (None for _ in range(4))
-        job_options_keywords, job_options_blocks, shortcut_keywords = '', '', ''
+        job_options_keywords, job_options_blocks, shortcut_keywords, dispersion = '', '', '', ''
```
alongd: What if we just do here `dispersion = self.dispersion`? (i.e., outside the `if self.software == 'gaussian'`)
I suggest we skip the `dispersion = ''` above
Then, it'll fix this for all ESS. `self.dispersion` is an empty string by default.

arc/job/job.py
```diff
@@ -638,7 +640,6 @@ def write_input_file(self):
             else:  # gaussian, molpro
                 restricted = ''
 
-        dispersion = ''
         if self.software == 'terachem':
```
alongd: Would you please also change the below block to `self.dispersion` instead of `dispersion`, and in line 1173 as well?

arc/job/job.py
amarkpayne: I agree, this is much better. I was wondering if there was a reason why we were using a local variable instead of `self.dispersion`. Done!

arc/job/job.py
amarkpayne: Done!

PR #304, opened 2020-02-06
Add option for running all optimizations as fine

### LINKED ISSUE #303: Add option to only perform a fine optimization
**Is your feature request related to a problem? Please describe.**
Usually, when one uses ARC, they need ARC to start from the very beginning and generate a plausible conformer for the molecule, and then use progressively higher levels of theory to optimize this geometry.

However, this is not always the case. For example, it is possible that the user has a very good geometry for the molecule already (for example, perhaps they have already performed a calculation for this species using a good DFT functional, but want to compare with another method). In this case, the user does not want to waste any computational time by starting from the very beginning.

ARC is fairly accommodating for this. For example, you can tell ARC not to perform conformer searches. One thing that ARC does not currently let you do though is bypass an initial optimization step before performing a fine optimization. *Ideally, the user should be able to tell ARC to go directly to the fine optimization in the event that they think the geometry supplied is very close.*

**Describe the solution you'd like**
I would like to add a keyword to ARC's input file called `fine_only` that if `True` will only perform a fine optimization.

**Additional context**
I am planning on implementing this myself, but I would appreciate any guidance as to which parts of ARC will likely need changing, as well as any comments about how this feature should best be implemented to best align with the vision for ARC going forward (if it is a feature that should be implemented in the first place).

amarkpayne:
fixes #303 

I found a solution that implements the option to run all optimizations as fine, and after testing it appears to work. It is sort of a hack around though, so let me know if there is a better way so that the flow of the code remains clear.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -180,6 +180,8 @@ class ARC(object):
         compute_rates (bool): Whether to compute rate coefficients for converged reactions.
         compute_transport (bool): Whether to compute transport properties for converged species.
         statmech_adapter (str): The statmech software to use.
+        fine_only (bool): If ``self.job_types['fine'] and not self.job_types['opt']`` ARC will not run optimization
+                          jobs without fine=True
     """
 
     def __init__(self, input_dict=None, project=None, arc_species_list=None, arc_rxn_list=None, level_of_theory='',
@@ -323,6 +325,12 @@ class ARC(object):
             self.ess_settings = check_ess_settings(ess_settings or global_ess_settings)
         if not self.ess_settings:
             self.determine_ess_settings()
+
+        # Determine if fine-only behavior is requested before determining chemistry for job types
+        self.fine_only = False
+        if self.job_types['fine'] and not self.job_types['opt']:
+            self.fine_only = True
+            self.job_types['opt'] = True  # Run the optimizations, self.fine_only will make sure that they are fine
 
         # execute regardless of new job or restart job
         self.determine_model_chemistry_for_job_types()  # all level of theory attributes should be dict after this call
@@ -582,7 +590,8 @@ class ARC(object):
                                    restart_dict=self.restart_dict, project_directory=self.project_directory,
                                    max_job_time=self.max_job_time, allow_nonisomorphic_2d=self.allow_nonisomorphic_2d,
                                    memory=self.memory, adaptive_levels=self.adaptive_levels,
-                                   n_confs=self.n_confs, e_confs=self.e_confs, dont_gen_confs=self.dont_gen_confs)
+                                   n_confs=self.n_confs, e_confs=self.e_confs, dont_gen_confs=self.dont_gen_confs,
+                                   fine_only=self.fine_only)
 
         save_yaml_file(path=os.path.join(self.project_directory, 'output', 'status.yml'), content=self.scheduler.output)
 
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -134,6 +134,7 @@ class Scheduler(object):
                         - 'method' (optional values: 'pcm' (default), 'cpcm', 'dipole', 'ipcm', 'scipcm')
                         -  'solvent' (values are strings of "known" solvents, see https://gaussian.com/scrf/,
                                       default is "water")
+        fine_only (bool): If ``True`` ARC will not run optimization jobs without fine=True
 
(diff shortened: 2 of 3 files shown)


alongd:
I see. Let me know what you think about the comment in #303? In general I think we should avoid accumulating case-specific arguments in main.

alongd:


```diff

Complexity increasing per file
==============================
- arc/main.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=4961352&bid=16393288)

amarkpayne:
I implemented your idea that if `fine == True` but `opt == False` that it would run optimizations using a fine level only. I wanted to implement this entirely in scheduler.py, but I ran into some problems because if `opt == False` functions in main.py such as `determine_model_chemistry_for_job_types` do not function properly. Instead of trying to go into these functions and add/modify if statements to catch these cases, I instead thought it would be more simplistic to set a `fine_only` attribute on the main ARC object, and if true, set `self.job_types['opt'] = True` so that the rest of the code behaves as expected.

This means that the ARC object gained an extra attribute (which is unfortunate) but it does avoid adding on syntax to the ARC input file as requested (i.e. fine_only is an attribute but not an argument of `__init__`). Let me know if this is acceptable to you. If not then we will have to do some searching to see where we need to add/modify statements that check both opt == True and fine == True.

Finally, I added some documentation for this feature, along with fixing a few typos that I found.

alongd:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=4961352&bid=16393288)

amarkpayne:
@alongd Are there any other changes you would like me to make?

alongd:
Do you think it'll work if instead of adding a new attribute ('fine_only') to Scheduler, we do the following?
```python
def run_opt_job(self, label):
    if 'opt' not in self.job_dict[label]:  # Check whether or not opt jobs have been spawned yet
        # we're spawning the first opt job for this species
        self.job_dict[label]['opt'] = dict()
    if self.species_dict[label].initial_xyz is None:
        raise SpeciesError(f'Cannot execute opt job for {label} without xyz (got None for Species.initial_xyz)')
    fine = self.job_types['fine'] and not self.job_types['opt']
    self.run_job(label=label, xyz=self.species_dict[label].initial_xyz, level_of_theory=self.opt_level,
                 job_type='opt', fine=fine)
```

amarkpayne:
> Do you think it'll work if instead of adding a new attribute (`fine_only`) to Scheduler, we do the following?
> 
> ```python
> def run_opt_job(self, label):
>     if 'opt' not in self.job_dict[label]:  # Check whether or not opt jobs have been spawned yet
>         # we're spawning the first opt job for this species
>         self.job_dict[label]['opt'] = dict()
>     if self.species_dict[label].initial_xyz is None:
>         raise SpeciesError(f'Cannot execute opt job for {label} without xyz (got None for Species.initial_xyz)')
>     fine = self.job_types['fine'] and not self.job_types['opt']
>     self.run_job(label=label, xyz=self.species_dict[label].initial_xyz, level_of_theory=self.opt_level,
>                  job_type='opt', fine=fine)
> ```

Although I agree that I would like to get rid of the flag, I am unsure if this solution will work because the function `run_opt_job` is usually only called if `self.job_types['opt']` is True. But if `self.job_types['opt']` is True how do we determine if we should run only fine optimizations?

There may be a solution here, but the limiting factor here is that a lot of code that needs to be run for ANY type of optimization job to work requires that `self.job_types['opt']` be True as far back as `arc/main.py` even before the scheduler is called. 

A possible solution would be to make ARC additionally check for `self.job_types['fine']` whenever it needs to determine if any type of opt job needs to be run, but then we will have to be careful to always code ARC this way going forward (i.e. this could be an easy vector for bugs to be introduced that could unnoticed for awhile unless we had a good functional test for fine optimizations).

An interesting solution would be to have these job types be classes in ARC, and have a fine job subclass an opt job class, but that would involve a lot of re-writing of course.

Perhaps we can discuss this more offline.

alongd:
OK. Could you rebase?

amarkpayne:
Done! Let me know if there is anything else I need to do.

PR #308, opened 2020-02-18
Add method for converting xyz coordinates from rmgpy conformer objects

amarkpayne:
Potentially useful when working with ArkaneSpecies objects, which stores xyz coordinates inside of rmgpy.statmech.Conformer objects

### DIFF
--- a/arc/species/converter.py
+++ b/arc/species/converter.py
@@ -15,7 +15,9 @@ from rdkit.Chem import rdMolTransforms as rdMT
 from arkane.common import symbol_by_number, mass_by_symbol, get_element_mass
 from rmgpy.exceptions import AtomTypeError
 from rmgpy.molecule.molecule import Atom, Bond, Molecule
+from rmgpy.quantity import ArrayQuantity
 from rmgpy.species import Species
+from rmgpy.statmech import Conformer
 
 from arc.common import get_logger, is_str_float, get_atom_radius
 from arc.exceptions import ConverterError, InputError, SanitizationError, SpeciesError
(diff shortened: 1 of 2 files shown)


arc/species/converterTest.py
```diff
@@ -589,6 +625,20 @@ def test_xyz_from_data(self):
         self.assertIsInstance(xyz_dict2['coords'], tuple)
         self.assertIsInstance(xyz_dict2['coords'][0], tuple)
 
+    def test_conformer_to_xyz_dict(self):
+        """Test the rmg_conformer_to_xyz function"""
+        xyz_dict = converter.rmg_conformer_to_xyz(self.conformer_12)
+        self.assertEqual(xyz_dict['symbols'], self.xyz_dict_12['symbols'])
+        self.assertEqual(xyz_dict['isotopes'], self.xyz_dict_12['isotopes'])
+        self.assertTrue(np.allclose(np.array(xyz_dict['coords']), np.array(self.xyz_dict_12['coords'])))
```
alongd: You can use `self.assertTrue(arc.common.almost_equal_coords_lists(xyz_dict, self.xyz_dict_12))` here, it checks the symbols and that the coords are alsoms equal (does not check isotopes). It's also OK to leave this as is.

arc/species/converter.py
```diff
@@ -318,6 +320,53 @@ def xyz_from_data(coords, numbers=None, symbols=None, isotopes=None):
...
+
+        We also assume that we can get the isotope number by rounding the mass
+
+    Args:
+        conformer (Conformer): An rmgpy.statmech.Conformer object containing the desired xyz coordinates
+
+    Returns:
+        dict: The ARC xyz format
+    """
+    symbols = tuple(symbol_by_number[n] for n in conformer.number.value)
```
alongd: Could you add an `ininstance` check for `conformer` and raise a `TypeError`?

arc/species/converter.py
```diff
@@ -318,6 +320,53 @@ def xyz_from_data(coords, numbers=None, symbols=None, isotopes=None):
...
+    Notes:
+        Only the xyz information will be supplied to the newly created Conformer object
+
+    Args:
+        xyz_dict (dict): The ARC dict xyz style coordinates
+
+    Returns:
+        Conformer: An rmgpy.statmech.Conformer object containing the desired xyz coordinates
+    """
+    mass_and_number = (get_element_mass(*args) for args in zip(xyz_dict['symbols'], xyz_dict['isotopes']))
```
alongd: Please start with `check_xyz_dict(xyz_dict)`, this way you are guaranteed that `xyz_dict` is of the correct format, plus if it happens to be a string it will be converted to a dict for you.

arc/species/converter.py
```diff
@@ -318,6 +320,53 @@ def xyz_from_data(coords, numbers=None, symbols=None, isotopes=None):
...
+    mass, number = zip(*mass_and_number)
```
alongd: nice!

arc/species/converterTest.py
amarkpayne: Thanks for the suggestion! Done

arc/species/converter.py
amarkpayne: Good idea, done!

arc/species/converter.py
amarkpayne: Done!

PR #263, opened 2019-12-10
Rotor troubleshoot freeze

dranasinghe:
I implemented a rotor troubleshoot routine where only the problematic dihedrals and bonds will be frozen. This is especially important when bond breaks during the rotor scan.
Most of the modifications were done to trsh.py. A new class called ScanInfo was added and it contains information about the scan. freeze_gic will find the necessary GIC need tobe freeze. There is code to combine reverse and forward scans but this part is not active yet.
I added unit test to scan_quality_check and trash_scan_job.
Please review the ScanInfo class and freeze_gic routine. If you have any questions or unclear code, do n't hesitate to message me.

Issue #42, opened 2019-01-14
A network error causes ARC to crush


alongd:
When the machine sending command to the server looses connectivity, ARC crashes wirh:
```python
"""---------------------------------------------------------------------------
gaierror                                  Traceback (most recent call last)
<ipython-input-5-f62c9e69185c> in <module>()
      1 arc0 = arc.ARC(project='Arc-cooptima-methylpropylether-cbsqb3', composite_method='cbs-qb3', rmg_species_list=[], arc_species_list=arc_species_list,ess_settings = {'gaussian': 'c3ddb', 'molpro': 'pharos', 'qchem': 'pharos'})
----> 2 arc0.execute()
/home/dranasinghe/Software/ARC/arc/main.pyc in execute(self)
    234                                    scan_level=self.scan_level, fine=self.fine, settings=self.settings,
    235                                    generate_conformers=self.generate_conformers, scan_rotors=self.scan_rotors,
--> 236                                    initial_trsh=self.initial_trsh)
    237         prc = Processor(project=self.project, species_dict=self.scheduler.species_dict, output=self.scheduler.output,
    238                         use_bac=self.use_bac, model_chemistry=self.model_chemistry)
/home/dranasinghe/Software/ARC/arc/scheduler.pyc in __init__(self, project, settings, species_list, composite_method, conformer_level, opt_level, freq_level, sp_level, scan_level, fine, generate_conformers, scan_rotors, initial_trsh)
    151                 self.species_dict[species.label].generate_conformers()
    152         self.timer = True
--> 153         self.schedule_jobs()
    154 
    155     def schedule_jobs(self):
/home/dranasinghe/Software/ARC/arc/scheduler.pyc in schedule_jobs(self)
    261                             and not self.job_dict[label]['scan'][job_name].job_id in self.servers_jobs_ids:
    262                         job = self.job_dict[label]['scan'][job_name]
--> 263                         successful_server_termination = self.end_job(job=job, label=label, job_name=job_name)
    264                         if successful_server_termination:
    265                             self.check_scan_job(label=label, job=job)
/home/dranasinghe/Software/ARC/arc/scheduler.pyc in end_job(self, job, label, job_name)
    325                          fine=job.fine, software=job.software, shift=job.shift, trsh=job.trsh, memory=job.memory,
    326                          conformer=job.conformer, ess_trsh_methods=job.ess_trsh_methods, scan=job.scan,
--> 327                          pivots=job.pivots, occ=job.occ)
[...]

PR #41, opened 2019-01-11
Added a YAML input and restart file feature 

alongd:
Input file example:
```yaml
project: project_name

species:
  - label: vinoxy
    smiles: C=C[O]
    multiplicity: 2
    charge: 0

  - label: OH
    xyz: |
      O       0.00000000    0.00000000   -0.12002167
      H       0.00000000    0.00000000    0.85098324

  - label: methylamine
    smiles: CN
    multiplicity: 1
    charge: 0

  - label: propene
    smiles: C=CC

  - label: N2H4
    smiles: NN

ess_settings: {'gaussian': 'pharos', 'molpro': 'rmg', 'qchem': 'pharos'}

scan_rotors: False

```

PR #321, opened 2020-02-25
1D non-directed rotor scan troubleshooting

xiaoruiDong:
This is a complete update and rewrite of #263. 
This PR aims to provide customized solutions to the 1D non-directed rotor scan. Currently, it only supports Gaussian scan results, but if any other ESS provides internal coordinate (IC) Infos, it is not difficult to implement this strategy. 

The methodology applied is based on the identification of abnormal behavior of internal coordinates (bond, angle, dihedral). Pretty intuitively, if any of the ICs have a sudden significant change or a difference between the initial and final conformer, these ICs tend to be relaxed too much, resulting in switching to a different conformer (not accounting the change of the scanning dihedral) or even different species (bond breaking) during the scan. This update is able to identify these changes by implementing thresholds (check If any of the changes exceed the threshold). Usually, the results from identification are redundant (you don't wish that many dihedrals!), so an algorithm is also implemented to yield a clean solution. This is important, especially when dealing with bond flipping dihedrals which share the same pivots as the scanning dihedrals.

This PR includes changes in 4 sections:
1. parser.py: Add parsers to allow parsing IC infos
2. common.py: Add functions to allow quick IC difference check and same pivot check
3. job/trsh.py: Update the methods.
4. scheduler.py: Changes based on the change of trsh.py

Unit tests are added for some of the modules. It is also tested on a 2k scan results set. Primarily, the method can successfully come up with solutions for each problematic job identified and the solutions make sense. But tests haven't been done to see if all the solution perfectly solve the problem. Even not, the new method now makes multiple-time troubleshooting possible and should solve the problems eventually.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -18,12 +18,11 @@ import sys
 import time
 import warnings
 import yaml
-from typing import Optional
+from typing import Any, List, Optional, Tuple, Union
 
 import numpy as np
+import pandas as pd
 import qcelemental as qcel
-
-from typing import Any, List, Optional, Tuple, Union
 
 from arkane.ess import ess_factory, GaussianLog, MolproLog, OrcaLog, QChemLog, TeraChemLog
 from rmgpy.molecule.element import get_element
@@ -1093,6 +1092,46 @@ def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
         est_memory = max_server_mem_gb * 1024
 
     return est_cpu, est_memory
+
+
+def check_torsion_change(torsions: pd.DataFrame,
+                         index_1: Union[int, str],
+                         index_2: Union[int, str],
+                         threshold: Union[float, int] = 20.0,
+                         delta: Union[float, int] = 0.0,
+                         ) -> pd.DataFrame:
+    """
+    Compare two sets of torsions (in DataFrame) and check if any entry has a
+    difference larger than threshold. The output is a DataFrame consisting of
+    ``True``/``False``, indicating which torsions changed significantly.
+
+    Args:
+        torsions (pd.DataFrame): A DataFrame consisting of multiple sets of torsions.
+        index_1 (Union[int, str]): The index of the first conformer.
+        index_2 (Union[int, str]): The index of the second conformer.
+        threshold (Union[float, int]): The threshold used to determine the difference significance.
+        delta (Union[float, int]): A known difference between torsion pairs. E.g.,
+                                   for the torsions to be scanned, the dihedral
+                                   differences are equal to the scan resolution.
+
+    Returns:
+        pd.DataFrame: a DataFrame consisting of ``True``/``False``, indicating
+                      which torsions changed significantly. ``True`` for significant
+                      change.
+    """
+    # First iteration without 180/-180 adjustment
+    change = (torsions[index_1] - torsions[index_2] - delta).abs() > threshold
+    # Apply 180/-180 adjustment to those shown significance
+    for label in change[change == True].index:
+        # a -180 / 180 flip causes different sign
+        if torsions.loc[label, index_1] * torsions.loc[label, index_2] < 0:
+            if torsions.loc[label, index_1] < 0 \
+              and abs(torsions.loc[label, index_1] + 360 - torsions.loc[label, index_2] - delta) < threshold:
+                change[label] = False
+            elif torsions.loc[label, index_2] < 0 \
+              and abs(torsions.loc[label, index_1] - 360 - torsions.loc[label, index_2] - delta) < threshold:
+                change[label] = False
+    return change
 
 
 def is_same_pivot(torsion1: list, torsion2: list) -> Optional[bool]:
--- a/arc/common.py
+++ b/arc/common.py
@@ -8,7 +8,7 @@ As such, it should not import any other ARC module (specifically ones that use t
 
 VERSION is the full ARC version, using `semantic versioning <https://semver.org/>`_.
 """
-
+import ast
 import datetime
 import logging
 import os
@@ -1135,17 +1135,21 @@ def check_torsion_change(torsions: pd.DataFrame,
     return change
 
 
-def is_same_pivot(torsion1: list, torsion2: list) -> Optional[bool]:
+def is_same_pivot(torsion1: Union[list, str],
+                  torsion2: Union[list, str],
+                  ) -> Optional[bool]:
     """
     Check if two torsions have the same pivots.
 
     Args:
-        torsion1 (list): The four atom indices representing the first torsion.
-        torsion2 (list): The four atom indices representing the second torsion.
+        torsion1 (Union[list, str]): The four atom indices representing the first torsion.
+        torsion2 (Union: [list, str]): The four atom indices representing the second torsion.
 
     Returns:
-        bool: ``True`` if two torsions share the same pivots.
+        Optional[bool]: ``True`` if two torsions share the same pivots.
     """
+    torsion1 = ast.literal_eval(torsion1) if isinstance(torsion1, str) else torsion1
+    torsion2 = ast.literal_eval(torsion2) if isinstance(torsion2, str) else torsion2
     if not (len(torsion1) == len(torsion2) == 4):
         return False
     if torsion1[1:3] == torsion2[1:3] or torsion1[1:3] == torsion2[1:3][::-1]:
--- a/arc/common.py
+++ b/arc/common.py
@@ -18,6 +18,7 @@ import sys
 import time
 import warnings
 import yaml
+from typing import Optional
 
 import numpy as np
 import qcelemental as qcel
@@ -1092,3 +1093,44 @@ def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
         est_memory = max_server_mem_gb * 1024
 
     return est_cpu, est_memory
+
+
+def is_same_pivot(torsion1: list, torsion2: list) -> Optional[bool]:
+    """
+    Check if two torsions have the same pivots.
+
+    Args:
+        torsion1 (list): The four atom indices representing the first torsion.
+        torsion2 (list): The four atom indices representing the second torsion.
+
+    Returns:
+        bool: ``True`` if two torsions share the same pivots.
+    """
+    if not (len(torsion1) == len(torsion2) == 4):
+        return False
+    if torsion1[1:3] == torsion2[1:3] or torsion1[1:3] == torsion2[1:3][::-1]:
+        return True
+
+
+def is_same_sequence_sublist(child_list, parent_list):
+    """
+    Check if the parent list has a sublist which is identical to the child list including the sequence.
+    Examples:
+        - child_list = [1,2,3], parent_list=[5,1,2,3,9] -> ``True``
+        - child_list = [1,2,3], parent_list=[5,6,1,3,9] -> ``False``
+
+    Args:
+        child_list (list): The child list (the pattern to search in the parent list).
+        parent_list (list): The parent list.
+
+    Returns:
+        bool: ``True`` if the sublist is in the parent list.
+    """
+    if len(parent_list) < len(child_list):
+        return False
+    if any([item not in parent_list for item in child_list]):
+        return False
+    for index in range(len(parent_list) - len(child_list) + 1):
+        if child_list == parent_list[index:index + len(child_list)]:
+            return True
+    return False
--- a/arc/common.py
+++ b/arc/common.py
@@ -1110,8 +1110,9 @@ def check_torsion_change(torsions: pd.DataFrame,
         index_1 (Union[int, str]): The index of the first conformer.
         index_2 (Union[int, str]): The index of the second conformer.
         threshold (Union[float, int]): The threshold used to determine the difference significance.
-        delta (Union[float, int]): A known difference between torsion pairs. E.g.,
-                                   for the torsions to be scanned, the dihedral
+        delta (Union[float, int]): A known difference between torsion pairs,
+                                   delta = tor[index_1] - tor[index_2].
+                                   E.g.,for the torsions to be scanned, the
                                    differences are equal to the scan resolution.
 
     Returns:
--- a/arc/common.py
+++ b/arc/common.py
@@ -1152,7 +1152,7 @@ def is_same_pivot(torsion1: list, torsion2: list) -> Optional[bool]:
         return True
 
 
-def is_same_sequence_sublist(child_list, parent_list):
+def is_same_sequence_sublist(child_list: list, parent_list: list) -> bool:
     """
     Check if the parent list has a sublist which is identical to the child list including the sequence.
     Examples:
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -1236,17 +1236,23 @@ def scan_quality_check(label: str,
 
         # 1.1 Find significant changes of internal coordinates
         threshold = preserve_param_in_scan_stable
-        step_num = int(360 / scan_res)
+        expected_step_num = int(360 / scan_res)
+        # 5 below referes to type, atoms, scan, redundant and initial guess
+        actual_step_num = scan_conformers.shape[1] - 5
+        step_num = min(expected_step_num, actual_step_num)
         changed_ic_dict = {}
         for index_1 in range(step_num + 1):
             if index_1 != 0:
                 # Compare the 'adjacent' conformers
                 index_2 = index_1 - 1
                 delta =  scan_res  # scan[index_1] - scan[index_2] = scan_res
-            else:
+            elif step_num == expected_step_num:
                 # Compare the first and the last conformer
                 index_2 = step_num
                 delta = 0
+            else:
+                # When the scan is not finished as desired
+                continue
             # Identify changes by type
             bond_change = (2 * (bonds[index_1] - bonds[index_2]) /
                           (bonds[index_1] + bonds[index_2])).abs() > threshold['bond']
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -1245,7 +1245,7 @@ def scan_quality_check(label: str,
             if index_1 != 0:
                 # Compare the 'adjacent' conformers
                 index_2 = index_1 - 1
-                delta =  scan_res  # scan[index_1] - scan[index_2] = scan_res
+                delta = scan_res  # scan[index_1] - scan[index_2] = scan_res
             elif step_num == expected_step_num:
                 # Compare the first and the last conformer
                 index_2 = step_num
@@ -1278,13 +1278,19 @@ def scan_quality_check(label: str,
         # 1.2 Check broken bond and any lowest conformation
         # Exclude those with boken bonds (different species)
         # Better to just freeze the broken bond when bond changing first happens
-        for ics in changed_ic_dict.values():
+        for conf_index, ics in changed_ic_dict.items():
             # R(X,Y) refers to bonds in ics
-            broken_bond = [ic for ic in ics if 'R' in ic]
-            if broken_bond:
+            broken_bonds = [ic for ic in ics if 'R' in ic]
+            if broken_bonds and conf_index != 0:
+                # Find the bond that changes the most, to avoid acompanied changes, like C-O transforms
+                # to C=O, which we don't want to freeze. If other bonds need to be frozen as well,
+                # it can be done in the following troubleshooting.
+                bonds = scan_conformers.loc[broken_bonds, :]
+                bond_change = (2 * (bonds[conf_index] - bonds[conf_index - 1]) /
+                              (bonds[conf_index] + bonds[conf_index - 1])).abs()
+                broken_bond_label = bond_change.sort_values().index[-1]  # the largest change
                 # Freeze the bonds, no further freezing other ics to prevent over-constraining
-                broken_bonds = [scan_conformers['atoms'][ic_label]
-                                for ic_label in list(set(broken_bonds))]
+                broken_bonds = [scan_conformers['atoms'][broken_bond_label]]
                 invalidation_reason = f'Bond ({broken_bonds}) broke during the scan.'
                 message = f'Rotor scan of {label} between pivots {pivots} has broken bonds: ' \
                           f'{broken_bonds}. ARC will attempt to troubleshoot this rotor scan.'
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -7,10 +7,18 @@ The ARC troubleshooting ("trsh") module
 
 import math
 import os
+from typing import Optional, Tuple
 
 import numpy as np
+import pandas as pd
 
-from arc.common import get_logger, determine_ess, estimate_orca_mem_cpu_requirement, is_str_float
+from arc.common import (check_torsion_change,
+                        determine_ess,
+                        estimate_orca_mem_cpu_requirement,
+                        get_logger,
+                        is_same_pivot,
+                        is_same_sequence_sublist,
+                        is_str_float)
 from arc.exceptions import SpeciesError, TrshError
 from arc.job.local import execute_command
 from arc.job.ssh import SSHClient
@@ -19,12 +27,17 @@ from arc.settings import (delete_command,
                           inconsistency_az,
                           list_available_nodes_command,
                           maximum_barrier,
+                          preserve_param_in_scan_stable,
                           rotor_scan_resolution,
                           servers,
                           submit_filename)
 from arc.species.converter import xyz_from_data, xyz_to_coords_list
 from arc.species.species import determine_rotor_symmetry
-from arc.parser import parse_normal_displacement_modes, parse_xyz_from_file
+from arc.parser import (parse_1d_scan_coords,
+                        parse_normal_displacement_modes,
+                        parse_scan_args,
+                        parse_scan_conformers,
+                        parse_xyz_from_file)
 
 
 logger = get_logger()
@@ -1000,14 +1013,20 @@ def scan_quality_check(label: str,
                        pivots: list,
                        energies: list,
                        scan_res: float = rotor_scan_resolution,
-                       used_methods: list = None):
+                       used_methods: Optional[list] = None,
+                       log_file: Optional[str] = None,
+                       ) -> Tuple[bool, str, str, dict]:
     """
     Checks the scan's quality:
+    1. Based on intermediate conformers if available:
     - Whether the initial and final points are consistent
     - whether it is relatively "smooth"
+    2. Based on the PES curve (if intermediate conformers are unavailable):
+    - Whether the initial and final points are consistent
+    - whether it is relatively "smooth"
+    3. Common
     - whether the optimized geometry indeed represents the minimum energy conformer
     - whether the barrier height is reasonable
-    Recommends whether or not to use this rotor using the 'successful_rotors' and 'unsuccessful_rotors' attributes.
 
     Args:
         label (str): The species label.
@@ -1015,74 +1034,210 @@ def scan_quality_check(label: str,
         energies (list): The scan energies in kJ/mol.
         scan_res (float, optional): The scan resolution in degrees.
         used_methods (list, optional): Troubleshooting methods already tried out.
+        log_file (str, optional): The path to the output file.
 
     Returns:
-        invalidate (bool): Whether to invalidate this rotor, ``True`` to invalidate.
-    Returns:
-        invalidation_reason (str): Reason for invalidating this rotor.
-    Returns:
-        message (str): Error or warning message.
-    Returns:
-        actions (list): Troubleshooting methods to apply, including conformational changes.
+        Tuple[bool, str, str, dict]:
+            - Whether to invalidate this rotor, ``True`` to invalidate.
+            - Reason for invalidating this rotor.
+            - Error or warning message.
+            - Troubleshooting methods to apply, including conformational changes.
 
     Todo:
         - adjust to ND
     """
     message, invalidation_reason = '', ''
     invalidate = False
-    actions = list()
+    actions = dict()
     used_methods = used_methods or list()
     energies = np.array(energies, np.float64)
 
-    # 1. Check rotor scan curve
-    # 1.1. Check consistency between initial and final points
-    if abs(energies[-1] - energies[0]) > inconsistency_az:
-        # initial and final points differ by more than `inconsistency_az` kJ/mol.
-        # seems like this rotor broke the conformer. Invalidate
-        invalidate = True
-        invalidation_reason = f'initial and final points are inconsistent by more than {inconsistency_az:.2f} kJ/mol'
-        message = f'Rotor scan of {label} between pivots {pivots} is inconsistent by more ' \
-                  f'than {inconsistency_az:.2f} kJ/mol between initial and final positions. ' \
-                  f'Invalidating rotor.\nenergies[0] = {energies[0]}, energies[-1] = {energies[-1]}'
-        logger.error(message)
-        actions = ['inc_res', 'freeze']
-        return invalidate, invalidation_reason, message, actions
+    # Check if the conformer based method is valid
+    conformer_based_check = False
+    if log_file:
+        try:
+            scan_conformers = parse_scan_conformers(log_file)
+        except NotImplementedError:
+            message = f'Rotor scan quality check using conformer internal coordinates ' \
+                      f'has not been implemented for current ESS. Using PES curve based ' \
+                      f'check for rotor scan of {label} between pivots {pivots}.'
+            logger.warning(message)
+        else:
+            conformer_based_check = True
 
-    # 1.2. Check consistency between consecutive points
-    for j in range(len(energies) - 1):
-        if abs(energies[j] - energies[j + 1]) > inconsistency_ab * np.max(energies):
-            # Two consecutive points on the scan differ by more than `inconsistency_ab` kJ/mol.
-            # This is a serious inconsistency. Invalidate
-            invalidate = True
-            invalidation_reason = f'Two consecutive points are inconsistent by more than ' \
-                                  f'{inconsistency_ab * max(energies):.2f} kJ/mol'
-            message = f'Rotor scan of {label} between pivots {pivots} is inconsistent ' \
-                      f'by more than {inconsistency_ab * max(energies):.2f} kJ/mol between two consecutive ' \
-                      f'points. Invalidating rotor.'
+    # 1. Check based on intermediate conformers
+    if conformer_based_check:
+        bonds = scan_conformers[scan_conformers['type'] == 'R']
+        angles = scan_conformers[scan_conformers['type'] == 'A']
+        non_scan_rotor = scan_conformers[(scan_conformers['type'] == 'D') \
+                                         & (scan_conformers['scan'] == False)]
+        scan_rotor = scan_conformers[scan_conformers['scan'] == True]
+
+        # 1.1 Find significant changes of internal coordinates
+        threshold = preserve_param_in_scan_stable
+        step_num = int(360 / scan_res)
+        changed_ic_dict = {}
+        for index_1 in range(step_num + 1):
+            if index_1 != 0:
+                # Compare the 'adjacent' conformers
+                index_2 = index_1 - 1
+                delta =  scan_res  # scan[index_1] - scan[index_2] = scan_res
+            else:
+                # Compare the first and the last conformer
+                index_2 = step_num
+                delta = 0
+            # Identify changes by type
+            bond_change = (2 * (bonds[index_1] - bonds[index_2]) /
+                          (bonds[index_1] + bonds[index_2])).abs() > threshold['bond']
+            angle_change = (angles[index_1] - angles[index_2]).abs() > threshold['angle']
+            non_scan_rotor_change = check_torsion_change(torsions=non_scan_rotor,
+                                                         index_1=index_1,
+                                                         index_2=index_2,
+                                                         threshold=threshold['torsion'])
+            scan_rotor_change = check_torsion_change(torsions=scan_rotor,
+                                                     index_1=index_1,
+                                                     index_2=index_2,
+                                                     threshold=threshold['torsion'],
+                                                     delta=delta)
+            # Summarize changes
+            change_sum = pd.concat([bond_change, angle_change,
+                                    non_scan_rotor_change, scan_rotor_change])
+            changed_ics = change_sum[change_sum == True].index.to_list()
+            # Save changes in the format of {conformer index: problematic ics}
+            if changed_ics:
+                invalidate = True
+                changed_ic_dict.update({index_1: changed_ics})
+
+        # 1.2 Check broken bond and any lowest conformation
+        broken_bonds = []
+        # Exclude those with boken bonds (different species)
+        for ics in changed_ic_dict.values():
+            # R(X,Y) refers to bonds in ics
+            any_broken_bond = [ic for ic in ics if 'R' in ic]
+            if any_broken_bond:
+                broken_bonds += any_broken_bond
+        if not broken_bonds:
+            # If no bond broke, ideally all conformers should be isomorphic.
+            # Switch to the lowest conformer
+            energy_diff = energies[0] - np.min(energies)
+            # Use tighter threshold to find lower conformer
+            if energy_diff >= 0.5 or energy_diff > 0.5 * (max(energies) - min(energies)):
+                invalidate = True
+                invalidation_reason = f'Another conformer for {label} exists which is ' \
+                                      f'{energy_diff:.2f} kJ/mol lower.'
+                message = f'Species {label} is not oriented correctly around pivots {pivots}, ' \
+                          f'searching for a better conformation...'
+                logger.info(message)
+                # Find the dihedrals in degrees of the lowest conformer:
+                min_index = np.argmin(energies)
+                conf_xyzs = parse_1d_scan_coords(log_file)
+                actions = {'change conformer': conf_xyzs[min_index]}
+                return invalidate, invalidation_reason, message, actions
+        else:
+            # Freeze the bonds, no further freezing other ics to prevent over-constraining
+            broken_bonds = [scan_conformers['atoms'][ic_label] for ic_label in list(set(broken_bonds))]
+            invalidation_reason = f'Bond ({broken_bonds}) broke during the scan.'
+            message = f'Rotor scan of {label} between pivots {pivots} has broken bonds: ' \
+                      f'{broken_bonds}. ARC will attempt to troubleshoot this rotor scan.'
             logger.error(message)
-            if ['inc_res'] not in used_methods:
-                actions = ['inc_res']
-            elif ['inc_res', 'freeze'] not in used_methods:
-                actions = ['inc_res', 'freeze']
+            actions = {'freeze': broken_bonds}
             return invalidate, invalidation_reason, message, actions
 
-    # 2. Check conformation:
-    energy_diff = energies[0] - np.min(energies)
-    if energy_diff >= 2 or energy_diff > 0.5 * (max(energies) - min(energies)):
-        invalidate = True
-        invalidation_reason = f'Another conformer for {label} exists which is {energy_diff:.2f} kJ/mol lower.'
-        message = f'Species {label} is not oriented correctly around pivots {pivots}. ' \
-                  f'Another conformer exists which is {energy_diff:.2f} kJ/mol lower. ' \
-                  f'searching for a better conformation...'
-        logger.info(message)
-        # Find the rotation dihedral in degrees to the closest minimum:
-        min_index = np.argmin(energies)
-        deg_increment = min_index * scan_res
-        actions = ['change conformer', pivots, deg_increment]
-        if actions in used_methods:
-            logger.error(f'Not troubleshooting a rotor with the same method: {actions}')
-            actions = list()
-        return invalidate, invalidation_reason, message, actions
+        # 1.3 Check consistency
+        if 0 in changed_ic_dict.keys() and len(changed_ic_dict) == 1:
+            # Smooth scan with different initial and final conformer
+            invalidation_reason = 'Inconsistent initial and final conformers'
+            message = f'Rotor scan of {label} between pivots {pivots} has inconsistent initial ' \
+                      f'and final conformers. Internal coordinates {changed_ic_dict[0]} are different. ' \
+                      f'ARC will attempt to troubleshoot this rotor scan.'
+            logger.error(message)
+            actions = {'freeze': [scan_conformers['atoms'][ic_label]
+                                  for ic_label in changed_ic_dict[0]]}
+            return invalidate, invalidation_reason, message, actions
+        elif len(changed_ic_dict) > 0:
+            # Not smooth scan
+            invalidation_reason = 'Significant difference observed between consecutive conformers'
+            message = f'Rotor scan of {label} between pivots {pivots} is inconsistent between ' \
+                      f'two consecutive conformers.\nInconsistent consecutive conformers and problematic ' \
+                      f'internal coordinates:'
+            changed_ic_label = []
+            for index, ics in changed_ic_dict.items():
+                if index > 0:  # Do not include the initial/final differences which may include more ics
+                    message += f'\nconformer #{index:>3d} / #{index+1:>3d}        '
+                    message += ', '.join(ics)
+                changed_ic_label += ics
+            message += '\nARC will attempt to troubleshoot this rotor scan.'
+            # list(set()) is used to remove duplicate labels
+            changed_ic_label = list(set(changed_ic_label))
+            logger.error(message)
+            actions = {'freeze': [scan_conformers['atoms'][ic_label]
+                                  for ic_label in changed_ic_label]}
+            return invalidate, invalidation_reason, message, actions
+
+    else:
+        # 2. Check rotor scan quality according to the PES curve
+        # 2.1. Check consistency between initial and final points
+        if abs(energies[-1] - energies[0]) > inconsistency_az:
+            # initial and final points differ by more than `inconsistency_az` kJ/mol.
+            # seems like this rotor broke the conformer. Invalidate
+            invalidate = True
+            invalidation_reason = f'initial and final points are inconsistent by more than {inconsistency_az:.2f} kJ/mol'
+            message = f'Rotor scan of {label} between pivots {pivots} is inconsistent by more ' \
+                      f'than {inconsistency_az:.2f} kJ/mol between initial and final positions. ' \
+                      f'Initial energy = {energies[0]}, final energy = {energies[-1]}. ARC will ' \
+                      f'attempt to troubleshoot this rotor scan.'
+            logger.error(message)
+            actions = {'inc_res': None, 'freeze': 'all'}
+            return invalidate, invalidation_reason, message, actions
+
+        # 2.2. Check consistency between consecutive points
+        for j in range(len(energies) - 1):
+            if abs(energies[j] - energies[j + 1]) > inconsistency_ab * np.max(energies):
+                # Two consecutive points on the scan differ by more than `inconsistency_ab` kJ/mol.
+                # This is a serious inconsistency. Invalidate
+                invalidate = True
+                invalidation_reason = f'Two consecutive points are inconsistent by more than ' \
+                                      f'{inconsistency_ab * max(energies):.2f} kJ/mol'
+                message = f'Rotor scan of {label} between pivots {pivots} is inconsistent by' \
+                          f'more than {inconsistency_ab * max(energies):.2f} kJ/mol between ' \
+                          f'two consecutive points. ARC will attempt to troubleshoot this rotor scan.'
+                logger.error(message)
+                # Propose a method
+                # Try increasing resolution firstly, and try increasing res. and freezing all
+                # torsions jointly, afterwards. 
+                # TODO: If we figure out that solely increasing res. is not effective,
+                # we can simplify the process to actions = {'inc_res': None, 'freeze': 'all'}
+                if any(['scan_res' in used_method for used_method in used_methods]):
+                    # Check if increasing scan resolution is ever applied
+                    if not any([used_method['scan_trsh'] != '' for used_method in used_methods]):
+                        # Case where freezing torisions has not been applied
+                        actions = {'inc_res': None, 'freeze': 'all'}
+                    else:
+                        # Since all torsions are frozen, there's not much we can do except increasing
+                        # scan resolution. But it is not that effective either. So stop and do nothing.
+                        pass
+                else:
+                    # Case where neither increasing scan resolution nor freezing
+                    # torisions has been applied
+                    actions = {'inc_res': None}
+                return invalidate, invalidation_reason, message, actions
+
+        # 2.3 Check energy and change conformation if needed:
+        energy_diff = energies[0] - np.min(energies)
+        if energy_diff >= 2 or energy_diff > 0.5 * (max(energies) - min(energies)):
+            invalidate = True
+            invalidation_reason = f'Another conformer for {label} exists which is {energy_diff:.2f} kJ/mol lower.'
+            message = f'Species {label} is not oriented correctly around pivots {pivots}. ' \
+                      f'Another conformer exists which is {energy_diff:.2f} kJ/mol lower. ' \
+                      f'searching for a better conformation...'
+            logger.info(message)
+            # Find the lowest conformer, and use the new conformer for further jobs.
+            # Since at this point, the scan has passed previous checks, the possibility
+            # to switch to a non-isomorphic conformer is low.
+            min_index = np.argmin(energies)
+            conf_xyzs = parse_1d_scan_coords(log_file)
+            actions = {'change conformer': conf_xyzs[min_index]}
+            return invalidate, invalidation_reason, message, actions
 
     # 3. Check the barrier height
     if (np.max(energies) - np.min(energies)) > maximum_barrier:
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -1270,38 +1270,37 @@ def scan_quality_check(label: str,
                 changed_ic_dict.update({index_1: changed_ics})
 
         # 1.2 Check broken bond and any lowest conformation
-        broken_bonds = []
         # Exclude those with boken bonds (different species)
+        # Better to just freeze the broken bond when bond changing first happens
         for ics in changed_ic_dict.values():
             # R(X,Y) refers to bonds in ics
-            any_broken_bond = [ic for ic in ics if 'R' in ic]
-            if any_broken_bond:
-                broken_bonds += any_broken_bond
-        if not broken_bonds:
-            # If no bond broke, ideally all conformers should be isomorphic.
-            # Switch to the lowest conformer
-            energy_diff = energies[0] - np.min(energies)
-            # Use tighter threshold to find lower conformer
-            if energy_diff >= 0.5 or energy_diff > 0.5 * (max(energies) - min(energies)):
-                invalidate = True
-                invalidation_reason = f'Another conformer for {label} exists which is ' \
-                                      f'{energy_diff:.2f} kJ/mol lower.'
-                message = f'Species {label} is not oriented correctly around pivots {pivots}, ' \
-                          f'searching for a better conformation...'
-                logger.info(message)
-                # Find the dihedrals in degrees of the lowest conformer:
-                min_index = np.argmin(energies)
-                conf_xyzs = parse_1d_scan_coords(log_file)
-                actions = {'change conformer': conf_xyzs[min_index]}
+            broken_bond = [ic for ic in ics if 'R' in ic]
+            if broken_bond:
+                # Freeze the bonds, no further freezing other ics to prevent over-constraining
+                broken_bonds = [scan_conformers['atoms'][ic_label]
+                                for ic_label in list(set(broken_bonds))]
+                invalidation_reason = f'Bond ({broken_bonds}) broke during the scan.'
+                message = f'Rotor scan of {label} between pivots {pivots} has broken bonds: ' \
+                          f'{broken_bonds}. ARC will attempt to troubleshoot this rotor scan.'
+                logger.error(message)
+                actions = {'freeze': broken_bonds}
                 return invalidate, invalidation_reason, message, actions
-        else:
-            # Freeze the bonds, no further freezing other ics to prevent over-constraining
-            broken_bonds = [scan_conformers['atoms'][ic_label] for ic_label in list(set(broken_bonds))]
-            invalidation_reason = f'Bond ({broken_bonds}) broke during the scan.'
-            message = f'Rotor scan of {label} between pivots {pivots} has broken bonds: ' \
-                      f'{broken_bonds}. ARC will attempt to troubleshoot this rotor scan.'
-            logger.error(message)
-            actions = {'freeze': broken_bonds}
+
+        # If no bond broke, ideally all conformers should be isomorphic.
+        # Switch to the lowest conformer
+        energy_diff = energies[0] - np.min(energies)
+        # Use tighter threshold to find lower conformer
+        if energy_diff >= 0.5 or energy_diff > 0.5 * (max(energies) - min(energies)):
+            invalidate = True
+            invalidation_reason = f'Another conformer for {label} exists which is ' \
+                                  f'{energy_diff:.2f} kJ/mol lower.'
+            message = f'Species {label} is not oriented correctly around pivots {pivots}, ' \
+                      f'searching for a better conformation...'
+            logger.info(message)
+            # Find the dihedrals in degrees of the lowest conformer:
+            min_index = np.argmin(energies)
+            conf_xyzs = parse_1d_scan_coords(log_file)
+            actions = {'change conformer': conf_xyzs[min_index]}
             return invalidate, invalidation_reason, message, actions
 
         # 1.3 Check consistency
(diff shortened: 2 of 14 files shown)


alongd:


```diff

Issues
======
- Added 16
           

Clones added
============
- arc/job/trsh.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=5069484&bid=16680380)

xiaoruiDong:
Unused import will be removed. I checked the travis test and found a problem related to the sequence of the list (obtain [b,a], expect[a,b], not equal). The error did not show up in my local test. Any advice?

alongd:
One of the tests passes, and another failed (and also, locally it did pass for you).
Looks like something in the code might be nondeterministic, let's take a look.

arc/common.py
```diff
@@ -907,3 +907,43 @@ def time_lapse(t0):
     else:
         d = ''
     return f'{d}{h:02.0f}:{m:02.0f}:{s:02.0f}'
+
+
+def is_same_pivot(dihed1, dihed2):
```
alongd: let's try to be consistent with the rest of ARC in terms of naming.
"dihedral" sounds like the value of the angle, let's use "torsion" for describing the atom indices defining the dihedral angle. (see, e.g., https://github.com/ReactionMechanismGenerator/ARC/blob/26d7daed170880a58669255512f66370109de173/arc/species/vectors.py#L177)
I use "scan" as a parameter name for 1-indexed torsions (like the input to Gaussian), and "torsion" as a parameter name for 0-indexed torsions. It helps to know how to work with each parameter just by its name

arc/common.py
```diff
@@ -907,3 +907,43 @@ def time_lapse(t0):
...
+    """
+    Check if two dihedrals have the same pivot.
+
+    Args:
+        dihed1 (list): The four atom indices representing the first torsion.
+        dihed2 (list): The four atom indices representing the second torsion.
+
+    Returns:
+        (Bool): True if two dihedrals share the same pivot.
```
alongd: I think there's no need for `( )` around `bool`, also I think it should be lowercase
Add "``True``" (please see the source for this comment, these are two weird quotation marks on each side) around `True`, it shows it in _italics_ in the API documentation this way.

arc/common.py
```diff
@@ -907,3 +907,43 @@ def time_lapse(t0):
...
+    Args:
+        child_list (list): The child list (the pattern to search in the parent list)
+        parent_list (list): The parent list
+
+    Returns:
+        (Bool): True if the sublist is in the parent list 
+    """
+    if len(parent_list) < len(child_list):
+        return False
+    for item in child_list:
```
alongd: consider
`if any([item not in parent_list for item in child_list])`

arc/parser.py
```diff
@@ -10,11 +10,12 @@
 
 import qcelemental as qcel
 import numpy as np
+import pandas as pd
 from arkane.exceptions import LogError
```
alongd: Add an empty line between general imports and Arkane/RMG

arc/parser.py
```diff
@@ -10,11 +10,12 @@
...
 from arkane.ess import GaussianLog, MolproLog, OrcaLog, QChemLog, TeraChemLog
 from arkane.util import determine_qm_software
 
-from arc.common import get_logger, determine_ess
+from arc.common import get_logger, determine_ess, is_same_pivot
```
alongd: do you mind sorting these alphabetically?

alongd:
Adding a screenshot of the error in case it passes when force-pushed:
image

arc/parser.py
```diff
@@ -718,3 +718,58 @@ def process_conformers_file(conformers_path):
...
+            if mode == 'search':
+                if (len(blks)) == block_count:
+                    break
+                else:
+                    match = search(head_pat, line)
+                    if match:
+                        tail_repeat = 0
+                        mode = 'read'
+                        blks.append([])
+                        blks[-1].append(line)
```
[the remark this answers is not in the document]
xiaoruiDong: blks is a list of lists of lines. E.g., `blks = [['good', 'morning',], ['good', 'bye'],]`. So `blks[-1].append` makes sure that line is apppend to the last block (E.g.,`['good', 'bye']` instead of `['good', 'morning',]`)

arc/parser.py
```diff
@@ -718,3 +718,58 @@ def process_conformers_file(conformers_path):
...
+                blks[-1].append(line)
+                match = search(tail_pat, line)
+                if match:
+                    tail_repeat += 1
+                    if tail_repeat == tail_count:
+                        mode = 'search'
+            line = f.readline()
+        # Remove the last incomplete search
+        if len(blks) > 0 and (tail_repeat != tail_count):
+            blks.pop()
```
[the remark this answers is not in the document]
xiaoruiDong: Since `blks` is a list of blocks, this pops the last block.

arc/parser.py
```diff
@@ -773,3 +773,47 @@ def search(x, y): return x in y
         if len(blks) > 0 and (tail_repeat != tail_count):
             blks.pop()
         return blks
+
+
+def parse_scan_args(file_path, software=None):
```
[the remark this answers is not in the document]
xiaoruiDong: My suggestion is keeping them separate, considering maintainance and flexibility.

arc/parser.py
```diff
@@ -773,3 +773,47 @@ def search(x, y): return x in y
...
+    """
+    Get the scan arguments, including which internal coordinates are scanning, which are frozen,
```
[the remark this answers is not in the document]
xiaoruiDong: You are right

arc/parser.py
```diff
@@ -773,3 +773,47 @@ def search(x, y): return x in y
...
+                          as well as step number, step size, number of atom
+    
+    Raises:
+        NotImplementedError: If the software is not supported
+    """
+    scan_args = {'scan': None, 'freeze': [],
+                 'step': 0, 'step_size': 0, 'n_atom': 0}
+    if software == 'gaussian':
+        scan_blk = parse_str_blocks(file_path, 'The following ModRedundant input section has been read:',
+                                    'Isotopes and Nuclear Properties', regex=False)[0][1:-1]
```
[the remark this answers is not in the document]
xiaoruiDong: `[1:-1]` exclude the first and the last lines, while `[1:]` only excludes the first line.

alongd:


```diff

Issues
======
- Added 1
           

Clones added
============
- arc/job/trsh.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=5069484&bid=16680380)

xiaoruiDong:
Thank you @alongd for the preliminary review. I know you are busy, so take your time to review the rest of the code. I think the indeterministic behavior is due to the use of `list(set())` (to remove duplicates) which loses sequence information, but the sequence has no importance for post-processing. So I just modify the way of testing.

alongd:


```diff

Issues
======
- Added 1
           

Clones added
============
- arc/job/trsh.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=5069484&bid=16680380)

alongd:


```diff

Issues
======
- Added 1
           

Clones added
============
- arc/job/trsh.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=5069484&bid=16680380)

alongd:


```diff

Issues
======
- Added 1
           

Clones added
============
- arc/job/trsh.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=5069484&bid=16680380)

arc/job/trsh.py
```diff
@@ -978,57 +1097,184 @@ def scan_quality_check(label, pivots, energies, scan_res=rotor_scan_resolution,
...
+                    else:
+                        # E.g., deg_abs = [([1,2,3,4], 10.0), ....]
+                        deg_abs.append((scan_conformers['atoms'][dihed_label], scan_conformers[min_index][dihed_label]))
+                        pruning += [d_l for d_l in changed_dihed_label
+                                    if is_same_pivot(deg_abs[-1][0], scan_conformers['atoms'][d_l])]
+                actions = ['change conformer', deg_abs]
+                return invalidate, invalidation_reason, message, actions
+        else:
+            bond_break = [scan_conformers['atoms'][ic_label] for ic_label in list(set(bond_break))]
+            invalidation_reason = f'Bond {bond_break} break during the scan.'
```
alongd: "broke" instead of "break"

arc/job/trsh.py
```diff
@@ -978,57 +1097,184 @@ def scan_quality_check(label, pivots, energies, scan_res=rotor_scan_resolution,
...
+            message = f'Rotor scan of {label} between pivots {pivots} has breaking bonds ' \
```
alongd: "broken" instead of "breaking"

arc/job/trsh.py
```diff
@@ -978,57 +1097,184 @@ def scan_quality_check(label, pivots, energies, scan_res=rotor_scan_resolution,
...
-        if abs(energies[j] - energies[j + 1]) > inconsistency_ab * np.max(energies):
-            # Two consecutive points on the scan differ by more than `inconsistency_ab` kJ/mol.
-            # This is a serious inconsistency. Invalidate
+        # 2.2 Check consistency
+        # Smooth scan with different initial and final conformer
+        if 0 in changed_ic_dict.keys() and len(changed_ic_dict) == 1:
+            invalidation_reason = 'Inconsistent initial and final conformers'
+            message = f'Rotor scan of {label} between pivots {pivots} has inconsistent initial ' \
+                      f'and final conformers. Internal coordinates {changed_ic_dict[0]} are different.' \
+                      f'Invalidating rotor.'
```
alongd: A general comment: I think it's incorrect to say "invalidating rotor" if we plan to troubleshoot it with some actions. "invalidation" is for rotors we don't know how to treat (if all trsh methods failed), and adding them to the statmech analysis will do more harm than good. Perhaps say "ARC will attempt to troubleshoot this rotor." instead.

arc/job/trsh.py
```diff
@@ -978,57 +1097,184 @@ def scan_quality_check(label, pivots, energies, scan_res=rotor_scan_resolution,
...
+            actions = ['freeze', [scan_conformers['atoms'][ic_label] for ic_label in changed_ic_dict[0]]]
+            return invalidate, invalidation_reason, message, actions
+        # Not smooth scan
+        elif len(changed_ic_dict) > 0:
+            invalidation_reason = 'Significant difference observed between consecutive conformers'
+            message = f'Rotor scan of {label} between pivots {pivots} is inconsistent ' \
+                      f'between two consecutive conformers. Invalidating rotor.'
+            changed_ic_label = []
+            for index, ics in changed_ic_dict.items():
+                if index > 0: # Do not include the initial/final differences which may include more ics
```
alongd: minor: two spaces between code and `#` comment marker

arc/job/trsh.py
```diff
@@ -978,57 +1097,184 @@ def scan_quality_check(label, pivots, energies, scan_res=rotor_scan_resolution,
...
+                          'points. Invalidating rotor.'.format(
+                          label=label, pivots=pivots, incons_ab=inconsistency_ab * max(energies))
+                logger.error(message)
+                if ['inc_res'] not in used_methods:
+                    actions = ['inc_res']
+                elif ['inc_res', 'all_freeze'] not in used_methods:
+                    actions = ['inc_res', 'all_freeze']
+                return invalidate, invalidation_reason, message, actions
+
+        # 2.3 Check conformation:
```
alongd: "# 2.3 Check energy and change conformation if needed"

arc/job/trshTest.py
```diff
@@ -448,6 +450,82 @@ def test_trsh_negative_freq(self):
         self.assertEqual(output_errors, list())
         self.assertEqual(output_warnings, list())
 
+    def test_scan_quality_check(self):
+        """Test scan quality check for 1D rotor"""
+        # Case 1: non-smooth scan which troubleshot once
+        case1 = {'label': 'CH2OOH',
+                 'pivots': [1, 2],
+                 'energies': np.array([1.05018884e-05, 5.19029816e-01, 1.98559728e+00, 4.14500528e+00,
```
alongd: do we need to specify the energies if we have the log file? Perhaps parse them instead? (same below)

arc/job/trshTest.py
```diff
@@ -448,6 +450,82 @@ def test_trsh_negative_freq(self):
...
+                                    4.08638049e+00, 1.93710692e+00, 5.03410717e-01, 4.72587962e-05,
+                                    4.84381094e-01, 1.88534257e+00, 3.99013492e+00, 6.48881000e+00,
+                                    8.97111510e+00, 1.09718851e+01, 1.20717543e+01, 1.20193047e+01,
+                                    1.08285302e+01, 8.73798121e+00, 6.18581944e+00, 3.66210759e+00,
+                                    1.60102194e+00, 3.24317495e-01, 1.95836032e-02, 7.01840746e-01,
+                                    2.24207974e+00, 4.38485256e+00, 6.80048572e+00, 9.11334105e+00,
+                                    1.09528319e+01, 1.20173539e+01, 1.21102100e+01, 1.11961343e+01,
+                                    9.36754720e+00, 6.92812177e+00, 4.32620415e+00, 2.05935807e+00,
+                                    5.33139252e-01, 0.00000000e+00]),
+                 'scan_res': 4.0,
```
alongd: there are 46 points, seems like a resolution of 8 degrees

arc/job/trshTest.py
```diff
@@ -448,6 +450,82 @@ def test_trsh_negative_freq(self):
...
+                 'log_file': os.path.join(arc_path, 'arc', 'testing', 'rotor_scans', 'CH2OOH.out'),
+                }
+        invalidate, invalidation_reason, message, actions = trsh.scan_quality_check(**case1)
+        self.assertTrue(invalidate)
+        self.assertEqual(
+            invalidation_reason, 'Significant difference observed between consecutive conformers')
+        expect_message = 'Rotor scan of CH2OOH between pivots [1, 2] is inconsistent between ' \
+                         'two consecutive conformers. Invalidating rotor. Between conformer ' \
+                         '#63 and #64, D3,D2 are changed significantly. Between conformer #80 ' \
+                         'and #81, D3,D2 are changed significantly.'
```
alongd: how come there are more than 80 conformers if there are only 46 energy points?

arc/parser.py
```diff
@@ -718,3 +720,241 @@ def process_conformers_file(conformers_path):
         else:
             line_index += 1
     return xyzs, energies
+
+
+def parse_str_blocks(file_path, head_pat, tail_pat, regex=True, tail_count=1, block_count=1):
```
alongd: Could you add type hints throughout this PR in the following format? (If for some functions it's not trivial, feel free to skip it!)
(for regex see https://docs.python.org/3/library/typing.html#typing.Match)
E.g.:
```python
from typing import Match

def parse_str_blocks(file_path: str,
                     head_pat: str or Match,
                     tail_pat: str or Match,
                     regex: bool = True,
                     tail_count: int = 1,
                     block_count: int = 1,
                     ) -> list:
```

arc/parser.py
```diff
@@ -718,3 +720,241 @@ def process_conformers_file(conformers_path):
...
+               for item in ['label', 'type', 'atoms', 'redundant', 'scan']}
+    scan_args = parse_scan_args(file_path)
+    max_atom_ind = scan_args['n_atom']
+    if isinstance(log, GaussianLog):
+        ic_info_block = parse_str_blocks(file_path, 'Initial Parameters', '-----------', regex=False,
+                                         tail_count=3)[0][5:-1]
+        for line in ic_info_block:
+            terms = line.split()
+            ic_dict['label'].append(terms[1])
+            ic_dict['type'].append(terms[1][0])  #'R: bond, A: angle, D: dihedral
```
alongd: minor: space after `#`

arc/parser.py
```diff
@@ -718,3 +720,241 @@ def process_conformers_file(conformers_path):
...
+            atom_inds = re.split(r'\(|\)|,', terms[2])[1:-1]
+            ic_dict['atoms'].append([int(atom_ind) for atom_ind in atom_inds])
+
+            # Identify redundant, cases like 5 atom angles or redundant atoms
+            if (ic_dict['type'][-1] == 'A' and len(atom_inds) > 3) \
+            and (ic_dict['type'][-1] == 'R' and len(atom_inds) > 2) \
```
alongd: fix indentation (two more "tabs")

arc/parser.py
```diff
@@ -718,3 +720,241 @@ def process_conformers_file(conformers_path):
...
+            and (ic_dict['type'][-1] == 'D' and len(atom_inds) > 4):
```
alongd: I think these `and`s should be `or`s
```python
if (ic_dict['type'][-1] == 'A' and len(atom_inds) > 3) \
        or (ic_dict['type'][-1] == 'R' and len(atom_inds) > 2) \
        or (ic_dict['type'][-1] == 'D' and len(atom_inds) > 4):
```

arc/parser.py
```diff
@@ -718,3 +720,241 @@ def process_conformers_file(conformers_path):
...
+                ic_dict['redundant'].append(True)
+            else:
+                weird_indices = [index for index in ic_dict['atoms']
+                                [-1] if index <= 0 or index > max_atom_ind]
```
alongd: does `[-1]` refer to `ic_dict['atoms']`? Try not to break them for better readability

alongd: what's the meaning of such indices in a Gaussian output file?

arc/parser.py
```diff
@@ -718,3 +720,241 @@ def process_conformers_file(conformers_path):
...
+    Raises:
+        NotImplementedError: If the software is not supported
+
+    Returns:
+        pd.DataFrame: A DataFrame containing the values of the internal coordinates
+    """
+    ic_dict = {item: [] for item in ['label', 'value']}
+    if software == 'gaussian':
+        for line in ic_block:
+            terms = line.split()
```
alongd: could you give a commented example of the format of the `line` this code in processing?

arc/parserTest.py
```diff
@@ -332,6 +332,34 @@ def test_process_conformers_file(self):
         spc6 = ARCSpecies(label='tst6', xyz=path3)
         self.assertEqual(len(spc6.conformers), 4)
 
+    def test_parse_str_blocks(self):
```
alongd: Can you also add tests for parse_scan_conformers(), parse_ic_values(), and parse_ic_info()?

arc/scheduler.py
```diff
@@ -2187,16 +2187,27 @@ def check_scan_job(self, label, job):
...
 
                 if len(actions):
                     # the rotor scan is problematic, troubleshooting is required
                     if 'change conformer' in actions:
                         # a lower conformation was found
-                        deg_increment = actions[2]
-                        self.species_dict[label].set_dihedral(
-                            scan=self.species_dict[label].rotors_dict[i]['scan'],
-                            deg_increment=deg_increment)
+                        if len(actions) == 3:
```
alongd: I think we should make `actions` a dictionary, otherwise this is a recipe for bugs. Do you mind converting it?

arc/species/species.py
```diff
@@ -1063,36 +1063,38 @@ def set_dihedral(self, scan, deg_increment=None, deg_abs=None, count=True, xyz=N
...
 
         Raises:
             InputError: If both ``deg_increment`` and ``deg_abs`` are None.
             RotorError: If the rotor could not be identified based on the pivots.
             TypeError: If ``deg_increment`` or ``deg_abs`` are of wrong type.
         """
-        if deg_increment is not None and not isinstance(deg_increment, float):
+        if deg_increment is None and deg_abs is None:
+            raise InputError('Either deg_increment or deg_abs must be specified.')
+        elif deg_increment is not None and not isinstance(deg_increment, float):
```
alongd: Please modify to `isinstance(deg_increment, (float, int))` (below as well, we can keep the error message with the `float` requirement.

alongd: not need for an `elif` after a `raise`

alongd:
Take a look at the Codacy report, it complains about an unused import

alongd:
I ran `COCOOC` on this branch, it seems to be oscillating when troubleshooting two of the rotors:
```
Found 5 possible rotors for COCOOC
Pivot list(s) for COCOOC: [[1, 2], [2, 3], [3, 4], [4, 5], [5, 6]]

...


Trying to troubleshoot rotor [3, 4] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
Running job scan_a40277 for COCOOC (pivots: [3, 4])
  Ending job scan_a40276 for COCOOC (run time: 0:12:13)
Error: Rotor scan of COCOOC between pivots [2, 3] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #25 and #26, D10 are changed significantly. Between conformer #26 and #27, D10 are changed significantly. Between conformer #32 and #33, D10 are changed significantly.
Trying to troubleshoot rotor [2, 3] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
Running job scan_a40278 for COCOOC (pivots: [2, 3])
  Ending job scan_a40277 for COCOOC (run time: 0:13:07)
Error: Rotor scan of COCOOC between pivots [3, 4] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #34 and #35, D10 are changed significantly. Between conformer #41 and #42, D10 are changed significantly.
Trying to troubleshoot rotor [3, 4] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
Running job scan_a40279 for COCOOC (pivots: [3, 4])
  Ending job scan_a40278 for COCOOC (run time: 0:12:18)
Error: Rotor scan of COCOOC between pivots [2, 3] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #25 and #26, D10 are changed significantly. Between conformer #26 and #27, D10 are changed significantly. Between conformer #32 and #33, D10 are changed significantly.
Trying to troubleshoot rotor [2, 3] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
Running job scan_a40280 for COCOOC (pivots: [2, 3])
  Ending job scan_a40279 for COCOOC (run time: 0:13:17)
Error: Rotor scan of COCOOC between pivots [3, 4] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #34 and #35, D10 are changed significantly. Between conformer #41 and #42, D10 are changed significantly.
Trying to troubleshoot rotor [3, 4] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
Running job scan_a40281 for COCOOC (pivots: [3, 4])
```

xiaoruiDong:
> I ran `COCOOC` on this branch, it seems to be oscillating when troubleshooting two of the rotors:
> 
> ```
> Found 5 possible rotors for COCOOC
> Pivot list(s) for COCOOC: [[1, 2], [2, 3], [3, 4], [4, 5], [5, 6]]
> 
> ...
> 
> 
> Trying to troubleshoot rotor [3, 4] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
> Running job scan_a40277 for COCOOC (pivots: [3, 4])
>   Ending job scan_a40276 for COCOOC (run time: 0:12:13)
> Error: Rotor scan of COCOOC between pivots [2, 3] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #25 and #26, D10 are changed significantly. Between conformer #26 and #27, D10 are changed significantly. Between conformer #32 and #33, D10 are changed significantly.
> Trying to troubleshoot rotor [2, 3] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
> Running job scan_a40278 for COCOOC (pivots: [2, 3])
>   Ending job scan_a40277 for COCOOC (run time: 0:13:07)
> Error: Rotor scan of COCOOC between pivots [3, 4] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #34 and #35, D10 are changed significantly. Between conformer #41 and #42, D10 are changed significantly.
> Trying to troubleshoot rotor [3, 4] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
> Running job scan_a40279 for COCOOC (pivots: [3, 4])
>   Ending job scan_a40278 for COCOOC (run time: 0:12:18)
> Error: Rotor scan of COCOOC between pivots [2, 3] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #25 and #26, D10 are changed significantly. Between conformer #26 and #27, D10 are changed significantly. Between conformer #32 and #33, D10 are changed significantly.
> Trying to troubleshoot rotor [2, 3] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
> Running job scan_a40280 for COCOOC (pivots: [2, 3])
>   Ending job scan_a40279 for COCOOC (run time: 0:13:17)
> Error: Rotor scan of COCOOC between pivots [3, 4] is inconsistent between two consecutive conformers. Invalidating rotor. Between conformer #34 and #35, D10 are changed significantly. Between conformer #41 and #42, D10 are changed significantly.
> Trying to troubleshoot rotor [3, 4] of COCOOC using ['freeze', [[3, 4, 5, 6]]]...
> Running job scan_a40281 for COCOOC (pivots: [3, 4])
> ```

Thank you for the feedback. To me, this seems to be that the frozen is not actually done. If possible, can you please screenshot the scan info of the input file of scan_a40279 and scan_a40278. D(3, 4, 5, 6) is expected to be frozen and no longer show up.

alongd:
Sure, here are the input files:
[scan_a40278.txt](https://github.com/ReactionMechanismGenerator/ARC/files/4325322/scan_a40278.txt)
[scan_a40279.txt](https://github.com/ReactionMechanismGenerator/ARC/files/4325323/scan_a40279.txt)

arc/parser.py
```diff
@@ -718,3 +718,58 @@ def process_conformers_file(conformers_path):
...
+            elif mode == 'read':
+                blks[-1].append(line)
+                match = search(tail_pat, line)
+                if match:
+                    tail_repeat += 1
+                    if tail_repeat == tail_count:
+                        mode = 'search'
+            line = f.readline()
+        # Remove the last incomplete search
+        if len(blks) > 0 and (tail_repeat != tail_count):
```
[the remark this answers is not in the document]
xiaoruiDong: For example, you only find the head, but no tail is found until the last line. Then, it may be desirable to remove this block, since it does not satisfy the search pattern.

arc/common.py
xiaoruiDong: @alongd Help! There is something that needs to be clarified. To me, [1, 2, 3, 4] and [5, 2, 3, 6] refers to the same torsion. But in scan troubleshooting, it is important to distinguish them, since they are not changed with the same degrees in some cases. Is that okay to call them torsion1 and torsion2? To me, they just refer to different dihedral angles, that's the reason I call them dihedrals.

alongd: Yes, it's OK to call them torsion1 and torsion2, even if they (may) represent the same torsion, since technically the answer the above defiinition.

arc/parser.py
xiaoruiDong: The role of these redundant atoms is not clear (to me). However, they are found in the internal coordinates. The weirdest case I found is that Gaussian used 5 atoms to define an angle that I even cannot figure out which angle it is.

arc/common.py
```diff
@@ -947,3 +947,75 @@ def time_lapse(t0):
...
+                                 a known dihedral differences between adjacent scanning conformers.
+
+    Returns:
+        pd.DataFrame: a DataFrame consisting of True/False, indicating which dihedrals are
+                      changing significantly. `True` for changing significantly.
+    """
+    # First iteration without 180/-180 adjustment
+    d_change = (diheds[ind2] - diheds[ind1] - delta).abs() > threshold
+    # Apply 180/-180 adjustment to those shown significance
+    for d_label in d_change[d_change == True].index:
```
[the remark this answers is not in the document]
xiaoruiDong: When there is only one column (with index) in the DataFrame., you can use this syntax as a filter function. It will return all of the entries which equal to `True`. So in this context, this will return all torsions which have significant differences. Look at this link about pandas syntax for filtering, http://scrapingauthority.com/pandas-dataframe-filtering/

arc/job/trsh.py
```diff
@@ -468,31 +469,143 @@ def trsh_scan_job(label, scan_res, scan, species_scan_lists, methods):
...
     if methods is None:
         raise TrshError('Expected to get a list of methods, got None.')
+
     scan_trsh = ''
     if 'freeze' in methods:
-        if scan not in species_scan_lists:
-            raise TrshError(f'Could not find the dihedral to troubleshoot for in the scan list of species {label}')
+        try:
+            scan_args = parse_scan_args(log_file)
+            already_frozen = scan_args['freeze']
```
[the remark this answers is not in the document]
xiaoruiDong: `scan_args` always has `freeze` but yes, you are right, it should be outside.

arc/job/trsh.py
```diff
@@ -468,31 +469,143 @@ def trsh_scan_job(label, scan_res, scan, species_scan_lists, methods):
...
+                        f'internal coordinate {ic} is given.')
+            methods[1] = [ic for ic in methods[1] if len(ic) == 4]
+            # 2. Treat dihedral shares the same pivot with the scanning rotor first
+            #    They cannot be frozen, so they need special treatments
+            pruning = []
+            same_pivot = []
+            # Find the one shares three common atoms with the scanning one
+            for ic in methods[1]:
+                if ic[1:] == scan[1:]:
+                    # scan = [1, 2, 3, 4],  ic = [5, 2, 3, 4], freeze[5, 1, 2, 3] to avoid flip
```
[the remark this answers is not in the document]
xiaoruiDong: In a real scan, `[1, 2, 3, 4]` does not change equally as `[5, 2, 3, 4]`. In most cases, the changes are quite close. However, if one of the pivot atom is a radical site, the difference can be quite huge yielding non-smoothness in the PES. However, we cannot freeze say `[5, 2, 3, 4]`, since we are rotating `[1, 2, 3, 4]`, otherwise `5` and `1` will overlap at some point and fail the scan job. So, I have to come up with a special treatment.

arc/job/trsh.py
```diff
@@ -468,31 +469,143 @@ def trsh_scan_job(label, scan_res, scan, species_scan_lists, methods):
...
+                    to_freeze.append(scan[1:] + [ic[-1]])
+                    # remove all [*, 4, 3, 5] or [4, 3, 5, *] or [*, 5, 3, 4] or [5, 3, 4, *]
+                    pruning += [dihed for dihed in methods[1]
+                                if (is_same_pivot(dihed, scan)
+                                    or is_same_sequence_sublist([ic[-1], scan[-2], scan[-1]], dihed)
+                                    or is_same_sequence_sublist([scan[-1], scan[-2], ic[-1]], dihed))]
+                    break
+                elif is_same_pivot(ic, scan):
+                    same_pivot.append(ic)
+            else:
```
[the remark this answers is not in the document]
xiaoruiDong: This `else` is for cases where no three-atom-identical torsion is found.  For this case, freeze both tops.

arc/job/trsh.py
```diff
@@ -978,57 +1097,184 @@ def scan_quality_check(label, pivots, energies, scan_res=rotor_scan_resolution,
...
+            changed_ics = change_sum[change_sum == True].index.to_list()
+            if changed_ics:
+                invalidate = True
+                changed_ic_dict.update({i: changed_ics})
+
+        # 2.1 Check lowest conformation
+        bond_break = []
+        # Exclude those with boken bonds (different species)
+        for ics in changed_ic_dict.values():
+            any_bond_break = [ic for ic in ics if 'R' in ic]
```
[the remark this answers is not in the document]
xiaoruiDong: Identified problematic internal coordinates are labeled as `R(X,Y)`, `A(X,Y,Z)` and `D(W,X,Y,Z)`. For all the cases I've ever seen, `R(X,Y)` always refers to a bond in a stable molecule (though TS may be tricky). So, if any bond 'R(X,Y)' are found changing significantly, the bond is almost certain to be broken.

xiaoruiDong:
@alongd, I finally corrected all the things you mentioned. Thank you for being so patient to look at my codes. I tried to make the followups clean. I guess most fixup commits are fine and clear, but the ones for trsh.py and schedular.py might be messy. I tried to add as many comments as I can, so for those who are not familiar with this part can understand the workflow. I also planned to add some comments to help you review. BTW, I also cherry-picked one of your commit in another branch, if it is not okay, please let me know. Regarding the test you have done, that is simply because the line of 'scan' has an extra `'\n'` since about 3 months ago, and I am not aware of that. By removing the starting `'\n'` in the `scan_trsh`, I think everything should work perfectly fine.

arc/parser.py
```diff
@@ -265,6 +265,40 @@ def parse_1d_scan_energies(path):
     return energies, angles
 
 
+def parse_1d_scan_coords(path):
```
xiaoruiDong: I  cherry-picked your commits so that I can directly get the conformer xyz instead of using set_dihedrals() to change the current xyz.

arc/job/trsh.py
```diff
@@ -626,6 +615,69 @@ def trsh_scan_job(label: str,
     return scan_trsh, scan_res
 
 
+def trsh_special_rotor(special_rotor: list,
+                       problematic_ic: list,
+                       special_type: str = 'scan',
+                       ) -> list:
+    """
+    Troubleshoot special rotor cases given all problematic torsional internal
+    coordinates. Special rotors include rotor to be scanned and rotor already frozen.
```
xiaoruiDong: If this is not clear, this method is not to troubleshoot 'the rotor to be scanned' or the 'rotor already frozen, it is to troubleshoot rotors which are relevant to these special rotors. Example: `scan = [1, 2, 3, 4]`, however, `scan_quality_check()` finds `[1, 2, 3, 5]` is problematic. We cannot freeze `[1, 2, 3, 5]` which will for sure lead to issues. By using this function, it will figure out which internal coordinates we need to actually freeze to solve the problem.

arc/job/trsh.py
```diff
@@ -25,14 +25,16 @@
                           inconsistency_az,
                           list_available_nodes_command,
                           maximum_barrier,
+                          preserve_param_in_scan_stable,
```
xiaoruiDong: Named something similar to what you used in another branch

arc/job/trsh.py
```diff
@@ -1147,23 +1149,22 @@ def scan_quality_check(label: str,
...
+            - Error or warning message.
+            - Troubleshooting methods to apply, including conformational changes. 
 
     Todo:
         - adjust to ND
     """
     message, invalidation_reason = '', ''
     invalidate = False
-    actions = list()
+    actions = dict()
```
xiaoruiDong: Changed to `dict` as you suggested

arc/job/trsh.py
```diff
@@ -1176,122 +1177,117 @@ def scan_quality_check(label: str,
         else:
             conformer_based_check = True
 
-    if conformer_based_check:
-        # 1. Check based on intermediate conformers
```
xiaoruiDong: Moved to `settings.py`

arc/job/trsh.py
```diff
@@ -1176,122 +1177,117 @@ def scan_quality_check(label: str,
...
-        # Threshold used for identify changes for bonds, angles and dihedrals
-        # Can be moved to settings.py later
-        th_bond = 0.1
-        th_angle = 10
-        th_dihed = 20
```
xiaoruiDong: Moved `settings.py`

arc/job/trsh.py
```diff
@@ -1176,122 +1177,117 @@ def scan_quality_check(label: str,
...
-                    if dihed_label in pruning:
-                        continue
-                    else:
-                        # E.g., deg_abs = [([1,2,3,4], 10.0), ....]
-                        deg_abs.append((scan_conformers['atoms'][dihed_label], scan_conformers[min_index][dihed_label]))
-                        pruning += [d_l for d_l in changed_dihed_label
-                                    if is_same_pivot(deg_abs[-1][0], scan_conformers['atoms'][d_l])]
-                actions = ['change conformer', deg_abs]
+                conf_xyzs = parse_1d_scan_coords(log_file)
+                actions = {'change conformer': conf_xyzs[min_index]}
```
xiaoruiDong: New behavior: using the xyz of the conformer.

arc/job/trsh.py
```diff
@@ -1312,18 +1309,30 @@ def scan_quality_check(label: str,
...
+                                      f'{inconsistency_ab * max(energies):.2f} kJ/mol'
+                message = f'Rotor scan of {label} between pivots {pivots} is inconsistent by' \
+                          f'more than {inconsistency_ab * max(energies):.2f} kJ/mol between ' \
+                          f'two consecutive points. ARC will attempt to troubleshoot this rotor scan.'
                 logger.error(message)
-                if ['inc_res'] not in used_methods:
-                    actions = ['inc_res']
-                elif ['inc_res', 'all_freeze'] not in used_methods:
-                    actions = ['inc_res', 'all_freeze']
+                # Propose a method
```
xiaoruiDong: `Used methods` are now changed to a list of `dict`s. The checking is a little bit complicated. We can simplify it to `actions = {'inc_res': None, 'freeze': 'all'}` in the future.

arc/job/trsh.py
```diff
@@ -1312,18 +1309,30 @@ def scan_quality_check(label: str,
...
+                    if not any([used_method['scan_trsh'] != '' for used_method in used_methods]):
+                        # Indicating freezing torisions is not applied
+                        actions = {'inc_res': None, 'freeze': 'all'}
+                    else:
+                        # Since all torsions are frozen, there's not much we can do except increasing
+                        # scan resolution. But it is not that effective either. So stop and do nothing. 
+                        pass
+                else:
+                    # Indicating neither increasing scan resolution nor freezing torisions is applied
+                    actions = {'inc_res': None}
```
xiaoruiDong: `Used methods` are now changed to a list of `dict`s. The checking is a little bit complicated. We can simplify it to `actions = {'inc_res': None, 'freeze': 'all'}` in the future.

arc/job/trsh.py
```diff
@@ -1332,13 +1341,12 @@ def scan_quality_check(label: str,
...
-            # Find the rotation dihedral in degrees to the closest minimum:
+            # Find the lowest conformer, and use the new conformer for further jobs.
+            # Since at this point, the scan has passed previous checks, the possibility
+            # to switch to a non-isomorphic conformer is low.
             min_index = np.argmin(energies)
-            deg_increment = min_index * scan_res
-            actions = ['change conformer', pivots, deg_increment]
-            if actions in used_methods:
-                logger.error(f'Not troubleshooting a rotor with the same method: {actions}')
-                actions = list()
```
xiaoruiDong: The check will be in `troubleshoot_scan_job`, which checks if the new conformer is used in the previous troubleshooting. I think this is more robust than check if this torsion has been rotated by XX degrees (equal `deg_increment` can still result in different conformers in different cases, if only the starting xyzs are not the same.

alongd:
@xiaoruiDong, I tried the branch out on some example, it seems to work very nicely, I'll uploade additional comments as I explore the output more.

Could you take a look at the failing tests?

Could you also add line breaks in the lengthy messages to make them readable, e.g.:
"Error: Rotor scan of s198 between pivots [4, 5] is inconsistent between two consecutive conformers. Between conformer # 6 and # 7, internal coordinates: D1, D2 changed significantly. Between conformer # 16 and # 17, internal coordinates: D7, D8, D9 changed significantly. ARC will attempt to troubleshoot this rotor scan."

alongd:
Also, we need to summarize the troubleshooting methods used per rotor so it is outputted for each species in the thermo library. Currently, we have the following description, for example:
"pivots: [1, 2], dihedral: [6, 1, 2, 3], rotor symmetry: 3, max scan energy: 7.34 kJ/mol".
The rotor description is set here: https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/statmech/arkane.py#L327

xiaoruiDong:
> @xiaoruiDong, I tried the branch out on some example, it seems to work very nicely, I'll upload additional comments as I explore the output more.
> 
> Could you take a look at the failing tests?
> 
> Could you also add line breaks in the lengthy messages to make them readable, e.g.:
> "Error: Rotor scan of s198 between pivots [4, 5] is inconsistent between two consecutive conformers. Between conformer # 6 and # 7, internal coordinates: D1, D2 changed significantly. Between conformer # 16 and # 17, internal coordinates: D7, D8, D9 changed significantly. ARC will attempt to troubleshoot this rotor scan."

Thank you for letting me aware of the test issues. It is an easy fix due to my recent change to the sequence of the input arguments.

For the error message, do you like the following pattern?
```
ERROR:arc:Rotor scan of  between pivots [2, 3] is inconsistent between two consecutive conformers.
Inconsistent consecutive conformers and problematicinternal coordinates:
conformer #  8/#  9        D21, D22, D23
conformer # 18/# 19        D21, D22, D23
conformer # 39/# 40        D21, D22, D23
ARC will attempt to troubleshoot this rotor scan.
```

arc/parser.py
alongd: no problem

arc/job/trsh.py
alongd: I see. Can you give this example in the docstring?

arc/job/trsh.py
alongd: Would you like to either make this change or add a `# Todo: ...` comment in the code?

arc/statmech/arkane.py
```diff
@@ -323,10 +323,20 @@ def generate_arkane_species_file(self,
...
                         max_e = None
+                    scan_trsh = ''
+                    scan_res = 360
+                    if 'trsh_methods' in species.rotors_dict[i]:
+                        for scan_trsh_method in species.rotors_dict[i]['trsh_methods']:
+                            if 'scan_trsh' in scan_trsh_method and len(scan_trsh) < len(scan_trsh_method['scan_trsh']):
+                                scan_trsh = scan_trsh_method['scan_trsh']
+                            if scan_res > scan_trsh_method['scan_res']:
+                                scan_res = scan_trsh_method['scan_res']
+                    scan_trsh = f'Troubleshot with the following constraints and {scan_res} degrees ' \
```
alongd, line 334: I think this line should be further indented to be under the `if` condition above

xiaoruiDong: Correct!!

arc/job/trsh.py
alongd: was this comment addressed?

arc/job/trsh.py
alongd: was this comment addressed?
(BTW, no need to write ""for developers" in docstings, all docstrings are for developers)

alongd:
@xiaoruiDong, what's the status of this PR? I'd like to add TS rotors, but waiting for this to be merged first.

xiaoruiDong:
@alongd Sorry for the delay. Actually, I don't quite get what you suggest for the two non-addressed comments.. 

For the first one, do you want me to give an example? something like: already frozen [1,2,3,4], but [1,2,3,5] flip? 
For the second one, you mentioned adding  'TODO:', but there isn't any to-do, I think. 

I will clean everything up, as soon as the above addressed. Thanks!

alongd:
No problem.

1. Regarding the "todo" comment (line 1344 in trsh.py):
You wrote that we can simplify the checking ("`Used methods` are now changed to a list of `dict`s. The checking is a little bit complicated. We can simplify it to `actions = {'inc_res': None, 'freeze': 'all'}` in the future.")
I suggested we either simplify this in this PR as you suggested, otherwise if you think this should be done in the future, add a "Todo" comment in that line to document your intentions.

2. Regarding the example comment (line 626 in trsh.py):
Yes, if you could add to the function docsting something like the below, it would be great.
"For example: If ``scan = [1, 2, 3, 4]``, ``scan_quality_check()`` might find ``[1, 2, 3, 5]`` which is problematic (we shouldn't freeze ``[1, 2, 3, 5]`` in such case). By using this function, ARC can figure out which internal coordinates we need to actually freeze to troubleshoot the scan."

alongd:
Can you take a look at the Codacy report for the minor stuff?
https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?bid=16680380&prid=5069484&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&

arc/job/trsh.py
xiaoruiDong: Examples are included in the docstring. Thanks.

arc/job/trsh.py
xiaoruiDong: TODO comment is added and 'developer's are removed. Thanks

xiaoruiDong:
@alongd  check is finished. Thanks

arc/parser.py
```diff
@@ -737,3 +774,267 @@ def process_conformers_file(conformers_path):
...
+        ic_info_block = parse_str_blocks(file_path, 'Initial Parameters', '-----------', regex=False,
+                                         tail_count=3)[0][5:-1]
+        for line in ic_info_block:
+            # Line example with split() indices:
+            # 0 1     2                        3              4         5       6            7
+            # ! R1    R(1, 2)                  1.3581         calculate D2E/DX2 analytically !
+            terms = line.split()
+            ic_dict['label'].append(terms[1])
+            ic_dict['type'].append(terms[1][0])  # 'R: bond, A: angle, D: dihedral
+            atom_inds = re.split(r'\(|\)|,', terms[2])[1:-1]
```
alongd, line 932: could `'\(|\)|,'` be replaced with `'[(),]'`?

arc/scheduler.py
```diff
@@ -2165,14 +2178,21 @@ def check_scan_job(self, label, job):
...
         """
-        # If the job has not converged, troubleshoot
-        if job.job_status[1]['status'] != 'done':
+        # If the job has not converged, troubleshoot ESS
+        # Besides, according to the experience, 'Internal coordinate error' cannot be handled by
+        # troubleshoot_ess() for scan jobs. It is usually related to bond or angle changes which
+        # causes the internal coordinates mess up in the middway of the scan. It can be resolved
+        # by conformer based scan troubleshooting method and its energies are readable.
+        if job.job_status[1]['status'] != 'done' \
+            and job.job_status[1]['error'] != 'Internal coordinate error':
```
alongd, line 2187: this line should have an additional indentation

arc/scheduler.py
```diff
@@ -2497,67 +2528,120 @@ def troubleshoot_negative_freq(self, label, job):
                 self.run_job(label=label, xyz=xyz, level_of_theory=self.conformer_level, job_type='conformer',
                              conformer=i)
 
-    def troubleshoot_scan_job(self, job, methods=None):
+    def troubleshoot_scan_job(self,
+                              job: Job,
+                              methods: Optional[dict]=None,
```
alongd, line 2533: minor: add spaces around `=`

arc/scheduler.py
```diff
@@ -2497,67 +2528,120 @@ def troubleshoot_negative_freq(self, label, job):
...
+            # If not succeed, we are in a situation that we find a lower conformer, but either
+            # this is a incorrect conformer or we have applied this troubleshooting before, but it
+            # didn't yield a good result.
+            self.delete_all_species_jobs(label)
+
+            new_xyz = methods['change conformer']
+            # Check if the same conformer is used in previous troubleshooting
+            for used_trsh_method in used_trsh_methods:
+                if 'change conformer' in used_trsh_method \
+                    and compare_confs(new_xyz, used_trsh_method['change conformer']):
```
alongd, line 2576: add indentation

arc/scheduler.py
```diff
@@ -2497,67 +2528,120 @@ def troubleshoot_negative_freq(self, label, job):
...
+                    # Find we have used this conformer for troubleshooting. Invalid the troubleshooting.
+                        logger.error(f'The change conformer method for {label} is invalid. ' \
```
alongd, line 2578: remove one indent

arc/scheduler.py
```diff
@@ -2497,67 +2528,120 @@ def troubleshoot_negative_freq(self, label, job):
...
+            except InputError as e:
+                logger.debug(f'Got invalid input for trsh_scan_job: {e}\nJob info:\n{job}')
             else:
-                species_scan_lists = [rotor_dict['scan'] for rotor_dict in self.species_dict[label].rotors_dict.values()]
-                scan_trsh, scan_res = trsh_scan_job(label=label, scan_res=job.scan_res, scan=job.scan,
-                                                    species_scan_lists=species_scan_lists, methods=methods)
-                self.run_job(label=label, xyz=job.xyz, level_of_theory=job.job_level_of_theory_dict, job_type='scan',
-                             scan=job.scan, pivots=job.pivots, scan_trsh=scan_trsh, scan_res=scan_res)
+                if scan_trsh or job.scan_res != scan_res \
+                    and {'scan_trsh': scan_trsh, 'scan_res': scan_res} not in used_trsh_methods:
```
alongd, line 2635: indentation

arc/scheduler.py
```diff
@@ -2497,67 +2528,120 @@ def troubleshoot_negative_freq(self, label, job):
...
+                    # Valid troubleshooting method for freezing or increasing resolution
+                    trsh_success = True
+                    actual_actions = {'scan_trsh': scan_trsh, 'scan_res': scan_res}
+                    self.run_job(label=label, xyz=job.xyz,
+                                 level_of_theory=job.job_level_of_theory_dict,
+                                 job_type='scan', scan=job.scan, pivots=job.pivots,
+                                 scan_trsh=scan_trsh, scan_res=scan_res)
+        return trsh_success, actual_actions
+
```
alongd, line 2644: please remove extra line break

arc/scheduler.py
```diff
@@ -2497,67 +2528,120 @@ def troubleshoot_negative_freq(self, label, job):
...
+        # Read used troubleshooting methods
+        for rotor in self.species_dict[label].rotors_dict.values():
+            if rotor['scan'] == job.scan:
+                used_trsh_methods = rotor['trsh_methods']
+                break
+
+        # A lower conformation was found
+        if 'change conformer' in methods:
+            # We will delete all of the jobs no matter we can successfully change to the conformer.
+            # If succeed, we have to cancel jobs to avoid conflictions
```
alongd, line 2566: typo: "conflicts" instead of "conflictions"

arc/common.py
```diff
@@ -1092,3 +1094,89 @@ def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
...
+
+    Returns:
+        pd.DataFrame: a DataFrame consisting of ``True``/``False``, indicating
+                      which torsions changed significantly. ``True`` for significant
+                      change.
+    """
+    # First iteration without 180/-180 adjustment
+    change = (torsions[index_1] - torsions[index_2] - delta).abs() > threshold
+    # Apply 180/-180 adjustment to those shown significance
+    for label in change[change == True].index:
```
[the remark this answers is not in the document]
xiaoruiDong: Actually, there are multiple steps. change is a table. The indexes are internal coordinates (IC) label, the value are `True` or `False`. 
`change[change == True]` will find all entries whose values are `True`
`change[change == True].index` is a list providing the IC labels of all these `True`s

arc/common.py
```diff
@@ -1092,3 +1094,89 @@ def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
...
+                                   E.g.,for the torsions to be scanned, the
+                                   differences are equal to the scan resolution.
+
+    Returns:
+        pd.DataFrame: a DataFrame consisting of ``True``/``False``, indicating
+                      which torsions changed significantly. ``True`` for significant
+                      change.
+    """
+    # First iteration without 180/-180 adjustment
+    change = (torsions[index_1] - torsions[index_2] - delta).abs() > threshold
```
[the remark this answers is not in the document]
xiaoruiDong: > If index_1 and index_2 could be strings (according to the typing above), this would crash
Actually `torsions` is not a list but a DataFrame that can take 'str' or 'value' depending on what the column name. Other words, if we define the IC column names as numbers, then indexes should be numbers, and if we define them as `str`s, then indexes should be `str`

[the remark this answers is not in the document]
xiaoruiDong: > I think the - delta should be outside the (...).abs()
or perhaps do: change = (torsions[index_1] - torsions[index_2]).abs() > delta + threshold

I think about it once, and I think it is better to include inside. Let me give you an example:
say [1,2,3,4] is being scanned at `step = 8` (which makes our `delta = 8`). I set default threshold as `20`. So, if delta is inside the parentheses, it means: 
```
abs(new_dihed - old_dihed - 8) > 20
# equivalent to
old_dihed - 12 < new_dihed < old_dihed + 28
```
otherwise, the boundary will be +- 28, which is less constrained. Considering the dihed is increasing, I think it is not a bad idea to constrain both sides differently.

arc/common.py
```diff
@@ -1092,3 +1094,89 @@ def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
...
+    # Apply 180/-180 adjustment to those shown significance
+    for label in change[change == True].index:
+        # a -180 / 180 flip causes different sign
+        if torsions.loc[label, index_1] * torsions.loc[label, index_2] < 0:
+            if torsions.loc[label, index_1] < 0 \
+               and abs(torsions.loc[label, index_1] + 360 - torsions.loc[label, index_2] - delta) < threshold:
+                change[label] = False
+            elif torsions.loc[label, index_2] < 0 \
+                 and abs(torsions.loc[label, index_1] - 360 - torsions.loc[label, index_2] - delta) < threshold:
```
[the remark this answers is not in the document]
xiaoruiDong: okay

arc/common.py
```diff
@@ -18,8 +18,10 @@
 import time
 import warnings
 import yaml
+from typing import Optional, Union
```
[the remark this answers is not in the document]
xiaoruiDong: I will

arc/job/trsh.py
```diff
@@ -501,49 +516,208 @@ def trsh_negative_freq(label: str,
...
+    3. increasing the scan resolution
 
     Args:
         label (str): The species label.
-        scan_res (int): The scan resolution in degrees.
+        scan_res (int or float): The scan resolution in degrees.
         scan (list): The four atom indices representing the torsion to be troubleshooted.
-        species_scan_lists (list): Entries are lists of four atom indices each representing a torsion.
-        methods (list): The troubleshooting method/s to try. Accepted values: 'freeze' and/or 'inc_res'.
+        scan_lists (list): Entries are the four-atom scan lists (1-indexed) of all torsions
```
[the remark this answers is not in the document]
xiaoruiDong: You are right

arc/job/trsh.py
```diff
@@ -501,49 +516,208 @@ def trsh_negative_freq(label: str,
...
+                        to_freeze_tmp = trsh_special_rotor(special_rotor=frozen_ic,
+                                                           problematic_ic=problematic_ic,
+                                                           special_type='frozen')
+                        to_freeze += to_freeze_tmp
+
+                # 1.4 For other dihedrals, if encountering torsions with the same pivots
+                #     just freeze the one first seen.
+                pruning = []
+                for torsion in problematic_ic:
+                    if torsion in pruning \
```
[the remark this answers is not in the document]
xiaoruiDong: Thank you !

arc/job/trsh.py
```diff
@@ -1000,89 +1174,242 @@ def scan_quality_check(label: str,
...
+            logger.warning(message)
+        else:
+            conformer_based_check = True
+
+    # 1. Check based on intermediate conformers
+    if conformer_based_check:
+        bonds = scan_conformers[scan_conformers['type'] == 'R']
+        angles = scan_conformers[scan_conformers['type'] == 'A']
+        non_scan_rotor = scan_conformers[(scan_conformers['type'] == 'D') \
+                                         & (scan_conformers['scan'] == False)]
```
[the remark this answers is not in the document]
xiaoruiDong: `scan_conformers` is similar to `torsions` mentioned above. Here, I am applying filters to scan_conformers. Let me briefly explain this line:

`non_scan_rotor = scan_conformers[(scan_conformers['type'] == 'D') & (scan_conformers['scan'] == False)]`

`scan_conformers` have many columns. This line filters based on two criteria: `'type'` and `'scan'`. Since `scan_conformers` has multiple rows, we cannot do `is` operation. For DataFrame, `scan_conformers['scan'] == False` is similar to the operation for np.NDArrays where each element will be compared. Besides, we should use `&` instead of using `and`, which I think is related to how DataFrame parses arguments.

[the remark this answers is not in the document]
xiaoruiDong: See the explanation above. Thanks.

arc/job/trsh.py
```diff
@@ -1000,89 +1174,242 @@ def scan_quality_check(label: str,
...
+        # Use tighter threshold to find lower conformer
+        if energy_diff >= 0.5 or energy_diff > 0.5 * (max(energies) - min(energies)):
+            invalidate = True
+            invalidation_reason = f'Another conformer for {label} exists which is ' \
+                                    f'{energy_diff:.2f} kJ/mol lower.'
+            message = f'Species {label} is not oriented correctly around pivots {pivots}, ' \
+                        f'searching for a better conformation...'
+            logger.info(message)
+            # Find the dihedrals in degrees of the lowest conformer:
+            min_index = np.argmin(energies)
```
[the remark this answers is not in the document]
xiaoruiDong: Let me describe what will be stored in the `actions`. 
1. I slightly modified when we save these `actions`, and now it is saved right after we find a troubleshooting strategy and ready to launch another job, (after `troubleshoot_scan_job()`). It used to be saved right after the `scan_quality_check()`. Although the thing generated from `scan_quality_check()` is also called `actions`, they are like raw actions, which will be tailored in the `troubleshoot_scan_job()`.

2.`actions` is a`dict` now, which will only have three kind of keys: `change conformer`, `scan_res` and `scan_trsh`. The value of `change conformer` is ARC's xyz dict. There is no `array`s. The `scan_res` will be either `float` or `int`, and `scan_trsh` is a `str`, they are just like what we had in the past.

arc/job/trsh.py
```diff
@@ -501,49 +516,208 @@ def trsh_negative_freq(label: str,
...
+                for torsion in problematic_ic:
+                    if torsion in pruning \
+                        or torsion in to_freeze \
+                        or torsion[::-1] in to_freeze:
+                            continue
+                    to_freeze.append(torsion)
+                    pruning += [ic for ic in problematic_ic if is_same_pivot(torsion, ic)]
+        
+        # 2. Freeze all torsions other than the rotor to be scanned
+        if methods['freeze'] == 'all':
```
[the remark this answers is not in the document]
xiaoruiDong: To make this function robust, I add an extra check in the `1.Freeze specific internal coordinates identified by scan_quality_check()` to see if we can successfully parse the scan arguments (scanned IC, frozen ICs, steps, etc). 
If succeed, `methods['freeze']` will be remained as a list. If not succeed, `methods['freeze']` will be changed to `'all'`, so that we can still do some sort of troubleshooting.

arc/job/trsh.py
```diff
@@ -1000,89 +1174,242 @@ def scan_quality_check(label: str,
...
+            message = f'Rotor scan quality check using conformer internal coordinates ' \
+                      f'has not been implemented for current ESS. Using PES curve based ' \
+                      f'check for rotor scan of {label} between pivots {pivots}.'
+            logger.warning(message)
+        else:
+            conformer_based_check = True
+
+    # 1. Check based on intermediate conformers
+    if conformer_based_check:
+        bonds = scan_conformers[scan_conformers['type'] == 'R']
```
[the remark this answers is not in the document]
xiaoruiDong: Actually, it is defined at line 1217. before the `try ... except ...`

arc/parser.py
```diff
@@ -281,6 +284,40 @@ def parse_1d_scan_energies(path):
     return energies, angles
 
 
+def parse_1d_scan_coords(path):
```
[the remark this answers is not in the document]
xiaoruiDong: If I understand correctly, `parse_1d_scan_coords()` returns xyz for optimized conformers, and 'parse_trajectory()` returns both optimized and non-optimized conformers. They can be merged, with an optional argument like 'optimized'. But I would suggest that we do it in another PR. I cherrypick `parse_1d_scan_coords()` from your commits and I just keep them intact. A change here may also increase your rebase and workload in TS rotor part.

arc/parser.py
```diff
@@ -737,3 +774,267 @@ def process_conformers_file(conformers_path):
...
+    """
+    log = ess_factory(fullpath=file_path)
+    scan_args = {'scan': None, 'freeze': [],
+                 'step': 0, 'step_size': 0, 'n_atom': 0}
+    if isinstance(log, GaussianLog):
+        try:
+            # g09, g16
+            scan_blk = parse_str_blocks(file_path, 'The following ModRedundant input section has been read:',
+                                        'Isotopes and Nuclear Properties', regex=False)[0][1:-1]
+        except:
```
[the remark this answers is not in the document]
xiaoruiDong: Yes, I agree it should be a more specific exception.

arc/parser.py
```diff
@@ -737,3 +774,267 @@ def process_conformers_file(conformers_path):
...
+                                  f'files, got {log}')
+    # Extract IC values for each conformer
+    conformers = []
+    for ind, ic_blk in enumerate(ic_blks):
+        ics = parse_ic_values(ic_blk[5:-1], software)
+        ics.rename(columns={'value': ind}, inplace=True)
+        conformers.append(ics)
+    # Concatenate ICs of conformers to a single table and remove redundant ICs
+    scan_conformers = pd.concat([scan_ic_info] + conformers, axis=1)
+    red_ind = scan_conformers[scan_conformers.redundant == True].index
```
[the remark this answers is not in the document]
xiaoruiDong: See explanation above

arc/parser.py
xiaoruiDong: Yes, your suggestion is clearer.

arc/scheduler.py
xiaoruiDong: It seems like my autopep8 using another rule, lol.

xiaoruiDong:
@alongd Alom the Rotor PR is updated. one of the checks is not passed because Travis dead. Is there a way to restart the check?

arc/common.py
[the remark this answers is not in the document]
alongd: Thanks for the explanation, I agree.
Is it possible that in a case like `abs(new_dihed - old_dihed - 8) > 20`, `new_dihed` will be lower than `old_dihed`? For example, if `old_dihed = 355` and `new_dihed = 355 + 8 = -3`?
Alternatively, if the user sets the dihedral increment in ARC's settings to be negative (users might want to scan a specific molecule in "reverse"?), then a situation where `new_dihed = 50` and `old_dihed = 65` is problematic as well: it should be within the 20 degree tolerance, but this calculation will determine otherwise.

arc/job/trsh.py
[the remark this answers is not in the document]
alongd: Sorry, I still didn't see where `scan_conformers` is defined. Line 1217 has:
```python
    conformer_based_check = False
```

arc/job/trsh.py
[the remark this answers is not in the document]
alongd: Thanks for the explanation!

alongd:
I restarted the test, they all passed.
I added one comment re the definition of `scan_conformers`, please take a look

arc/common.py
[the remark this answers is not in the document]
xiaoruiDong: > Is it possible that in a case like abs(new_dihed - old_dihed - 8) > 20, new_dihed will be lower than old_dihed? For example, if old_dihed = 355 and new_dihed = 355 + 8 = -3?

For this case, the block below `if torsions.loc[label, index_1] * torsions.loc[label, index_2] < 0:` is aimed to solve this problem. If any has a sign flipping due to -180 <=> 180 transition, this function can handle them.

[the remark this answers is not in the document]
xiaoruiDong: > Alternatively, if the user sets the dihedral increment in ARC's settings to be negative (users might want to scan a specific molecule in "reverse"?), then a situation where new_dihed = 50 and old_dihed = 65 is problematic as well: it should be within the 20 degree tolerance, but this calculation will determine otherwise.

Actually, `delta` is set as the same as `scan_res`, so if people do the reverse scan, they need to set `scan_res` negative, and `delta` will be negative automatically. Back to the example,
```
old_dihed = 65
step_size = -8
new_dihed_expect = 57
new_dihed = 50
```
50 and 57 have a difference of 7 within 20 degrees threshold, so this should not be problematic

In our code, if the following inequality is True, `abs(new_dihed - old_dihed - delta) > 20`, then we think the change is too huge and problematic. The inequality is equivalent to
```
new_dihed < old_dihed + 20 + delta
new_dihed > old_dihed - 20  + delta
# Plug-in our numbers
new_dihed < old_dihed + 20 + delta
new_dihed > old_dihed - 20  + delta
=>
new_dihed < old_dihed + 12
new_dihed > old_dihed - 28
=>
new_dihed < 65 + 20 - 8 = 77
new_dihed > 65 - 20 - 8 = 37
```
So it also works for negative resolution

arc/job/trsh.py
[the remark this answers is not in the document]
xiaoruiDong: Yep, the logic is if line 1220 `scan_conformers = parse_scan_conformers(log_file)` works okay, then it make `conformer_based_check = True`. Otherwise, `conformer_based_check = False`. For cases of `False` , line 1230 `if conformer_based_check:` will find out and won't go further to line 1231.

xiaoruiDong:
@alongd  I responded to the comments. Please find them out

PR #326, opened 2020-02-28
Minor: Limit the default num of confs to generate

xiaoruiDong:
This commit limits the default number of conformers to generate. The previous default is only limited by energy threshold which can cause hundreds of conformer jobs to run.


alongd:
@xiaoruiDong, can you update the unit test?

xiaoruiDong:
@alongd updated. Thanks for checking the unittest.

Issue #328, opened 2020-02-28
ROO Conformer cannot be generated by RDkit


xiaoruiDong:
With current master dd8582, ROO (molecules with O[O] group, e.g., CO[O]) conformers cannot be generated, (and further jobs will be just canceled). While I remember it is not a problem in the previous version. I trace down the functions and figure out that `Chem.AllChem.MMFFOptimizeMolecule` from RDkit cannot optimize these molecules.

A short example:
![image](https://user-images.githubusercontent.com/35771233/75582465-a2d93100-5a39-11ea-8fea-67bdba6f2f2f.png)
where RETURNS: 0 if the optimization converged, -1 if the forcefield could
not be set up, 1 if more iterations are required.
 
Although there are options `try_ob` to try the Openbabel package, but ARC's default process does not try Openbabel (`try_ob=False`). I will try `try_ob=True` and see if it works.

xiaoruiDong:
#306 related

Issue #306, opened 2020-02-16
Troubleshoot conformer generation for radicals


alongd:
Some species cannot be embedded, e.g.: `COCO[O]`
If they happen to be radicals, generate conformers for the RH species, then remove the corresponding H atom from the coordinates

alongd:
Same behaviour for `OOCOCC=[N]`

PR #332, opened 2020-03-02
Use OpenBabel force field to optimize the geoms if RDkit fails

xiaoruiDong:
Related to #306 and #328.

RDKit fails to optimize certain types of molecules (e.g., ROO) using force field. And currently, in ARC, Openbabel force field and conformer search functions are no longer used, due to incompatibility, causes ARC cannot handle these molecules. 

This PR allows force field optimization for molecules like ROOs. It uses randomly generated conformers (by RDKit) and optimizes their geometries using OpenBabel. RDKit is used because I didn't find a random conformer generation function in OpenBabel. Moreover, Openbabel optimizer is slower than RDKit's optimizer makes it a fall back option.

The original OpenBabel methods `openbabel_force_field` and `mix_rdkit_and_openbabel_force_field`, in case, if we want to further investigate on them.

I saw openbabel is prohibited in previous PRs, however since it is a method covering more molecules. Though it is way slower (~10x), it is still worth adding as a fallback method.

### DIFF
--- a/arc/species/conformers.py
+++ b/arc/species/conformers.py
@@ -447,7 +447,7 @@ def conformers_combinations_by_lowest_conformer(label, mol, base_xyz, multiple_t
         list: New conformer combinations, entries are conformer dictionaries.
     """
     base_energy = get_force_field_energies(label, mol, num_confs=None, xyz=base_xyz,
-                                           force_field=force_field, optimize=True)[1]
+                                           force_field=force_field, optimize=True, suppress_warning=True)[1]
     if len(base_energy) == 0:
         return list()
     else:
@@ -544,7 +544,7 @@ def generate_all_combinations(label, mol, base_xyz, multiple_tors, multiple_samp
     else:
         # no multiple torsions (all torsions are symmetric or no torsions in the molecule), this is a trivial case
         energy = get_force_field_energies(label, mol, num_confs=None, xyz=base_xyz, force_field=force_field,
-                                          optimize=True)[1][0]
+                                          optimize=True, suppress_warning=True)[1][0]
         new_conformers.append({'index': len_conformers + len(new_conformers),
                                'xyz': base_xyz,
                                'FF energy': energy,
@@ -658,7 +658,8 @@ def change_dihedrals_and_force_field_it(label, mol, xyz, torsions, new_dihedrals
         xyz = converter.str_to_xyz(xyz)
 
     if torsions is None or new_dihedrals is None:
-        xyz, energy = get_force_field_energies(label, mol=mol, xyz=xyz, optimize=True, force_field=force_field)
+        xyz, energy = get_force_field_energies(label, mol=mol, xyz=xyz, optimize=True,
+                                               force_field=force_field, suppress_warning=True)
         return xyz, energy
 
     xyzs, energies = list(), list()
@@ -677,7 +678,7 @@ def change_dihedrals_and_force_field_it(label, mol, xyz, torsions, new_dihedrals
                 xyz_dihedrals = converter.set_rdkit_dihedrals(conf, rd_mol, torsion_0_indexed, deg_abs=dihedral)
         if force_field != 'gromacs':
             xyz_, energy = get_force_field_energies(label, mol=mol, xyz=xyz_dihedrals, optimize=True,
-                                                    force_field=force_field)
+                                                    force_field=force_field, suppress_warning=True)
             if energy and xyz_:
                 energies.append(energy[0])
                 if optimize:
@@ -769,7 +770,7 @@ def determine_number_of_conformers_to_generate(label, heavy_atoms, torsion_num, 
     if mol is None and xyz is not None:
         mol = converter.molecules_from_xyz(xyz)[1]
     if mol is not None and xyz is None:
-        xyzs = get_force_field_energies(label, mol, num_confs=1)[0]
+        xyzs = get_force_field_energies(label, mol, num_confs=1, suppress_warning=True)[0]
         xyz = xyzs[0] if len(xyzs) else None
     if mol is not None and xyz is not None:
         num_chiral_centers = get_number_of_chiral_centers(label, mol, xyz=xyz, just_get_the_number=True)
@@ -1018,7 +1019,14 @@ def get_torsion_angles(label, conformers, torsions):
     return torsion_angles
 
 
-def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='MMFF94s', optimize=True, try_ob=True):
+def get_force_field_energies(label: str,
+                             mol: Molecule,
+                             num_confs: int = None,
+                             xyz: dict = None,
+                             force_field: str = 'MMFF94s',
+                             optimize: bool = True,
+                             try_ob: bool = True,
+                             suppress_warning: bool = False) -> (list, list):
     """
     Determine force field energies using RDKit.
     If ``num_confs`` is given, random 3D geometries will be generated. If xyz is given, it will be directly used instead.
@@ -1032,6 +1040,7 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
         force_field (str, optional): The type of force field to use.
         optimize (bool, optional): Whether to first optimize the conformer using FF. True to optimize.
         try_ob (bool, optional): Whether to try OpenBabel if RDKit fails. ``True`` to try, ``True`` by default.
+        suppress_warning (bool, optional): Wheter to suppress warning of using OpenBabel. ``True`` to suppress, ``False`` by default.
 
     Raises:
         ConformerError: If conformers could not be generated.
@@ -1046,8 +1055,9 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
         rd_mol = embed_rdkit(label, mol, num_confs=num_confs, xyz=xyz)
         xyzs, energies = rdkit_force_field(label, rd_mol, force_field=force_field, optimize=optimize)
     if not len(xyzs) and force_field.lower() in ['gaff', 'mmff94', 'mmff94s', 'uff', 'ghemical'] and try_ob:
-        logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
-                       f'This is often slower, and prohibits ARC from using all features of the conformers module.')
+        if not suppress_warning:
+            logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
+                           f'This is often slower.')
         xyzs, energies = openbabel_force_field_on_rdkit_conformers(
                                             label, rd_mol, force_field=force_field, optimize=optimize)
     if not len(xyzs):
--- a/arc/species/conformers.py
+++ b/arc/species/conformers.py
@@ -1347,18 +1347,15 @@ def read_rdkit_embedded_conformer_i(rd_mol, i, rd_index_map=None):
     return xyz_dict
 
 
-def rdkit_force_field(label, rd_mol, mol=None, force_field='MMFF94s', optimize=True, try_ob=False):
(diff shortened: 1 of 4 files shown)


arc/species/conformers.py
```diff
@@ -1060,6 +1060,53 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
     return xyzs, energies
 
 
+def openbabel_force_field_on_rdkit_conformers(label, rd_mol, force_field='MMFF94s', optimize=True):
+    """
+    Optimize RDKit conformers by OpenBabelusing a force field (MMFF94 or MMFF94s are recommended).
```
alongd: minor: space after OpenBabel

arc/species/conformers.py
```diff
@@ -1060,6 +1060,53 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
...
+    Args:
+        label (str): The species' label.
+        rd_mol (RDKit RDMol): The RDKit molecule with embedded conformers to optimize.
+        force_field (str, optional): The type of force field to use.
+        optimize (bool, optional): Whether to first optimize the conformer using FF. True to optimize.
+
+    Returns:
+        list: Entries are optimized xyz's in a dictionary format.
+    Returns:
+        list: Entries are float numbers representing the energies.
```
alongd: in what units?

arc/species/conformers.py
```diff
@@ -1060,6 +1060,53 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
...
+    # Set up Openbabel input and output format
+    obconversion = ob.OBConversion()
+    obconversion.SetInAndOutFormats('xyz', 'xyz')
+    # Set up Openbabel force field
+    ff = ob.OBForceField.FindForceField(force_field)
+    symbols = [rd_atom.GetSymbol() for rd_atom in rd_mol.GetAtoms()]
+    for i in range(rd_mol.GetNumConformers()):
+        # Convert RDKit conformer to xyz string
+        conf = rd_mol.GetConformer(i)
+        xyz_str = str(conf.GetNumAtoms()) + '\n\n'
```
alongd: please change to `f'{conf.GetNumAtoms()}\n\n'`

arc/species/conformers.py
```diff
@@ -1369,7 +1366,7 @@ def rdkit_force_field(label, rd_mol, mol=None, force_field='MMFF94s', optimize=T
     for i in range(rd_mol.GetNumConformers()):
         if optimize:
             v, j = 1, 0
-            while v and j < 200:
+            while v == 1 and j < 200:  # v == 1: continue, v == 0: enough steps, v == -1: unable to set up
```
alongd: thanks!

arc/species/conformers.py
```diff
@@ -1030,28 +1031,31 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
...
+        xyzs, energies = rdkit_force_field(label, rd_mol, force_field=force_field, optimize=optimize)
     if not len(xyzs) and force_field.lower() in ['gaff', 'mmff94', 'mmff94s', 'uff', 'ghemical'] and try_ob:
-        xyzs, energies = mix_rdkit_and_openbabel_force_field(label,
-                                                             mol,
-                                                             num_confs=num_confs,
-                                                             xyz=xyz,
-                                                             force_field=force_field,
-                                                             try_ob=try_ob)
+        logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
+                       f'This is often slower, and prohibits ARC from using all features of the conformers module.')
```
alongd: is it still prohibiting ARC from using all the features of this module?

arc/species/conformers.py
```diff
@@ -1030,28 +1031,31 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
...
+        xyzs, energies = openbabel_force_field_on_rdkit_conformers(
+                                            label, rd_mol, force_field=force_field, optimize=optimize)
+        # xyzs, energies = mix_rdkit_and_openbabel_force_field(label,
```
alongd: remove?

arc/main.py
```diff
@@ -442,7 +442,7 @@ def from_dict(self, input_dict, project=None, project_directory=None):
         self.memory = input_dict['job_memory'] if 'job_memory' in input_dict else self.memory
         self.bath_gas = input_dict['bath_gas'] if 'bath_gas' in input_dict else None
         self.solvent = input_dict['solvent'] if 'solvent' in input_dict else None
-        self.n_confs = input_dict['n_confs'] if 'n_confs' in input_dict else None
+        self.n_confs = input_dict['n_confs'] if 'n_confs' in input_dict else 10
```
alongd: was this already merged?

arc/species/conformers.py
```diff
@@ -658,7 +658,7 @@ def change_dihedrals_and_force_field_it(label, mol, xyz, torsions, new_dihedrals
         xyz = converter.str_to_xyz(xyz)
 
     if torsions is None or new_dihedrals is None:
-        xyz, energy = get_force_field_energies(label, mol=mol, xyz=xyz, optimize=True, force_field=force_field)
+        xyz, energy = get_force_field_energies(label, mol=mol, xyz=xyz, optimize=True, force_field=force_field, suppress_warning=True)
```
alongd: minor: line too long

arc/species/conformers.py
```diff
@@ -1018,7 +1018,7 @@ def get_torsion_angles(label, conformers, torsions):
     return torsion_angles
 
 
-def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='MMFF94s', optimize=True, try_ob=True):
+def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='MMFF94s', optimize=True, try_ob=True, suppress_warning=False):
```
alongd: line too long

arc/species/conformers.py
```diff
@@ -1046,8 +1047,9 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
         rd_mol = embed_rdkit(label, mol, num_confs=num_confs, xyz=xyz)
         xyzs, energies = rdkit_force_field(label, rd_mol, force_field=force_field, optimize=optimize)
     if not len(xyzs) and force_field.lower() in ['gaff', 'mmff94', 'mmff94s', 'uff', 'ghemical'] and try_ob:
-        logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
-                       f'This is often slower, and prohibits ARC from using all features of the conformers module.')
+        if not suppress_warning:
+            logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
+                        f'This is often slower, and prohibits ARC from using all features of the conformers module.')
```
alongd: indentations don't match

arc/species/conformers.py
```diff
@@ -1046,8 +1047,9 @@ def get_force_field_energies(label, mol, num_confs=None, xyz=None, force_field='
         rd_mol = embed_rdkit(label, mol, num_confs=num_confs, xyz=xyz)
         xyzs, energies = rdkit_force_field(label, rd_mol, force_field=force_field, optimize=optimize)
     if not len(xyzs) and force_field.lower() in ['gaff', 'mmff94', 'mmff94s', 'uff', 'ghemical'] and try_ob:
-        logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
-                       f'This is often slower, and prohibits ARC from using all features of the conformers module.')
+        if not suppress_warning:
+            logger.warning(f'Using OpenBabel instead of RDKit as a fall back method to generate conformers for {label}. '
```
alongd: why would we like to suppress this warning? If it takes that longer, the user should know why, and have an option to turn it off

alongd:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ReactionMechanismGenerator/ARC/pullRequest?prid=5105764&bid=16791171)

xiaoruiDong:
@alongd Thank you for the review, everything is addressed. 

To 'suppress this warning' is to avoid this warning showing too many times since `get_force_field_energies` is called many times in the `generate_conformers`. All warnings generated in the `deduce_nwe_conformers` is now suppressed. It is a disaster when there is a force field free rotor and thousands of combinations are predicted.

image

The warning is now only generated once and looks like above (' and prohibits ARC from using all features of the conformers module.' is removed).

There is a problem with the unit test, which I passed locally, even after I correct `almost_equal_coord_lists`. I am not sure if it is because of the random conformer generation step (highly likely) or different results from optimization.

alongd:
The fixup commit removed `atol` from the test. Could it be related?
Can you update your environment and test locally again?

xiaoruiDong:
@alongd Thanks for the advice, I updated my environment, but nothing changed. I could not find the Openbabel version on Travis (mine is 2.4.1).

Besides, I did some tests and trials and finally narrowed down the issue. The difference is due to the behavior of openbabel optimization function. In the unit test, I tried to print all of the results, and I found that the same input geometry (OBmol object) yields different final geometry after OB optimization with the same number of iteration (2000) using different machines. For each xyz coordinate value, the difference is about O(1e-3). However, the energy difference is trivial (O(1e-9) both atol and rtol). I cannot track further into OpenBabel, and I don't know if things like the compilers is the cause. Maybe, it is okay to not compare the geometry in the unit test? since energy is close enough?

alongd:
I see. You could instead test the energy and just the `xyz['symbols']` part of the geometry.

xiaoruiDong:
Thank you. The unit test is modified.

arc/species/conformers.py
alongd: why would we like to suppress this warning? If it takes that longer, the user should know why, and have an option to turn it off

arc/species/conformersTest.py
```diff
@@ -501,6 +501,49 @@ def test_openbabel_force_field(self):
...
+                                    (-1.71119, 0.70925, -0.32507),
+                                    (-1.2765, -0.99769, -0.58128),
+                                    (1.18277, -0.81419, 0.37478),
+                                    (0.64876, 0.50323, -1.12562),
+                                    (0.59336, 1.16044, 0.53412),
+                                    (2.51642, -0.38757, 0.25036))}]  
+        self.assertEqual(len(energies), 2)
+        self.assertAlmostEqual(energies[0], 0.2446742680965306, 4)
+        self.assertAlmostEqual(energies[1], -0.7460169699282158, 4)
+        self.assertTrue(almost_equal_coords_lists(xyzs, expected_xyzs, atol=1e-5))
```
alongd: I see we are testing the entire xyz, not only the symbols. Did I miss a fix, or does it work now?

xiaoruiDong: It is likely you are reviewing an older commit? The changes to the unit test are in the last commit. Thanks.

alongd:
I see: The commit "BugFix: Test OB FF on RDKit" fixes the commit "Tests: openbabel_force_field_on_rdkit_conformers".
Let's refer to something as a "BugFix" if it fixes a bug on master, and as a "fixup" (to be squashed) if it fixes something within the same PR.
Would you mind organizing (squashing) the commits, taking a look to see there aren't any other comments, and rebasing?

xiaoruiDong:
@alongd Thank you for articulating the usage of 'BugFix'. With no further change, I cleaned all the commits and squashed all of the fixups.

alongd:
Thanks! The PR is in excellent condition.

One last request before we merge: Could you please add a unit test that calls `conformers.generate_conformers()` with a species for which RDKit fails to generate conformers, and show that now they are being generated?

xiaoruiDong:
> Thanks! The PR is in excellent condition.
> 
> One last request before we merge: Could you please add a unit test that calls `conformers.generate_conformers()` with a species for which RDKit fails to generate conformers, and show that now they are being generated?

Unit test is added.

PR #333, opened 2020-03-02
Return a status dictionary from ARC.execute

amarkpayne:
Could be useful for scripting large batches of jobs with the API to quickly identify jobs that need further attention.

The returned dictionary has species labels as keys, and a Boolean where True means that the job converged


alongd:
Thanks @amarkpayne! Could you please rebase?

alongd:
We need to add to the docstring of `execute`
```python

Returns:
    dict: Keys are species labels, values are Booleans representing species convergence.
```

And a returned type hint to the function definition:
```python
def execute(self) -> dict:
```

PR #334, opened 2020-03-03
BugFix: Limit the default num of confs to generate

xiaoruiDong:
This is a follow-up fix for 35b68b. all n_confs are set to 10 as default.

The change in the main is the critical one, while other changes don't really have an impact.


alongd:
Thanks for this fix! Fixing the `from_dict` in main is crucial, otherwise we won't enforce the default when running ARC from an input file.
Do you think we should also change the default of `n_confs` [here](https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/species/conformers.py#L103) as well?

PR #338, opened 2020-03-04
BugFix: Don't re-run a scan job after changing the conformer

alongd:
ARC will now not re-run the same scan (from the wrong conformer) when troubleshooting a scan.


alongd:
Thanks @xiaoruiDong !

PR #339, opened 2020-03-04
If a restart job has the same sp and opt, set the sp path from geometry

### LINKED ISSUE #335: Restart is broken if opt and sp are the same level of theory
**Describe the bug**
I was running an ARC job where the opt and sp level were the same, which crashed for an unrelated reason. When I fixed the issue and restarted the job, I got the following error:

```
Not running an sp job for spcs267 at b3lyp/def2tzvp empiricaldispersion=gd3bj, since the optimization was done at the same level of theory. Using the optimization output for parsing the sp energy.
  File "../../ARC.py", line 73, in <module>
    main()
  File "../../ARC.py", line 67, in main
    arc_object.execute()
  File "/home/ampayne/RMG/ARC/arc/main.py", line 591, in execute
    fine_only=self.fine_only, main_arc_object=self)
  File "/home/ampayne/RMG/ARC/arc/scheduler.py", line 453, in __init__
    self.run_sp_job(species.label)
  File "/home/ampayne/RMG/ARC/arc/scheduler.py", line 1025, in run_sp_job
    for opt_job_name, opt_job in self.job_dict[label]['opt'].items():
KeyError: 'opt'
```

The code in question can be seen below:

https://github.com/ReactionMechanismGenerator/ARC/blob/a3de5883b3d1db8599fbcf7118834f300cd30059/arc/scheduler.py#L1013-L1021

I am pretty sure what is going in is that `opt` is no longer in  `self.job_dict[label]` because this job was already complete when the job restarted, so it never gets added to the job dictionary.

I believe that this line is only needed so that `self.output[label]['paths']['sp']` can be set in `post_sp_actions`. I got around this by setting this with `self.output[label]['paths']['sp'] = self.output[label]['paths']['geo']`. My question is whether or not this is an acceptable solution in general. If so I'll push a commit and PR

amarkpayne:
fixes #335 

I'll test this out to confirm. Hopefully this is a general solution that will work no matter what the situation is.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -989,10 +989,19 @@ class Scheduler(object):
                         f'since the optimization was done at the same level of theory. '
                         f'Using the optimization output for parsing the sp energy.')
             recent_opt_job_name, recent_opt_job = 'opt_a0', None
-            for opt_job_name, opt_job in self.job_dict[label]['opt'].items():
-                if int(opt_job_name.split('_a')[-1]) > int(recent_opt_job_name.split('_a')[-1]):
-                    recent_opt_job_name, recent_opt_job = opt_job_name, opt_job
-            self.post_sp_actions(label=label, job=recent_opt_job)
+            if 'opt' in self.job_dict[label]:
+                for opt_job_name, opt_job in self.job_dict[label]['opt'].items():
+                    if int(opt_job_name.split('_a')[-1]) > int(recent_opt_job_name.split('_a')[-1]):
+                        recent_opt_job_name, recent_opt_job = opt_job_name, opt_job
+                self.post_sp_actions(label=label, sp_path=os.path.join(recent_opt_job.local_path, 'output.out'))
+
+            # If opt is not in the job dictionary, the likely explanation is this job has been restarted
+            elif 'geo' in self.output[label]['paths']:  # Then just use this path directly
+                self.post_sp_actions(label=label, sp_path=self.output[label]['paths']['geo'])
+
+            else:
+                raise RuntimeError(f'Unable to set the path for the sp job for species {label}')
+            
             return
         if 'sp' not in self.job_dict[label]:  # Check whether or not single point jobs have been spawned yet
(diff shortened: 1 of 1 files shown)


alongd:
You are right, in `post_sp_actions`, the `job` argument is only used for setting the path via `job.local_path`.
To make it cleaner, let's change the `post_sp_actions()` argument `job` to `sp_path`, and just pass the path instead of the Job object. In other words, let's only set `self.output[label]['paths']['sp']` in `post_sp_actions` instead of in several places.

This will require changing `self.post_sp_actions(label, job)` to `self.post_sp_actions(label, sp_path=os.path.join(job.local_path, 'output.out'))` under `check_sp_job()` as well.

Do you think this makes sense?

amarkpayne:
I agree, this is much cleaner code. I have force pushed these changes, let me know what you think.

arc/scheduler.py
```diff
@@ -985,10 +985,17 @@ def run_sp_job(self, label):
...
-                if int(opt_job_name.split('_a')[-1]) > int(recent_opt_job_name.split('_a')[-1]):
-                    recent_opt_job_name, recent_opt_job = opt_job_name, opt_job
-            self.post_sp_actions(label=label, job=recent_opt_job)
+            if 'opt' in self.job_dict[label]:
+                for opt_job_name, opt_job in self.job_dict[label]['opt'].items():
+                    if int(opt_job_name.split('_a')[-1]) > int(recent_opt_job_name.split('_a')[-1]):
+                        recent_opt_job_name, recent_opt_job = opt_job_name, opt_job
+                self.post_sp_actions(label=label, sp_path=os.path.join(recent_opt_job.local_path, 'output.out'))
+
+            else:  # This is likely because the job restarted. See if we can set the sp path from the geometry path
```
alongd: Can you turn this into an `elif`? (with a final `else raise`)

amarkpayne: Done!

PR #341, opened 2020-03-05
BugFix: Correctly trsh rotor conformer change

alongd:
This is a continuation to PR #338, where as a result of the code organization the variable `i` was not defined.


alongd:
Thanks.
I agree - I have a scheme in mind, let's meet to discuss it

Issue #344, opened 2020-03-06
Figures are retained open during job


xiaoruiDong:
This is related to plotter.py. When running an ARC job where many figures are to be plotted (e.g., thinking of many species with many rotors), all matplotlib figures are retained in the RAM, and potentially this will influence the performance. I got some warnings like the below one:
```
/home/xiaorui/ARC/arc/plotter.py:993: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  plt.figure(figsize=(4, 3), dpi=120)
```
I will add a PR to solve this.

PR #345, opened 2020-03-06
Minor: explicitly close figures created by plotter

xiaoruiDong:
Related to #344.

Without closing figures may cause us memory issues in large ARC jobs (e.g., many species with many rotors). This PR force to close figures at the end of each plot function. A quick test is also done for plotting in the notebook, and this added close function will not influence showing figures in the notebook.

### DIFF
--- a/arc/plotter.py
+++ b/arc/plotter.py
@@ -200,6 +200,7 @@ def plot_3d_mol_as_scatter(xyz, path=None, plot_h=True, show_plot=True, name='',
             os.makedirs(path)
         image_path = os.path.join(path, f'scattered_balls_structure_{name}.png')
         plt.savefig(image_path, bbox_inches='tight')
+    plt.close(fig=fig)
 
 
 def check_xyz_species_for_drawing(xyz=None, species=None):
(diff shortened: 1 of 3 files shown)


alongd:
Thanks! Does it make figures show up in the notebook before ARC terminates (currently they don't)?

xiaoruiDong:
> Thanks! Does it make figures show up in the notebook before ARC terminates (currently they don't)?

Thanks for reviewing this PR. I don't know what is a good way to test the feature you mentioned. I just did a simple trial (figure below). To me, this seems that the figures are plotted immediately. Please let me know if there is a good way to test. BTW, I also add `%matplotlib inline` in the notebook.

image

xiaoruiDong:
@alongd 
image
This is what I got from running the demos. Add `%matplotlib inline` is necessary to display these figures. I will add another commit.

alongd:
Thanks for the screenshots!
A full test might be to run an ARC job in a notebook with _many_ species with rotors, and see that we no longer get the error message. But that's intensive. Instead, it might be sufficient to just run ARC for two species (with rotors) in a notebook, and see that the scan images are displayed immediately. Could you do the latter?

xiaoruiDong:
@alongd Test is done. Figures can be shown. (example below)

image

alongd:
Thanks again! Could you rebase this branch?

xiaoruiDong:
@alongd Thank you! Rebased.

PR #349, opened 2020-03-09
Add job submission script for Orca on Pharos

amarkpayne:
Supports parallelization across multiple cores now that openmpi is installed on the server


amarkpayne:
@oscarwumit I am tagging you in case you are interested in trying this out in the future.

PR #351, opened 2020-03-12
[WIP] Add TS search capabilities to ARC

alongd:
Added an ABS/Adapter/Factory design pattern for TS search algorithms.
Currently only user guesses and family-based heuristics are implemented.
Added training data for H Abstraction.


alongd:
superseded by #504

Issue #346, opened 2020-03-06
Bug: cannot upload files due to '[]' in the folder name


xiaoruiDong:
I don't know whether this is common. I usually use SMILES as the species label (in the automatically generated input file), and sometimes, if there are similar species, I run into issues submitting files, even if SSH connection is okay. I found it may be related to '[]' in the folder name. Here is an example.

![image](https://user-images.githubusercontent.com/35771233/76111313-0caf8880-5fae-11ea-92e2-609a3956314a.png)
Probably use `mkdir -p "{}"` in ARC whenever add new paths? I will try to fix this issue.

PR #355, opened 2020-03-13
A large rewriting to the SSHClient

xiaoruiDong:
This change is originally to solve #346, but soon becoming a rewriting to the whole class. 

What this PR solves and how it solves:
1. In the original issue #346, ARC cannot deal with path has '[]' in it and identifies this kind of problem as connection failure (which is not). 
Solution: make sure remote dir exists (otherwise create it); make sure `""` used when calling `mkdir` or `cd`; Rewrite upload and download file function (actually twice, no need to carefully review the first time change). Add a connection check before calling upload and download function (more details later).

2. Slightly better security. Personally, I don't think directly calling this `send_command_to_server` in other modules is a good idea. It is not readable enough. And it also provides large flexibility to other modules. I don't know much about hacking, but it is not a bad idea to somehow constraint what we can do with this module.
Solution: make this function private and create functions like `create_dir`, `list_available_nodes`, etc.

3. Better handling ssh and sftp. Previously, we will create multiple paramiko SSH and SFTP Client during using this module. Potential problems: (1) increasing connection load and more sensitive to the stability of the connection. (2) different sessions, for some reason, are not synchronized immediately. (3) repetition of creating and closing SSH and SFTP blocks in most of the methods. 
Solution: (1) add these two private attributes _ssh and _sftp. We will only deal with them. (2) convert SSHClient to a Context Manager like class, so people can use `with ... as ...` to use it. Using this syntax not only results in a shorter code but also has better error handling, making sure the clients are closed, even errors are raised.

4. Connection check. Since we are using a single ssh and sftp client. It is better to check their connectivity before using related functions. And some modules have their own connection check, which sometimes will see other issues as connection problems.
Solution: Add a stand-alone connection check decorator to check the connection before each call. If not well connected, then simply reconnect. If still unsuccess, follow the original approach "sleep and try again".

How to review, some of the early changes are less important. More focus can be put on commits since `Add two private attribute _ssh and _sftp`. Once go through by commits, better have a overview directly look at the files changed.

This module has been tested by running the demo thermo job in the ipython folder. I didn't find any abnormalities. Since it heavily related to server communication and we don't have a test server, unit tests are not included.

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -1593,7 +1593,7 @@ end
                     elif 'terachem' in esss:
                         self.software = 'terachem'
                 elif self.job_type in ['conformer', 'opt', 'freq', 'optfreq', 'sp',
-                                     'directed_scan']:
+                                       'directed_scan']:
                     if self.method == 'hf':
                         if 'gaussian' in esss:
                             self.software = 'gaussian'
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -1229,13 +1229,11 @@ end
 
     def _upload_submit_file(self):
         ssh = SSHClient(self.server)
-        ssh.send_command_to_server(command='mkdir -p {0}'.format(self.remote_path))
         remote_file_path = os.path.join(self.remote_path, submit_filename[servers[self.server]['cluster_soft']])
         ssh.upload_file(remote_file_path=remote_file_path, file_string=self.submit)
 
     def _upload_input_file(self):
         ssh = SSHClient(self.server)
-        ssh.send_command_to_server(command='mkdir -p {0}'.format(self.remote_path))
         if self.input is not None:
             remote_file_path = os.path.join(self.remote_path, input_filename[self.software])
             ssh.upload_file(remote_file_path=remote_file_path, file_string=self.input)
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -1228,38 +1228,38 @@ end
                 self._upload_check_file(local_check_file_path=self.checkfile)
 
     def _upload_submit_file(self):
-        ssh = SSHClient(self.server)
         remote_file_path = os.path.join(self.remote_path, submit_filename[servers[self.server]['cluster_soft']])
-        ssh.upload_file(remote_file_path=remote_file_path, file_string=self.submit)
+        with SSHClient(self.server) as ssh:
+            ssh.upload_file(remote_file_path=remote_file_path, file_string=self.submit)
 
     def _upload_input_file(self):
-        ssh = SSHClient(self.server)
-        if self.input is not None:
-            remote_file_path = os.path.join(self.remote_path, input_filename[self.software])
-            ssh.upload_file(remote_file_path=remote_file_path, file_string=self.input)
-        for up_file in self.additional_files_to_upload:
-            if up_file['source'] == 'path':
-                local_file_path = up_file['local']
-            elif up_file['source'] == 'input_files':
-                local_file_path = input_files[up_file['local']]
-            else:
-                raise JobError('Unclear file source for {0}. Should either be "path" of "input_files", '
-                               'got: {1}'.format(up_file['name'], up_file['source']))
-            ssh.upload_file(remote_file_path=up_file['remote'], local_file_path=local_file_path)
-            if up_file['make_x']:
-                ssh.change_mode(mode='+x', path=up_file['name'], remote_path=self.remote_path)
-        self.initial_time = ssh.get_last_modified_time(
-            remote_file_path=os.path.join(self.remote_path, submit_filename[servers[self.server]['cluster_soft']]))
+        with SSHClient(self.server) as ssh:
+            if self.input is not None:
+                remote_file_path = os.path.join(self.remote_path, input_filename[self.software])
+                ssh.upload_file(remote_file_path=remote_file_path, file_string=self.input)
+            for up_file in self.additional_files_to_upload:
+                if up_file['source'] == 'path':
+                    local_file_path = up_file['local']
+                elif up_file['source'] == 'input_files':
+                    local_file_path = input_files[up_file['local']]
+                else:
+                    raise JobError(f'Unclear file source for {up_file["name"]}. Should either be "path" or '
+                                   f'"input_files", got: {up_file["source"]}')
+                ssh.upload_file(remote_file_path=up_file['remote'], local_file_path=local_file_path)
+                if up_file['make_x']:
+                    ssh.change_mode(mode='+x', path=up_file['name'], remote_path=self.remote_path)
+            self.initial_time = ssh.get_last_modified_time(
+                remote_file_path=os.path.join(self.remote_path, submit_filename[servers[self.server]['cluster_soft']]))
 
     def _upload_check_file(self, local_check_file_path=None):
         if self.server != 'local':
-            ssh = SSHClient(self.server)
             remote_check_file_path = os.path.join(self.remote_path, 'check.chk')
-            local_check_file_path = os.path.join(self.local_path, 'check.chk') if remote_check_file_path is None\
+            local_check_file_path = os.path.join(self.local_path, 'check.chk') if local_check_file_path is None\
                 else local_check_file_path
             if os.path.isfile(local_check_file_path) and self.software.lower() == 'gaussian':
-                ssh.upload_file(remote_file_path=remote_check_file_path, local_file_path=local_check_file_path)
-                logger.debug('uploading checkpoint file for {0}'.format(self.job_name))
+                with SSHClient(self.server) as ssh:
+                    ssh.upload_file(remote_file_path=remote_check_file_path, local_file_path=local_check_file_path)
+                logger.debug(f'uploading checkpoint file for {self.job_name}')
         else:
             # running locally, just copy the check file to the job folder
             new_check_file_path = os.path.join(self.local_path, 'check.chk')
@@ -1269,67 +1269,67 @@ end
         """
         Download ESS output, orbitals check file, and the Gaussian check file, if relevant.
         """
-        ssh = SSHClient(self.server)
+        with SSHClient(self.server) as ssh:
 
-        # download output file
-        remote_file_path = os.path.join(self.remote_path, output_filename[self.software])
-        ssh.download_file(remote_file_path=remote_file_path, local_file_path=self.local_path_to_output_file)
-        if not os.path.isfile(self.local_path_to_output_file):
-            raise JobError('output file for {0} was not downloaded properly'.format(self.job_name))
-        self.final_time = ssh.get_last_modified_time(remote_file_path=remote_file_path)
+            # download output file
+            remote_file_path = os.path.join(self.remote_path, output_filename[self.software])
+            ssh.download_file(remote_file_path=remote_file_path, local_file_path=self.local_path_to_output_file)
+            if not os.path.isfile(self.local_path_to_output_file):
+                raise JobError(f'output file for {self.job_name} was not downloaded properly')
+            self.final_time = ssh.get_last_modified_time(remote_file_path=remote_file_path)
 
-        # download orbitals FChk file
-        if self.job_type == 'orbitals':
-            remote_file_path = os.path.join(self.remote_path, 'input.FChk')
-            ssh.download_file(remote_file_path=remote_file_path, local_file_path=self.local_path_to_orbitals_file)
-            if not os.path.isfile(self.local_path_to_orbitals_file):
-                logger.warning('Orbitals FChk file for {0} was not downloaded properly '
-                               '(this is not the Gaussian formatted check file...)'.format(self.job_name))
+            # download orbitals FChk file
+            if self.job_type == 'orbitals':
+                remote_file_path = os.path.join(self.remote_path, 'input.FChk')
+                ssh.download_file(remote_file_path=remote_file_path, local_file_path=self.local_path_to_orbitals_file)
+                if not os.path.isfile(self.local_path_to_orbitals_file):
+                    logger.warning(f'Orbitals FChk file for {self.job_name} was not downloaded properly '
+                                   f'(this is not the Gaussian formatted check file...)')
 
-        # download Gaussian check file
-        if self.software.lower() == 'gaussian':
-            remote_check_file_path = os.path.join(self.remote_path, 'check.chk')
-            ssh.download_file(remote_file_path=remote_check_file_path, local_file_path=self.local_path_to_check_file)
-            if not os.path.isfile(self.local_path_to_check_file):
-                logger.warning('Gaussian check file for {0} was not downloaded properly'.format(self.job_name))
+            # download Gaussian check file
+            if self.software.lower() == 'gaussian':
+                remote_check_file_path = os.path.join(self.remote_path, 'check.chk')
+                ssh.download_file(remote_file_path=remote_check_file_path, local_file_path=self.local_path_to_check_file)
+                if not os.path.isfile(self.local_path_to_check_file):
+                    logger.warning(f'Gaussian check file for {self.job_name} was not downloaded properly')
 
-        # download Orca .hess hessian file generated by frequency calculations
-        # Hessian is useful when the user would like to project rotors
-        if self.software.lower() == 'orca' and self.job_type == 'freq':
-            remote_hess_file_path = os.path.join(self.remote_path, 'input.hess')
-            ssh.download_file(remote_file_path=remote_hess_file_path, local_file_path=self.local_path_to_hess_file)
-            if not os.path.isfile(self.local_path_to_hess_file):
-                logger.warning(f'Orca hessian file for {self.job_name} was not downloaded properly')
+            # download Orca .hess hessian file generated by frequency calculations
+            # Hessian is useful when the user would like to project rotors
+            if self.software.lower() == 'orca' and self.job_type == 'freq':
+                remote_hess_file_path = os.path.join(self.remote_path, 'input.hess')
+                ssh.download_file(remote_file_path=remote_hess_file_path, local_file_path=self.local_path_to_hess_file)
+                if not os.path.isfile(self.local_path_to_hess_file):
+                    logger.warning(f'Orca hessian file for {self.job_name} was not downloaded properly')
 
-        # download Lennard_Jones data file
-        if self.software.lower() == 'onedmin':
-            remote_lj_file_path = os.path.join(self.remote_path, 'lj.dat')
-            ssh.download_file(remote_file_path=remote_lj_file_path, local_file_path=self.local_path_to_lj_file)
-            if not os.path.isfile(self.local_path_to_lj_file):
-                logger.warning('Lennard-Jones data file for {0} was not downloaded properly'.format(self.job_name))
+            # download Lennard_Jones data file
+            if self.software.lower() == 'onedmin':
+                remote_lj_file_path = os.path.join(self.remote_path, 'lj.dat')
+                ssh.download_file(remote_file_path=remote_lj_file_path, local_file_path=self.local_path_to_lj_file)
+                if not os.path.isfile(self.local_path_to_lj_file):
+                    logger.warning(f'Lennard-Jones data file for {self.job_name} was not downloaded properly')
 
-        # download molpro log file (in addition to the output file)
-        if self.software.lower() == 'molpro':
-            remote_log_file_path = os.path.join(self.remote_path, 'input.log')
-            local_log_file_path = os.path.join(self.local_path, 'output.log')
-            ssh.download_file(remote_file_path=remote_log_file_path, local_file_path=local_log_file_path)
-            if not os.path.isfile(local_log_file_path):
-                logger.warning('Could not download Molpro log file for {0} '
-                               '(this is not the output file)'.format(self.job_name))
-
-        # download terachem files (in addition to the output file)
-        if self.software.lower() == 'terachem':
-            base_path = os.path.join(self.remote_path, 'scr')
-            filenames = ['results.dat', 'output.molden', 'charge.xls', 'charge_mull.xls', 'optlog.xls', 'optim.xyz',
-                         'Frequencies.dat', 'I_matrix.dat', 'Mass.weighted.modes.dat', 'moments_of_inertia.dat',
-                         'output.basis', 'output.geometry', 'output.molden', 'Reduced.mass.dat', 'results.dat']
-            for filename in filenames:
-                remote_log_file_path = os.path.join(base_path, filename)
-                local_log_file_path = os.path.join(self.local_path, 'scr', filename)
+            # download molpro log file (in addition to the output file)
+            if self.software.lower() == 'molpro':
+                remote_log_file_path = os.path.join(self.remote_path, 'input.log')
+                local_log_file_path = os.path.join(self.local_path, 'output.log')
                 ssh.download_file(remote_file_path=remote_log_file_path, local_file_path=local_log_file_path)
-            xyz_path = os.path.join(base_path, 'optim.xyz')
-            if os.path.isfile(xyz_path):
-                self.local_path_to_xyz = xyz_path
+                if not os.path.isfile(local_log_file_path):
+                    logger.warning(f'Could not download Molpro log file for {self.job_name} ' \
+                                   f'(this is not the output file)')
+
+            # download terachem files (in addition to the output file)
+            if self.software.lower() == 'terachem':
+                base_path = os.path.join(self.remote_path, 'scr')
+                filenames = ['results.dat', 'output.molden', 'charge.xls', 'charge_mull.xls', 'optlog.xls', 'optim.xyz',
+                             'Frequencies.dat', 'I_matrix.dat', 'Mass.weighted.modes.dat', 'moments_of_inertia.dat',
+                             'output.basis', 'output.geometry', 'output.molden', 'Reduced.mass.dat', 'results.dat']
+                for filename in filenames:
+                    remote_log_file_path = os.path.join(base_path, filename)
+                    local_log_file_path = os.path.join(self.local_path, 'scr', filename)
+                    ssh.download_file(remote_file_path=remote_log_file_path, local_file_path=local_log_file_path)
+                xyz_path = os.path.join(base_path, 'optim.xyz')
+                if os.path.isfile(xyz_path):
+                    self.local_path_to_xyz = xyz_path
 
     def run(self):
         """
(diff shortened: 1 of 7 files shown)


arc/job/job.py
```diff
@@ -1225,39 +1225,37 @@ def write_input_file(self):
...
+                    ssh.change_mode(mode='+x', path=up_file['name'], remote_path=self.remote_path)
+            self.initial_time = ssh.get_last_modified_time(
+                remote_file_path=os.path.join(self.remote_path, submit_filename[servers[self.server]['cluster_soft']]))
 
     def _upload_check_file(self, local_check_file_path=None):
         if self.server != 'local':
-            ssh = SSHClient(self.server)
             remote_check_file_path = os.path.join(self.remote_path, 'check.chk')
-            local_check_file_path = os.path.join(self.local_path, 'check.chk') if remote_check_file_path is None\
+            local_check_file_path = os.path.join(self.local_path, 'check.chk') if local_check_file_path is None\
```
xiaoruiDong: I think there is a typo in previous PR, since it is not reasonable to check whether `remote_check_file_path` is `None` given the context.

arc/job/job.py
```diff
@@ -1349,15 +1347,9 @@ def run(self):
         logger.debug('writing input file...')
         self.write_input_file()
         if self.server != 'local':
-            ssh = SSHClient(self.server)
             logger.debug('submitting job...')
             # submit_job returns job server status and job server id
-            try:
-                self.job_status[0], self.job_id = ssh.submit_job(remote_path=self.remote_path)
-            except IndexError:
```
xiaoruiDong: For reviewer: Since connection is checked, I don't really think the current script will raise IndexError. But if needed, I can keep it.

arc/job/ssh.py
```diff
@@ -49,11 +49,19 @@ def __init__(self, server=''):
         self.key = servers[server]['key']
         logging.getLogger("paramiko").setLevel(logging.WARNING)
 
-    def send_command_to_server(self, command, remote_path=''):
+    def _send_command_to_server(self, command: str, remote_path: str='') -> (list, list):
```
alongd: Thanks for adding type hints! I'm also only learning the right way to do it, and I hope this repo moves in the right direction. Once we standardize our approach, we'll ask for it in every new PR.

In cases like this (returning two or more items), we should do the following:

Import `typing` (a standard python library):
```python
from typing import Tuple
```

Add the type hint, e.g.:
```python
def _send_command_to_server(self,
                            command: str,
                            remote_path: str = '',
                            ) -> Tuple[list, list]:
```

And in the docsting do:
```
Returns:
    Tuple[list, list]: A list of lines of standard output stream, and a list of lines of standard error stream.
```
or

```
Returns:
    Tuple[list, list]:
        - A list of lines of the standard output stream.
        - A list of lines of the standard error stream.
```

We need to see how the docsting comes out in the documentation and decide which of the above formats to adhere to. I think the second one is clearer, perhaps we could compile the documentation locally to verify that it looks OK.

xiaoruiDong: @alongd Thank you for introducing type hint into ARC and teach me how to do it correctly. I tried my best to clean up the type hints. Please take a look at the new commit.

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
                 return '', 'ssh timed-out after two trials'
         stdout = stdout.readlines()
         stderr = stderr.readlines()
-        ssh.close()
         return stdout, stderr
 
-    def upload_file(self, remote_file_path, local_file_path='', file_string=''):
+    def upload_file(self,
+                    remote_file_path: str,
+                    local_file_path: str = '',
```
alongd: local_file_path: Optional[str] = '',

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
+                    file_string: str = '',
+                    ) -> None:
         """
-        Upload `local_file_path` or the contents of `file_string` to `remote_file_path`.
-        Either `file_string` or `local_file_path` must be given.
+        A modulator method of _upload_file(). Upload a local file or contents
```
alongd: what is `_upload_file()`?

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
+            raise ServerError(f'Could not write file {remote_file_path} on {self.server}. ')
 
-    def _download_file(self, remote_file_path, local_file_path):
+    def download_file(self,
+                      remote_file_path: str,
+                      local_file_path: str,
+                      ) -> None:
         """
-        Download a file from `remote_file_path` to `local_file_path`.
+        A modulator function of _download_file(). Download a file from the server.
```
alongd: what's `_upload_file()`?

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
+
+        Args:
+            remote_file_path (str): The remote path to be downloaded from.
+            local_file_path (str): The local path to be downloaded to.
+
+        Raises:
+            ServerError: If the file cannot be downloaded with maximum times to try
```
alongd: where is the maximum times to try implemented?

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
-            if result == 'connection error':
-                logger.error(f'ARC is sleeping for {sleep_time * i} min before re-trying, '
-                             f'please check your connectivity.')
-                logger.info('ZZZZZ..... ZZZZZ.....')
-                time.sleep(sleep_time * i * 60)  # in seconds
-            else:
-                i = 1000
-            i += 1
-        return result
+        A modulator method of _check_job_status(). Check job's status.
```
alongd: what is `_check_job_status()`?

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
+            str: Possible statuses: `before_submission`, `running`, `errored on node xx`,
+                 `done`, and `errored: ...`
         """
-        cmd = check_status_command[servers[self.server]['cluster_soft']] + ' -u ' + servers[self.server]['un']
-        stdout, stderr = self.send_command_to_server(cmd)
+        cmd = check_status_command[servers[self.server]['cluster_soft']] \
+              + ' -u ' + servers[self.server]['un']
+        stdout, stderr = self._send_command_to_server(cmd)
+        # Status line formats:
+        # pharos: '540420 0.45326 xq1340b    user_name       r     10/26/2018 11:08:30 long1@node18.cluster'
```
alongd: could you change pharos to OGE and rmg to Slurm?

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
-            if servers[self.server]['cluster_soft'].lower() == 'oge':
+            if cluster_soft.lower() == 'oge':
                 job_id = int(stdout[0].split()[2])
-            elif servers[self.server]['cluster_soft'].lower() == 'slurm':
+            elif cluster_soft.lower() == 'slurm':
                 job_id = int(stdout[0].split()[3])
             else:
-                raise ValueError(f'Unrecognized cluster software {servers[self.server]["cluster_soft"]}')
+                raise ValueError(f'Unrecognized cluster software '
+                                 f'{servers[self.server]["cluster_soft"]}')
```
alongd: replace `servers[self.server]["cluster_soft"]` with `cluster_soft`

arc/job/ssh.py
```diff
@@ -250,60 +315,170 @@ def connect(self):
...
             else:
                 logger.debug(f'Successfully connected to {self.server} at the {times_tried} trial.')
-                return sftp, ssh
+                return
             time.sleep(interval)
         raise ServerError(f'Could not connect to server {self.server} even after {times_tried} trials.')
 
-    def try_connecting(self):
-        """A helper function for connecting via paramiko, returns the `sftp` and `ssh` objects"""
+    def _connect(self) -> Tuple[paramiko.sftp_client.SFTPClient, paramiko.SSHClient]:
```
alongd: nice type hints implementation!

arc/job/ssh.py
```diff
@@ -250,60 +315,170 @@ def connect(self):
...
+        """
+        Connect via paramiko, and open a SSH session as well as a SFTP session.
+
+        Returns:
+            Tuple[paramiko.sftp_client.SFTPClient, paramiko.SSHClient]:
+                - A SFTP client used to perform remote file operations.
```
alongd: "An SFTP client..."

arc/job/ssh.py
```diff
@@ -250,60 +315,170 @@ def connect(self):
...
+            Tuple[paramiko.sftp_client.SFTPClient, paramiko.SSHClient]:
+                - A SFTP client used to perform remote file operations. 
+                - A high-level representation of a session with an SSH server.
+        """
         ssh = paramiko.SSHClient()
         ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         ssh.load_system_host_keys(filename=self.key)
         try:
-            ssh.connect(hostname=self.address, username=self.un)
+            ssh.connect(hostname=self.address, username=self.un, banner_timeout=200)
```
alongd: what's `banner_timeout`?

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
...
+        self._sftp = None
+        self._ssh = None
         logging.getLogger("paramiko").setLevel(logging.WARNING)
 
-    def send_command_to_server(self, command, remote_path=''):
+    def __enter__(self) -> 'SSHClient':
+        self.connect()
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback) -> None:
```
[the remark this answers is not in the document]
xiaoruiDong: Reference for context manager: https://book.pythontips.com/en/latest/context_managers.html. __exit__() needs 3 arguments: exception_type, exception value and traceback as default.

arc/job/ssh.py
```diff
@@ -47,200 +51,261 @@ def __init__(self, server=''):
         self.address = servers[server]['address']
         self.un = servers[server]['un']
         self.key = servers[server]['key']
+        self._sftp = None
+        self._ssh = None
         logging.getLogger("paramiko").setLevel(logging.WARNING)
 
-    def send_command_to_server(self, command, remote_path=''):
+    def __enter__(self) -> 'SSHClient':
```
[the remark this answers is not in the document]
xiaoruiDong: Yes, they are called by `with`. Please take a look at the reference. You can find more info by searching 'Context Manager python` if you find it is not clear.

arc/job/ssh.py
xiaoruiDong: That is my mistake. `_upload_file()` and `_download_file()` is no longer needed.

arc/job/ssh.py
xiaoruiDong: Sorry for the confusion. This is also an obsolete comment I forgot to modified.

arc/job/ssh.py
xiaoruiDong: It relates to several issues (errors) I've seen in the past, e.g.,
`SSHException: Error reading SSH protocol banner`

This happens, if the server accepts the connection but the ssh daemon doesn't respond in 15 seconds (default in paramiko). It can be due to network congestion, faulty switches, etc... .The common solution is enlarging the timeout time and if still not work, waiting and trying again later, and that strategy is what I actually implemented in SSHClient.

xiaoruiDong:
@alongd, thank you for the review. I have made changes according to your suggestions.

arc/job/ssh.py
alongd: Thanks! Can you add this comment here?
`#  If the server accepts the connection but the SSH daemon doesn't respond in 15 seconds (default in paramiko) due to network congestion, faulty switches, etc..., common solution is to enlarging the timeout variable.`

arc/job/submit.py
```diff
@@ -41,12 +41,12 @@
 cd $WorkDir
 . $g09root/g09/bsd/g09.profile
 
-cp $SubmitDir/input.gjf .
-cp $SubmitDir/check.chk .
+cp "$SubmitDir/input.gjf" .
```
alongd, line 44: why where `"` added?
modifying submit.py will cause merge conflicts for many users, let's only make essential modifications.

xiaoruiDong: It is essential from my point of view. As I state in the issue, sometimes when we are using smiles as labels (which I think is a common choice), sometimes we cannot create/ change to the correct directory unless we use `"` (`'` doesn't work).

alongd:
@xiaoruiDong, please rebase

alongd:
@xiaoruiDong, could you squash the fixups?

xiaoruiDong:
@alongd rebased and squashed. Thanks for the patience!

xiaoruiDong:
@alongd  checks are done

alongd:
@xiaoruiDong, this is next in line to be merged. I only want to test it in action (most of my ARC jobs use local rather than SSH, I'll use another computer for SSHing into the RMG server). Please rebase again. Thanks

xiaoruiDong:
@ Rebased, waiting for checks

arc/job/ssh.py
```diff
@@ -47,161 +55,202 @@ def __init__(self, server=''):
...
+        self.connect()
+        return self
+
+    def __exit__(self, exc_type, exc_value, exc_traceback) -> None:
+        self.close()
+
+    def check_connections(function: Callable[..., Any]) -> Callable[..., Any]:
```
alongd, line 69: why is the first parameter `function` and not `self`?

arc/job/ssh.py
```diff
@@ -47,161 +55,202 @@ def __init__(self, server=''):
...
         """
-        Send commands to server. `command` is either a sting or an array of string commands to send.
-        If remote_path is not an empty string, the command will be executed in the directory path it points to.
-        Returns lists of stdout, stderr corresponding to the commands sent.
+        A decorator to check SSH connections
+        """
+        def decorator(*args, **kwargs) -> Any:
```
alongd, line 73: would you like to add a docsting?

arc/job/ssh.py
```diff
@@ -47,161 +55,202 @@ def __init__(self, server=''):
...
+            return function(*args, **kwargs)
+        return decorator
+
+    @check_connections
+    def _send_command_to_server(self, 
+                                command: Union[str, list], 
+                                remote_path: Optional[str] = '',
+                                ) -> Tuple[list, list]:
+        """
+        A wapper for exec_command in paramiko.SSHClient. Send commands to the server.
```
alongd, line 95: typo: "wrapper"

arc/job/ssh.py
```diff
@@ -47,161 +55,202 @@ def __init__(self, server=''):
...
+            _, stdout, stderr = self._ssh.exec_command(command)
+        except Exception as e:  # SSHException: Timeout opening channel.
+            logger.debug(f'ssh timed-out in the first trial. Got:{e}')
             try:  # try again
-                _, stdout, stderr = ssh.exec_command(command)
-            except:
+                _, stdout, stderr = self._ssh.exec_command(command)
+            except Exception as e:
+                logger.debug(f'ssh timed-out after two trials. Got:{e}')
                 return '', 'ssh timed-out after two trials'
```
alongd, line 126: this line should be changed as well to return the same types as the final `return`

arc/job/ssh.py
```diff
@@ -47,161 +55,202 @@ def __init__(self, server=''):
...
+    def check_running_jobs_ids(self) -> list:
         """
-        Return a list of ``int`` representing job IDs of all jobs submitted by the user on a server
+        Check all jobs submitted by the user on a server.
+
+        Returns:
+            list: A list of job IDs
         """
         running_jobs_ids = list()
         cmd = check_status_command[servers[self.server]['cluster_soft']] + ' -u $USER'
```
alongd, line 256: `self.send_command_to_server(cmd)` below isn't functional

arc/job/ssh.py
```diff
@@ -254,66 +320,182 @@ def connect(self):
...
+        """
+        if self._sftp is not None:
+            self._sftp.close()
+        if self._ssh is not None:
+            self._ssh.close()
+
+    @check_connections
+    def get_last_modified_time(self, 
+                               remote_file_path: str,
+                               ) -> datetime.datetime:
```
alongd, line 363: should be `Union[datetime.datetime, None]`

arc/job/ssh.py
```diff
@@ -254,66 +320,182 @@ def connect(self):
...
 
-def write_file(sftp, remote_file_path, local_file_path='', file_string=''):
-    """
-    Write a file. If `file_string` is given, write it as the content of the file.
-    Else, if `local_file_path` is given, copy it to `remote_file_path`.
+        Args:
+            remote_path (str, optional): The directory path at which the command will be executed.
+        """
+        command = f'ls -alF'
+        return self._send_command_to_server(command, remote_path)
```
alongd, line 387: looks like this returns two lists, but the return type hint is one list, could you check?

arc/job/ssh.py
```diff
@@ -344,7 +526,8 @@ def check_job_status_in_stdout(job_id, stdout, server):
             raise ValueError(f'Unknown cluster software {servers[server]["cluster_soft"]}')
 
 
-def delete_all_arc_jobs(server_list, jobs=None):
+
```
alongd, line 529: could you remove the extra empty line?

arc/job/ssh.py
```diff
@@ -361,19 +544,19 @@ def delete_all_arc_jobs(server_list, jobs=None):
...
+                s = re.search(r' a\d+', status_line)
+                if s is not None:
+                    job_id = s.group()[1:]
+                    if job_id in jobs or jobs is None:
+                        if servers[server]['cluster_soft'].lower() == 'slurm':
+                            server_job_id = status_line.split()[0]
+                            ssh.delete_job(server_job_id)
+                            print(f'deleted job {job_id} ({server_job_id} on server)')
+                        elif servers[server]['cluster_soft'].lower() == 'oge':
+                            ssh.delete_job(job_id)
```
alongd, line 559: do you mind changing `ssh.delete_job(job_id)` to `ssh.delete_job(int(job_id))`?

xiaoruiDong:
> Looks like send_command_to_server on line 257 in ssh.py should have a prefix underscore _send_command_to_server.
BTW, this PR made send_command_to_server a private/protected method (by adding the underscore), but the delete_all_arc_jobs do access this protected method.

Thank you Alon, I am sure that they are mistakes from my recent squash or rebase. I will fix them. Besides, I will respond to your comments and make everything fixup. I will leave the fixup commits until it is ready to merge.

arc/job/ssh.py
xiaoruiDong: This is a decorator, which doesn't behave like a normal class method. Actually, it can be moved to the place above this SSHClient class. If there is a place ARC used to save decorators, I can also put it there if you like.

arc/job/ssh.py
xiaoruiDong: I will provide a more detailed docstring

arc/job/ssh.py
xiaoruiDong: Thanks.

arc/job/ssh.py
xiaoruiDong: Sure thing. I will make them as two lists, just as stdout and stderr.

arc/job/ssh.py
xiaoruiDong: I am thinking of move delete all jobs on a single server to be a method of SSHClient and `delete_all_arc_jobs` will be clearer.

arc/job/job.py
```diff
@@ -1228,40 +1228,38 @@ def write_input_file(self):
...
+            if self.input is not None:
+                remote_file_path = os.path.join(self.remote_path, input_filename[self.software])
+                ssh.upload_file(remote_file_path=remote_file_path, file_string=self.input)
+            for up_file in self.additional_files_to_upload:
+                if up_file['source'] == 'path':
+                    local_file_path = up_file['local']
+                elif up_file['source'] == 'input_files':
+                    local_file_path = input_files[up_file['local']]
+                else:
+                    raise JobError(f'Unclear file source for {up_file["name"]}. Should either be "path" of ' \
```
alongd, line 1246: the `\` is redundant between parenthesis (same for line 1286 and additional occurrences below)

alongd, line 1246: could you please replace `either be "path" of` with `either be "path" or`?

alongd, line 1246: could you please replace `either be "path" of` with `either be "path" or`?

arc/job/job.py
```diff
@@ -1271,67 +1269,67 @@ def _download_output_file(self):
...
-                self.local_path_to_xyz = xyz_path
+                if not os.path.isfile(local_log_file_path):
+                    logger.warning(f'Could not download Molpro log file for {self.job_name} ' \
+                                   f'(this is not the output file)')
+
+            # download terachem files (in addition to the output file)
+            if self.software.lower() == 'terachem':
+                base_path = os.path.join(self.remote_path, 'scr')
+                filenames = ['results.dat', 'output.molden', 'charge.xls', 'charge_mull.xls', 'optlog.xls', 'optim.xyz',
+                            'Frequencies.dat', 'I_matrix.dat', 'Mass.weighted.modes.dat', 'moments_of_inertia.dat',
```
alongd, line 1324: could you please indent these two lines by another white space (so that both `'` are in the same column)?
same for line 1596

arc/job/ssh.py
```diff
@@ -47,161 +55,202 @@ def __init__(self, server=''):
...
+        # SLURM: '14428     debug xq1371m2   user_name  R 50-04:04:46      1 node06'
         if stderr:
             logger.info('\n\n')
             logger.error(f'Could not check status of job {job_id} due to {stderr}')
-            return 'connection error'
+            return f'errored: {stderr}'
         return check_job_status_in_stdout(job_id=job_id, stdout=stdout, server=self.server)
 
-    def delete_job(self, job_id):
+    def delete_job(self, job_id: int) -> None:
```
alongd, line 238: please change to `job_id: Union[str, int]`

arc/job/ssh.py
alongd: I also prefer to use `Optional` for arguments, but I think it makes more sense to use `Union[x, None]` for `return` because the function always returns something (it's not optional), whereas an argument can me omitted if it's truly optional. It's like the difference between passing `None` to an argument, and not passing any value to an optional argument that will initialize as `None`. However, if you'd still like to use `Optional` it's OK with me, technically it's correct.

arc/job/ssh.py
alongd: OK

arc/job/ssh.py
xiaoruiDong: Thanks. I understand your points. I have seen people do `Optional` way for returns, but haven't seen the other way (but I believe there could be a lot of people do this way). That's the main reason I use `Optional`. Thank you for understanding.

arc/job/job.py
xiaoruiDong: True!

xiaoruiDong: Good point!

arc/job/job.py
xiaoruiDong: I am sorry for leaving this kind of issue. I am using VS Code and its autopep8 always gives a not-so-perfect alignment. I will try my best to avoid them.

arc/job/ssh.py
xiaoruiDong: Sure!

xiaoruiDong:
@alongd This PR is updated and the checks are mostly passed.

alongd:
I ran the minimal example on this branch, and got the following trace:
```
Running job opt_a33920 for H2
  Ending job opt_a33920 for H2 (run time: 0:00:22)
Running job opt_a33921 for H2 (fine opt)
  Ending job opt_a33921 for H2 (run time: 0:00:49)

Optimized geometry for H2 at wb97xd/def2tzvp:
H       0.37207800    0.00000000    0.00000000
H      -0.37207800    0.00000000   -0.00000000
Running job freq_a33922 for H2
Running job sp_a33923 for H2
  Ending job freq_a33922 for H2 (run time: 0:00:57)
  File "/home/alongd/Code/ARC/arc/job/ssh.py", line 200, in download_file
    localpath=local_file_path)
  File "/home/alongd/anaconda3/envs/arc_env/lib/python3.7/site-packages/paramiko/sftp_client.py", line 802, in get
    size = self.getfo(remotepath, fl, callback)
  File "/home/alongd/anaconda3/envs/arc_env/lib/python3.7/site-packages/paramiko/sftp_client.py", line 778, in getfo
    file_size = self.stat(remotepath).st_size
  File "/home/alongd/anaconda3/envs/arc_env/lib/python3.7/site-packages/paramiko/sftp_client.py", line 493, in stat
    t, msg = self._request(CMD_STAT, path)
  File "/home/alongd/anaconda3/envs/arc_env/lib/python3.7/site-packages/paramiko/sftp_client.py", line 813, in _request
    return self._read_response(num)
  File "/home/alongd/anaconda3/envs/arc_env/lib/python3.7/site-packages/paramiko/sftp_client.py", line 865, in _read_response
    self._convert_status(msg)
  File "/home/alongd/anaconda3/envs/arc_env/lib/python3.7/site-packages/paramiko/sftp_client.py", line 894, in _convert_status
    raise IOError(errno.ENOENT, text)
FileNotFoundError: [Errno 2] No such file

During handling of the above exception, another exception occurred:

  File "/home/alongd/Code/ARC//ARC.py", line 66, in <module>
    main()
  File "/home/alongd/Code/ARC//ARC.py", line 62, in main
    arc_object.execute()
  File "/home/alongd/Code/ARC/arc/main.py", line 594, in execute
    fine_only=self.fine_only)
  File "/home/alongd/Code/ARC/arc/scheduler.py", line 440, in __init__
    self.schedule_jobs()
  File "/home/alongd/Code/ARC/arc/scheduler.py", line 521, in schedule_jobs
  File "/home/alongd/Code/ARC/arc/scheduler.py", line 782, in end_job
    job.determine_job_status()  # also downloads output file
  File "/home/alongd/Code/ARC/arc/job/job.py", line 1386, in determine_job_status
    self._check_job_ess_status()  # populates self.job_status[1], and downloads the output file
  File "/home/alongd/Code/ARC/arc/job/job.py", line 1496, in _check_job_ess_status
    self._download_output_file()  # also downloads the check file and orbital file if exist
  File "/home/alongd/Code/ARC/arc/job/job.py", line 1315, in _download_output_file
    ssh.download_file(remote_file_path=remote_log_file_path, local_file_path=local_log_file_path)
  File "/home/alongd/Code/ARC/arc/job/ssh.py", line 204, in download_file
    raise ServerError(f'Could not download file {remote_file_path} from {self.server}. ')
arc.exceptions.ServerError: Could not download file runs/ARC_Projects/minimal/H2/sp_a33923/input.log from rmg. 
```

ARC is trying to download the `input.log` molpro log file from the server, but that file has been renamed to `input.out` (determined by the `output_filename` dict in settings). @xiaoruiDong, could you take a look at the cause for this behaviour?

xiaoruiDong:
@alongd Sorry for the error. I haven't seen this before. I will take a look.

xiaoruiDong:
@alongd I catch it. I think it is due to the file output name of molpro in `job._download_output_file` is hardcoded as 'output.log' (even before I changed this function). I can fix it by following the settings.

PR #362, opened 2020-03-25
Add more specific error message when over requesting server resources

kspieks:


kspieks:
Sorry, I just rebased

PR #366, opened 2020-04-02
Modified delete.py to accept command line arguments 

### LINKED ISSUE #354: Automatically delete all ARC jobs for a specific project
ARC has a utility to automatically delete all currently running jobs on all servers (for a specific user) - https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/utils/delete.py.

It should be modified such that an optional keyword with a project name (either via API or the command-line) would make it delete running jobs for a specified project only. While at it, we could add support to delete jobs from a specific server via command-line arguments as well (already supported via the API).

alongd:
fixes #354

ARC can now delete all ARC-spawned jobs from selected servers and projects.
To delete all ARC jobs, type in a terminal in the ARC code folder after activating `arc_env`:

    python arc/utils/delete.py -a

You can also request to delete jobs from a specific server by specifying its name after the `-s` flag:

    python arc/utils/delete.py -s server1 -a

To delete jobs from a specific ARC project, pass the project's name after the `-p` flag:

    python arc/utils/delete.py -p project1

Alternatively (since project names might be long and not always shown in full when requesting the server job status),
one can supply an ARC job ID, and ALL jobs related to the project of the given job ID will be deleted
(NOT only the given job!):

    python arc/utils/delete.py -j a_54836

Note that either a `-a`, a `-p`, or a `-j` flag must be given.
All flags can be combined with the optional `-s` flag.

### DIFF
--- a/arc/utils/delete.py
+++ b/arc/utils/delete.py
@@ -9,19 +9,86 @@ Using this module might results in deletion of jobs running on all servers ARC h
 Use this only if you are certain in what you're doing to avoid deleting valuable jobs and information loss.
 """
 
+import argparse
+import csv
+import os
+
+from arc.exceptions import InputError
 from arc.job.local import delete_all_local_arc_jobs
 from arc.job.ssh import delete_all_arc_jobs
-from arc.settings import servers
+from arc.settings import arc_path, servers
+
+
+def parse_command_line_arguments(command_line_args=None):
+    """
+    Parse command-line arguments.
+
+    Args:
+        command_line_args: The command line arguments.
+    """
+
+    parser = argparse.ArgumentParser(description='ARC delete util')
+
+    parser.add_argument('-p', '--project', type=str, nargs=1, default='',
+                        metavar='Project', help='the project for which all jobs wil be deleted')
+    parser.add_argument('-j', '--job', type=str, nargs=1, default='',
+                        metavar='Job', help='the job the belongs to a project for which all jobs wil be deleted')
+    parser.add_argument('-s', '--server', type=str, nargs=1, default='',
+                        metavar='Server', help='the server name from which to delete jobs')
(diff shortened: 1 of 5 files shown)


arc/utils/delete.py
```diff
@@ -9,19 +9,86 @@
...
+    Delete ARC jobs according to the command line arguments specifications.
     """
-    server_list = [server for server in servers.keys() if server != 'local']
-    delete_all_arc_jobs(server_list=server_list)
-    if 'local' in servers:
-        delete_all_local_arc_jobs()
+
+    args = parse_command_line_arguments()
+
+    if not all and not args.project and not args.job:
```
xiaoruiDong, line 60: Not clear to me where `all` is defined.

arc/utils/delete.py
```diff
@@ -9,19 +9,86 @@
...
+                        metavar='Job', help='the job the belongs to a project for which all jobs wil be deleted')
+    parser.add_argument('-s', '--server', type=str, nargs=1, default='',
+                        metavar='Server', help='the server name from which to delete jobs')
+    parser.add_argument('-a', '--all', type=bool, nargs=1, default=False,
+                        metavar='All', help='delete all ARC jobs')
+
+    args = parser.parse_args(command_line_args)
+    if args.job:
+        if args.job[0][0] == 'a':
+            args.job = args.job[0][1:]
```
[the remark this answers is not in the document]
xiaoruiDong: I see, thanks

arc/utils/delete.py
alongd: you're right, it should be `args.all`, thanks!

alongd:
Thanks, @xiaoruiDong, I force-pushed with that fix

PR #368, opened 2020-04-02
Parse geometry from a Gaussian log file of a large molecule

alongd:
Parsing xyz from a Gaussina log file with more than 50 atoms where the `iop(2/9=2000)` keyword is not specified.
A test was added.


arc/parser.py
```diff
@@ -172,7 +172,24 @@ def parse_geometry(path):
...
     except LogError:
         logger.debug(f'Could not parse xyz from {path}')
+
+        # try parsing Gaussian standard orientation instead of the input orientation parsed by Arkane
+        lines = _get_lines_from_file(path)
+        xyz_str = ''
+        for i in range(len(lines)):
+            if 'Standard orientation:' in lines[i]:
+                xyz_str = ''
+                j = i + 5
```
oscarwumit, line 182: Are we sure we are always skipping 4 lines? What if Gaussian changed their log file format? Looking at the log file, since the 'center number' of the first atom has to be '1', maybe we should add a check here to ensure the first line we parse has `line.split()[0] == 1`

oscarwumit:
Thanks. I add one comment on parsing the xyz from the log file. Please take a look.

alongd:
@oscarwumit, can you take a second look? Let me know if you have further comments, otherwise please merge. Thanks!

oscarwumit:
Looks good. Thanks.

PR #370, opened 2020-04-04
Check atom balance

### LINKED ISSUE #369: An ARCReaction should check stoichiometry of reactants/TS/products
**Is your feature request related to a problem? Please describe.**
ARC currently doesn't alert the user if a reaction isn't defined correctly in terms of atom balance.

**Describe the solution you'd like**
Upon initialization, an ARCReaction object should check the reactants and products wells, and any user given TS guess for atom balance.

alongd:
fixes #369 

Check atom balance of Species and Reactions as they are defined by users

### DIFF
--- a/arc/reactionTest.py
+++ b/arc/reactionTest.py
@@ -11,14 +11,17 @@ from rmgpy.reaction import Reaction
 from rmgpy.species import Species
 
 import arc.rmgdb as rmgdb
+from arc.exceptions import ReactionError
 from arc.reaction import ARCReaction
 from arc.settings import default_ts_methods
+from arc.species import ARCSpecies
 
 
 class TestARCReaction(unittest.TestCase):
     """
     Contains unit tests for the ARCSpecies class
     """
+
     @classmethod
     def setUpClass(cls):
         """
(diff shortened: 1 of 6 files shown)


arc/reactionTest.py
```diff
@@ -119,6 +122,20 @@ def test_determine_multiplicity(self):
         self.rxn4.determine_rxn_multiplicity()
         self.assertEqual(self.rxn4.multiplicity, 3)
 
+    def test_check_atom_balance(self):
+        """Test the Reaction check_atom_balance method"""
+        rxn1 = ARCReaction(reactants=['CH4', 'OH'], products=['CH3', 'H2O'])
+        rxn1.r_species = [ARCSpecies(label='CH4', smiles='C'), ARCSpecies(label='OH', smiles='[OH]')]
```
dranasinghe, line 128: I tested the following reaction "CH4 + OH + H2O <=> CH3 + H2O + H2O" technically it should be balance. Could you take a look at the modifications I did to unit tests?  
`        rxn1 = ARCReaction(reactants=['CH4', 'OH','H2O'], products=['CH3', 'H2O','H2O'])
        rxn1.r_species = [ARCSpecies(label='CH4', smiles='C'), ARCSpecies(label='OH', smiles='[OH]'), ARCSpecies(label='H2O', smiles='O')]
        rxn1.p_species = [ARCSpecies(label='CH3', smiles='[CH3]'), ARCSpecies(label='H2O', smiles='O'), ARCSpecies(label='H2O', smiles='O')]
        self.assertTrue(rxn1.check_atom_balance())`

alongd:
Thanks for looking at the PR, and for the suggestion!
I implemented your test suggestion of having the same species on one side of the reaction. The `count` parameter by which entries that are added to `r_well` and `p_well` are multiplied should address such cases.
I modified the test a bit, `products` indeed has two H2O species labels, but `p_species` points to the actual objects calculated, so it holds only one H2O species (these are the actual species for which QM jobs were spawned, in this example we will only need to calculate one H2O species, but we'll use its E0 twice for the reaction).
Let me know if you have any additional comments

alongd:
@dranasinghe, I added the bug fix for multiple occurrences of species in a reaction well to this PR since it requires features introduced here (counting species occurrences from the reaction label).
Please take a look.
Could you try restarting your ARC project on this branch and see if you get a reasonable rate now?

PR #373, opened 2020-04-07
Improved troubleshoot methods for issues related to Orca job memory and cpu allocations

oscarwumit:
1. ARC now estimates memory and cpu cores needed for an Orca job using heuristics based on the number of heavy atoms in a species.
2. ARC now parse Orca version 4.1.x log files for troubleshooting correctly.
   - original code was written for Orca version 4.2.x. The format of the error message in the log files is different from Orca version 4.1.x. Specifically, in 4.2.x, Orca gives estimations of memory and cpu in the log files while in 4.1.x it does not. 
   - ARC now does not crash if it cannot find estimated memory and cpu information from log files.
3. ARC now attempts to troubleshoot GTOInt error.
   - Based on our experience, GTOInt error seems to mostly related to memory allocation. 
4. Tests are updated with new example failed Orca log files.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -985,16 +985,33 @@ def is_notebook() -> bool:
 
 def is_str_float(value: str) -> bool:
     """
-    Check whether a string represents a number.
+    Check whether a string can be converted to a floating number.
 
     Args:
         value (str): The string to check.
 
     Returns:
-        bool: ``True`` if it does, ``False`` otherwise.
+        bool: ``True`` if it can, ``False`` otherwise.
     """
     try:
         float(value)
+        return True
+    except ValueError:
+        return False
+
+
+def is_str_int(value: str) -> bool:
+    """
+    Check whether a string can be converted to an integer.
+
+    Args:
+        value (str): The string to check.
+
+    Returns:
+        bool: ``True`` if it can, ``False`` otherwise.
+    """
+    try:
+        int(value)
         return True
     except ValueError:
         return False
--- a/arc/common.py
+++ b/arc/common.py
@@ -22,7 +22,7 @@ import yaml
 import numpy as np
 import qcelemental as qcel
 
-from typing import List, Tuple, Union
+from typing import Any, List, Optional, Tuple, Union
 
 from arkane.ess import GaussianLog, MolproLog, OrcaLog, QChemLog, TeraChemLog
 from arkane.util import determine_qm_software
@@ -39,7 +39,9 @@ logger = logging.getLogger('arc')
 VERSION = '1.1.0'
 
 
-def initialize_job_types(job_types, specific_job_type=''):
+def initialize_job_types(job_types: dict,
+                         specific_job_type: str = '',
+                         ) -> dict:
     """
     A helper function for initializing job_types.
     Returns the comprehensive (default values for missing job types) job types for ARC.
@@ -99,7 +101,7 @@ def initialize_job_types(job_types, specific_job_type=''):
     return job_types
 
 
-def determine_ess(log_file):
+def determine_ess(log_file: str) -> str:
     """
     Determine the ESS to which the log file belongs.
 
@@ -124,7 +126,7 @@ def determine_ess(log_file):
                      f'Gaussian, Molpro, Orca, QChem, or TeraChem.')
 
 
-def check_ess_settings(ess_settings=None):
+def check_ess_settings(ess_settings: Optional[dict] = None) -> dict:
     """
     A helper function to convert servers in the ess_settings dict to lists
     Assists in troubleshooting job and trying a different server
@@ -163,7 +165,11 @@ def check_ess_settings(ess_settings=None):
     return settings
 
 
-def initialize_log(log_file, project, project_directory=None, verbose=logging.INFO):
+def initialize_log(log_file: str,
+                   project: str,
+                   project_directory: Optional[str] = None,
+                   verbose: int = logging.INFO,
+                   ) -> None:
     """
     Set up a logger for ARC.
 
@@ -226,7 +232,9 @@ def get_logger():
     return logger
 
 
-def log_header(project, level=logging.INFO):
+def log_header(project: str,
+               level: int = logging.INFO,
+               ) -> None:
     """
     Output a header containing identifying information about ARC to the log.
 
@@ -259,7 +267,9 @@ def log_header(project, level=logging.INFO):
     logger.info(f'Starting project {project}')
 
 
-def log_footer(execution_time, level=logging.INFO):
+def log_footer(execution_time: str,
+               level: int = logging.INFO,
+               ) -> None:
     """
     Output a footer for the log.
 
@@ -312,7 +322,7 @@ def get_git_branch():
 
 
 def read_yaml_file(path: str,
-                   project_directory: str = None,
+                   project_directory: Optional[str] = None,
                    ) -> dict or list:
     """
     Read a YAML file (usually an input / restart file, but also conformers file)
@@ -537,7 +547,7 @@ def get_single_bond_length(symbol1: str,
     return 2.5
 
 
-def determine_symmetry(xyz):
+def determine_symmetry(xyz: dict) -> Tuple[int, int]:
     """
     Determine external symmetry and chirality (optical isomers) of the species.
 
@@ -547,7 +557,9 @@ def determine_symmetry(xyz):
     Returns:
         int: The external symmetry number.
     Returns:
-        int: 1 if no chiral centers are present, 2 if chiral centers are present.
+        Tuple[int, int]:
+            - The external symmetry number.
+            - ``1`` if no chiral centers are present, ``2`` if chiral centers are present.
     """
     atom_numbers = list()  # List of atomic numbers
     for symbol in xyz['symbols']:
@@ -635,7 +647,9 @@ def extermum_list(lst: list,
         return max([entry for entry in lst if entry is not None])
 
 
-def sort_two_lists_by_the_first(list1, list2):
+def sort_two_lists_by_the_first(list1: List[Union[float, int, None]],
+                                list2: List[Union[float, int, None]],
+                                ) -> Tuple[List[Union[float, int]], List[Union[float, int]]]:
     """
     Sort two lists in increasing order by the values of the first list.
     Ignoring None entries from list1 and their respective entries in list2.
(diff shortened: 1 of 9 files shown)


arc/common.py
```diff
@@ -1017,3 +1019,47 @@ def time_lapse(t0):
...
+
+    if server in servers:
+        max_server_mem_gb = servers[server].get('memory', None)
+        max_server_cpus = servers[server].get('cpus', None)
+    else:
+        logger.debug(f'Cannot find server {server} in settings.py')
+        consider_server_limits = False
+
+    if consider_server_limits and max_server_mem_gb is not None:
+        try:
```
alongd, line 1046: Could you use `if isinstance(param, str) and not is_str_float(param)` instead of a try-except, to avoid code duplicity?
Could you also implement an `is_str_int()` function in common? I see it might be useful for the blocks you just added to trsh.py.

arc/common.py
```diff
@@ -1017,3 +1019,47 @@ def time_lapse(t0):
...
+        except ValueError:
+            logger.debug(f'CPU {max_server_cpus} specifications in settings.py contains non-numerical values.')
+
+    est_cpu = 2 + num_heavy_atoms * 4
+    if consider_server_limits and max_server_cpus and est_cpu > max_server_cpus:
+        est_cpu = max_server_cpus
+
+    est_memory = 2000.0 * est_cpu
+    if consider_server_limits and max_server_mem_gb and est_memory > 0.8 * max_server_mem_gb * 1024:
+        est_memory = 0.8 * max_server_mem_gb * 1024  # This 80% limit is the same limit imposed in job.set_cpu_and_mem()
```
alongd, line 1063: maybe skip the 0.8 factor. If in the future we'd change it in job.py, we'll have to trace it here as well.

arc/job/trsh.py
```diff
@@ -223,42 +223,65 @@ def determine_ess_status(output_path: str,
                     keywords = ['SCF']
                     for j, info in enumerate(reverse_lines):
                         if 'Please increase MaxCore' in info:
-                            estimated_mem = info.split()[-2]  # e.g., Please increase MaxCore to more than: 289 MB
-                            keywords.append('Memory')
-                            line = reverse_lines[j + 3].rstrip()  # e.g., Error (ORCA_SCF): Not enough memory available!
-                            error = f'Orca suggests to increase per cpu core memory to {estimated_mem} MB.'
-                            break
+                            try:
```
alongd, line 226: which line will raise the exception? Could you wrap only that specific code with the try-except block? In cases (below?) where it's the `float()` function, consider using `common.is_str_float` for cleaner code. Same comment for the blocks below.

arc/common.py
```diff
@@ -22,6 +22,8 @@
 import numpy as np
 import qcelemental as qcel
 
+from typing import List, Tuple, Union
```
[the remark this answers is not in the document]
oscarwumit: `List` and `Union` are commonly used for annotating functions. The real issue is that we are not annotating all the functions. Instead of removing these imports, I will add correct annotations to functions in this file.

arc/common.py
```diff
@@ -1017,3 +1019,47 @@ def time_lapse(t0):
     else:
         d = ''
     return f'{d}{h:02.0f}:{m:02.0f}:{s:02.0f}'
+
+
+def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
```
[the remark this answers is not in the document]
oscarwumit: The reason I put this function here is that it can also be used for estimating Orca job memory and cpu before spawning the job for the first time instead of using this function during troubleshoot. However, estimating job memory and cpu in advance requires more thoughts. Regarding open a new job module, we should discuss it. For now, let's put this function here.

arc/common.py
```diff
@@ -1017,3 +1019,47 @@ def time_lapse(t0):
...
+                                      server: str = '',
+                                      consider_server_limits: bool = False) -> Tuple[int, float]:
+    """
+    Estimates memory and cpu requirements for an Orca job.
+
+    :param num_heavy_atoms: The number of heavy atoms in the species.
```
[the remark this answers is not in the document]
oscarwumit: ok

arc/common.py
```diff
@@ -1017,3 +1019,47 @@ def time_lapse(t0):
     else:
         d = ''
     return f'{d}{h:02.0f}:{m:02.0f}:{s:02.0f}'
+
+
+def estimate_orca_mem_cpu_requirement(num_heavy_atoms: int,
+                                      server: str = '',
+                                      consider_server_limits: bool = False) -> Tuple[int, float]:
```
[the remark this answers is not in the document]
oscarwumit: sure

arc/common.py
oscarwumit: good idea

arc/common.py
oscarwumit: I think a better way is we store all the empirical parameters in settings or some other file, then we import these parameters when we use them in each module. If we want to update these parameters, we will then only need to update one file. The problem here is if I do not specify 0.8 here, I think it will cause problem in job, as job will not allow total memory to exceed 80% of the node memory.

arc/job/trsh.py
oscarwumit: I will wrap one line only. I tried rewriting using `common.is_str_float`, but that requires if else statements that lead to similar number of lines compared to try except.

oscarwumit:
I made the changes you suggested. Please take another look @alongd. Thanks.

arc/job/trsh.py
alongd: It's not just the number of lines, it's code maintenance. This is not a great example, because the function is very short, but if it's OK I'd prefer using the function where relevant.

arc/common.py
alongd: It's OK to store it in settings.
It shouldn't cause a problem in Job, Job corrects (reduces) the mem if it's higher than 80%*max_mem. So we don't ave to implement it here as well. WHat do you say?

arc/common.py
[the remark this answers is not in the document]
alongd: awesome, thanks!!

arc/common.py
oscarwumit: I agree we don't have to implement the limit here. I changed it in a fix up commit.

arc/job/trsh.py
oscarwumit: I have already modified this part so that the try except only warps one line. I think here try except makes sense, as we need to convert the string to float. After using the `is_str_float` function, we still need to convert the string to float, so why don't we just accomplish checking + conversion in one line?

oscarwumit:
Please let me know if you have additional comments, or should I squash and rebase. Thanks :-)

arc/common.py
```diff
@@ -122,7 +126,7 @@ def determine_ess(log_file):
                      f'Gaussian, Molpro, Orca, QChem, or TeraChem.')
 
 
-def check_ess_settings(ess_settings=None):
+def check_ess_settings(ess_settings: dict = None) -> dict:
```
alongd, line 129: should be `Optional[dict]` (need to import Optional from typing)
Same for anywhere below where an argument can be `None` (not everywhere it ha s a default value)

arc/common.py
```diff
@@ -224,7 +232,9 @@ def get_logger():
     return logger
 
 
-def log_header(project, level=logging.INFO):
+def log_header(project: str,
+               level: int =logging.INFO,
```
alongd, line 236: minor: space after `=`

arc/common.py
```diff
@@ -535,7 +547,7 @@ def get_single_bond_length(symbol1: str,
     return 2.5
 
 
-def determine_symmetry(xyz):
+def determine_symmetry(xyz: dict) -> Tuple[int, int]:
     """
     Determine external symmetry and chirality (optical isomers) of the species.
```
alongd, line 553: Could you also change the `Returns` below?
should be:
```
Returns:
    Tuple[int, int]:
        - The external symmetry number.
        - ``1`` if no chiral centers are present, ``2`` if chiral centers are present.
```

arc/common.py
```diff
@@ -837,7 +853,7 @@ def determine_model_chemistry_type(method: str or dict) -> str:
     return model_chemistry_class
 
 
-def format_level_of_theory_inputs(level_of_theory: str or dict):
+def format_level_of_theory_inputs(level_of_theory: str or dict) -> Tuple[dict, str]:
```
alongd, line 856: could you replace `str or dict` with `Union[str, dict]`?

arc/common.py
```diff
@@ -1017,3 +1050,43 @@ def time_lapse(t0):
...
+    """
+    Estimates memory and cpu requirements for an Orca job.
+
+    Args:
+        num_heavy_atoms (int): The number of heavy atoms in the species.
+        server (str): The name of the server where Orca runs.
+        consider_server_limits (bool):  Try to give realistic estimations.
+
+    Returns:
+        A tuple of the amount of total memory (MB) and cpu cores required for the Orca job of given species.
```
alongd, line 1068: Could you add `Tuple[int, float]:`?

arc/common.py
```diff
@@ -22,6 +22,8 @@
 import numpy as np
 import qcelemental as qcel
 
+from typing import Any, List, Tuple, Union
```
alongd, line 25: thanks for adding type hints!

arc/common.py
```diff
@@ -998,6 +1014,23 @@ def is_str_float(value: str) -> bool:
         return False
 
 
+def is_str_int(value: str) -> bool:
```
alongd, line 1017: is this function tested? if not, would you mind adding a small test?

arc/settings.py
```diff
@@ -172,3 +172,10 @@
 inconsistency_az = 5    # maximum allowed inconsistency (kJ/mol) between initial and final rotor scan points. Default: 5
 inconsistency_ab = 0.3  # maximum allowed inconsistency between consecutive points in the scan given as a fraction
 #  of the maximum scan energy. Default: 30%
+
+# Default job memory, cpu, time settings
+default_job_settings = {'job_total_memory_gb': 14,
```
alongd, line 177: thanks!!

arc/common.py
oscarwumit: ok

arc/common.py
oscarwumit: fixed

arc/common.py
oscarwumit: done

arc/common.py
oscarwumit: ok

arc/common.py
oscarwumit: done

arc/common.py
oscarwumit: ok

oscarwumit:
Thanks for the comment. I have made the changes, squashed them, and rebased onto master. @alongd

PR #374, opened 2020-04-13
BugFix: Make sure Job.total_job_memory_gb is an int before dumping into the restart dict

alongd:
Otherwise, a numpy value could be stored (from troubleshooting Orca jobs perhaps?), which does not load safely in YAML.


oscarwumit:
Good catch. Thanks. We should only store `float` and `int` in restart YAML files, is that correct? I was not aware of the load safety issue.

PR #375, opened 2020-04-14
Update maxstep settings for Gaussian opt jobs

oscarwumit:
1. Remove maxstep=5 for regular (not fine) TS opt jobs
2. Add maxstep = 5 for non TS fine opt jobs (wells)

'maxstep=5' is too conservative, which makes calculation really expensive. For regular opt TS job, using Gaussian default value maxstep=30 is more efficient. For opt fine job, 'maxstep=5' is kept.

In addition, 'maxstep=5' is also added for opt fine non-ts jobs (wells).

### DIFF
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -771,7 +771,7 @@ wf,spin={spin},charge={charge};}}
         if self.job_type in ['conformer', 'opt']:
             if self.software == 'gaussian':
                 if self.is_ts:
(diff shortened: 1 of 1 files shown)


alongd:
We care about this only for TSs, not for wells?

oscarwumit:
@alongd Thanks for the comment. After consulting with @dranasinghe, the `maxstep=5` keyword is added to non TS opt jobs as well.

PR #382, opened 2020-04-23
BugFix: Check if_str_float before converting a string to float

oscarwumit:
Orca log files have different formats, the previous implementation failed when the information needed in the Orca log file changed location. This PR fixed this issue by first checking if the string read from log file is a float before trying to convert it to float. Also added a test.

Previously, the troubleshooting method expects the following output from Orca:

               ***  Starting incremental Fock matrix formation  ***
  0   -189.1598341022   0.000000000000 0.01426203  0.00045574  0.3200117 0.7000
  1   -189.1954365019  -0.035602399743 0.01167587  0.00035104  0.2256806 0.7000

It turns out that for some Orca jobs, the output is different:

               ***  Starting incremental Fock matrix formation  ***
                      *** Initiating the Newton-Raphson procedure ***
                      *** Removing any level shift *** 
ITER      Energy       Delta-E        Grad      Rot      Max-DP    RMS-DP
  0   -359.75006651 -359.7500665143  0.000003  0.000000  0.000007  0.000000

The previous troubleshooting algorithm failed because it did not expect the second case. The updated algorithm will read lines until it finds the energy it needs.

### DIFF
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -10,7 +10,7 @@ import os
 
 import numpy as np
 
-from arc.common import get_logger, determine_ess, estimate_orca_mem_cpu_requirement
+from arc.common import get_logger, determine_ess, estimate_orca_mem_cpu_requirement, is_str_float
 from arc.exceptions import SpeciesError, TrshError
 from arc.job.local import execute_command
 from arc.job.ssh import SSHClient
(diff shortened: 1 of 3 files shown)


arc/job/trsh.py
```diff
@@ -199,7 +199,10 @@ def determine_ess_status(output_path: str,
                     # not done yet, things can still go wrong (e.g., SCF energy might blow up)
                     for j, info in enumerate(forward_lines):
                         if 'Starting incremental Fock matrix formation' in info:
-                            scf_energy_initial_iteration = float(forward_lines[j + 1].split()[1])
+                            while not is_str_float(forward_lines[j + 1].split()[1]):
+                                j += 1
+                            else:
```
alongd, line 204: no need for the `else` here

oscarwumit: Thanks.

oscarwumit:
Thanks, I have made the changes. Please take another look. @alongd

alongd:
Could you rebase?

oscarwumit:
Just rebased. @alongd

alongd:
Can you take a look at the commits? Something seems wrong (there are 5 commits)

oscarwumit:
@alongd It's fixed now. I accidentally pulled when I push the commits using git Karen.

PR #390, opened 2020-05-06
SSH improvement phase2

xiaoruiDong:
This is the second phase of SSH and server communication improvement. A major impetus is that I found the behavior of `paramiko` is slightly different now. 

In a previous version, when I tried to download a non-existing file with `sftp.get()`, there will be no error raised but an empty file at the `localpath` I assigned. For the version I used (I recently updated my arc environment, it is now the latest stable`2.7`), the behavior of `paramiko` is changed. It will raise an `IOError` (no such file). We tried to catch `IOError` in ARC, but due to some of the intermediate step is problematic (raise A another type of Error when catching this error, or have no catches at all), but we can only deal with part of the issues. Besides, if the ARC dead weirdly (say ARC has created an input file but hasn't submitted it to the server when my computer crashed), the current ARC cannot restart the job.

Anyway, this is a patch to the previous SSH Improvement to make the system more robust. However, this PR leaves the case when the quota is reached. (Proposal: We can delete finished jobs done within the scope of the project). It is aimed to solve in the future PR.


arc/job/job.py
```diff
@@ -1472,13 +1472,18 @@ def _get_additional_job_info(self):
...
+        Possible statuses: `initializing`, `running`, `errored`, `done`.
         """
         if self.server != 'local':
             with SSHClient(self.server) as ssh:
-                return ssh.check_job_status(self.job_id)
+                server_status = ssh.check_job_status(self.job_id)
         else:
-            return check_job_status(self.job_id)
+            server_status = check_job_status(self.job_id)
+        # `server_status` should only be 'initializing' / 'running' / 'errored' / 'done'
```
alongd, line 1536: don't we need to know the node for troubleshooting by switching nodes?

arc/job/ssh.py
```diff
@@ -187,14 +187,16 @@ def download_file(self,
...
-            # This doesn't have a real impact now to avoid screwing up ESS trsh
-            # but introduce an opportunity for better troubleshooting.
-            # The current behavior is that if the remote path does not exist
-            # an empty file will be created at the local path
+            # For paramiko 2.7.1 and higher, if the remote file does not existed, then
+            # IOError ('[Errno 2] No such file) will be raised when calling `sftp.get()`
+            # The behavior in the previous versions is creating an empty file at `localpath`
             logger.debug(
                 f'{remote_file_path} does not exist on {self.server}.')
+            raise InputError(f'{remote_file_path} does not exist on {self.server}.')
```
alongd, line 198: I think we should raise `ServerError`
Are these errors being captured, or will they crash ARC?

arc/job/ssh.py
```diff
@@ -197,13 +197,24 @@ def download_file(self,
...
         try:
             self._sftp.get(remotepath=remote_file_path,
                            localpath=local_file_path)
-        except IOError:
+        except IOError as e:
+            # Catch IOError other than 'no such file' error
+            logger.debug(
+                f'Got an IOError ({e}) when trying to download file {remote_file_path}'
+                f'from {self.server}')
+            raise ServerError(f'Could not download file {remote_file_path} from {self.server}.'
```
alongd, line 209: add a line break at the end of  this line for readability when this is being printed. Same below

arc/job/ssh.py
```diff
@@ -216,6 +216,15 @@ def download_file(self,
             raise ServerError(f'Could not download file {remote_file_path} from {self.server}.'
                               f'Got: {e}.')
 
+        # Check if the file is correctly downloaded
+        if not os.path.isfile(local_file_path):
+            logger.debug(f'Cannot find the downloaded file {local_file_path}. SFTP seems to work, '
+                         f'but the file {remote_file_path} is not actually downloaded.')
```
alongd, line 222: "was not"

arc/job/job.py
```diff
@@ -1439,7 +1439,11 @@ def determine_job_status(self):
 
     def _get_additional_job_info(self):
         """
-        Download the additional information of stdout and stderr from the server.
+        Get the additional information of stdout and stderr according to the 
+        queue software outputs.
+
+        Returns:
```
alongd, line 1479: while revisiting this method, please add a type hint for the returned value

arc/job/job.py
```diff
@@ -1434,6 +1434,17 @@ def determine_job_status(self):
                                                           'time limit.'
                             self.job_status[1]['line'] = ''
                             break
+                        # example:
+                        # slurmstepd: *** JOB 12219482 CANCELLED AT 2020-05-03T01:25:00 DUE TO NODE node027 FAILURE ***
+                        if 'cancelled' in line.lower() and 'due to node' in line.lower() and 'failure' in line.lower():
+                            logger.warning(f'Looks like the job was cancelled on {self.server} due to node faliure. '
```
alongd, line 1444: line break at the end

arc/job/trsh.py
```diff
@@ -1158,7 +1158,7 @@ def trsh_job_on_server(server: str,
         insert_line_num = 5
     else:
         # Other software?
-        logger.denug(f'Unknown cluster software {cluster_soft} is encountered when '
+        logger.debug(f'Unknown cluster software {cluster_soft} is encountered when '
```
alongd, line 1161: let's change this to `logger.error`

alongd:
@xiaoruiDong, setting up a dummy server for the tests is an interesting point. If you have time, could you look into it? Perhaps we can see how paramiko tests itself.

alongd:
@xiaoruiDong, what's the status of this PR?

xiaoruiDong:
> @xiaoruiDong, what's the status of this PR?

@alongd Hi Alon, I am sorry that I haven't had time look at the dummy server things. For the branch, I've been using this branch to run my jobs for a while. And it works okay.

alongd:
OK, please rebase.
It could be nice to add these in the future if any of us ever have some free time... I know we're all tied up.

arc/job/job.py
xiaoruiDong: Based on the current workflow, changing node actually happens in the function `Job.troubleshoot_server()`. (line 1535 in arc/job/job.py). A (possible) available is picked by `trsh_job_on_server` in this method. We don't need to know the current node, since it is saved in `self.server_nodes`, we don't need to know what nodes are available in prior, since `trsh_job_on_server` will find one. The `server_status` here is only used to assign a value to `self.job_status[0]` and does not play a role in switching nodes. However, if I miss something, please let me know. Thanks!

arc/job/ssh.py
xiaoruiDong: I think this is a good question. The answer depends on how we see this situation. If we regard this function as a single API, and we are using this API to download stuff. Then if `remote_file_path` does not exist on the server, it simply means we input invalid arguments, it is not due to a malfunction of this method/class/connection. However, in an ARC job, this function is run since we think there should be a path there, and we want to download it, but the path actually does not exist. Still, I would say this is an error that is not due to a malfunction of this method/class/connection, so I think it is inappropriate to say it is a `ServerError`. It is more likely that some previous step goes into an undesired way. I would say let some upper-level function catch the `InputError` and re-raise this error as a `JobError`. And this is what I did in the `Job` (in `Job._download_output_file`). Then, this `JobError` will be caught by `job.determine_job_status` (two levels upper), and `job.determine_job_status` will analyze whether we can fix this or not. If we can, we submit a new job; if we cannot, we mark this job as `errored`.

arc/job/job.py
xiaoruiDong: Yeah, I did. See line 1446. The returned item is a `str` which can be either empty str `''` or very long str which combines all of the lines from the files downloaded (basically, files like *.out and *.err).

xiaoruiDong:
Both Travis tests fail due to RMG issue (#2011, https://github.com/ReactionMechanismGenerator/RMG-Py/issues/2011), which is not related to changes in this PR. Tests need to rerun when that issue is solved.

alongd:
@xiaoruiDong , please rebase again and tag me so I'll merge. Thanks.

PR #393, opened 2020-05-07
Allow user to input paths to previous jobs

### LINKED ISSUE #347: How to use frequencies from old ARC jobs
**Describe what you are trying to achieve**
I am currently trying to calculate some species thermochemistry using DLPNO in ARC. I have previously calculated the geometry/frequency for the species at a different level of theory though. Is there a way I can notify ARC of this information so that it can use it during the Arkane calculation?

There are two scenarios where I can see this happening, so this may change the approach.

1. When the geometry and frequency have been calculated by a prior ARC job
2. When the geometry and frequency have been calculated outside of ARC

amarkpayne:
fixes #347 

There are many times when a user might want to calculate a species at a higher sp level using a previously calculated geometry (for example, when updating the reference database). When this is the case, it does not make sense for ARC to recalculate the geometry. ARC will submit the new sp job, but cannot submit the Arkane job because it does not know the paths to the old geometry/frequency job.

The additions in this PR allow the user to pass a dictionary with species labels as keys, and `geo` and `freq` paths in a dictionary as values. ARC then adds these species paths if given to the output dictionary before processing the species.

### DIFF
--- a/arc/processor.py
+++ b/arc/processor.py
@@ -49,6 +49,7 @@ def process_arc_project(statmech_adapter: str,
                         lib_long_desc: str = '',
                         rmg_database: Type[RMGDatabase] = None,
                         compare_to_rmg: bool = True,
+                        previous_job_paths: dict = None,
                         ) -> None:
     """
     Process an RMG project, generate thermo and rate coefficients using statistical mechanics (statmech).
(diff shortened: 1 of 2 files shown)


amarkpayne:
@alongd let me know what you think of the implementation. Also, let me know where this should be added to the documentation.

I have tested this out in my own work, and I think it is working as desired

amarkpayne:
@alongd hold off on merging this, there is one additional change that I would like to make

arc/processor.py
```diff
@@ -75,7 +76,22 @@ def process_arc_project(statmech_adapter: str,
...
+            if spcs_label in previous_job_paths.keys():
+                if 'geo' in previous_job_paths[spcs_label].keys():
+                    geo_path = previous_job_paths[spcs_label]['geo']
+                    output_dict[spcs_label]['paths']['geo'] = geo_path
+                    print(f'Geometry path {geo_path} added for species {spcs_label}')
+                if 'freq' in previous_job_paths[spcs_label].keys():
+                    freq_path = previous_job_paths[spcs_label]['freq']
+                    output_dict[spcs_label]['paths']['freq'] = freq_path
+                    print(f'Frequency path {freq_path} added for species {spcs_label}')
+
```
xiaoruiDong, line 94: Just curious, when shall we use different logs for `geo` and `freq`? I used to thought we can use results from `freq` for both `freq` and `geo`.

alongd: I think we use `freq` only for both geo and freqs, but that's done at the adapter level, processor stores all our paths and passes them along.

xiaoruiDong: So, it seems like, we don't have to assign `geo` and `freq` separately here?

amarkpayne: I can just pass `freq` if you want as long as there is no reason why the user we need to specify a separate `geo` file (I can't think of any reason) and as long as ARC won't freak out about not having a file for `geo`

alongd: let's accept both, but if `geo` isn't given, override it with freq. Does that sound reasonable?

PR #395, opened 2020-05-12
Fix missing path debug

amarkpayne:
I caught a typo in the messages the Arkane adapter returns when file paths are missing. While I was at it, I figured I would quickly refactor this code so that all missing file paths could be displayed if needed.

If you don't like the refactoring, I am fine with excluding all but the typo fix.


alongd:
@amarkpayne, what is the status of this PR?

PR #397, opened 2020-05-23
Adapting the level of theory to Arkane's LevelOfTheory class

alongd:
This is a temporary fix, eventually we'd like to exclusively use Arkane's LOT representations


alongd:
replaced by #399

alongd:
replaced by #399

PR #399, opened 2020-05-26
Organize the level of theory representation in ARC

alongd:
Added a `Level` class that converts easily to the new Arkane `LevelOfTheory` class to determine freq scale factor and energy corrections.
This class also organizes the different functions related to the level of theory.

This PR also introduces significant changes to `main`, resulting in changes to ARC's API. In particular:
- removed input/restart file load and argument checks, now an input file and the API are indistinguishable in `main` (removed the from_dict method).
- non-backward compatible API changes:
  - `species` (was `arc_species_list`),
  - `reactions` (was `arc_rxn_list`),
  - `arkane_level_of_theory` (was `model_chemistry`, now only refers to the sp level, not freq level),
  - removed `job_additional_options` and `job_shortcut_keywords`, they are included in the `Level` object as an `args` argument
  - removed `solvation`, now included in the `Level` object which can be specified per job  type
  - `thermo_adapter` and `kinetics_adapter` were added, replacing the general `statmech_adapter`
  - `bac_type` specifies `'p'` or `'m'` BAC type (could be `None` for not using BACs), replaced the boolean flag `use_bac`
- changed how level of theories per job type are handled


arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+    """
+    Uniquely defines the settings used for a quantum calculation.
+
+    Args:
+        repr (str, dict, optional): A dictionary or a simple string representation of the level of theory,
+                                    e.g. "wb97xd/def2-tzvp", or {'method': 'b3lyp', 'basis': '6-31g'}.
+        method (str, optional): Quantum chemistry method.
+        basis (str, optional): Basis set.
+        auxiliary_basis (str, optional): Auxiliary basis set for correlated methods.
+        cabs (str, optional): Complementary auxiliary basis set for F12 calculations.
```
oscarwumit, line 29: what about orca's `normalPNO`, `tightPNO` options? Are we going to include that in `args` or we should make an attribute for that?

oscarwumit: Maybe we also need an attribute for empirical dispersion of DFT, sometimes dispersion is not included in the method name. like b3lyp with gd3bj.

arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+
+class Level(object):
+    """
+    Uniquely defines the settings used for a quantum calculation.
+
+    Args:
+        repr (str, dict, optional): A dictionary or a simple string representation of the level of theory,
+                                    e.g. "wb97xd/def2-tzvp", or {'method': 'b3lyp', 'basis': '6-31g'}.
+        method (str, optional): Quantum chemistry method.
+        basis (str, optional): Basis set.
```
oscarwumit, line 27: @dranasinghe once told me that some researchers use their own basis. Maybe the basis attribute can also point to a basis set file?

arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+        auxiliary_basis (str, optional): Auxiliary basis set for correlated methods.
+        cabs (str, optional): Complementary auxiliary basis set for F12 calculations.
+        method_type (str, optional): The level of theory method type (DFT, wavefunction, force field, semi-empirical, or composite).
+        software (str, optional): Quantum chemistry software.
+        software_version (Union[int, float, str], optional): Quantum chemistry software version.
+        solvation_method (str, optional): Solvation method.
+        solvent (str, optional): Solvent.
```
oscarwumit, line 34: What about co-solvents with different vol/mol fraction? Maybe consider something like:
[('methanol', 0.9), ('water', 0.1), ('fraction_type', 'volume[mole]')]
or make a dictionary?

arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+        solvation_scheme_level (Level, optional): A Level class representing the level of theory to calculate a
+                                                  solvation energy correction at.
```
oscarwumit, line 36: Another thought on solvation. For PCM methods, it's possible to specify a single dielectric constant. For SMD methods, it's possible to specify the six solvation parameters in an ESS's input file (or maybe let the ESS read some external file with these parameters like how costume basis works). Maybe we should have some attributes to store these solvent properties/info?

arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+        if self.software is not None:
+            str_ += f', software: {self.software}'
+            if self.software_version is not None:
+                str_ += f', software_version: {self.software_version}'
+        if self.args is not None:
+            str_ += f', args: {self.args}'
+        if self.method_type is not None:
+            str_ += f' ({self.method_type})'
+        return str_
+
```
oscarwumit, line 131: for dlpno, info like `normalPNO`, `tightPNO` might also be useful to print

arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+                self.software = 'qchem'
+            elif 'orca' in supported_ess:
+                self.software = 'orca'
+            elif 'terachem' in supported_ess:
+                self.software = 'terachem'
+        if self.software is None and job_type == 'irc':
+            if 'gaussian' not in supported_ess:
+                raise ValueError(f'Could not find Gaussian to run {job_type}')
+            self.software = 'gaussian'
+        if self.software is None and self.method == 'hf':
```
oscarwumit, line 315: Please check out #398 
 and https://github.com/ReactionMechanismGenerator/ARC/pull/398/commits/db67a5693c9d7da83eb1fe8e89cc68a7c95ad4b0
We should aim to replace long lines of if-else rules with simple functions for maintainability and extensibility.

arc/level.py
```diff
@@ -0,0 +1,397 @@
...
+        Determine the type of a model chemistry:
+        DFT, wavefunction, force field, semi-empirical, or composite
+        """
+        wave_function_methods = ['hf', 'cc', 'ci', 'mp2', 'mp3', 'cp', 'cep', 'nevpt', 'dmrg', 'ri', 'cas', 'ic', 'mr',
+                                 'bd', 'mbpt']
+        semiempirical_methods = ['am', 'pm', 'zindo', 'mndo', 'xtb', 'nddo']
+        force_field_methods = ['amber', 'mmff', 'dreiding', 'uff', 'qmdff', 'gfn', 'gaff', 'ghemical', 'charmm', 'ani']
+        # all composite methods supported by Gaussian
+        composite_methods = ['cbs-4m', 'cbs-qb3', 'rocbs-qb3', 'cbs-apno', 'w1u', 'w1ro', 'w1bd', 'g1', 'g2', 'g3',
+                             'g4', 'g2mp2', 'g3mp2', 'g3b3', 'g3mp2b3', 'g4mp2', 'cbs-qb3-paraskevas']
```
oscarwumit, line 236: Are we using `cbs-qb3-paraskevas` anywhere in ARC? I think this should be assigned to `arkane_model_chemistry` but not `level_of_theory` as it's not really a method keyword that ESS understands, it's just a flag for letting Arkane know which BAC to use. I think we should not mix `cbs-qb3-paraskevas` with actual qm methods.

arc/level.py
amarkpayne: Currently this is implemented in args. The args variable was meant for things like tolerances, which tightPNO sort of falls into.

Adding an attribute for dispersion I think could be a good idea. Right now b3lyp with gd3bj is listed as `method=b3lypd3bj`, but having it as a different attribute is more flexible. This would need to be implemented on RMG-Py

alongd: I think that can all go into `args`.
Maybe we can also deprecate the `job_shortcut_keywords` argument.
We could expand `args` to be more like `job_additional_options` which specified whether to add a key word to a line or a block.
What do you think?

arc/level.py
alongd: Interesting point. We'll need to know how to direct the ESS to use that file. Or do you simply add that file to a certain folder, and specify the file name as the basis set? (I think that's how TeraChem works)

arc/level.py
alongd: Will the ESS know how to interpret it?

arc/level.py
alongd: Excellent point, I agree. What would you recommend? (and, would you know how to implement in Gaussian?)

arc/level.py
alongd: if that's in `args`, it gets printed (see the tests for an example with a DFT dispersion)

arc/level.py
alongd: OK

arc/level.py
alongd: You're right

amarkpayne:
I haven't gotten a chance to look at the code too closely yet, but I am concerned about this implementation easily diverging from the one in RMG-Py. Of course, as you noted there is a problem because the RMG-Py class discards hyphens (it does this so that the methods are not necessarily software dependent).

I wonder if there is a way that we can use the RMG-Py objects but build an interface to them such that issues like hyphens are handled when ARC needs them. This shouldn't be an issue when trying to convert ARC representations to LOT (discarding hyphens is easy), but I think there should be a solution to go the other way as well.

One possible idea: What if we implement adapters between LOT and each quantum chemistry software? Some of these software adapters might be very simple if the software does not care about hyphens.

@cgrambow also has an interesting idea about using the __repr__ method to store things like `wb97X-D` with hyphens but that allow interfacing with Arkane. I'll let him comment here about it as well.

We should discuss this tomorrow during subgroup if you want, or setup a Zoom meeting

cgrambow:
Adding on to what @amarkpayne is saying, I also don't think it's necessary to implement yet another class for dealing with levels of theory; rather we could try to make the level of theory classes in RMG-Py suitable for ARC. This is what I was thinking:
- Implement `LevelOfTheory` and `CompositeLevelOfTheory` such that if you pass it, for example,  `method='wB97X-D'`, it will store the "non-standardized" string `wB97X-D` as the `method` attribute in `LevelOfTheory`.
- `LevelOfTheory.__repr__` can remain unchanged such that it will create the string representation based on the string that you pass it (e.g., the representation for the above example would be `LevelOfTheory(method='wB97X-D')`), which allows differentiating versions with/without hyphens by manipulating the attribute values.
- Re-implement the default `__eq__` and `__hash__` methods of `LevelOfTheory` such that they convert attributes to their standardized representations prior to comparisons. For example, this means that `LevelOfTheory(method='wB97X-D') == LevelOfTheory(method='wb97xd')` would evaluate to `True` and that `LevelOfTheory(method='wB97X-D')` and `LevelOfTheory(method='wb97xd')` are identical keys in dictionaries.

arc/level.py
oscarwumit: I think we should not mix level_of_theory with job_options, for this might gets level_of_theory out of hand quickly (too many variants). Instead, maybe we can consider the following

(1) make level_of_theory (LOT) object only contain essential info like method, basis etc. The software attribute of LOT can be a list, effectively representing compatible ESS.

e.g., Gaussian, Q-chem, Terachem all have b3lyp. Maybe we only really need one LOT for b3lyp at the higher up level with self.compatible_software = [Gaussian, Q-chem, Terachem]. Then we deal with ESS specific implementations using other modules (see point 2).

(2) implement factories to process specific job settings based on different ESS. Doing so free us from populating `args` with ESS specific settings. 

e.g., Gaussian, Q-chem, Terachem all allow the user to control DFT grid for b3lyp, but their keywords are different. Instead of creating three LOT objects with different `args` for each ESS, maybe we can create objects like `Gaussian_settings`, `Q-chem_settings`, and factories like `set_up_gaussian_job`, `set_up_qchem_job` that takes LOT and corresponding settings to convert the info to whatever it's needed in the input files.

arc/level.py
oscarwumit: Maybe we should just store the path to the basis set file here. And we can use a particular ESS factory to process it so that it can be used for that particular ESS.

arc/level.py
oscarwumit: In COSMO-RS, it accepts co-solvents with mol/vol fraction defined. But other ESS might not know how to interpret this. Maybe we should again store the info, and later let ESS factories to do the hard work of converting to correct input file format.

This information might also be useful for Arkane and  TCKDB purposes. Say someone computes a rate using a cosolvent, we should store this info in the RMG Arrhenius object, and TCKDB should be aware of such info.

arc/level.py
oscarwumit: Yes, @dranasinghe, @yunsiechung, and I have experience using costume PCM and SMD parameters in Gaussian.

oscarwumit:
@amarkpayne I think implementing adapters between LOT and each quantum chemistry software is a good idea and a potential way to clean up some long if-else rules in ARC. Also, we should also think about within each ESS, how to treat different job types. 

Currently, ARC uses if-else to treat job_type and ESS together, which is really hard to maintain.

Maybe to generate a job we can do: LOT --> ESS adaptor --> method for creating the input file for the required job_type for the particular ESS

arc/level.py
yunsiechung: @oscarwumit I have SMD and PCM solvent parameters for some species. We can store them and have Arc read it when it wants it?

oscarwumit: @yunsiechung @alongd Maybe we should store these parameters in TCKDB and let ARC reads it.

yunsiechung: @oscarwumit just let me know what kind of format it should be, and I can give you the file

cgrambow:
@oscarwumit @alongd Check out the modifications I just pushed to the `modify_lot` branch (https://github.com/ReactionMechanismGenerator/RMG-Py/tree/modify_lot) and see if they accomplish what you need. Hyphens and spaces are no longer dropped. This only occurs implicitly when comparing level of theory objects.

alongd:
@cgrambow @amarkpayne , sounds good!
We'll also need to implement additional methods and arguments to `LevelOfTheory` (see this updated PR):
- a clean dict representation to save in ARC's restart file
- a from_dict method
- `solvation_scheme_level` - when you do the solvation correction at a different level
- `method_type` - specifying whether a method is DFT, WF, MM etc.
- make `args` a dict with `keyword` and `block` keys
I'm happy to add all of these to Arkane.

arc/level.py
alongd: Let's store it in a YAML format either in ARC or Arkane (wherever the class ends up implemented)

arc/level.py
alongd: What do you mean LOT variants? The objects don't exist simultaneously if a user specifies a certain level (which is intended to be used in a specific ESS), the may have in mind ESS-specific keywords that should go along with this object. I don't think it complicates it, we just need to convert `args` from a tuple/list to a dict with the `keyword` and `block` keys. Let's talk offline to decide how to proceed.

alongd:
I take my above comment back:  There's no need to add all that stuff to Arkane which isn't being used in the RMG repo, instead we should set up a class in ARC that inherits from `LevelOfTheory` (made possible by @cgrambow's recent commit - let's PR it), and augment it with additional arguments and methods.

cgrambow:
@alongd I'm working on getting a PR up for the changes I made. Do you need the string representation of `LevelOfTheory` (called with `repr` or `str` and defined in `LevelOfTheory.__repr__`) to return the "nonstandard" representation, i.e., with hyphens, etc., or can that stay the same way as before?

alongd:
@oscarwumit, could you review the Level class again? Let's talk tomorrow.
@xiaoruiDong, could you take a look at this PR?

There are some API changes, see the edited first comment for details. I plan to modify the examples and scripts as needed, and modify the documentation.

arc/level.py
```diff
@@ -0,0 +1,506 @@
...
+        """
+        Determine the type of a model chemistry:
+        DFT, wavefunction, force field, semi-empirical, or composite
+        """
+        wave_function_methods = ['hf', 'cc', 'ci', 'mp2', 'mp3', 'cp', 'cep', 'nevpt', 'dmrg', 'ri', 'cas', 'ic', 'mr',
+                                 'bd', 'mbpt']
+        semiempirical_methods = ['am', 'pm', 'zindo', 'mndo', 'xtb', 'nddo']
+        force_field_methods = ['amber', 'mmff', 'dreiding', 'uff', 'qmdff', 'gfn', 'gaff', 'ghemical', 'charmm', 'ani']
+        # all composite methods supported by Gaussian
+        composite_methods = ['cbs-4m', 'cbs-qb3', 'cbs-qb3-paraskevas', 'rocbs-qb3', 'cbs-apno', 'w1u', 'w1ro', 'w1bd',
```
oscarwumit, line 394: 'cbs-qb3-paraskevas' is not really a method supported by gaussian

alongd: correct, but we need this here to comply with the definition in Arkane:
image

oscarwumit: I see. Then we need to make sure to convert this to `cbs-qb3` when we are writing the Gaussian input file. Btw, why don't we utilize the `arkane_level_of_theory` attribute when giving arkane this information? I am thinking in the ARC's input file, an user can specify cbs-qb3 as `method` and then cbs-qb3-paraskevas as `arkane_level_of_theory`. If we make `method` cbs-qb3-paraskevas, then isn't `arkane_level_of_theory` becomes redundant in this case?

alongd: This is exactly the intention, to define:
```python
composite_method = 'CBS-QB3'
arkane_level_of_theory = 'CBS-QB3-Paraskevas'
```
However, since `arkane_level_of_theory ` is in itself a `Level` object, not a `LevelOfTheory` object, we need to recognize it as a supported method in Gaussian (otherwise, we won't be able to pass the software info to Arkane w/o the user having to explicitly specify that).

I agree there might be a confusion where users might define `composite_method = 'CBS-QB3-Paraskevas'`. This isn't my intention, but I'll add a quick conversion in Job to treat cases like that so we always pass just 'CBS-QB3' to Gaussian.

arc/level.py
alongd: `dispersion` was made a separate argument as @amarkpayne suggested, perhaps could be updated in RMG-Py as well. As Discussed with @oscarwumit, job-specific keywords are now stored in args. Note that ARC's args are different that RMG's args (dict vs. list), a conversion method was implemented.

arc/level.py
alongd: Should be considered as a future feature

arc/level.py
alongd: As discussed with @oscarwumit, this too should be considered as a future feature.

arc/species/converter.py
```diff
@@ -1666,14 +1669,15 @@ def compare_confs(xyz1: dict,
 
 
 def ics_to_scan_constraints(ics: list,
-                            software: str = 'gaussian',
+                            software: Optional[str] = 'gaussian',
```
xiaoruiDong, line 1672: I don't think we should use `Optional` when the default value is not `None`

arc/utils/scaleTest.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-# encoding: utf-8
-
```
xiaoruiDong, line 3: ?

arc/job/trsh.py
```diff
@@ -535,9 +539,9 @@ def trsh_scan_job(label: str,
         scan_list (list): Entries are the four-atom scan lists (1-indexed) of all torsions
                            (without duplicate pivots) in this species.
         methods (dict): The troubleshooting method/s to try.
-                        Example:
-                        {'inc_res': None,
-                         'freeze': 'all' or [[1, 2, 3, 4], ...]}
+                        Example::
```
xiaoruiDong, line 540: double `:`

arc/job/trsh.py
```diff
@@ -1098,7 +1102,7 @@ def trsh_conformer_isomorphism(software: str,
 
 def trsh_job_on_server(server: str,
                        job_name: str,
-                       job_id: str,
+                       job_id: int,
```
xiaoruiDong, line 1107: Actually, this can be `Union[int, str]`. Please also change the docstring.

arc/level.py
```diff
@@ -0,0 +1,529 @@
...
+
+
+logger = get_logger()
+
+
+class Level(object):
+    """
+    Uniquely defines the settings used for a quantum calculation.
+    Its a dataclass inheriting from ``RMGObject``.
+
```
xiaoruiDong, line 26: Since all of the arguments are optional, can we add some instructions such as "Either repr or method must be specified", so that people have a better sense of how to use this API.

arc/level.py
```diff
@@ -0,0 +1,529 @@
...
+        self.lower()
+
+        if self.method_type is None:
+            self.deduce_method_type()
+        if self.dispersion is not None and self.method_type not in ['dft', 'composite']:
+            raise ValueError(f'Dispersion is only allowed for DFT (or composite) methods, got {self.dispersion} '
+                             f'for {self.method} which is a {self.method_type}')
+        if self.software is None:
+            # it wasn't set bu the user, try determining it
```
xiaoruiDong, line 108: typo: `by`

ARC.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-# encoding: utf-8
-
```
xiaoruiDong, line 3: ?

arc/mainTest.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-# encoding: utf-8
-
```
xiaoruiDong, line 3: ?

arc/job/ssh.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-# encoding: utf-8
-
```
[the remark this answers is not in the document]
xiaoruiDong: I found a lot of shebangs are removed? Is there a reason to do so?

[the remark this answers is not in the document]
alongd: A sheabang tells a Unix-like OS what mechanism to use to parse a file as an executable. Since all files in ARC except for ARC.py are not executable, we don't need these lines. I removed them in most files I touched during this PR, I agree it's inconsistent, I'll add another commit to remove them from additional non-executable files. I'll leave them in our test files which are executable.

arc/job/trsh.py
alongd: The double `:` makes the literal block underneath look good in the compiled docs.

For example:
code: https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/species/converter.py#L128
docs: https://reactionmechanismgenerator.github.io/ARC/api/converter.html#arc.species.converter.xyz_to_str

see: https://docutils.sourceforge.io/docs/user/rst/cheatsheet.txt

arc/job/trsh.py
alongd: OK

arc/level.py
alongd: excellent point! will do

arc/level.py
alongd: I agree. Let's add this one as well in a future PR. I'd like to get this PR merged to sort out the current incompatibility with RMG.

arc/species/converter.py
alongd: I'm uncertain either.
From [the docs](https://docs.python.org/3/library/typing.html#typing.Optional) I understand that "if an explicit value of `None` is allowed, the use of Optional is appropriate". The don't relate to situations where `None` is allowed, but it's not the default. PyCharm, on the other hand (for what it's worth) will flag cases where we pass `None` to an argument not defined as Optional. I'm keeping it to reduce these flagging by PyCharm, if you strongly feel otherwise I'm also OK with removing it.

arc/utils/scale.py
```diff
@@ -148,17 +149,17 @@ def calculate_truhlar_scaling_factors(zpe_dict: dict,
...
-        level_of_theory (str): The frequencies level of theory.
+        level (str): A string representation of the frequencies level of theory.
 
     Returns:
         float: The scale factor for the vibrational zero-point energy (lambda ZPE) as defined in reference [2].
     """
     unconverged = [key for key, val in zpe_dict.items() if val is None]
     if len(unconverged):
-        logger.info(f'\n\nWarning: Not all species in the standard set have converged at the {level_of_theory} '
+        logger.info(f'\n\nWarning: Not all species in the standard set have converged at the {level} '
```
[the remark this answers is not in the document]
alongd: I wanted to add two clean `\n` w/o the "Warning:" prefix. The alternative would be to replace this line with three logger calls (`info('\n\n')`, `warning(message)`, `info('\n\n')`)

PR #401, opened 2020-06-06
Feature: Enhance conformer comparison with RMSD score

oscarwumit:
Enable ARC to better compare two conformers using root-mean-square deviation (RMSD) of their distance matrices. Added function that uses RMSD to filter/cluster a large number of conformers for both TS and non-TS geometries.


arc/species/converter.py
```diff
@@ -1656,13 +1657,19 @@ def compare_confs(xyz1: dict,
...
         bool: Whether the two conformers have almost equal atom distances. ``True`` if they do.
     """
     xyz1, xyz2 = check_xyz_dict(xyz1), check_xyz_dict(xyz2)
     dmat1, dmat2 = xyz_to_dmat(xyz1), xyz_to_dmat(xyz2)
-    return almost_equal_lists(dmat1, dmat2, rtol=rtol, atol=atol)
+    rmsd = None
+    if rmsd_score:
+        # distance matrix is symmetric, only need the upper triangular part to compute rmse
+        rmsd = calc_rmsd(np.triu(dmat1), np.triu(dmat2))
+    return almost_equal_lists(dmat1, dmat2, rtol=rtol, atol=atol), rmsd
```
alongd, line 1671: Do you think we'll need both outputs simultaneously?
Maybe make the flag return either a bool or RMSD?

oscarwumit: OK

oscarwumit:
Thanks for the comments, please take a look at the fixup commits. @alongd

oscarwumit:
@alongd  I just added a function to cluster conformers using RMSD. Please review that as well. Thanks.

xiaoruiDong:
Looks good to me. @alongd, can you restart the PR travis test? Thanks.

arc/species/converter.py
```diff
@@ -1644,7 +1644,8 @@ def compare_confs(xyz1: dict,
                   xyz2: dict,
                   rtol: float = 1e-5,
                   atol: float = 1e-5,
-                  ) -> bool:
+                  rmsd_score: bool = False,
+                  ) -> Union[Optional[float], bool]:
```
alongd, line 1648: Is the `float` really "`Optional`"? Can it be `None`?

arc/species/converter.py
```diff
@@ -1656,13 +1657,21 @@ def compare_confs(xyz1: dict,
         xyz2 (dict): Conformer 2.
         rtol (float): The relative tolerance parameter (see Notes).
         atol (float): The absolute tolerance parameter (see Notes).
+        rmsd_score (bool): Whether to output a root-mean-square deviation score of the two distance matrices.
 
+    Returns:
```
alongd, line 1662: Should be:
```
Returns:
    Union[Optional[float], bool]:  # reconsider "Optional"
        - If ``rmsd_score`` is ``False`` (default): Whether the two conformers have almost equal atom distances.
          ``True`` if they do.
        - If ``rmsd_score`` is ``True``: The RMSD score of two distance matrices.
```

arc/species/converter.py
```diff
@@ -1692,6 +1692,27 @@ def calc_rmsd(x: np.array,
     rmsd = np.sqrt(sqr_sum/n)
     return float(rmsd)
 
+def cluster_confs_by_rmsd(xyzs: Iterable[dict],
```
alongd, line 1695: Maybe do `Iterable[Dict[str, tuple]]`

arc/species/converter.py
```diff
@@ -1692,6 +1692,27 @@ def calc_rmsd(x: np.array,
...
+                          rmsd_threshold: float = 1e-2,
+                          ) -> Tuple[dict]:
+    """
+    Cluster conformers using RMSD of distance matrices.
+
+    Args:
+        xyzs (Iterable): Conformers with the same atom orders.
+        rmsd_threshold (float): The minimum RMSD to consider two conformers as distinctive
```
alongd, line 1703: I think "as distinct" rather than "as distinctive"

arc/species/converter.py
```diff
@@ -1692,6 +1692,27 @@ def calc_rmsd(x: np.array,
...
+                                (i.e., if rmsd > rmsd_threshold, then two conformers are considered distinctive).
+
+    Returns:
+        Tuple[dict]: Conformers with distinctive geometries.
```
alongd, line 1707: `Tuple[Dict[str, tuple]]`

arc/species/converter.py
```diff
@@ -1692,6 +1692,27 @@ def calc_rmsd(x: np.array,
...
+    """
+    xyzs = tuple(xyzs)
+    distinct_xyzs = [xyzs[0]]
+    for xyz in xyzs:
+        rmsd = [compare_confs(xyz, distinct_xyz, rmsd_score=True) for distinct_xyz in tuple(distinct_xyzs)]
```
alongd, line 1712: rename to or `rmsd_list`

arc/species/converter.py
```diff
@@ -1692,6 +1692,27 @@ def calc_rmsd(x: np.array,
...
+        if all([xyz_rmsd > rmsd_threshold for xyz_rmsd in tuple(rmsd)]):
```
alongd, line 1713: Maybe rename `xyz_rmsd` to `rmsd`?

alongd, line 1713: Why is it important to do `tuple(rmsd)`?

arc/species/converter.py
```diff
@@ -1692,6 +1692,27 @@ def calc_rmsd(x: np.array,
...
+
+    Args:
+        xyzs (Iterable): Conformers with the same atom orders.
+        rmsd_threshold (float): The minimum RMSD to consider two conformers as distinctive
+                                (i.e., if rmsd > rmsd_threshold, then two conformers are considered distinctive).
+
+    Returns:
+        Tuple[dict]: Conformers with distinctive geometries.
+    """
+    xyzs = tuple(xyzs)
```
alongd, line 1709: Why is it important to convert it into a tuple? The function never attempts to modify `xyzs`

arc/species/converter.py
alongd, line 1695: I'd like to rethink together what is the intended behaviour of this function.
Do you think we should really "cluster" them, i.e., identify clusters and all conformers that belong to a single cluster, then **average the coordinates**?

Think of the following case:
```
RMSD_threshold = 0.5
RMSD(C0, C1) = 0.4
RMSD(C1, C2) = 0.4, RMSD(C0, C2) = 1
RMSD(C2, C3) = 0.4
RMSD(C3, C4) = 0.4, RMSD(C0, C4) = 1, RMSD(C0, C2) = 1
RMSD(C4, C5) = 7
```

Will this algorithm probably end up with [C0, C2, C4, C5]?
What is the intended behaviour? Is it not to identify C0-C4 as a single well, and return [<A0-4>, C5]?
(where <A0-4> is the average of conformers C0, C1, C2, C3, C4)

alongd, line 1695: Where would we like to implement these functions? In `converter` or in `conformers`?
I'd like the engines to be in `conformers`, and the drivers to be anywhere convenient that makes sense.

arc/species/converter.py
```diff
@@ -1664,6 +1664,25 @@ def compare_confs(xyz1: dict,
...
+              ) -> float:
+    """
+    Compute the root-mean-square deviation between two matrices.
+
+    Args:
+        x (np.array): Matrix 1.
+        y (np.array): Matrix 2.
+
+    Returns:
+        float: the RMSD score of two matrices.
```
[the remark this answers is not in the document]
oscarwumit: ok

arc/species/converter.py
oscarwumit: you are right. it can't be none

arc/species/converter.py
oscarwumit: thanks

arc/species/converter.py
oscarwumit: ok

arc/species/converter.py
oscarwumit: ok

arc/species/converter.py
oscarwumit: ok

arc/species/converter.py
oscarwumit: ok

arc/species/converter.py
oscarwumit: ok

oscarwumit: you are right, it's not needed here. since we are doing list operations here, I convert it to tuple just to be safe.

arc/species/converter.py
oscarwumit: Because the input xyzs is an `iterable` object, meaning that it can be a list or tuple or something else. Here I convert it to be `tuple` so that we know what to expect later in the code. Also if you leave it as an `iterable`, PyCharm will complain when we do `xyzs[0]` because "Class 'Iterable' does not define '__getitem__', so the '[]' operator cannot be used on its instances"

arc/species/converter.py
oscarwumit: Let's leave the function in converter for now.

oscarwumit: Thanks for the discussion tonight, I have updated the doc string to indicate where this function is intended to be used.

oscarwumit:
@alongd Thanks for the comments and discussion. Please take a look at the changes.

arc/species/converter.py
```diff
@@ -1656,14 +1657,71 @@ def compare_confs(xyz1: dict,
...
+
+def cluster_confs_by_rmsd(xyzs: Iterable[Dict[str, tuple]],
+                          rmsd_threshold: float = 1e-2,
+                          ) -> Tuple[Dict[str, tuple]]:
+    """
+    Cluster conformers with the same atom orders using RMSD of distance matrices. Work for both TS and non-TS conformers.
+
+    Intended for finding structurally distinct conformers from a pool of conformers.
+    Suitable scenarios:
+        1. filter a pool of randomly generated conformers.
```
alongd, line 1704: let's remove the first point here

oscarwumit: ok

oscarwumit:
Thanks for the comments. I have updated the doc string. @alongd

PR #405, opened 2020-06-17
Minor: Optional output zmat after modify_coords

oscarwumit:
Zmat contains important connectivity information that xyz does not. Giving the option to output zmat instead of xyz after coordinates modification is useful to ensure connectivity information is preserved.

Zmat is particularly useful when `modify_coords` is called multiple times such as changing two dihedrals of a species by calling the function twice. Using xyz in this case may cause bugs as ARC needs to perceive connectivity from xyz twice, and the perceived connectivity might be inconsistent.

### DIFF
--- a/arc/species/converter.py
+++ b/arc/species/converter.py
@@ -847,6 +847,7 @@ def modify_coords(coords: Dict[str, tuple],
                   mol: Optional[Molecule] = None,
                   index: int = 0,
                   fragments: Optional[List[List[int]]] = None,
+                  output_zmat: bool = False,
                   ) -> Dict[str, tuple]:
     """
     Modify either a bond length, angle, or dihedral angle in the given coordinates.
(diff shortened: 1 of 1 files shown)


arc/species/converter.py
```diff
@@ -884,6 +886,8 @@ def modify_coords(coords: Dict[str, tuple],
 
     Returns:
         dict: The respective cartesian (xyz) coordinates reflecting the desired modification.
+    Returns:
+        dict: The internal (zmat) coordinates reflecting the desired modification.
```
xiaoruiDong, line 890: Minor: @alongd, I am not familiar with the convention. Do we write separate `Returns`, or we merge them together?

arc/species/converter.py
```diff
@@ -946,7 +950,10 @@ def modify_coords(coords: Dict[str, tuple],
             # a list of the two corresponding parameters was be returned
             zmat['vars'][param[0]] = new_value - sum(zmat['vars'][par] for par in param[1:])
         new_xyz = zmat_to_xyz(zmat=zmat)
-    return new_xyz
+    if output_zmat:
+        return zmat
+    else:
+        return new_xyz
```
xiaoruiDong, line 956: Minor: I suggest to put line 952 into the `else` section.

arc/species/converter.py
alongd: We should have just one returned type, it could be a composite type, e.g., https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/species/conformers.py#L739

oscarwumit:
Thanks for the suggestions. Please take another look. This PR is rebased.

alongd:
@oscarwumit, can you look at the tests?

alongd:
@oscarwumit , what's the status of this PR?

PR #408, opened 2020-07-08
BugFix: Only parsing the last frequencies from Gaussian log file

oscarwumit:
In certain gaussian jobs, frequencies are printed more than once. Previously the parser will erroneously record the duplicated frequencies. Now only the last frequencies output by Gaussian is recorded. A test is also added.

### DIFF
--- a/arc/parser.py
+++ b/arc/parser.py
@@ -47,6 +47,9 @@ def parse_frequencies(path: str,
         with open(path, 'r') as f:
             line = f.readline()
             while line != '':
+                # this line intends to only capture the last occurrence of the frequencies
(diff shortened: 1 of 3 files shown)


arc/parser.py
```diff
@@ -47,6 +47,8 @@ def parse_frequencies(path: str,
         with open(path, 'r') as f:
             line = f.readline()
             while line != '':
+                if 'and normal coordinates' in line:
```
alongd, line 50: Could you just add a comment that this lines intends to only capture the last occurence of the frequencies?

oscarwumit: ok

PR #419, opened 2020-08-16
Added an option to use external submit scrips

alongd:
Sometimes it's desired to use submit scripts stored outside of the ARC repository (e.g., when ARC is installed for a group of users on a server, and each user would like to customize the scripts used by ARC).
It is now possible to do so by passing a YAML file containing the desired scripts into the ``external_submit_scripts`` argument.
Documentation and an example were added.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -658,7 +658,7 @@ def sort_two_lists_by_the_first(list1: List[Union[float, int, None]],
     """
     Sort two lists in increasing order by the values of the first list.
     Ignoring None entries from list1 and their respective entries in list2.
-    The function was written in this format rather the more pytonic ``zip(*sorted(zip(list1, list2)))`` style
+    The function was written in this format rather the more pythonic ``zip(*sorted(zip(list1, list2)))`` style
     to accommodate for dictionaries as entries of list2, otherwise a
     ``TypeError: '<' not supported between instances of 'dict' and 'dict'`` error is raised.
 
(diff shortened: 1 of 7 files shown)


dranasinghe:
I would like to know your option on checking if path exit and if true use that setting.py and submit.py script. 
`arc_path = os.getenv("HOME")+'/.arc/'
if os.path.exists(arc_path +'settings.py'):
    sys.path.insert(1,arc_path)
    from settings import arc_path as default_path
else:
    from arc.settings import arc_path as default_path`
My idea is w check if .arc folder exits. we can have setting .py and submit.py there. if it exist we add .arc to PYTONPATH

alongd:
I didn't entirely understand the new suggestion, we should discuss it.
For example, do you think that `if os.path.exists(arc_path +'settings.py')` will ever be False? We must have the setting.py file for many other reasons.

dranasinghe:
If you don't have settings.py then it will be false and use the
arc.settings.

xiaoruiDong:
I think Duminda's idea is using some sort of second/external settings? In my view, it may be ideal to have a folder named `/template` and put all templates (things in submit.py and inputs.py) there. We can provide some default templates and people can also add their templates. Or something like `/template/default` and `/template/user_defined`, then if there are files with the same name in both folders, ARC will use the external one priorly. But whatever way we chose to go, we can still keep the changes Alon made here.

dranasinghe:
I agree!

alongd:
Closed in favor of #420.
Thank you all for the comments!

PR #431, opened 2020-10-10
Initiated jobs

dranasinghe:
completed_jobs.csv and initiated_jobs.csv  were located in ARC which can causes permission issues. I moved csv files to .arc folder. 
found a bug in job.py; replaced "local" with server names form settings.py

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -35,6 +35,7 @@ from arc.imports import settings
 logger = logging.getLogger('arc')
 
 arc_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))  # absolute path to the ARC folder
+local_arc_path = os.path.join(os.getenv("HOME"), '.arc')
 
 VERSION = '1.1.0'
 
--- a/arc/job/job.py
+++ b/arc/job/job.py
@@ -10,7 +10,7 @@ import shutil
 from pprint import pformat
 from typing import Dict, Optional, Union
 
-from arc.common import arc_path, get_logger
+from arc.common import arc_path, get_logger, local_arc_path
 from arc.exceptions import JobError, InputError
 from arc.imports import settings, input_files, submit_scripts
 from arc.job.local import (get_last_modified_time,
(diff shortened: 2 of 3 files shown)


arc/common.py
```diff
@@ -35,6 +35,7 @@
 logger = logging.getLogger('arc')
 
 arc_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))  # absolute path to the ARC folder
+local_arc_path = os.path.join(os.getenv("HOME"), '.arc')
```
alongd, line 38: We have this in https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/imports.py#L14
No need to have it here

arc/job/job.py
```diff
@@ -10,7 +10,7 @@
 from pprint import pformat
 from typing import Dict, Optional, Union
 
-from arc.common import arc_path, get_logger
+from arc.common import arc_path, get_logger, local_arc_path
```
alongd, line 13: import `local_arc_path` from `arc.imports`

arc/job/job.py
```diff
@@ -424,7 +424,7 @@ def _set_job_number(self):
         """
         Used as the entry number in the database, as well as the job name on the server.
         """
-        csv_path = os.path.join(arc_path, 'initiated_jobs.csv')
+        csv_path = os.path.join(local_arc_path, 'initiated_jobs.csv')
```
alongd, line 427: this means we'll always have a `.arc` folder locally. So by default, If I run from two machines on the same server, I'll get different job counters. I think this isn't the desired behaviour. Instead, I think we should only use the local arch path for the CSV files if the respective file (not folder) exists.

alongd (commented):
Thanks! Please take a look at the comments

arc/job/job.py
dranasinghe: . arc is per user. If I am not mistaken if I run from two different machines I get two different job counts. I was thinking about how it working on the delta. there are multiple users using the same ARC repo. Therefore the counter is for everyone.  I was afraid a permission error might occur when multiple people write to the same file. The CSV file was created automatically. Are you suggesting that we create a CSV file in the .arc folder and if the csv file exists then write to the .arc folder?

alongd: Yes, check if it exists in .arc first. Other users would like to keep a songle counter while submitting from different machines onto a single server.

PR #438, opened 2020-12-15
GCN TS

kspieks:
This PR furthers the goal of allowing ARC to automatically find TS and refine kinetic parameters in a high throughput fashion. It incorporates an updated version of the GCN published by Pattanaik et al. to quickly predict initial guesses for TS structures. These guesses will be further optimized at the level of theory provided by the user. Thanks to Lucky and Xiaorui for your collaboration :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,10 @@ jobs:
         - cd AutoTST
         - export PYTHONPATH=$PYTHONPATH:$(pwd)
         - cd ..
+        - cd ARC 
+        - make gcn-travis
         - conda install -y -c conda-forge codecov
         - conda list
-        - cd ARC
       script:
         - make test
         - codecov --token=f259713a-7f1d-4e9c-b140-bb3bb371d3ef
--- a/arc/job/ssh.py
+++ b/arc/job/ssh.py
@@ -275,11 +275,11 @@ class SSHClient(object):
         cmd = check_status_command[servers[self.server]['cluster_soft']] + ' -u $USER'
         stdout = self._send_command_to_server(cmd)[0]
         for i, status_line in enumerate(stdout):
-            if servers['local']['cluster_soft'].lower() == 'slurm' and i > 0:
+            if servers[self.server]['cluster_soft'].lower() == 'slurm' and i > 0:
                 running_jobs_ids.append(int(status_line.split()[0]))
-            elif servers['local']['cluster_soft'].lower() in ['oge', 'sge'] and i > 1:
+            elif servers[self.server]['cluster_soft'].lower() in ['oge', 'sge'] and i > 1:
                 running_jobs_ids.append(int(status_line.split()[0]))
-            elif servers['local']['cluster_soft'].lower() == 'pbs' and i > 4:
+            elif servers[self.server]['cluster_soft'].lower() == 'pbs' and i > 4:
                 running_jobs_ids.append(int(status_line.split('.')[0]))
         return running_jobs_ids
 
@@ -308,13 +308,13 @@ class SSHClient(object):
                 if 'Requested node configuration is not available' in line:
                     logger.warning(f'User may be requesting more resources than are available. Please check server '
                                    f'settings, such as cpus and memory, in ARC/arc/settings/settings.py')
-        elif servers['local']['cluster_soft'].lower() in ['oge', 'sge'] and 'submitted' in stdout[0].lower():
+        elif servers[self.server]['cluster_soft'].lower() in ['oge', 'sge'] and 'submitted' in stdout[0].lower():
             job_id = int(stdout[0].split()[2])
             job_status = 'running'
-        elif servers['local']['cluster_soft'].lower() == 'slurm' and 'submitted' in stdout[0].lower():
+        elif servers[self.server]['cluster_soft'].lower() == 'slurm' and 'submitted' in stdout[0].lower():
             job_id = int(stdout[0].split()[3])
             job_status = 'running'
-        elif servers['local']['cluster_soft'].lower() == 'pbs':
+        elif servers[self.server]['cluster_soft'].lower() == 'pbs':
             job_id = int(stdout[0].split('.')[0])
             job_status = 'running'
         else:
--- a/arc/reaction.py
+++ b/arc/reaction.py
@@ -4,6 +4,7 @@ A module for representing a reaction.
 
 from typing import List, Optional
 
+import numpy as np
 from qcelemental.exceptions import ValidationError
 from qcelemental.models.molecule import Molecule as QCMolecule
 
@@ -14,7 +15,7 @@ import arc.rmgdb as rmgdb
 from arc.common import extermum_list, get_logger
 from arc.exceptions import ReactionError, InputError
 from arc.imports import settings
-from arc.species.converter import xyz_to_str
+from arc.species.converter import str_to_xyz, xyz_to_str
 from arc.species.species import ARCSpecies, check_atom_balance
 
 
@@ -670,3 +671,29 @@ class ARCReaction(object):
             data = products.align(ref_mol=reactants, verbose=verbose)[1]
             atom_map = data['mill'].atommap.tolist()
         return atom_map
+
+    def get_mapped_product_xyz(self):
+        """
+        Uses the mapping from product onto reactant to create a new ARC Species for the mapped product.
+        For now, only functional for A <=> B reactions.
+
+        Returns:
+            Tuple[dict, ARCSpecies]:
+                dict: Mapped product atoms in ARC dictionary format.
+                ARCSpecies: ARCSpecies object created from mapped coordinates.
+        """
+        product = self.p_species[0]
+        mapping = self.atom_map
+        original_xyz_str = xyz_to_str(product.get_xyz())
+        original_xyz = np.array(original_xyz_str.split('\n'))
+        mapped_xyz = '\n'.join(original_xyz[mapping].tolist())
+        # create another product with the mapped atoms since ARCSpecies objects preserve this ordering
+        # in RDKit objects and in RMG-Py Molecule objects
+        mapped_product = ARCSpecies(label=product.label,
+                                    smiles=product.mol.smiles,
+                                    multiplicity=product.multiplicity,
+                                    charge=product.charge,
+                                    xyz=mapped_xyz,
+                                    )
+        mapped_xyz = str_to_xyz(mapped_xyz)
+        return mapped_xyz, mapped_product
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -352,6 +352,17 @@ class Scheduler(object):
                             rmg_reaction=rxn.rmg_reaction,
                         )
                     )
+                # now generate TSGuess objects for the other methods
+                for method in rxn.ts_methods:
+                    if 'user guess' not in method:
+                        ts_guess = TSGuess(
+                            method=method,
+                            rmg_reaction=rxn.rmg_reaction,
+                            arc_reaction=rxn,
+                            project_dir=self.project_directory
+                        )
+                        ts_guess.execute_ts_guess_method()
+                        ts_species.ts_guesses.append(ts_guess)
                 rxn.check_atom_balance()
 
         for species in self.species_list:
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -6,6 +6,8 @@ If the species is a transition state (TS), its ``ts_guesses`` attribute will hav
 import datetime
 import numpy as np
 import os
+from rdkit import Chem
+import time
 from typing import List, Optional, Tuple, Union
 
 import rmgpy.molecule.element as elements
@@ -48,7 +50,7 @@ from arc.species.converter import (check_isomorphism,
                                    xyz_from_data,
                                    xyz_to_str)
 from arc.species.vectors import calculate_distance
-from arc.ts import atst
+from arc.ts import atst, gcn
 
 
 logger = get_logger()
@@ -1732,6 +1734,7 @@ class TSGuess(object):
     - NEB (not implemented)
     - Kinbot: requires the RMG family, only works for H,C,O,S (not implemented)
     - AutoTST
+    - GCN
 
     Args:
         method (str, optional): The method/source used for the xyz guess.
@@ -1742,8 +1745,10 @@ class TSGuess(object):
         family (str, optional): The RMG family that corresponds to the reaction, if applicable.
         xyz (dict, str, optional): The 3D coordinates guess.
         rmg_reaction (Reaction, optional): An RMG Reaction object.
+        arc_reaction (ARCReaction, optional): An ARC Reaction object.
         ts_dict (dict, optional): A dictionary to create this object from (used when restarting ARC).
-        energy (float): Relative energy of all TS conformers in kJ/mol.
+        energy (float, optional): Relative energy of all TS conformers in kJ/mol.
+        project_dir (str, optional): Folder path for the project: the input file path or ARC/Projects/project-name.
 
     Attributes:
         initial_xyz (dict): The 3D coordinates guess.
@@ -1755,12 +1760,14 @@ class TSGuess(object):
                              (product label, product xyz).
         family (str): The RMG family that corresponds to the reaction, if applicable.
         rmg_reaction (Reaction): An RMG Reaction object.
+        arc_reaction (ARCReaction, optional): An ARC Reaction object.
         t0 (float): Initial time of spawning the guess job.
         execution_time (str): Overall execution time for the TS guess method.
         success (bool): Whether the TS guess method succeeded in generating an XYZ guess or not.
         energy (float): Relative energy of all TS conformers in kJ/mol.
         index (int): An index corresponding to the conformer jobs spawned for each TSGuess object.
                      Assigned only if self.success is ``True``.
+        project_dir (str): Folder path for the project: the input file path or ARC/Projects/project-name.
 
     """
     def __init__(self,
@@ -1770,8 +1777,11 @@ class TSGuess(object):
                  family: Optional[str] = None,
                  xyz: Optional[Union[dict, str]] = None,
                  rmg_reaction: Optional[Reaction] = None,
+                 arc_reaction: Optional = None,
                  ts_dict: Optional[dict] = None,
-                 energy: Optional[float] = None):
+                 energy: Optional[float] = None,
+                 project_dir: Optional[str] = None,
+                 ):
 
         if ts_dict is not None:
             # Reading from a dictionary
@@ -1786,6 +1796,7 @@ class TSGuess(object):
             self.process_xyz(xyz)  # populates self.initial_xyz
             self.success = None
             self.energy = energy
+            self.project_dir = project_dir
             self.method = method.lower() if method is not None else 'user guess'
             if 'user guess' in self.method:
                 if self.initial_xyz is None:
@@ -1795,10 +1806,11 @@ class TSGuess(object):
             self.reactants_xyz = reactants_xyz if reactants_xyz is not None else list()
             self.products_xyz = products_xyz if products_xyz is not None else list()
             self.rmg_reaction = rmg_reaction
+            self.arc_reaction = arc_reaction
             self.family = family
             # if self.family is None and self.method.lower() in ['kinbot', 'autotst']:
             #     raise TSError('No family specified for method {0}'.format(self.method))
-        if not ('user guess' in self.method or 'autotst' in self.method
+        if not ('user guess' in self.method or 'autotst' in self.method or 'gcn' in self.method
                 or self.method in ['user guess'] + [tsm.lower() for tsm in default_ts_methods]):
             raise TSError('Unrecognized method. Should be either {0}. Got: {1}'.format(
                           ['User guess'] + default_ts_methods, self.method))
@@ -1900,6 +1912,8 @@ class TSGuess(object):
             self.kinbot()
         elif self.method == 'autotst':
             self.autotst()
+        elif self.method == 'gcn':
+            self.gcn()
         else:
             raise TSError('Unrecognized method. Should be either {0}. Got: {1}'.format(
                           ['User guess'] + default_ts_methods, self.method))
@@ -1921,6 +1935,52 @@ class TSGuess(object):
             self.initial_xyz = None
         else:
             self.initial_xyz = atst.autotst(rmg_reaction=self.rmg_reaction, reaction_family=self.family)
+
+    def gcn(self):
+        """
+        Determine a TS guess using a graph convolutional network.
+
+        Reactant and product must each be a single fragment.
+        """
+        # check that this is an isomerization reaction i.e. only one reactant and one product
+        num_reactants = len(self.arc_reaction.r_species)
+        num_products = len(self.arc_reaction.p_species)
+
+        if num_reactants > 1:
+            logger.error(f'Error while using GCN with reactants: {self.arc_reaction.r_species}.\n'
+                         f'Isomerization reactions must have only 1 reactant.')
+        elif num_products > 1:
+            logger.error(f'Error while using GCN with products: {self.arc_reaction.p_species}.\n'
+                         f'Isomerization reactions must have only 1 product.')
+        else:
+            # run GCN
+            reactant = self.arc_reaction.r_species[0]
+            _, r_mol = rdkit_conf_from_mol(reactant.mol, reactant.get_xyz())
+
+            # GCN requires an atom-mapped reaction so map the product atoms onto the reactant atoms
+            _, mapped_product = self.arc_reaction.get_mapped_product_xyz()
+            _, p_mol = rdkit_conf_from_mol(mapped_product.mol, mapped_product.get_xyz())
+
+            # write input files for GCN to the TS project folder
+            ts_path = os.path.join(self.project_dir, 'calcs', 'TSs', self.arc_reaction.ts_label, 'GCN')
+            os.makedirs(ts_path, exist_ok=True)
+
+            r_path = os.path.join(ts_path, 'reactant.sdf')
+            w = Chem.SDWriter(r_path)
+            w.write(r_mol)
+            w.close()
+
+            p_path = os.path.join(ts_path, 'product.sdf')
+            w = Chem.SDWriter(p_path)
+            w.write(p_mol)
+            w.close()
+
+            start = time.time()
+            ts_xyz_dict = gcn.gcn(ts_path)
+            end = time.time()
+            self.execution_time = end - start  # seconds
+            self.success = False if ts_xyz_dict is None else True
+            self.initial_xyz = ts_xyz_dict
 
     def qst2(self):
         """
--- a/arc/ts/gcn.py
+++ b/arc/ts/gcn.py
@@ -1,0 +1,61 @@
+"""
+A module for calling TS-GEN
+
+Notes for possible future developments:
+- Currently, ARC will only use the network for unimolecular reactions (A --> B). The network was trained on both A --> B
+  and A --> B + C reactions. However, the products must be oriented as if they just reacted. They can't be arbitrarily
+  placed in 3D space. ARC would have to automatically orient the two molecules correctly before using the network for
+  more reaction families.
+
+- The network was trained on geometries optimized at ωB97X-D3 with a def2-TZVP basis set. For now, ARC runs the GCN
+  using force field optimized geometries from RDKit. The TS guesses have looked reasonable, though in the future, it may
+  be beneficial to use the final QM optimized geometries as input to the network.
+
+- This network is not symmetric so it's possible that reversing the reaction direction could impact the convergence of
+  the TS guess. For example, it might be beneficial to call the network on A --> B and on B --> A to generate two TS
+  guesses, though the impact on convergence has not been studied.
+
+"""
+
+import os
+import subprocess
+
+from arc.common import get_logger
+from arc.common import TS_GCN_PATH, TS_GCN_PYTHON
+from arc.species.converter import str_to_xyz
+
+logger = get_logger()
+
+
+def gcn(ts_path,
+        ):
+    """
+    Generates a TS guess using the updated graph convolutional network originally published by Pattanaik et al.
+    https://chemrxiv.org/articles/Genereting_Transition_States_of_Isomerization_Reactions_with_Deep_Learning/12302084
+    This function writes the TS guess as an xyz file to the corresponding TS directory.
+
+    Args:
+        ts_path (str): Path to the inputs for the network.
+
+    Returns:
+         ts_xyz_dict (dict): ARC xyz dictionary of TS guess
+    """
+
+    # run the GCN as a subprocess
+    p = subprocess.run(f'{TS_GCN_PYTHON} {os.path.join(TS_GCN_PATH, "inference.py")} '
+                       f'--r_sdf_path {os.path.join(ts_path, "reactant.sdf")} '
+                       f'--p_sdf_path {os.path.join(ts_path, "product.sdf")} '
+                       f'--ts_xyz_path {os.path.join(ts_path, "TS.xyz")} ',
+                       shell=True)
+    # if subprocess ran successfully, read the TS structure into ARC's dictionary format
+    if p.returncode == 0:
+        ts_xyz_dict = str_to_xyz(os.path.join(ts_path, "TS.xyz"))
+    # otherwise, log the error
+    else:
+        logger.error(f'GCN subprocess did not give a successful return code:\n'
+                     f'Got return code: {p.returncode}\n'
+                     f'stdout: {p.stdout}\n'
+                     f'stderr: {p.stderr}')
+        ts_xyz_dict = None
+
+    return ts_xyz_dict
--- a/devtools/install_gcn.sh
+++ b/devtools/install_gcn.sh
@@ -1,0 +1,12 @@
+# temporarily change directory to install software
+pushd ..
+
+# clone the repo in the parent directory
+git clone https://github.com/ReactionMechanismGenerator/TS-GCN
+cd TS-GCN
+
+# create the environment
+bash create_env.sh
+
+# Restore original directory
+popd
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -41,6 +41,13 @@ Install dependencies
   Make sure to add RMG-Py to your PATH and PYTHONPATH variables as explained in RMG's documentation.
 - If you'd like to use `AutoTST <https://github.com/ReactionMechanismGenerator/AutoTST>`_ in ARC (optional),
   clone it in a separate folder and add it to your PYTHONPATH just as well.
+- If you'd like to use a pre-trained graph convolutional network to quickly predict TS guesses for further optimization,
+  run `make gcn` from ARC's home directory, which clones `TS-GCN <https://github.com/ReactionMechanismGenerator/TS-GCN.git>`_
+  under the same parent folder as the ARC repository. It also sets up the environment to run the network.
+  The original network was published by `Pattanaik et al. <https://chemrxiv.org/articles/Genereting_Transition_States_of_Isomerization_Reactions_with_Deep_Learning/12302084>`_
+  using TensorFlow 1.4 and Python 2.7. This repo uses the same architecture with various minor upgrades, more training data,
+  and is translated into PyTorch Geometric and Python 3.7.
+
 - Create the Anaconda environment for ARC (after changing the directory to the
   installation folder by, e.g., ``cd ~/Code/ARC/``)::
 
--- a/Makefile
+++ b/Makefile
@@ -7,3 +7,8 @@
 test test-unittests:
 	nosetests --nocapture --nologcapture --all-modules --verbose --with-coverage --cover-inclusive --cover-package=arc --cover-erase --cover-html --cover-html-dir=testing/coverage
 
+gcn:
+	bash devtools/install_gcn.sh
+
+gcn-travis:
+	bash devtools/install_gcn_travis.sh
--- a/arc/common.py
+++ b/arc/common.py
@@ -38,6 +38,13 @@ arc_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))  # absolu
 
 VERSION = '1.1.0'
 
+# define default values for using the optional GCN to predict TS guesses
+# default assumption is that TS-GCN is installed in the same parent folder as the ARC repository
+TS_GCN_PATH = os.path.join(os.path.dirname(arc_path), 'TS-GCN')
+# default environment name for this repo is `ts_gcn`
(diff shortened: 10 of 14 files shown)


arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+    ts_gen_v2_path = os.path.join(os.path.dirname(arc_path), 'ts_gen_v2')
+    # define path to the corresponding TS folder
+    TS_path = os.path.join(project_dir, 'calcs', 'TSs', arc_reaction.ts_label)
+
+    # run the GCN as a subprocess
+    subprocess.run(f'conda activate ts_gen_v2;python {os.path.join(ts_gen_v2_path, "inference.py")} '
+                   f'--r_sdf_path {os.path.join(TS_path, "reactant.sdf")} '
+                   f'--p_sdf_path {os.path.join(TS_path, "product.sdf")} '
+                   f'--ts_xyz_path {os.path.join(TS_path, "TS.xyz")} ',
+                   shell=True)
```
xiaoruiDong, line 50: I am thinking we can do `p = subprocess.run ....` and check `p.returncode` to see if any error raises during the execution (return code = 0 is all good). If any, we can raise errors to the upper level and let the upper level set `self.success=False`

xiaoruiDong (commented):
Kevin, thank you for adding this wonderful feature! I add some comments. Besides, let me know your thoughts about possible unit tests (no need to be in this PR though).

arc/scheduler.py
```diff
@@ -352,6 +352,17 @@ def __init__(self,
                             rmg_reaction=rxn.rmg_reaction,
                         )
                     )
+                # now generate TSGuess objects for the other methods
+                for i, method in enumerate(rxn.ts_methods):
+                    if method != 'user guess':
```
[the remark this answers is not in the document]
kspieks: Thank you for catching this. I updated it

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+To use this network
+1) clone `ts_gen_v2` repo into the same level as ARC via
+`git clone https://github.com/kspieks/ts_gen_v2.git`
+
+2) Add `ts_gen_v2` to your PYTHONPATH
+
+3) Create the `ts_gen_v2` conda environment by running `bash devtools/create_env.sh`
+from the repo's home directory. This sets up the environment and installs PyTorch Geometric.
+https://pytorch-geometric.readthedocs.io/en/latest/notes/installation.html
+"""
```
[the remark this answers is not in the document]
kspieks: Looks like installation instructions for AutoTST are under "install dependencies" in `/docs/source/installation.rst`. I added instructions for cloning the GCN here, though I'm happy to move those instructions to `advanced.rst` or `tools.rst`. Just let me know :)

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
...
+        Chem.AllChem.EmbedMolecule(p_mol)
+        p_path = os.path.join(ts_path, 'product.sdf')
+        w = Chem.SDWriter(p_path)
+        w.write(p_mol)
+        w.close()
+
+        start = time.time()
+        ts_xyz_dict = gcn.gcn(self.project_dir,
+                              self.arc_reaction,
+                              )
```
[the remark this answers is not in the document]
kspieks: I pushed a draft of this. Let me know if this looks better

arc/ts/gcn.py
kspieks: Good idea. I pushed a fix up commit that uses the return code from the subprocess and implements a try except block as you suggested

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
...
+                                    multiplicity=product.multiplicity,
+                                    charge=product.charge,
+                                    xyz=mapped_xyz,
+                                    )
+        p_mol = to_rdkit_mol(mapped_product.mol)
+
+        # write input files for GCN to the TS project folder
+        ts_path = os.path.join(self.project_dir, 'calcs', 'TSs', self.arc_reaction.ts_label)
+        os.makedirs(ts_path)
+        Chem.AllChem.EmbedMolecule(r_mol)
```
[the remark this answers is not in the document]
kspieks: Thanks that seems like a useful function. Which `xyz` should this use? 

It won't have access to the QM optimized geometry at this stage since ARC calls TS guess methods at the start. If we use `Species.get_xyz()`, that tries to generate an RDKit conformer that is low in energy (but not necessarily the lowest). I'm having trouble finding [documentation](https://www.rdkit.org/docs/source/rdkit.Chem.AllChem.html) for `EmbedMolecule` so I'm unsure if that already does a cheap conformer scan. If `EmbedMolecule` already does a conformer scan, I think using `Species.get_xyz()` and then `rdkit_conf_from_mol(mol, xyz)` would be redundant? Please double check my understanding and correct me if I'm wrong.

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+        project_dir (str): Folder path for the project: the input file path or ARC/Projects/project-name.
+        arc_reaction (ARCReaction, optional): An ARC Reaction object.
+
+    Returns:
+         ts_xyz_dict (dict): ARC xyz dictionaru of TS guess
+    """
+    # the GCN repo must be cloned into the level as ARC
+    ts_gen_v2_path = os.path.join(os.path.dirname(arc_path), 'ts_gen_v2')
+    # define path to the corresponding TS folder
+    TS_path = os.path.join(project_dir, 'calcs', 'TSs', arc_reaction.ts_label)
```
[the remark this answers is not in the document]
kspieks: Thanks I added these variables to `settings.py` so the user can customize them.

kspieks:
Thanks for the helpful feedback. I pushed some fixup commits that try to address your comments. If you think the changes are sufficient, I can squash them before merging.

Regarding unit tests, I'm confused what unit tests we would add since I didn't think it was ARC's responsibility to test other repos, such as AutoTST, GCN, pyGSM, KinBot, etc?

docs/source/installation.rst
```diff
@@ -41,6 +41,16 @@ Install dependencies
   Make sure to add RMG-Py to your PATH and PYTHONPATH variables as explained in RMG's documentation.
 - If you'd like to use `AutoTST <https://github.com/ReactionMechanismGenerator/AutoTST>`_ in ARC (optional),
   clone it in a separate folder and add it to your PYTHONPATH just as well.
+- If you'd like to use a pre-trained graph convolutional network to quickly predict TS guesses for further optimization,
+  clone `ts_gen_v2 <https://github.com/kspieks/ts_gen_v2.git>`_ into the same level as ARC.
```
alongd, line 45: What is meant by "the same level as ARC"?
Also, can we automate this in a Makefile?

kspieks: I intended to say in the same directory as ARC, like how we commonly clone ARC, RMG-Py, and RMG-database in the same directory for convenience

Sorry what would the make file do? Just run the two steps of clone the GCN repo and run the bash script to create the corresponding environment?

arc/ts/gcn.py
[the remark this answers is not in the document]
xiaoruiDong: Yep, your suggestion makes sense.

arc/species/species.py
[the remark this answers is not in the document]
xiaoruiDong: `EmbedMolecule` doesn't do any conformer scan. Roughly speaking, the conformer is generated simply based on some built-in atom-bond distance criteria + some randomness. Therefore, the generated conformers are just reasonable ones, instead of optimized ones. Look at [document](https://www.rdkit.org/docs/source/rdkit.Chem.rdDistGeom.html?highlight=embedmolecule#rdkit.Chem.rdDistGeom.EmbedMolecule) which should be the actual API being called.
Back to the question, I think whether we want to follow the current workflow or we will modify the workflow, we should use `rdkit_conf_from_mol`, since it allows us to embed a `RDKit Conformer` to the `RDKit Mol` with the provided `xyz`. Note, since Species.get_xyz() will return `Species.final_xyz` (if it is assigned), ideally we don't need to modify these codes if the workflow is changed. We will provide `reactant.get_xyz()` for `r_mol` and `mapped_xyz` for `p_mol`. Does this make sense?

arc/ts/gcn.py
```diff
@@ -37,18 +23,18 @@ def gcn(project_dir,
...
-                   shell=True)
+    p = subprocess.run(f'conda activate {TS_GEN_V2_ENV_NAME};python {os.path.join(TS_GEN_V2_PATH, "inference.py")} '
+                       f'--r_sdf_path {os.path.join(ts_path, "reactant.sdf")} '
+                       f'--p_sdf_path {os.path.join(ts_path, "product.sdf")} '
+                       f'--ts_xyz_path {os.path.join(ts_path, "TS.xyz")} ',
+                       shell=True)
 
-    ts_xyz_dict = str_to_xyz(os.path.join(TS_path, "TS.xyz"))
+    ts_xyz_dict = str_to_xyz(os.path.join(ts_path, "TS.xyz"))
+    return_code = p.returncode
```
xiaoruiDong, line 39: Instead of `return` a `return_code` by this function, I am actually thinking of checking the `return_code` before `ts_xyz_dict` being assigned. If `p.returncode` is not 0, it is possible that `TS.xyz` may not exist right? So if return code == 0, we read the `xyz`, else we raise an error to the upper level. And we don't have to `return` `return_code`.

arc/species/species.py
```diff
@@ -1987,12 +1987,25 @@ def gcn(self):
...
+                                               )
+            if return_code == 0:
+                self.success = True
+                self.initial_xyz = ts_xyz_dict
+            else:
+                self.success = False
+                logger.error(f'GCN method did not give a successful return code:\n'
+                             f'Got return code: {return_code}'
+                             )
+
```
xiaoruiDong, line 2002: See the other comment, I think we don't need to have `if...else...` here. Better to have it in `gcn.gcn()`.

arc/species/species.py
```diff
@@ -1987,12 +1987,25 @@ def gcn(self):
...
+        except Exception as e:
```
xiaoruiDong, line 2003: `Exception` sounds too general to me. Why don't we create a GCNException which is more specific?

xiaoruiDong:
> Thanks for the helpful feedback. I pushed some fixup commits that try to address your comments. If you think the changes are sufficient, I can squash them before merging.
> 
> Regarding unit tests, I'm confused what unit tests we would add since I didn't think it was ARC's responsibility to test other repos, such as AutoTST, GCN, pyGSM, KinBot, etc?

Yes, you are right. We don't test other repo. Is there a simple way to test if this piece of code works? I don't have any thoughts therefore I ask for your input. But if there is no good proposal, we can leave it for future development.

alongd:
We don't test other repos in the sence that we shouldn't check for a particular XYZ result from GCN. But we should include in ARC's tests calls to GCN, checking that the received output is reasonable (e.g., looking at the number and type of atoms in the response from GCN).
The downside is of course that if for some reason GCN's master branch is buggy, all of our tests will fail. While this is manageable (tentatively checking out a previous commit for the test), the up side is that we make sure this feature works continually. 
I would suggest adding some simple tests.

kspieks:
Thanks @alongd and @xiaoruiDong for the feedback. I pushed some fix up commits and added a unit test to ensure that ARC can run the GCN. Can you look over the changes when you get a chance? I’ll remove the unused import to address the lgtm when squashing the fix up commits at the end

arc/settings/settings.py
```diff
@@ -200,4 +201,5 @@
 arc_path = os.path.abspath(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
 TS_GEN_V2_PATH = os.path.join(os.path.dirname(arc_path), 'ts_gen_v2')
 # default environment name for this repo is `ts_gen_v2`
-TS_GEN_V2_ENV_NAME = 'ts_gen_v2'
+TS_GEN_V2_PYTHON = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(sys.executable))),
+                                'ts_gen_v2', 'bin', 'python')
```
xiaoruiDong, line 205: I guess can we add some guidance on how to set this variable?

kspieks: This variable redirects to use the python executable for the GCN's conda environment. Help me make sure I understand your question, but I don't think we need to change how this variable is set since it will work if the user is running ARC from the `arc_env`, `t3_env`, or `rmg_env` (or any other environment).

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
+"""
+A GCN to predict TS guesses for isomerization reactions.
```
alongd: Please add the initials of GCN

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+The original network was published using TensorFlow 1.4 and Python 2.7
+Citation: Pattanaik, L.; Ingraham, J.; Grambow, C.; Green, W. H. 2020.
```
alongd: is there a DOI?

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+This file uses `ts_gen_v2`, which uses the same architecture with minor upgrades
+and translated into PyTorch Geometric and Python 3.7
+
+To use this network
+1) clone `ts_gen_v2` repo into the same level as ARC via
+`git clone https://github.com/kspieks/ts_gen_v2.git`
```
alongd: Is there a reason not to place the repo under the ReactionMechanismGenerator GitHub organization?

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+
+2) Add `ts_gen_v2` to your PYTHONPATH
+
+3) Create the `ts_gen_v2` conda environment by running `bash devtools/create_env.sh`
```
alongd: Which environment should be used when running ARC with ts_gen_v2? Does the ts_gen_v2 env include all packages required by ARC?
Alternatively, I'd like to have all libraries required for ts_gen_v2 included in arc_env. If you need to install PyTorch, please add the relevant commands to execute the `.sh` into the MakeFile. Let's make our env a one stop shop.

arc/ts/gcn.py
alongd: Is it necessary to use a subprocess? Can't we call it or submit it as a job?

arc/ts/gcn.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Args:
+        project_dir (str): Folder path for the project: the input file path or ARC/Projects/project-name.
+        arc_reaction (ARCReaction, optional): An ARC Reaction object.
+
+    Returns:
+         ts_xyz_dict (dict): ARC xyz dictionaru of TS guess
+    """
+    # the GCN repo must be cloned into the level as ARC
+    ts_gen_v2_path = os.path.join(os.path.dirname(arc_path), 'ts_gen_v2')
```
alongd: I prefer not to have a git repo within a git repo, if possible. If we clone ad add it to our path, can we simply import it instead?

arc/species/species.py
```diff
@@ -1742,8 +1746,10 @@ class TSGuess(object):
         family (str, optional): The RMG family that corresponds to the reaction, if applicable.
         xyz (dict, str, optional): The 3D coordinates guess.
         rmg_reaction (Reaction, optional): An RMG Reaction object.
+        arc_reaction (ARCReaction, optional): An ARC Reaction object.
         ts_dict (dict, optional): A dictionary to create this object from (used when restarting ARC).
-        energy (float): Relative energy of all TS conformers in kJ/mol.
+        energy (float, optional): Relative energy of all TS conformers in kJ/mol.
+        project_dir (str, optional): Folder path for the project: the input file path or ARC/Projects/project-name.
```
alongd, line 1751: the "input file path" or the "folder path"?

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
         else:
             self.initial_xyz = atst.autotst(rmg_reaction=self.rmg_reaction, reaction_family=self.family)
 
+    def gcn(self):
+        """
+        Determine a TS guess using a graph convolutional network.
+
+        Reactant must be a single fragment. Product can be 1 or 2 molecules.
```
alongd: please replace "molecules" with "species" because of the connotations we have with the RMG objects (of course that they are the same in English)

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
...
+        """
+        # check that this is an isomerization reaction i.e. only one reactant and one product
+        num_reactants = len(self.arc_reaction.r_species)
+        num_products = len(self.arc_reaction.p_species)
+        if num_reactants > 1:
+            raise TSError(f'Error while using GCN with reactants: {self.arc_reaction.r_species}'
+                          f'Isomerization reactions must have only 1 reactant.')
+        if num_products > 1:
```
alongd, line 1951: I thought the above docstring said that we could have 2 product species. Which is correct?

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
...
+                                    smiles=product.mol.smiles,
+                                    multiplicity=product.multiplicity,
+                                    charge=product.charge,
+                                    xyz=mapped_xyz,
+                                    )
+        p_mol = to_rdkit_mol(mapped_product.mol)
+
+        # write input files for GCN to the TS project folder
+        ts_path = os.path.join(self.project_dir, 'calcs', 'TSs', self.arc_reaction.ts_label)
+        os.makedirs(ts_path)
```
alongd: will this throw an error of the folder exists? If so, can we make sure ARC won't crush here if restarting and the folder exists?

arc/species/species.py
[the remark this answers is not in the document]
alongd: I agree with @xiaoruiDong

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
         else:
             self.initial_xyz = atst.autotst(rmg_reaction=self.rmg_reaction, reaction_family=self.family)
 
+    def gcn(self):
```
alongd, line 1939: A general question: is it worthwhile in terms of success rates of the method to call this function twice, each time from a different reaction direction (replacing reactants with products)? Or is the GCN method symmetrical and doesn't care about the direction in which it is called?

arc/species/species.py
```diff
@@ -1922,6 +1937,64 @@ def autotst(self):
...
+        w.close()
+
+        start = time.time()
+        ts_xyz_dict = gcn.gcn(self.project_dir,
+                              self.arc_reaction,
+                              )
+        end = time.time()
+        self.initial_xyz = ts_xyz_dict
+        self.success = True
+        self.execution_time = end - start   # seconds
```
alongd, line 1999: will common.time_lapse() help here? It returns a string, so maybe not

arc/scheduler.py
```diff
@@ -352,6 +352,17 @@ def __init__(self,
                             rmg_reaction=rxn.rmg_reaction,
                         )
                     )
+                # now generate TSGuess objects for the other methods
```
alongd, line 355: I think we might be calling TS methods too early here. We still haven't optimized the reactants and products (and perhaps we'll find out that a different conformer should be used, or that one of them isn't physical).
This should be done in the main scheduler loop, after checking that the reactant and product are optimized.

arc/job/ssh.py
```diff
@@ -275,11 +275,11 @@ def check_running_jobs_ids(self) -> list:
         cmd = check_status_command[servers[self.server]['cluster_soft']] + ' -u $USER'
         stdout = self._send_command_to_server(cmd)[0]
         for i, status_line in enumerate(stdout):
-            if servers['local']['cluster_soft'].lower() == 'slurm' and i > 0:
```
alongd, line 278: Thanks for this catch. I think it's a bug (local should be used in 'local' not in 'ssh') - could you change the commit message to BugFix and remove the commit description about hardcoding?

docs/source/installation.rst
alongd: Maybe say "under the same parent folder as the ARC repository"?
Yes, do you think it would be convenient for users to simply type `make` and have GCN and other future features we have automatically installed correctly?

alongd (commented):
Thanks for this awesome contribution!
Please see some comments attached.
Di you run a test case actually using ARC to generate a rate constant from scartch with GCN?

alongd:
@kspieks, some of my comments were on outdated commits. If @xiaoruiDong doesn't mind, perhaps squash and organize the commits before we take a final look at the PR?

arc/species/species.py
kspieks: Thanks for catching this. The original network was only trained on unimolecular reactions. We trained the updated network on both 1 --> 1 and 1 --> 2 reactions, so the network should be able to handle 1 --> 2. However, if there are 2 products, they must be oriented as if they just reacted (and if 2 reactants, they must be oriented as if they're about to react). They can't be arbitrarily placed in 3D space. This requirement makes it tricky for the network to operate on non unimolecular reactions in practice; I'm not sure how ARC would automatically orient the molecules correctly. I will update the above docstring to say the network can be used for unimolecular reactions only

arc/ts/gcn.py
kspieks: The `ts_gen_v2_env` requires PyTorch and PyTorch Geometric, but I _think_ adding these packages to the `arc_env` would be sufficient to run the network from `arc_env`. However, I kept the `ts_gen_v2_env` separate from the `arc_env` since I wasn't sure if we wanted ARC's environment to always get larger as we allow it to control other software. I'm imagining an example where a TS search method or other software requires package versions that are different than what ARC uses, which would prevent ARC's env from being a one stop shop. Would you prefer me to try making `arc_env` accommodate the GCN or should we keep them separate in case of possible future conflict?

alongd: Good point. I would suggest for now to include all libraries in a single `arc_env`, but please add a comment to the environment.yml file noting that these packages belong to GCN. In the future if this becomes an issue, we could have several sets of environment files to suite different needs.

arc/species/species.py
kspieks: I had copied this description from the `Scheduler` docstring in `scheduler.py`. I believe this refers to the path to the input file (`PATH/TO/INPUT` as opposed to `PATH/TO/INPUT/input.yml`) or the project folder? Do you want me to update this description and the one in `Scheduler` to be more clear?

arc/job/ssh.py
kspieks: Sure thing. Can you let me know if the commit message is clearer now?

arc/species/species.py
```diff
@@ -1922,6 +1936,68 @@ def autotst(self):
...
+        if num_products > 1:
+            raise TSError(f'Error while using GCN with products: {self.arc_reaction.p_species}'
+                          f'Isomerization reactions must have only 1 product.')
+        reactant = self.arc_reaction.r_species[0]
+        product = self.arc_reaction.p_species[0]
+        _, r_mol = rdkit_conf_from_mol(reactant.mol, reactant.get_xyz())
+
+        # GCN requires an atom-mapped reaction so map the product atoms onto the reactant atoms
+        # todo: update ARC's atom-mapping to a more accurate method
+        mapping = self.arc_reaction.get_atom_map()
```
alongd, line 1960: These lines are nice and useful. Could you turn them into a function in common.py with tests?
The function could return both xyz and ARCSpecies.

arc/species/species.py
```diff
@@ -1922,6 +1936,68 @@ def autotst(self):
...
+        # create another product with the mapped atoms since ARCSpecies objects preserve this ordering in RDKit objects
+        mapped_product = ARCSpecies(label=product.label,
+                                    smiles=product.mol.smiles,
+                                    multiplicity=product.multiplicity,
+                                    charge=product.charge,
+                                    xyz=mapped_xyz,
+                                    )
+        _, p_mol = rdkit_conf_from_mol(mapped_product.mol, mapped_product.get_xyz())
+
+        # write input files for GCN to the TS project folder
```
alongd, line 1973: Can we relocate this section into the gcn.gcn() function? So GCN will just get an `arc_reaction` object and treat it all the way?

arc/ts/gcn.py
```diff
@@ -0,0 +1,62 @@
...
+    Args:
+        project_dir (str): Folder path for the project: the input file path or ARC/Projects/project-name.
+        arc_reaction (ARCReaction, optional): An ARC Reaction object.
+
+    Returns:
+         ts_xyz_dict (dict): ARC xyz dictionaru of TS guess
+    """
+
+    # define path to the corresponding TS folder
+    ts_path = os.path.join(project_dir, 'calcs', 'TSs', arc_reaction.ts_label)
```
alongd, line 44: I think the processing should be done in another subfolder (e.g., `GCN`)

arc/ts/gcn.py
```diff
@@ -0,0 +1,62 @@
...
+    p = subprocess.run(f'{TS_GEN_PYTHON} {os.path.join(TS_GEN_PATH, "inference.py")} '
+                       f'--r_sdf_path {os.path.join(ts_path, "reactant.sdf")} '
+                       f'--p_sdf_path {os.path.join(ts_path, "product.sdf")} '
+                       f'--ts_xyz_path {os.path.join(ts_path, "TS.xyz")} ',
+                       shell=True)
+    # if subprocess ran successfully, read the TS structure into ARC's dictionary format
+    if p.returncode == 0:
+        ts_xyz_dict = str_to_xyz(os.path.join(ts_path, "TS.xyz"))
+    # otherwise, raise an error
+    else:
```
alongd, line 56: Instead of raising an error, can we log the error and return `None`?

docs/source/installation.rst
```diff
@@ -41,6 +41,16 @@ Install dependencies
   Make sure to add RMG-Py to your PATH and PYTHONPATH variables as explained in RMG's documentation.
 - If you'd like to use `AutoTST <https://github.com/ReactionMechanismGenerator/AutoTST>`_ in ARC (optional),
   clone it in a separate folder and add it to your PYTHONPATH just as well.
+- If you'd like to use a pre-trained graph convolutional network to quickly predict TS guesses for further optimization,
```
alongd, line 44: So do users still need to clone things, or do we automate it with a make command?

arc/settings/settings.py
```diff
@@ -193,3 +195,11 @@
     'job_time_limit_hrs': 120,
     'job_max_server_node_memory_allocation': 0.8,  # e.g., at most 80% node memory will be used
 }
+
+# define default values for using the optional GCN to predict TS guesses
+# default assumption is that TS-GEN is installed in the same parent folder as the ARC repository
+arc_path = os.path.abspath(os.path.dirname(os.path.dirname(os.path.dirname(__file__))))
```
alongd, line 201: We have arc_path defined in common. DO we need to redefine it?
https://github.com/ReactionMechanismGenerator/ARC/blob/238176204d10d6af88d5789b35e80f15b06d9e09/arc/common.py#L37

arc/settings/settings.py
```diff
@@ -193,3 +195,11 @@
...
+TS_GEN_PATH = os.path.join(os.path.dirname(arc_path), 'TS-GEN')
```
alongd, line 202: I think this sets `TS_GEN_PATH` to point to `.../ARC/TS-GEN`
From our talk I understood that this repository will be located outside the ARC repo.

Also, why not call the folder `GCN`? Wi'll have other repos for generating TS's, so `TS-GEN` is too general

.travis.yml
```diff
@@ -45,6 +45,12 @@ jobs:
         - cd AutoTST
         - export PYTHONPATH=$PYTHONPATH:$(pwd)
         - cd ..
+        - git clone https://github.com/ReactionMechanismGenerator/TS-GEN
```
alongd, line 48: could you instead add a bash script that clones GCN, and use it both here and in the Makefile as `make gcn`?
Then we also don't need to add GCN to the python path in ARC's common/settings

alongd (commented):
Thanks for the modifications! I added some comments, please take a look.

arc/species/species.py
```diff
@@ -1922,6 +1936,68 @@ def autotst(self):
...
+        Reactant and product must each be a single fragment.
+        """
+        # check that this is an isomerization reaction i.e. only one reactant and one product
+        num_reactants = len(self.arc_reaction.r_species)
+        num_products = len(self.arc_reaction.p_species)
+        if num_reactants > 1:
+            raise TSError(f'Error while using GCN with reactants: {self.arc_reaction.r_species}'
+                          f'Isomerization reactions must have only 1 reactant.')
+        if num_products > 1:
+            raise TSError(f'Error while using GCN with products: {self.arc_reaction.p_species}'
```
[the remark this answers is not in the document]
kspieks: Currently, ARC will crash if there are 2 reactants or 2 products. Would you rather we log the error and set `self.initial_xyz = None` instead? Perhaps

```
if num_reactants > 1:
    logger.error(f'Error while using GCN with reactants: {self.arc_reaction.r_species}.\n'
                 f'Isomerization reactions must have only 1 reactant.')
    self.initial_xyz = None
elif num_products > 1:
    logger.error(f'Error while using GCN with products: {self.arc_reaction.p_species}.\n'
                 f'Isomerization reactions must have only 1 product.')
    self.initial_xyz = None
else:
    # run GCN
```

arc/species/species.py
kspieks: Sorry I wasn't clear yesterday. From `main` in `ARC.py`, `project_directory` is defined as either the input file path or the `ARC/Projects/project-name`. I think the description accurately describes these two options so I'm confused how to update the description.

arc/species/species.py
```diff
@@ -1922,6 +1936,68 @@ def autotst(self):
...
+
+        Reactant and product must each be a single fragment.
+        """
+        # check that this is an isomerization reaction i.e. only one reactant and one product
+        num_reactants = len(self.arc_reaction.r_species)
+        num_products = len(self.arc_reaction.p_species)
+        if num_reactants > 1:
+            raise TSError(f'Error while using GCN with reactants: {self.arc_reaction.r_species}'
+                          f'Isomerization reactions must have only 1 reactant.')
+        if num_products > 1:
```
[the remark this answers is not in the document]
kspieks: From line 1948, we currently raise the same `TSError` if the number of reactants is > 1. As discussed below, let me know if you'd prefer ARC to log, rather than raise, the error.

arc/species/species.py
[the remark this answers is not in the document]
kspieks: Thanks for catching this typo. I'll updated this

arc/ts/gcn.py
```diff
@@ -0,0 +1,62 @@
...
+Notes for possible future developments:
+- Currently, ARC will only use the network for unimolecular reactions (A --> B). The network was trained on both A --> B
+  and A --> B + C reactions. However, the products must be oriented as if they just reacted. They can't be arbitrarily
+  placed in 3D space. ARC would have to automatically orient the two molecules correctly before using the network for
+  more reaction families.
+
+- The network was trained on geometries optimized at ωB97X-D3 with a def2-TZVP basis set. For now, ARC runs the GCN
+  using force field optimized geometries from RDKit. The TS guesses have looked reasonable, though in the future, it may
+  be beneficial to use the final QM optimized geometries as input to the network.
+
```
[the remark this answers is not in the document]
kspieks: Thanks and sorry I'd forgotten about this idea. I'll add it now. Lucky and I agree that the impact of reaction direction on TS convergence is unclear and has not been studied.

arc/ts/gcn.py
kspieks: Sure thing. I'll update this

arc/settings/settings.py
kspieks: I had tried adding `from arc.common import arc_path` to `settings.py` to avoid redefining this variable. Unfortunately, this leads to circular logic that causes ARC to crash since `common.py` already imports `settings`. I think redefining `arc_path` is necessary here, but let me know if you see a better way around this.

arc/settings/settings.py
kspieks: Yes we're on the same page. By joining the directory name of `arc_path`, the variable points to `.../RMG/TS-GEN`. 

In hindsight, I see how `TS-GEN` is too general. How about `TS-GEN-GCN` or `TS-GCN` or something along those lines? I think `GCN` might be too general in case graph networks with other purposes (besides generating TS guesses) are added to the RMG organization?

arc/species/species.py
kspieks: Since `common.py` can't import any ARC modules due to circular logic, I'm a bit confused how to move these lines (that use `xyz_to_str` and `ARCSpecies`) to `common.py`. I could try creating this as a function in `converter.py`? Alternatively, are we okay to move this to `gcn.gcn()` along with the code requested by the below comment?

arc/species/species.py
kspieks: Absolutely, I'll work on that now :)

arc/ts/gcn.py
kspieks: Sure thing. I can change this to `ts_path = os.path.join(project_dir, 'calcs', 'TSs', arc_reaction.ts_label, 'GCN')`.

arc/species/species.py
kspieks: Moving this to `gcn.gcn()` won't work either because that requires `gcn` to import `ARCSpecies`, but `species.py` already imports `gcn`, leading to circular logic that causes ARC to crash. I propose leaving this code (and the code discussed below) under `TSGuess.gcn()`. I'm happy to discuss alternatives so feel free to slack/zoom me :)

alongd: You're right. Perhaps it should be a method under ARCReaction. 
We already have `get_atom_map`, let's make a `get_mapped_product_xyz` method that calls `get_atom_map` and returns xyz. It's OK if for now `get_mapped_product_xyz` will only be functional for A <=> B reactions.

arc/settings/settings.py
alongd: TS-GCN is good.

Why is the path `.../RMG/TS-GEN`? Did you mean `.../ARC/TS-GEN`? Why RMG, and more generally, why clone a repo within a repo?

arc/settings/settings.py
alongd: Perhaps move all of this to common? Or, are these settings often modified by the user?

arc/species/species.py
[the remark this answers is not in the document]
alongd: Yes, it's better not to crush (say you have 100's of rxns, and one turns out buggy, ARC should continue running for the others

arc/species/species.py
alongd: You're good, my confusion, sorry

arc/settings/settings.py
kspieks: Sounds good. I'll change the name of the repo to `TS-GCN` and will update the environment name to be `ts_gcn`, rather than `ts_gen`, so it matches.

Sorry for the confusion. I was not clear. In this example, the `RMG` directory is an arbitrary parent directory for all repos; `RMG` is not hardcoded. I shouldn't have typed `RMG` above because it gives the appearance of being hardcoded. 
The intention is for RMG-Py, RMG-database, ARC, T3, AutoTST, and TS-GCN to all be in the same directory. This is not necessary, but seems commonly done by group members and is the convention of our `.travis.yml` files, so I was sticking with this convention. TS-GCN is not going to be cloned within the ARC repo. The bash script installs it in the same parent directory containing ARC. Let me know if this is clearer and please follow up with any questions.

arc/settings/settings.py
kspieks: Moving these variables to common should also works. I'd anticipate users to use the default settings, which means the variables wouldn't need to be modified. I'll pushed this change today.

arc/species/species.py
kspieks: Okay I'll add this a method to ARCReaction and add a corresponding test. I'm thinking the unit test will be short and simple since this method relies on `get_atom_map`, which already has many tests. Also if it's okay, I'm thinking of just returning the ARCSpecies since users can call the `get_xyz()` method (and then use any functions from `converter.py` to change the format) if the xyz coordinates are desired.

arc/settings/settings.py
alongd: Thanks for the clarification!

arc/species/species.py
alongd: If in the future we expand this method to also deal with multi-product reactions, then we shouldn't return an ARCSpecies, but rather a combined XYZ for the products. I think it's best to return just the XYZ to avoid changing later functionalities that depend on this method.

kspieks: Okay we can return the xyz coordinates. Do you want the xyz coordinates in ARC dictionary format or just as a string?

Also, I need a mapped RDKit molecule in preparation for the GCN. So I think I'll make this method return both the coordinates and the ARCSpecies object since I can convert the mapped ARCSpecies object to an RDKit molecule using `rdkit_conf_from_mol` to preserve the atom ordering.

alongd: ARC dictionary format would be the best.
OK

arc/reaction.py
```diff
@@ -670,3 +671,28 @@ def get_atom_map(self, verbose: int = 0) -> Optional[List[int]]:
             data = products.align(ref_mol=reactants, verbose=verbose)[1]
             atom_map = data['mill'].atommap.tolist()
         return atom_map
+
+    def get_mapped_product_xyz(self):
+        """
+        Uses the mapping from product onto reactant to create a new ARC Species for the mapped product.
+        For now, only functional for A <=> B reactions.
+
+        Returns:
```
alongd, line 680: We modified the docstring syntax for "Returns:", if the method returns more than one variable, then it returns a tuple. The new syntax compiles better in the auto-docs-deployment. See here for example: https://github.com/ReactionMechanismGenerator/ARC/blob/master/arc/species/species.py#L2020

(lines 2020 -> 2021 😄 )

devtools/install_gcn.sh
```diff
@@ -0,0 +1,4 @@
+cd ..
```
alongd, line 1: Could you add before this script (and in install_gcn_travis.sh):
```
# Temporarily change directory to $HOME to install software
pushd .
```
and after it:

```

# Restore original directory
popd
```
(see, e.g., [here](https://github.com/ReactionMechanismGenerator/T3/blob/master/devtools/travis_ci_before_install.sh))

alongd (commented):
Looks great!! just a couple minor comment

arc/reaction.py
kspieks: Thanks for letting me know. I'll update the docstring

devtools/install_gcn.sh
kspieks: Sure thing. I used `pushd ..` to be more condense than doing `push .` and then `cd ..`. Let me know if this is okay.

alongd (approved):
Thanks for all the work you've put into this PR! Looking great!

PR #440, opened 2020-12-24
Add unit test for MolGraph.get_formula after fixing bug in same 

mefuller:
(and one other little typo in a comment)

### DIFF
--- a/arc/species/speciesTest.py
+++ b/arc/species/speciesTest.py
@@ -31,7 +31,7 @@ from arc.species.species import (ARCSpecies,
                                  check_xyz,
                                  determine_rotor_symmetry,
                                  determine_rotor_type)
-
+from arc.species.xyz_to_2d import MolGraph
 
 class TestARCSpecies(unittest.TestCase):
     """
@@ -1375,6 +1375,18 @@ class TestTSGuess(unittest.TestCase):
         self.assertEqual(tsg.method, 'autotst')
         self.assertTrue(isinstance(tsg.rmg_reaction, Reaction))
 
+    def test_xyz_to_2d_get_formula(self):
+        xyz_arb = {'symbols': ('H', 'C', 'H', 'H', 'O', 'N', 'O'),
+                 'isotopes': (1, 13, 1, 1, 16, 14, 16),
+                 'coords': ((-1.0, 0.0, 0.0),
+                            (0.0, 0.0, 0.0),
+                            (0.0, -1.0, 0.0),
+                            (0.0, 1.0, 0.0),
+                            (1.0, 0.0, 0.0),
+                            (2.0, 0.0, 0.0),
+                            (3.0, 0.0, 0.0),)}
+        mol_graph_1 = MolGraph(symbols=xyz_arb['symbols'], coords=xyz_arb['coords'])
+        self.assertEqual(mol_graph_1.get_formula(), 'CH3NO2')
 
 if __name__ == '__main__':
     unittest.main(testRunner=unittest.TextTestRunner(verbosity=2))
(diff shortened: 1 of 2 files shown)


arc/species/speciesTest.py
```diff
@@ -580,9 +580,9 @@ def test_process_xyz(self):
         """Test the process_xyz() function"""
         # the last four H's in xyz hve a leading TAB character which is removed when processing
         xyz = """
-
```
alongd, line 583: these "tabs" were added intentionally (perhaps deleted here unintentionally?)
the idea is to make sure ARC is robust in reading in XYZ's even if those were pasted from an online source which often has "tab" chars. Same for the lines below.
Could you remove these changes from the commit? ("squash")

arc/species/speciesTest.py
```diff
@@ -1375,6 +1375,19 @@ def test_from_dict(self):
         self.assertEqual(tsg.method, 'autotst')
         self.assertTrue(isinstance(tsg.rmg_reaction, Reaction))
 
+    def test_xyz_to_2d_get_formula(self):
+        from arc.species.xyz_to_2d import MolGraph
```
alongd, line 1379: Could you relocate the import line to the top of the module? (we sort them there alphabetically)
This way additional unit tests we'll add will also have access to `MolGraph`

arc/species/xyz_to_2d.py
```diff
@@ -17,7 +17,7 @@
 
 class Atom(object):
     """
-    Represents an atopm in a molecular graph.
+    Represents an atom in a molecular graph.
```
alongd, line 20: thanks!

arc/species/xyz_to_2d.py
```diff
@@ -144,13 +144,14 @@ def get_formula(self):
         formula = ''
         if 'C' in list(elements.keys()):
             count = elements['C']
+            print(count)
```
alongd, line 147: a debug leftover?

arc/species/xyz_to_2d.py
```diff
@@ -144,13 +144,14 @@ def get_formula(self):
...
             formula += 'C{:d}'.format(count) if count > 1 else 'C'
             del elements['C']
             if 'H' in list(elements.keys()):
                 count = elements['H']
                 formula += 'H{:d}'.format(count) if count > 1 else 'H'
                 del elements['H']
-        keys = elements.keys()
+        keys = list(elements.keys())
```
alongd, line 154: nice catch!

alongd (commented):
Thanks for this contribution!
Please take a look at some minor comments I added

arc/species/speciesTest.py
mefuller: Is there an easy way to do line-by-line modifications here?

mefuller: Looks like my linter ate those and never told me...

arc/species/xyz_to_2d.py
mefuller: the function did not work without this fix

arc/species/speciesTest.py
mefuller: The easiest solution was to put back tabs (not sure why they show as different length - I thought I checked this) - I hope this is still ok

alongd: You can perhaps reset soft (git reset --soft official/master) to remove the commits but keep all the code modifications (so the changes will only be staged but uncommited), then select just the particular lines you'd like to commit, e.g. using [git cola](https://git-cola.github.io/) gui.

arc/species/xyz_to_2d.py
alongd: It's probably a fossil of the Python 2 -> 3 transition. Thanks for the fix!

alongd (approved):
Thanks for this contribution!

PR #449, opened 2021-03-19
Constrain conda channels in environment.yml

kspieks:
When creating a new `arc_env`, conda currently does not solve the environment after 45 minutes. I killed the task since it should not take this long. I propose making `environment.yml` more specific. This PR essentially copies the changes Professor West recently added on RMG-Py by constraining the conda channels in [PR #2089](https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2089). After making the changes, conda can create the environment in just under 15 minutes.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,55 +2,52 @@ name: arc_env
 channels:
   - defaults
   - rmg
-  - rdkit
-  - cantera
   - anaconda
   - omnia
-  - pytorch
   - conda-forge
 dependencies:
   - cairo
   - cairocffi
-  - cantera >=2.3.0
+  - rmg::cantera >=2.3.0
   - cclib >=1.6
-  - chemprop
+  - rmg::chemprop
   - codecov
   - coolprop
   - coverage
   - cython >=0.25.2
   - ffmpeg
-  - gprof2dot
+  - rmg::gprof2dot
   - graphviz
   - h5py
   - jinja2
   - jupyter
-  - lpsolve55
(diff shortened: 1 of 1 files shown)


xiaoruiDong:
Thank you for fixing the environment setup, Kevin! looks good to me!

alongd:
Thanks!!
Would you know why RDKit is installed from the RMG channel and not from the rdkit channel?

kspieks:
Before making these changes, both the `rmg_env` and `arc_env` use RDKit from the rmg channel so this PR (and the PR by Professor West) preserve that behavior. I'm unsure why we do not use the rdkit channel

rwest:
> Would you know why RDKit is installed from the RMG channel and not from the rdkit channel?

I'd like us to move towards using other official packages, preferably from a common channel like conda-forge, but I think that may require us rebuilding some of our conda binaries (which is long overdue).  The goal of https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2089 (and this PR I guess) was just to help anaconda reach the current solution in less time.

kspieks:
That's correct. Another student had tried to create the `arc_env` yesterday and apparently conda had not resolved the dependencies, even after 6 hours... This PR brings the time down to 15 minutes, allowing us to have a workable solution while we discuss other long-term plans, such as creating a common conda channel and updating the binaries (as mentioned above :) )

alongd (approved):
I see. Thanks!

PR #452, opened 2021-04-28
add clean to makefile

mefuller:
`make clean` now available to wipe out items created locally during testing, `__pycache__` folders; `PHONY` added to protect arguments

### DIFF
--- a/.github/workflows/cont_int.yml
+++ b/.github/workflows/cont_int.yml
@@ -44,6 +44,7 @@ jobs:
         make gcn-cpu
         cd ..
         echo $(pwd)
+        conda activate arc_env
         conda list
 
     - name: Lint with flake8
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -12,3 +12,9 @@ gcn:
 
 gcn-cpu:
 	bash devtools/install_gcn_cpu.sh
+
+clean:
+	find -type d -name __pycache__ -exec rm -rf {} +
+	rm -rf testing
```
alongd, line 20: I'm not certain the test folders should be here. ARC should delete al testing folders after it's done testing. It is possible that if a test fails, then the testing folder persists, but is this what we're addressing?

alongd (commented):
Looking good. I added one question.
Could you rebase so the tests will pass?

Makefile
mefuller: Yes

mefuller:
I broke something during rebase - see #454

PR #453, opened 2021-04-30
Fix docs auto-deploy

alongd:
Changed Py to 3.7 and other general modifications


mefuller (approved):
All looks good to me.
I definitely want to track down why we're stuck at Python 3.7, but that's a project for another day

mefuller (approved):

PR #454, opened 2021-05-03
add clean and PHONY to makefile

mefuller:
I am terrible at this and royally broke something when rebasing, so this is a new PR attempting to be the old PR (#452)...
`make clean` now available to wipe out items created locally during testing, `__pycache__ `folders; `PHONY` added to protect arguments
`make clean` deletes the folders it does to clean up from failed tests since these would otherwise persist - if tests run successfully and folders aren't present, `make clean` will silently do nothing (`-f` option)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@
 #
 ################################################################################
 
+.PHONY : clean gcn gcn-cpu test test-unittests
+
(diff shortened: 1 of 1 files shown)


alongd (approved):
Thanks!!

PR #455, opened 2021-05-28
Add commit strings

kspieks:
ARC currently logs the git commit string for the ARC repository. Since ARC relies on code from RMG-Py and RMG-database, this PR allows ARC to log the git commit strings for these repos as well. I think it would be particularly informative when using ARC to run Arkane with specified AECs + BACs to obtain thermo or rates so that the `arc.log` records the commit hash of all repositories involved in the calculation.

Within `common.py`, should we rename `arc_path` to `ARC_PATH` to indicate the variable is constant and keep it consistent with PEP 8?

This PR also includes a minor change for updating the message printed after obtaining a frequency factor for a new LoT. Previously, users were directed to `RMG-Py/arkane/statmech.py`; the message now points users to `RMG-database/input/quantum_corrections/data.py`

Happy to hear any feedback :)

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -24,6 +24,7 @@ import pandas as pd
 import qcelemental as qcel
 
 from arkane.ess import ess_factory, GaussianLog, MolproLog, OrcaLog, QChemLog, TeraChemLog
+import rmgpy
 from rmgpy.molecule.element import get_element
 from rmgpy.qm.qmdata import QMData
 from rmgpy.qm.symmetry import PointGroupCalculator
@@ -34,7 +35,12 @@ from arc.imports import settings
 
 logger = logging.getLogger('arc')
 
-arc_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))  # absolute path to the ARC folder
+# absolute path to the ARC folder
+arc_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+# absolute path to RMG-Py folder
+RMG_PATH = os.path.abspath(os.path.dirname(os.path.dirname(rmgpy.__file__)))
+# absolute path to RMG-database folder
+RMG_DATABASE_PATH = os.path.abspath(os.path.dirname(rmgpy.settings['database.directory']))
 
 VERSION = '1.1.0'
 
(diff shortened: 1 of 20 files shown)


arc/common.py
```diff
@@ -36,6 +37,9 @@
 
 arc_path = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))  # absolute path to the ARC folder
```
alongd, line 38: Yes, we can change to `ARC_PATH` as you suggested

arc/common.py
```diff
@@ -36,6 +37,9 @@
...
 
+RMG_PATH = os.path.abspath(os.path.dirname(os.path.dirname(rmgpy.__file__)))  # absolute path to RMG-Py folder
```
alongd, line 40: Currently ARC does require RMG to be installed from source. But if one day we'll allow RMG's binaries, will this still work?

arc/common.py
```diff
@@ -265,16 +269,23 @@ def log_header(project: str,
...
-    branch_name = get_git_branch()
-    if head != '' and date != '':
-        logger.log(level, 'The current git HEAD for ARC is:')
-        logger.log(level, f'    {head}\n    {date}')
-    if branch_name and branch_name != 'master':
-        logger.log(level, f'    (running on the {branch_name} branch)\n')
-    else:
-        logger.log(level, '\n')
+
+    paths_dict = {'ARC': None,
```
alongd, line 273: please add a comment that the git commit/branch functions use ARC_PATH by default
alternatively, we could pass ARC_PATH here as well for readability

arc/utils/scale.py
```diff
@@ -241,12 +241,12 @@ def summarize_results(lambda_zpes: list,
             text += f'(execution time: {execution_time})\n'
             logger.info(text)
             f.write(text)
-            arkane_formats.append(f"                 '{level}': {harmonic_freq_scaling_factor:.3f},  # [4]\n")
-        logger.info(arkane_text)
-        f.write(arkane_text)
-        for arkane_format in arkane_formats:
-            logger.info(arkane_format)
-            f.write(arkane_format)
+            database_formats.append(f"                 '{level}': {harmonic_freq_scaling_factor:.3f},  # [4]\n")
```
alongd, line 244: Spaces were added in the string to make copy/paste seamless in accordance with the indentation in Arkane. Could you check the indentation in `RMG-database/input/quantum_corrections/data.py` and make modifications to leading spaces if necessary?

alongd (commented):
Thanks!!
Please see some minor comments

arc/utils/scale.py
kspieks: Thanks for letting me know. I updated the number of leading spaces so it is consistent with the indentation in RMG-database

arc/common.py
kspieks: Good point. I'll pass in `ARC_PATH` for readability

arc/common.py
kspieks: Good question. This would only work if RMG has been cloned somewhere on the computer. If ARC allows RMG's binaries in the future, we could import (or copy) this [function](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L2377-L2395) from RMG-Py and implement the same if else logic to [report the binary version](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/rmg/main.py#L1902-L1928) instead of the commit hash if the `RMG_PATH` and `RMG_DATABASE_PATH` cannot be found.

alongd (approved):
Thanks!!

PR #463, opened 2021-09-08
Added a function to remove the 1st atom in a Z matrix

alongd:
This will be useful for combining zmats in TS search heuristics.
Tests added.

Additional commits had to be added to make the tests pass (regardless of the original content of this PR):
- Don't attempt to load files from the RMGDatabase as libraries (recently a comment text file was added, and ARC mistakenly attempts to read it as a library)
- The new QCElemental version is giving unexpected errors for fragments when using atom mapping. Soon a better atom mapping algorithm will be implemented, for now keep the working version of QCElemental
- Modify how atom IDs are being tested


arc/species/zmatTest.py
```diff
@@ -1614,6 +1614,62 @@ def test_up_param(self):
...
+        expected_zmat_2 = {'symbols': ('H',), 'coords': ((None, None, None),),
+                           'vars': {}, 'map': {0: 0}}
+        self.assertEqual(zmat.remove_1st_atom(zmat_2), expected_zmat_2)
+
+        zmat_3 = {'symbols': ('H', 'C', 'C', 'H', 'H', 'H', 'H', 'H'),
+                  'coords': ((None, None, None), ('R_1_0', None, None), ('R_2_1', 'A_2_1_0', None),
+                             ('R_3_1', 'A_3_1_2', 'D_3_1_2_0'), ('R_4_1', 'A_4_1_2', 'D_4_1_2_3'),
+                             ('R_5_2', 'A_5_2_1', 'D_5_2_1_4'), ('R_6_2', 'A_6_2_1', 'D_6_2_1_5'),
+                             ('R_7_2', 'A_7_2_1', 'D_7_2_1_6')),
+                  'vars': {'R_1_0': 1.0940775789443724, 'R_2_1': 1.5120487296562577,
```
mefuller, line 1641: Could we revise this test to have 0 referenced in the original coordinates such that it is not deleted by removing the atom zero, bond 1, angle 2, and dihedral 3? Or is there some check on the construction of ZMats that prevents this case (e.g. A_7_2_1 becomes A_7_2_0)
In this case, the expected ZMat would be a priori undefined? A cursory read of the code doesn't make it clear to me what should happen in this case

mefuller (approved):
Approve as currently written, but question/comment about possible malformed ZMats

arc/species/zmatTest.py
mefuller: Sounds fine

PR #464, opened 2021-09-21
A collection of minor style modification commits

alongd:
These are minor modifications to ARC accumulated while working on other features.

### DIFF
--- a/arc/job/trshTest.py
+++ b/arc/job/trshTest.py
@@ -550,7 +550,7 @@ class TestTrsh(unittest.TestCase):
                  'scan_res': 4.0,
                  'used_methods': None,
                  'log_file': log_file,
-                }
+                 }
         invalidate, invalidation_reason, message, actions = trsh.scan_quality_check(**case1)
         self.assertTrue(invalidate)
         self.assertEqual(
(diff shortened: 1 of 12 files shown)


arc/species/speciesTest.py
```diff
@@ -239,9 +239,9 @@ def test_conformers(self):
         self.assertEqual(len(self.spc11.conformers), 3)
 
         xyz12 = """C       0.00000000    0.00000000    0.00000000
-H       1.07008000   -0.14173100    0.00385900
-H      -0.65776100   -0.85584100   -0.00777700
-H      -0.41231900    0.99757300    0.00391900"""
+                   H       1.07008000   -0.14173100    0.00385900
+                   H      -0.65776100   -0.85584100   -0.00777700
+                   H      -0.41231900    0.99757300    0.00391900"""
         spc12 = ARCSpecies(label='CH3', smiles='[CH3]', xyz=xyz12)
```
xiaoruiDong, line 245: This will cause undesirable behavior, adding extra spaces starting from the second line.

xiaoruiDong (commented):
@alongd Thank you for cleaning up these codes! I really appreciate it. There is only one modification that I think needs further correction. I also find you remove a few debug logs. I don't have strong opinions on that but generally think they might be helpful in testing and debugging.

arc/job/trshTest.py
```diff
@@ -550,16 +550,15 @@ def test_scan_quality_check(self):
...
         self.assertTrue(invalidate)
         self.assertEqual(
             invalidation_reason, 'Significant difference observed between consecutive conformers')
         expect_message = 'Rotor scan of CH2OOH between pivots [1, 2] is inconsistent between ' \
                          'two consecutive conformers.\nInconsistent consecutive conformers and ' \
-                         'problematic internal coordinates:\nconformer # 63 / # 64        D3, D2' \
-                         '\nconformer # 80 / # 81        D3, D2\nARC will attempt to troubleshoot' \
-                         ' this rotor scan.'
+                         'problematic internal coordinates:\nconformer # 80 / # 81        D3, D2\n' \
+                         'ARC will attempt to troubleshoot this rotor scan.'
```
xiaoruiDong, line 561: This change causes the test to fail.

alongd:
@xiaoruiDong, Thanks for pointing out the cause for the error! I was too focused on reformatting text that I didn't notice that this must not be changed. Regarding the spaces in xyz starting from the 2nd line, that's OK, ARC strips every xyz line it reads. Please take another look once the tests pass, feel free to merge if it all looks OK. Thanks!

xiaoruiDong (approved):

PR #466, opened 2021-09-21
Added various functions in suport of the new (upcomming) job module

alongd:
This PR is another in a chain meant to assist managing a large future PR that re-writes the job module and adds various other modifications to atom mapping and scheduler.

Here, functions that could be taken as is without too much lost of context were added along with tests. The functions here are mainly concerned with species, reactions, and rmgdb.


arc/common.py
```diff
@@ -1245,3 +1222,107 @@ def convert_list_index_0_to_1(_list: Union[list, tuple], direction: int = 1) ->
...
+                  props=atom_dict['props'],
+                  ) for atom_dict in representation['atoms']]
+    mol.atoms = atoms
+    for i, atom_1 in enumerate(atoms):
+        for atom_2_id, bond_order in representation['atoms'][i]['edges'].items():
+            for atom_2 in atoms:
+                if atom_2.id == atom_2_id:
+                    break
+            bond = Bond(atom_1, atom_2, bond_order)
+            mol.add_bond(bond)
```
xiaoruiDong, line 1302: I guess each bond will be added twice? 

Also, Instead of making atoms a `list`, can we make it a `dict` as a map from `id` to `Atom`. So we can make the cost of the loop reduce from O(N^2) to O(N)?

My proposal
```
atoms = {atom_dict['id']: Atom(...) for atom_dict in representation['atoms']}
mol.atoms = list(atoms.values()]
for i, atom_1 in atoms.items():
    for atom_2_id, bond_order in representation['atoms'][i]['edges'].items():
        if atom_2_id > i:
            bond = Bond(atom_1, atoms[atom_2_id], bond_order)
            mol.add_bond(bond)
```

arc/main.py
```diff
@@ -51,6 +52,16 @@
...
+class StatmechEnum(str, Enum):
+    """
+    The supported statmech software adapters.
+    The available adapters are a finite set.
+    """
+    arkane = 'arkane'
+    # mesmer = 'mesmer'
+    # mess = 'mess'
+
+
```
xiaoruiDong, line 64: Excuse me, can you briefly explain the benefit of using `Enum class` here?

arc/reaction.py
```diff
@@ -324,10 +324,18 @@ def set_label_reactants_products(self, species_list: Optional[List[ARCSpecies]]
                     self.products = products.split(self.plus)
                 else:
                     self.products = [products]
-                if not self.r_species and self.reactants is not None:
-                    self.r_species = [spc for spc in species_list if spc.label in self.reactants]
-                if not self.p_species and self.products is not None:
-                    self.p_species = [spc for spc in species_list if spc.label in self.products]
+                if species_list is not None:
+                    if self.reactants is not None and self.products is not None:
```
xiaoruiDong, line 328: I think this is redundant since there is no way for them to be `None`?

arc/reaction.py
```diff
@@ -818,22 +826,51 @@ def get_mapped_product_xyz(self):
...
+        elif len(self.r_species) == 2:
+            xyz_dict = {'symbols': tuple(), 'isotopes': tuple(), 'coords': tuple()}
+            for reactant in self.r_species:
+                xyz = reactant.get_xyz()
+                xyz_dict['symbols'] += xyz['symbols']
+                xyz_dict['isotopes'] += xyz['isotopes']
+                xyz_dict['coords'] += xyz['coords']
+        else:
+            raise NotImplementedError(f'Reactants xyz is only implemented for up to 2 reactants, '
+                                      f'got {len(self.r_species)}.')
```
xiaoruiDong, line 867: I don't see a reason why `len(self.r_species)>2` is not supported by this function? Unless the upper layer is trying to catch this error. I suggest we simply not raise this error, so we don't need to change this specific function once ARC supports trimolecular reactions.

```
xyz_dict = {'symbols': tuple(), 'isotopes': tuple(), 'coords': tuple()}
for reactant in self.r_species:
      xyz = reactant.get_xyz()
      xyz_dict['symbols'] += xyz['symbols']
      xyz_dict['isotopes'] += xyz['isotopes']
      xyz_dict['coords'] += xyz['coords']
```

arc/reaction.py
alongd: I agree with your comment.
It wasn't directly introduced in this commit, so I'll add another commit that makes sure `.reactant`, `.products`, `.r_species`, and `.p_species` are lists when they are created / assigned, and only compares their `len()` instead of checking for `None`.

arc/common.py
alongd: I tried different versions of this, and it always made a wrongly defined Molecule instance, until I implemented it the RMG way of looping twice, e.g., [here](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/master/rmgpy/molecule/graph.pyx#L367). I learned the hard way that bonds must be added in both directions. I can't say I got to the bottom of it, but an O(N) loop didn't do the trick.

arc/main.py
alongd: A `thermo_adapter` and/or a `kinetics_adapter` are supplied by the user.
(Currently we don't really use it in input files since ARC only supports Arkane for statmech and that's the default, but in the future we plan to support Mesmer and MESS.)
The benefit of using an `Enum` class is just an elegant way to validate user inputs. We pass the user statmech adapter string through this Enum class, and if it's not something pre-defined by us in this class, it throws an error (I hope it's a meaningful error message).

alongd:
@xiaoruiDong, thanks for the quick review!
I think I addressed all suggestions either as comments or fixups.
Could you please take another look?

alongd:
Added a few fixups based on LGTM's comments

arc/common.py
xiaoruiDong: I see. Thanks for sharing that knowledge!  I guess the following should work, where I just avoid iterating the list of atoms for finding out the atom_2 (by `if atom_2.id == atom_2_id`) to reduce the computational complexity. Each bond should still be added twice.

```
atoms = {atom_dict['id']: Atom(...) for atom_dict in representation['atoms']}
mol.atoms = list(atoms.values()]
for i, atom_1 in atoms.items():
    for atom_2_id, bond_order in representation['atoms'][i]['edges'].items():
        bond = Bond(atom_1, atoms[atom_2_id], bond_order). 
        mol.add_bond(bond)
```

arc/species/converter.py
```diff
@@ -362,18 +382,21 @@ def xyz_from_data(coords, numbers=None, symbols=None, isotopes=None) -> dict:
...
-                           indices: List[int],
+                           indices: Optional[List[int]],
                            ) -> dict:
     """
     Sort the tuples in an xyz dict according to the given indices.
 
     Args:
         xyz_dict (dict): The Cartesian coordinates.
-        indices (List[int]): Entries are 0-indices of the desired order.
+        indices (Optional[List[int]]): Entries are 0-indices of the desired order.
```
xiaoruiDong, line 392: This seems a bit confusing to me. `indices` is `Optional` but we actually set no default value (`None`) for it.

xiaoruiDong (commented):
Thank @alongd  for your quick feedback. I added three more comments regarding your latest change.

arc/reaction.py
```diff
@@ -256,8 +254,6 @@ def from_dict(self,
         self.label = reaction_dict['label'] if 'label' in reaction_dict else ''
         self.multiplicity = reaction_dict['multiplicity'] if 'multiplicity' in reaction_dict else None
         self.charge = reaction_dict['charge'] if 'charge' in reaction_dict else 0
-        self.reactants = reaction_dict['reactants'] if 'reactants' in reaction_dict else None
-        self.products = reaction_dict['products'] if 'products' in reaction_dict else None
```
xiaoruiDong, line 260: This still needs to be kept, otherwise raise errors in Test ARCReaction.from_dict()

arc/common.py
alongd: I like that it is a dict now, and we don't have to loop again to find atom 2 by its ID.
Two modifications that I have to make:

1. The `i` we get from `for i, atom_1 in atoms.items():` is the atom ID now, and is not the same `i` we need for `for atom_2_id, bond_order in representation['atoms'][i]['edges'].items():` which should just be a counter. So I left the outer loop unchanged, `for i, atom_1 in enumerate(atoms.values()):`.

2. I'm hesitating to implement `mol.atoms = list(atoms.values()]`. Since atoms is a dict now, we shouldn't rely on it to keep the atom order. I added another entry to the representation of atom order to restore it here.

arc/species/converter.py
alongd: We call `sort_xyz_using_indices()` and pass the reaction `.atom_map` property to `indices`. If something went wrong with atom mapping, then `rxn.atom_map` is `None`. There's no need in a default value here, because we always send something to `indices`. But since we added a check below to verify it is not None, the typing should reflect that it's technically possible to pass `None`.

alongd:
@xiaoruiDong, thanks for your valuable comments!
I added fixup commits, let me know once I can squash them.

xiaoruiDong:
> @xiaoruiDong, thanks for your valuable comments!
> I added fixup commits, let me know once I can squash them.

@alongd, thanks! Looks good. There is one unit test about `test determine chirality` that still fails. I am not aware of the cause of it. Can you briefly comment on that?  I will approve the PR once it is justified.

alongd:
Thanks! I'll squash the fixups and debug the remaining issue.

alongd:
@xiaoruiDong, [this commit](https://github.com/alongd/ARC/commit/c2c04068cd97efb812c1725baa56771fbde31cbf) caused the conformers unit test to fail. I dropped it from this PR and I'll rethink its implementation separately (perhaps add a flag and a condition).
If all tests pass, feel free to merge. Thanks again for the valuable inputs, always a pleasure :)

xiaoruiDong (approved):
Thanks, @alongd. The PR now looks good! Thank you for your effort!

PR #475, opened 2021-10-08
Update the environment to reflect the addition of RMS into RMG

xiaoruiDong:
Recently RMG added new dependencies related to RMS. Add those dependencies according to the environment file in RMG.


alongd:
irrelevant, the env was fixed and separated from Julia now

PR #490, opened 2022-03-06
BugFix: Consider the migrating H atom when atom mapping intra-H migration reactions

alongd:
Previously, if a H atom in an intra-H migration reaction was the last atom in the reactant, it wasn't included in the atom map.
A fix was added along with a test. Additional minor changes were added as well.


kfir4444 (commented):
Hey Alon,
Thanks for the PR, looks really good. Did you ever run into the edge case where flipping the map gives the correct result? Flipping the reaction upon unsuccessful mapping might be a good idea in general, we might want think about implementing that as a "safety mechanism" in the rest of the mapping functions.

kfir4444 (approved):

alongd:
Thanks - no, I never ran into this case. Originally I thought that this would be the solution to the bug this PR fixes. After the bug was fixed by another (simpler) commit, I left the original check as well to be extra safe, but it could be just over-engineering.

PR #493, opened 2022-03-31
Issue491 solved

kfir4444:
A correction for issue 491, not yet complete for all edge cases, but does solves the issue, at least.


alongd:
@kfir4444, can you rename the PR's title and add some description of what it does, so it is searchable regardless of the issue it links to? And BTW, you can link to the issue itself by pasting the full link in the description, GitHub will shorten it for you

PR #507, opened 2022-05-01
CI with mamba

kfir4444:
Work In Progress, trying to make CI work with mamba. Please do not merge.

### DIFF
--- a/.github/workflows/cont_int.yml
+++ b/.github/workflows/cont_int.yml
@@ -1,11 +1,16 @@
-name: continuous integration
+name: CI
 
 on:
   push:
+    branches:
+      - '*'
   pull_request:
     branches:
       - main
     types: [opened, synchronize, reopened, ready_for_review, review_requested]
+  schedule:
+    - cron: '0 0 * * *'
+
 
 jobs:
   build:
@@ -17,16 +22,21 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Set up miniconda Python 3.7
+    - name: Setup Mambaforge Python 3.7
       uses: conda-incubator/setup-miniconda@v2
       with:
         environment-file: environment.yml
+        miniforge-variant: Mambaforge
+        miniforge-version: latest
         activate-environment: arc_env
         python-version: 3.7
-        auto-activate-base: false
+        use-mamba: true
+
+    - name: Update environment
+      run: mamba env update -n arc_env -f environment.yml
 
     - name: Install codecov
-      run: conda install -y -c conda-forge codecov
+      run: mamba install -y -c conda-forge codecov
 
     - name: Install dependencies
       run: |
@@ -49,13 +59,14 @@ jobs:
 
     - name: Lint with flake8
       run: |
-        conda install flake8
+        mamba install flake8
+        mamba config --append channels anaconda
         # stop the build if there are Python syntax errors or undefined names
         python -m flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         python -m flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 
-    - name: Test with nosetests
+    - name: Test with python unittest
       run: |
         cd ..
         cd RMG-Py
--- a/arc/species/zmatTest.py
+++ b/arc/species/zmatTest.py
@@ -1151,7 +1151,9 @@ class TestZMat(unittest.TestCase):
                                    (-3.871932295381926, -0.9306597576029849, -2.5400987960833166),
                                    (-3.440082669618663, 0.6565731451487455, -1.8462094217817948)]
         expected_ordered_symbols = ['C', 'C', 'H', 'H', 'H', 'H', 'H', 'H', 'C', 'C', 'O', 'O', 'H', 'H', 'H', 'H', 'H']
-        self.assertEqual(ordered_coords, expected_ordered_coords)
+        for coords1, coords2 in zip(expected_ordered_coords, ordered_coords):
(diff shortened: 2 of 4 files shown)


alongd:
try `self.assertTrue(almost_equal_coords_lists(new_xyz, expected_xyz))`

arc/species/zmatTest.py
```diff
@@ -1151,7 +1151,8 @@ def test_zmat_to_coords(self):
                                    (-3.871932295381926, -0.9306597576029849, -2.5400987960833166),
                                    (-3.440082669618663, 0.6565731451487455, -1.8462094217817948)]
         expected_ordered_symbols = ['C', 'C', 'H', 'H', 'H', 'H', 'H', 'H', 'C', 'C', 'O', 'O', 'H', 'H', 'H', 'H', 'H']
-        self.assertEqual(ordered_coords, expected_ordered_coords)
+        for coord1,coord2 in zip(expected_ordered_coords,ordered_coords):
+            self.assertAlmostEqual(coord1,coord2)
```
alongd, line 1155: Replace this (and in L1666) with
`self.assertTrue(almost_equal_coords_lists(coord1, coord2))`

Of course, you'll need to do add at the top: `from arc.common import almost_equal_coords_lists`

devtools/install_autotst.sh
```diff
@@ -0,0 +1,29 @@
+# temporarily change directory to install software, and move one directory up in the tree
```
alongd, line 1: install_autotst.sh shouldn't be part of this commit

.github/workflows/cont_int.yml
```diff
@@ -70,7 +70,7 @@ jobs:
 
     - name: Lint with flake8
       run: |
-        conda install flake8
+        mamba install flake8
```
alongd, line 73: I think we should squash all cont_int.yml commits together

arc/species/converterTest.py
```diff
@@ -4248,7 +4248,7 @@ def test_modify_coords(self):
         new_xyz = converter.modify_coords(coords=xyz1, indices=indices, new_value=new_val,
                                           modification_type=modification_type, mol=mol1)
         self.assertTrue(almost_equal_coords_lists(new_xyz, expected_xyz))
-        self.assertAlmostEqual(converter.get_zmat_param_value(coords=new_xyz, indices=indices, mol=mol1), new_val)
+        self.assertAlmostEqual(converter.get_zmat_param_value(coords=new_xyz, indices=indices, mol=mol1), new_val, 5)
```
alongd, line 4251: Please change the commit message, currently it says "Merge branch 'main' into CI_with_mamba "

alongd (commented):
Congrats that this branch passes all tests!!
It will be interesting to try re-running the tests and see whether the cache works now and it takes less time. Currently it takes ~75 min (used to be ~140 min) to run CI tests online.
I added some comments, please take a look

kfir4444:
@alongd it seems that the CI/build test on PR is failing, due to an issue with a test marked with `@work_in_progress`. I will try to push again and see if the problem returns, but if not, we might need to consider this PR, especially with the warning [here](https://github.com/ReactionMechanismGenerator/ARC/actions/runs/2263993967) and [here](https://github.com/ReactionMechanismGenerator/ARC/actions/runs/2263994060).

alongd:
The problem seems to be with GCN, I don't think it's an issue in the current PR, I'm having GCN problems as well with the Job PR #504.

.github/workflows/cont_int.yml
```diff
@@ -1,11 +1,18 @@
...
+      - '*'
   pull_request:
     branches:
       - main
     types: [opened, synchronize, reopened, ready_for_review, review_requested]
+  schedule:
+    - cron: '0 0 * * *'
+
+env:
+  CACHE_NUMBER: 1  # increase to reset conda env cache manually
```
alongd, line 15: This should be 0... If the test only passes (once) with a value of 1, it means that the cache probably doesn't work.
Before merging, we should report how long (approximately) it takes to run a test for the first time (i.e., changing `CACHE_NUMBER` to some random number to have a new cache), and how long it takes to run it for the 2nd time with the same `CACHE_NUMBER` value by restarting the test.
If the test repeatedly fails for the second time per `CACHE_NUMBER` value, then should stop using the cache (modify this file) if we cannot solve it, the mamba speed-up is already a nice feature.

.github/workflows/cont_int.yml
```diff
@@ -17,16 +24,30 @@ jobs:
...
-        auto-activate-base: false
+        use-mamba: true
+
+    - name: Set cache date
+      run: echo "DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
+
+    - uses: actions/cache@v2
+      with:
+        path: /usr/share/miniconda3/envs/my-env
+        key: linux-64-conda-${{ hashFiles('environment.yml') }}-${{ env.DATE }}-${{ env.CACHE_NUMBER }}
```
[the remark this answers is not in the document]
alongd: Can you take a look at the comment above?

.github/workflows/cont_int.yml
```diff
@@ -36,6 +57,7 @@ jobs:
         git clone https://github.com/ReactionMechanismGenerator/RMG-Py
         cd RMG-Py
         export PYTHONPATH=$PYTHONPATH:$(pwd)
+        conda install -c anaconda cython
```
alongd, line 60: conda or mamba?

alongd (commented):
Thanks for taking care of this PR! Please see some comments below. Feel free to squash the commits.

alongd (approved):
Looking great, thanks for taking the time to sort this out!
Crossing figures that rebasing `job` on top of this will do the trick :)

PR #508, opened 2022-05-01
Fixed issue 506, cyclic scisson

kfir4444:
Added the option of exiting with one species, if mol_splits has only one item in it, due to scission of a cyclic molecule. In addition, added a test.

### DIFF
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -1693,7 +1693,8 @@ class ARCSpecies(object):
             indices (tuple): The atom indices between which to cut (1-indexed, atoms must be bonded).
 
         Returns: list
-            The scission-resulting species.
+            The scission-resulting species, a list of either one or two species, if the scissored location is linear,
+            or one if the scission is in a cycle.
         """
         if any([i < 1 for i in indices]):
             raise SpeciesError(f'Scissors indices must be larger than 0 (1-indexed). Got: {indices}.')
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -1742,7 +1742,16 @@ class ARCSpecies(object):
             logger.warning(f'Scissors were requested to remove a non-single bond in {self.label}.')
         mol_copy.remove_bond(bond)
         mol_splits = mol_copy.split()
-        if len(mol_splits) == 2:
+        if len(mol_splits) == 1: # If cutting leads to only one split, then the split is cyclic.
+            spc1 = ARCSpecies(label=self.label + '_BDE_' + str(indices[0] + 1) + '_' + str(indices[1] + 1) + '_cyclic',
+                    mol=mol_splits[0],
+                    multiplicity=mol_splits[0].multiplicity,
+                    charge=mol_splits[0].get_net_charge(),
+                    compute_thermo=False,
+                    e0_only=True)
+            spc1.generate_conformers()
+            return [spc1]
+        elif len(mol_splits) == 2:
             mol1, mol2 = mol_splits
         else:
             logger.warning(f'Could not split {self.label} between indices {indices}.')
--- a/arc/species/speciesTest.py
+++ b/arc/species/speciesTest.py
@@ -1733,7 +1733,6 @@ class TestTSGuess(unittest.TestCase):
         mol_graph_1 = MolGraph(symbols=xyz_arb['symbols'], coords=xyz_arb['coords'])
         self.assertEqual(mol_graph_1.get_formula(), 'CH3NO2')
 
-    @work_in_progress
     def test_gcn(self):
         """
         Test that ARC can call the GNN to make TS guesses for further optimization.
--- a/arc/species/speciesTest.py
+++ b/arc/species/speciesTest.py
@@ -1309,6 +1309,14 @@ H       1.11582953    0.94384729   -0.10134685"""
                                                                       charge=spc.charge)[1]))
         self.assertTrue(any(spc.mol.to_smiles() == 'CO[NH]' for spc in spc_list))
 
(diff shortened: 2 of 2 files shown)


alongd:
Could you separate the test into a different commit? See the guidelines for suggested commit style: https://github.com/ReactionMechanismGenerator/ARC/wiki/ARC-Contributor-Guidelines#commits-style

arc/species/species.py
```diff
@@ -1741,7 +1741,16 @@ def _scissors(self, indices: tuple) -> list:
             logger.warning(f'Scissors were requested to remove a non-single bond in {self.label}.')
         mol_copy.remove_bond(bond)
         mol_splits = mol_copy.split()
-        if len(mol_splits) == 2:
+        if len(mol_splits) == 1: #If cutting leads to only one split, then the split is cyclic.
```
alongd, line 1744: According to Python's style guide, we should have two spaces before an `#` of an inline comment, then one additional space after the `#`, e.g.:
```python
if len(mol_splits) == 1  # comment
```

See the style guide: https://peps.python.org/pep-0008/#inline-comments

arc/species/species.py
```diff
@@ -1741,7 +1741,16 @@ def _scissors(self, indices: tuple) -> list:
             logger.warning(f'Scissors were requested to remove a non-single bond in {self.label}.')
```
alongd, line 1741: Can you describe the returned value better?
In the docsting we have

```
Returns: list
            The scission-resulting species.
```

Perhaps say that we can get either one or two ARCSpecies entries in the list, and perhaps what each length means?

arc/species/speciesTest.py
```diff
@@ -1309,6 +1309,15 @@ def test_scissors(self):
                                                                       charge=spc.charge)[1]))
         self.assertTrue(any(spc.mol.to_smiles() == 'CO[NH]' for spc in spc_list))
 
+        cycle = ARCSpecies(label="cycle",smiles = "C(1)CC(1)")
```
alongd, line 1312: minor: space after `,`

arc/species/speciesTest.py
```diff
@@ -1309,6 +1309,15 @@ def test_scissors(self):
...
+        cycle.bdes = [(1,2)]
```
alongd, line 1313: minor: space after `,`

arc/species/speciesTest.py
```diff
@@ -1309,6 +1309,15 @@ def test_scissors(self):
...
+        cycle.final_xyz = cycle.get_xyz()
+        cycle_scissors = cycle.scissors()
+        cycle_scissors[0].mol.update()
+        self.assertTrue(cycle_scissors[0].mol.is_isomorphic(ARCSpecies(label="check",smiles ="[CH2+]C[CH2+]").mol))
```
alongd, line 1317: also add an assertion of the length of the returned result?

arc/species/speciesTest.py
```diff
@@ -1309,6 +1309,15 @@ def test_scissors(self):
...
+        
+
```
alongd, line 1319: minor: too many empty lines (I think should only be one)

alongd (commented):

arc/species/speciesTest.py
```diff
@@ -1309,6 +1309,14 @@ def test_scissors(self):
                                                                       charge=spc.charge)[1]))
         self.assertTrue(any(spc.mol.to_smiles() == 'CO[NH]' for spc in spc_list))
 
+        cycle = ARCSpecies(label="cycle",smiles= "C(1)CC(1)")
+        cycle.bdes = [(1, 2)]
+        cycle.final_xyz = cycle.get_xyz()
+        cycle_scissors = cycle.scissors()
+        cycle_scissors[0].mol.update()
+        self.assertTrue(cycle_scissors[0].mol.is_isomorphic(ARCSpecies(label="check",smiles ="[CH2+]C[CH2+]").mol))
```
alongd, line 1317: It won't work without the `+` in the SMILES?

arc/species/speciesTest.py
```diff
@@ -1315,7 +1315,7 @@ def test_scissors(self):
         cycle_scissors = cycle.scissors()
         cycle_scissors[0].mol.update()
         self.assertTrue(cycle_scissors[0].mol.is_isomorphic(ARCSpecies(label="check",smiles ="[CH2+]C[CH2+]").mol))
-        self.assertEqual(len(cycle_scissors),2)
+        self.assertEqual(len(cycle_scissors),1)
```
alongd, line 1318: Please squash into the previous commit that added this line
Also, could you add a space after the `,`?

alongd (commented):
Thanks! Looks good and I'm happy to see the tests passing. Please see two minor comments

arc/species/speciesTest.py
kfir4444: No, this is the updated molecule, otherwise it is a molecule with smiles smils 'CCC' and atom list of 3 carbons and 6 hydrogens.

arc/species/speciesTest.py
```diff
@@ -1733,7 +1733,6 @@ def test_xyz_perception(self):
         mol_graph_1 = MolGraph(symbols=xyz_arb['symbols'], coords=xyz_arb['coords'])
         self.assertEqual(mol_graph_1.get_formula(), 'CH3NO2')
 
-    @work_in_progress
```
alongd, line 1728: This is very weird that we need to delete it here and add it in another PR.
I wonder if that means that in the CI-mamba PR we lost the ability to properly install TS-GCN.

alongd (approved):
Perfect, thanks for this fix!

PR #514, opened 2022-06-02
BugFix: Pass parameter name correctly in plotter

alongd:
Corrected the `show_atom_indices` argument passed within `draw_structure()` to `show_sticks()`


kfir4444 (approved):
Great! helps avoid confusions with the atom mapping features.

PR #515, opened 2022-06-04
Some enhancements for identifying TSs

alongd:
- BugFix: Initialize Species.ts_checks as dict when reading the species from a dictionary (or an input file)
- Added a `check_rxn_e0_by_spc()` method to Scheduler , so it can be called when needed, and executed only                                                                                        when appropriate
- Trsh confs only if requested to
- BugFix: Process rxn_copy for kinetics in ts E0 check
-  Changed symmetry scr_dir to be under HOME, since users don't always have access to root (e.g., on Atlas)
- Added tests


alongd:
@kfir4444, I see now that the last two commits for `Disproportionation` should not be merged, they were just for some trials, I'll remove them from this PR

kfir4444 (commented):
Great PR again @alongd!
Please look at the comments on `is_isomorphic`, these might cause some troubles in the future.
Also, it is currently untested, could you please add that test?

alongd:
Thanks for the comments! I added a fixup commit to is_isomorphic(), added a check that a non-TS species label is legal, and added tests.
Let me know if the modifications look good, and I'll squash the fixup before merging.

kfir4444 (approved):
Thanks for the corrections @alongd!

PR #516, opened 2022-06-07
Improved directed scans

alongd:
The feature of running scans directed by ARC (as opposed to scans managed bv the ESS) is improved (it suffered some functionality loss by the job PR #504)

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -482,6 +482,25 @@ def globalize_path(string: str,
     return string
 
 
+def delete_check_files(project_directory: str):
+    """
+    Delete ESS checkfiles. They usually take up lots of space and are not needed after ARC terminates.
+    Pass ``True`` to the ``keep_checks`` flag in ARC to avoid deleting check files.
+
+    Args:
+        project_directory (str): The path to the ARC project folder.
+    """
+    logged = False
+    calcs_path = os.path.join(project_directory, 'calcs')
+    for (root, _, files) in os.walk(calcs_path):
+        for file_ in files:
+            if os.path.splitext(file_)[1] == '.chk' and os.path.isfile(os.path.join(root, file_)):
+                if not logged:
+                    logger.info('\ndeleting all check files...\n')
+                    logged = True
+                os.remove(os.path.join(root, file_))
+
+
 def string_representer(dumper, data):
     """
     Add a custom string representer to use block literals for multiline strings.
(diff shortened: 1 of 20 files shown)


arc/common.py
```diff
@@ -482,6 +482,25 @@ def globalize_path(string: str,
     return string
 
 
+def delete_check_files(project_directory: str):
+    """
+    Delete ESS checkfiles. They usually take up lots of space and are not needed after ARC terminates.
+    Pass ``True`` to the ``keep_checks`` flag in ARC to avoid deleting check files.
```
kfir4444, line 488: Does this required here? I do not think users commonly check `common.py`

arc/common.py
```diff
@@ -482,6 +482,25 @@ def globalize_path(string: str,
     return string
 
 
+def delete_check_files(project_directory: str):
```
kfir4444, line 485: Untested function

arc/job/adapters/ts/gcn_ts.py
```diff
@@ -37,6 +37,8 @@
 
 TS_GCN_PYTHON = settings['TS_GCN_PYTHON']
 
+DIHEDRAL_INCREMENT = 10
```
kfir4444, line 40: Why is this only here and not in `autotst_ts.py` and the others?

arc/job/adapters/ts/heuristics.py
```diff
@@ -155,7 +157,9 @@ def __init__(self,
         self.conformer = conformer
         self.constraints = constraints or list()
         self.cpu_cores = cpu_cores
+        self.dihedral_increment = dihedral_increment or DIHEDRAL_INCREMENT
```
kfir4444, line 160: Here the `DIHEDRAL_INCREMENT = 30`

kfir4444 (commented):
Thanks @alongd for this PR!
please see some comments.

arc/job/adapters/ts/heuristics.py
alongd: Yes, in heuristics it is used differently than in GCN, and usually this default value is being used (i.e., ARC doesn't alter it)

arc/common.py
alongd: This function was copied as is from main to common.
I think that advanced users could look at it before running something so destructive.

PR #520, opened 2022-06-28
Allowed a reaction object to have more than 3 reactants or products

MichalKesl:
Allowed a reaction object to have more than 3 reactants or products if TS guess is given


alongd (approved):
Thanks for the contribution, @MichalKesl!

PR #522, opened 2022-07-07
Changed SO2 to hydrazine in thermo demo

kfir4444:
This because the Adjlist that was given is for hydrazin, and not SO2


alongd (approved):
Thanks!

PR #527, opened 2022-07-20
Ts improvements

alongd:
This PR contains several modifications to improve TS handing in ARC.

- Plot TSGuesses and highlight the chosen TS correctly
- Improved conformers / transition state guess handling upon restarting ARC
- Better treatment of ts energy checks
- Clustering TS guesses based on similarity of geometry
- Additional tweaks

Tests were added


arc/common.py
```diff
@@ -801,6 +801,33 @@ def get_extremum_index(lst: list,
...
+
+    Returns:
+        Optional[float]: The result.
+    """
+    if any(entry is None or not isinstance(entry, (int, float)) for entry in lst):
+        return None
+    if multipliers is None:
+        return sum(lst)
+    result = 0
+    for i in range(len(lst)):
```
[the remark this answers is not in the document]
kfir4444: I tries it with `numpy.dot`, and it is ~half the time

ipython/Tools/TS Guesses.ipynb
```diff
@@ -0,0 +1,546 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "5bae17f3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from arc.imports import settings\n",
```
kfir4444, line 45: I think it's better practice to put the imports before the actual code.

arc/checks/ts.py
```diff
@@ -174,14 +177,14 @@ def check_rxn_e0(reaction: 'ARCReaction',
         Optional[bool]: Whether the test failed and the scheduler should switch to a different TS guess,
                         ``None`` if the test couldn't be performed.
     """
-    for spc_label in reaction.reactants + reaction.products + [reaction.ts_label]:
-        folder = 'rxns' if species_dict[spc_label].is_ts else 'Species'
-        freq_path = os.path.join(project_directory, 'output', folder, spc_label, 'geometry', 'freq.out')
-        if not os.path.isfile(freq_path) and not species_dict[spc_label].is_monoatomic():
+    for spc in reaction.r_species + reaction.p_species + [reaction.ts_species]:
```
kfir4444, line 180: Why `reaction.ts_species` is not a list and the others are? it should be a list for consistency.

arc/species/converter.py
```diff
@@ -1884,8 +1884,8 @@ def compare_zmats(z1, z2, r_tol=0.01, a_tol=2, d_tol=2, verbose=False, symmetric
 
 def compare_confs(xyz1: dict,
                   xyz2: dict,
-                  rtol: float = 1e-5,
-                  atol: float = 1e-5,
+                  rtol: float = 0.01,
```
kfir4444, line 1887: This is a large accuracy change, why?

kfir4444 (commented):
Great job @alongd!
Please see comments.

arc/checks/ts.py
alongd: We may have several reactants / products, but we always have a single transition state object

arc/species/converter.py
alongd: I note that in the commit message:

Modified rtol and atol in converter compare_confs(). Specifically, it is important to increase atol (here to 0.1 by default), since we shouldn't bother about <0.1 A differences between coords

alongd:
Thanks, @kfir4444, for the quick response
I replied to the comments and updated the iPy notebook (and force pushed).
Let me know if you have additional comments, otherwise feel free to merge once the tests pass

arc/common.py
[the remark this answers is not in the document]
kfir4444: of course you can! I did something like:
`multipliers = np.array(multipliers + [1]*(len(lst)-len(multipliers))`

arc/checks/ts.py
kfir4444: yes, but if we would have one reactants, we would still put it in a list...

alongd: The `ARCReaction` object wass constructed in a way to have a `list` for reactants/products to *allow* more than one entry, but have the `.ts_species` not as a `list`, since there's currently no scenario where there are more than a single TS. If you think it should be otherwise, I'm happy to discuss. But this is the object we're working with here, not something this PR introduces

alongd:
@kfir4444, let me know if you have additional comments, otherwise I think this PR is ready to be merged

kfir4444 (approved):
Excellent!

PR #530, opened 2022-07-24
Queue TS guesses

alongd:
This PR accomplishes two things. First, it facilitates running TS guess Job adapters to be submitted on the server's queue system and properly parsed. Then it implements this for GCN (KinBot and AutoTST to follow later).
Tests were added to the GCN Job adapter.


kfir4444 (approved):
Great job @alongd

PR #538, opened 2022-08-08
Added TorchANI

alongd:
This PR adds a TorchANI Job adapter for opt/sp/freq jobs
It is currently a draft PR since the environment must be fixed: we should execute TorchANI in a separate env than ARC's


arc/job/adapters/torch_ani.py
```diff
@@ -0,0 +1,278 @@
...
+        content.update(settings)
+        save_yaml_file(path = os.path.join(self.local_path, "input.yml"), content=content)
+
+    def check_settings(self, settings):
+        options = {"model" : ['ani1ccx', 'ani1x', 'ani2x'],
+                   "device" : ["cpu", "cuda"],
+                   "engine" : ["bfgs", "scipyfminbfgs", "scipyfmingc"]}
+        for key in options.keys():
+            if key in settings.keys() and not settings[key] in options[key]:
+                return False
```
alongd, line 191: maybe add a logging.error message per `return False` in this function, so users know what went wrong in their settings? Below you raise a general error. adding another one here will be more informative/specific

arc/job/adapters/torch_ani.py
```diff
@@ -0,0 +1,278 @@
...
+
+    def execute_incore(self):
+        """
+        Execute a job incore.
+        """
+        self._log_job_execution()
+
+        if self.level is not None:
+            if self.level.args is not None:
+                if any(key in self.level.args for key in ["model", "device", "fmax", "engine", "steps"]):
```
alongd, line 235: When we have consecutive `if` statements, and none of them has an `else`, it's better practice, I think, to combine them into a single `if` statement with `and`s

arc/job/adapters/torch_ani.py
```diff
@@ -0,0 +1,278 @@
...
+                    temp_dict = {"model" : self.level.args["model"],
+                                 "device" : self.level.args["device"],
+                                 "fmax" : self.level.args["fmax"],
```
alongd, line 238: This will crash if the user gives only some of the keywords in `self.level.args`
Can you add tests for different scenarios of defining this adapter?

arc/job/adapters/torch_ani.py
```diff
@@ -0,0 +1,278 @@
...
+                            species=species,
+                            testing=testing,
+                            times_rerun=times_rerun,
+                            torsions=torsions,
+                            tsg=tsg,
+                            xyz=xyz,
+                            )
+
+
+    def write_input_file(self, settings) -> None:
```
alongd, line 175: can you add typehints for `settings`? it's a dict, but can it also be `None`?

arc/job/adapters/torch_ani.py
```diff
@@ -0,0 +1,278 @@
...
+
+        self.write_input_file(tani_default_options_dict)
+        commands = ['source ~/.bashrc',
+                   f'{TANI_PYTHON} {TANI_SCRIPT_PATH} '
+                   f'--yml_path {self.local_path}']
+        command = '; '.join(commands)
+        output = subprocess.run(command, shell=True, executable='/bin/bash')
+        if output.returncode:
+            logger.warning(f'Torch ANI subprocess ran and did not '
+                           f'give a successful return code for {self.job_id}.\n'
```
alongd, line 257: do we have a `job_id` if we didn't run on the server? Maybe use `job_name` (I think, check)

arc/job/adapters/torch_ani.py
```diff
@@ -0,0 +1,278 @@
...
+                           f'Got return code: {output.returncode}\n'
+                           f'stdout: {output.stdout}\n'
+                           f'stderr: {output.stderr}')
+        try:
+            output = read_yaml_file(path=os.path.join(self.local_path, "output.yml"))
+        except FileNotFoundError:
+            logger.error("Could not find output file")
+        self.sp, self.force, self.opt_xyz, self.freqs = output["sp"] if "sp" in output.keys() else None,\
```
alongd, line 265: if the above block is unsuccessful (exception raised), then ARC will crash here ungracefully because `output` is not defined.
Also, I suggest using `os.path.isfile` instead of a `try` - `except`

arc/job/adapters/torch_ani_test.py
```diff
@@ -0,0 +1,228 @@
...
+                                    project_directory=os.path.join(ARC_PATH, 'arc', 'testing', 'test_TorchANIAdapter_8'),
+                                    species=[ARCSpecies(label='EtOH', smiles='CCO')],
+                                    )
+        cls.job_9 = TorchANIAdapter(execution_type='incore',
+                                    job_type='force',
+                                    project='test_1',
+                                    project_directory=os.path.join(ARC_PATH, 'arc', 'testing', 'test_TorchANIAdapter_1'),
+                                    species=[ARCSpecies(label='EtOH', smiles='CCO')],
+                                    )
+        cls.job_10 = TorchANIAdapter(execution_type='incore',
```
alongd, line 93: what's the difference between `job_10` and `job_6`?

arc/job/adapters/torch_ani_test.py
```diff
@@ -0,0 +1,228 @@
...
+                                    )
+        cls.job_8 = TorchANIAdapter(execution_type='incore',
+                                    job_type='sp',
+                                    project='test_8',
+                                    project_directory=os.path.join(ARC_PATH, 'arc', 'testing', 'test_TorchANIAdapter_8'),
+                                    species=[ARCSpecies(label='EtOH', smiles='CCO')],
+                                    )
+        cls.job_9 = TorchANIAdapter(execution_type='incore',
+                                    job_type='force',
+                                    project='test_1',
```
alongd, line 89: update project names

arc/job/adapters/torch_ani_test.py
```diff
@@ -0,0 +1,228 @@
...
+        self.job_4.execute()
+        self.assertEqual(self.job_4.opt_xyz['symbols'], ('C', 'C', 'O', 'H', 'H', 'H', 'H', 'H', 'H'))
+
+    def test_run_vibrational_analysis(self):
+        """Test the run_vibrational_analysis() method."""
+        self.job_10.execute()
+        results = read_yaml_file(path=self.job_10.local_path_to_output_file)
+        self.assertEqual(list(results.keys()), ['force_constants', 'freqs', 'hessian', 'modes', 'reduced_masses'])
+        self.assertGreater(len(results['force_constants']), 0)
+        self.assertGreater(len(results['freqs']), 0)
```
alongd, line 144: The size of freqs (and the dimensions of the hessian) are deterministic, you can test them directly

arc/job/adapters/torch_ani_test.py
```diff
@@ -0,0 +1,228 @@
...
+        self.assertGreater(len(results['hessian'][0][0]), 0)
+        self.assertGreater(len(results['modes'][0][0]), 0)
+        self.assertGreater(len(results['reduced_masses']), 0)
+
+    def test_run_freq(self):
```
alongd, line 149: can you also test freqs for a TS, and see that you get a reasonable imaginary freq? (if tani at all supports TSs?)

arc/job/adapters/torch_ani_test.py
```diff
@@ -0,0 +1,228 @@
...
+        self.assertAlmostEqual(self.job_6.freqs[-1], 3756.57643, places=3)
+        self.assertGreater(self.job_7.freqs[-1], 3800)
+
+    def test_check_settings(self):
+        """Test the test_check_settings method"""
+        dummy = self.job_1
+        self.assertFalse(dummy.check_settings(settings={"model":"Kinetic Filter and Imputation Regression"}))
+        self.assertFalse(dummy.check_settings(settings={"device":"gpu"}))
+        self.assertFalse(dummy.check_settings(settings={"engine":"Knowledge-based Approach for Linear And Nonlinear optimization"}))
+        self.assertFalse(dummy.check_settings(settings={"fmax":"0.001"}))
```
alongd, line 164: why is this False?
Can you also test cases where the expected result is True?

arc/job/adapters/torch_ani_test.py
```diff
@@ -0,0 +1,228 @@
...
+                    'model': 'ani2x',
+                    'device': 'cpu',
+                    'fmax': 0.001,
+                    'engine': 'bfgs',
+                    'steps': None}
+        self.assertEqual(content, expected)
+        self.job_10.write_input_file(settings = tani_default_options_dict)
+        self.assertTrue(os.path.isfile(os.path.join(self.job_10.local_path, "input.yml")))
+        content = read_yaml_file(os.path.join(self.job_10.local_path, "input.yml"))
+        content = read_yaml_file(os.path.join(self.job_10.local_path, "input.yml"))
```
alongd, line 194: this line looks redundant

devtools/install_torchani.sh
```diff
@@ -0,0 +1,16 @@
...
+if command -v mamba 2>/dev/null; then
+    COMMAND_PKG=mamba
+elif command -v conda >/dev/null 2>&1; then
+    COMMAND_PKG=conda
+else
+    echo "Error: mamba or conda is not installed. Please download and install mamba or conda - we strongly recommend mamba"
+    exit 1
+fi
+$COMMAND_PKG clean -a -y
+echo creating tani_env
```
alongd, line 15: nice script!
would you like to add a `conda deactivate` command before cratering a new env?

devtools/install_torchani.sh
```diff
@@ -0,0 +1,16 @@
...
+$COMMAND_PKG env create -f devtools/tani.yml
```
alongd, line 16: this assumes that the current work dir is `ARC`. Is this a valid assumption?

arc/job/adapters/scripts/tani_script.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    Return the element number of an element.
+    
+    Args:
+        element (str): the element symbol. For example: "H", "Br"
+    
+    Returns:
+        The element number. For the example: 1, 35
+    """
+    try:
```
alongd, line 60: can do:
````python
elements.get(element, None)

File: arc/job/adapters/scripts/tani_script.py [hunk 16]
```diff
@@ -0,0 +1,373 @@
...
+            if len(atom_indices) == 2:
+                bonds.append([constraint[1], atom_indices])
+            if len(atom_indices) == 3:
+                angles.append([constraint[1], atom_indices])
+            if len(atom_indices) == 4:
+                dihedrals.append([constraint[1], atom_indices])
+        constraints = FixInternals(bonds=bonds, angles_deg=angles, dihedrals_deg=dihedrals)
+        atoms.set_constraint(constraints)
+
+    engine_list = ['bfgs', 'scipyfminbfgs', 'scipyfmincg']
```
alongd, line 122: maybe add `'SciPyFminBFGS'` as well, it's the default, but if the user inputs something else, then this won't be one of the options

arc/job/adapters/scripts/tani_script.py
```diff
@@ -0,0 +1,373 @@
...
+    engine_dict = {'bfgs': BFGS, 'scipyfminbfgs': SciPyFminBFGS, 'scipyfmincg': SciPyFminCG}
+    for opt_engine_name in engine_set:
+        opt_engine = engine_dict[opt_engine_name]
+        opt = opt_engine(atoms, logfile=None)
+        try:
+            opt.run(fmax=fmax, steps=steps)
+        except (Converged, NotImplementedError, OptimizerConvergenceError):
+            pass
+        else:
+            break
```
alongd, line 133: can you report back which engine was successful? even in just a plain `print`

arc/job/adapters/scripts/tani_script.py
```diff
@@ -0,0 +1,373 @@
...
+    species = torch.tensor(atoms.get_atomic_numbers(), device=device, dtype=torch.long).unsqueeze(0)
+    coordinates = torch.from_numpy(atoms.get_positions()).unsqueeze(0).requires_grad_(True)
+    masses = torchani.utils.get_atomic_masses(species)
+    energies = model.double()((species, coordinates)).energies
+    hessian = torchani.utils.hessian(coordinates, energies=energies)
+    freqs, modes, force_constants, reduced_masses = torchani.utils.vibrational_analysis(masses, hessian, mode_type='MDU')
+    freqs = freqs.numpy()
+    freqs = freqs[~np.isnan(freqs)]
+    results = {'hessian': hessian.tolist(),
+                'freqs': freqs.tolist(),
```
alongd, line 161: minor: can you sort out the indent (remove the extra space)?

arc/job/adapters/scripts/tani_script.py
```diff
@@ -0,0 +1,373 @@
...
+    if model is None:
+        model = torchani.models.ANI2x(periodic_table_index=True).to(device)
+
+    xyz = input_dict["xyz"]
+    opt_xyz = None
+    if job_type == 'sp':
+        sp = run_sp(xyz=xyz, device=device, model=model)
+        save_output_file(path = str(args.yml_path), key="sp", val=sp)
+
+    if job_type == 'force':
```
alongd, line 350: change `if`s to `elif`s, we know that only one of them is possible, this way the software won't even try the others once it hits one of them (not that it's very expensive to evaluate them, just for good practice)

arc/settings/settings.py
```diff
@@ -195,6 +195,11 @@
     'global': {'keyword': {'scf_convergence': 'TightSCF',
                            'dlpno_threshold': 'normalPNO'}},
 }
+tani_default_options_dict = {"model" : "ani2x",
+                             "device" : "cpu",
+                             "fmax" : 0.001,
+                             "engine" : "bfgs",
```
alongd, line 201: can you add comments here for what's the meaning of `bfgs`, for example, and what are the other options (also for the other values in this dict)

arc/parser.py
```diff
@@ -726,6 +727,12 @@ def parse_xyz_from_file(path: str) -> Optional[Dict[str, tuple]]:
                     start_parsing = True
     elif 'out' in file_extension or 'log' in file_extension:
         xyz = parse_geometry(path)
+    elif "yml" in file_extension or 'yaml' in file_extension:
+        contant = read_yaml_file(path=path)
```
alongd, line 731: typo: content

alongd (commented):
Great, thanks!!
I added a few comments and questions.
What do you think might be missing from this PR? (as you wrote on Slack)

devtools/install_torchani.sh
kfir4444: should we? you can activate an environment while in another environment.

devtools/install_torchani.sh
kfir4444: I think so, otherwise you cannot activate the `make install-torchani` command.
I can add a comment about that though

arc/job/adapters/torch_ani_test.py
kfir4444: Nope, they are defined as float or int. I can make a comment about that in the settings, and maybe raise an error when it happens.

arc/job/adapters/scripts/tani_script.py
kfir4444: doesn't work, because of jobs like `optfreq`

arc/job/adapters/scripts/tani_script.py
kfir4444: tried it, doesn't work

arc/job/adapters/torch_ani_test.py
kfir4444: I tried it naively and it didn't find any imaginary or negative frequency. I will leave the test for the future, maybe we need to specify something else

arc/job/adapters/torch_ani_test.py
kfir4444: No difference, but they are used in two different tests, so to prevent them clashing I made a new one.

arc/job/adapters/torch_ani.py
alongd: Still, if the `os.path.isfile(os.path.join(self.local_path, "output.yml"))` condition is evaluated to `False`, this line will cause ARC to crash - `output` will not be a dict and will not have these keys

arc/job/adapters/torch_ani_test.py
alongd: OK. Can you add tests with correct values where you assert that the `check_settings` method returns `True`?

arc/job/adapters/scripts/tani_script.py
alongd: It should work, see e.g.: https://stackoverflow.com/questions/11041405/why-dict-getkey-instead-of-dictkey
what error do you get?

arc/job/adapters/scripts/tani_script.py
alongd: So just leave the last one as `if`, but the 3 middle ones can be `elif`s

arc/job/adapters/scripts/tani_script.py
```diff
@@ -0,0 +1,372 @@
...
+    if job_type in ['opt', 'conformers', 'optfreq']:
+        constraints = input_dict["constraints"] if "constraints" in input_dict.keys() else None
+        opt_xyz = run_opt(xyz=xyz, constraints=constraints, fmax=input_dict["fmax"], model=model,
+                          steps=input_dict["steps"] if "steps" in input_dict.keys() else None, engine=input_dict["engine"])
+        save_output_file(path = str(args.yml_path), key="opt_xyz", val=opt_xyz)
+        sp = run_sp(xyz=opt_xyz, device=device, model=model)
+        save_output_file(path = str(args.yml_path), key="sp", val=sp)
+
+    if job_type in ['freq', 'optfreq']:
+
```
alongd, line 362: minor: can remove this extra line

alongd (commented):
Looking good! Added a few comments

alongd (commented):
Looking good, thanks!!

Issue #535, opened 2022-08-04
Running a job when the path contains spaces crushes.


kfir4444:
**Is your feature request related to a problem? Please describe.**
Trying to run arc as excusable in a directory with spaces in the name, causes failures.

**Describe the solution you'd like**
Whenever a job directory is called, if the path has spaces. replace them with a back slash, meaning:
`path = P rojects/ runs`
Then it will become:
 `path = P\ rojects/ runs`

**For users**
If you are a user, and have encountered this issue, please refrain from running jobs from folders with spaces until this issue is resolved.

PR #541, opened 2022-08-10
Issue #535 solved 

MichalKesl:
[](https://github.com/ReactionMechanismGenerator/ARC/issues/535)
If path has spaces the function will modify the name and  add backslash before the space.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -37,12 +37,10 @@ from rmgpy.qm.symmetry import PointGroupCalculator
 from arc.exceptions import InputError, SettingsError
 from arc.imports import home, settings
 
-
 if TYPE_CHECKING:
     from rmgpy.reaction import Reaction
     from rmgpy.species import Species
     from arc.reaction import ARCReaction
-
 
 logger = logging.getLogger('arc')
 logging.getLogger('matplotlib.font_manager').disabled = True
@@ -55,7 +53,6 @@ RMG_PATH = os.path.abspath(os.path.dirname(os.path.dirname(rmgpy.__file__)))
 RMG_DATABASE_PATH = os.path.abspath(os.path.dirname(rmgpy.settings['database.directory']))
 
 VERSION = '1.1.0'
-
 
 default_job_types, servers = settings['default_job_types'], settings['servers']
 
@@ -638,7 +635,8 @@ def get_bonds_from_dmat(dmat: np.ndarray,
             A list of tuple entries, each represents a bond and contains sorted atom indices.
     """
     if len(elements) != dmat.shape[0] or len(elements) != dmat.shape[1] or len(dmat.shape) != 2:
-        raise ValueError(f'The dimensions of the DMat {dmat.shape} must be equal to the number of elements {len(elements)}')
+        raise ValueError(
+            f'The dimensions of the DMat {dmat.shape} must be equal to the number of elements {len(elements)}')
     bonds, bonded_hydrogens = list(), list()
     charges = charges or [0] * len(elements)
     # Heavy atoms
@@ -684,7 +682,8 @@ def determine_symmetry(xyz: dict) -> Tuple[int, int]:
     # Coords is an N x 3 numpy.ndarray of atomic coordinates in the same order as `atom_numbers`.
     coords = np.array(xyz['coords'], np.float64)
     unique_id = '0'  # Just some name that the SYMMETRY code gives to one of its jobs.
-    scr_dir = os.path.join(home, 'tmp', 'symmetry_scratch')  # Scratch directory that the SYMMETRY code writes its files in.
+    scr_dir = os.path.join(home, 'tmp',
+                           'symmetry_scratch')  # Scratch directory that the SYMMETRY code writes its files in.
     if not os.path.exists(scr_dir):
         os.makedirs(scr_dir)
     symmetry = optical_isomers = 1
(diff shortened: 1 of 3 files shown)


rwest:
If you do want to just replace all the ` `  with `\ ` then you could consider  `folder_name.replace(' ','\ ')` instead of your split and recombine routine.

```python
In [3]: folder_name = 'foo bar  baz'

In [4]: folder_name.replace(' ','\ ')
Out[4]: 'foo\\ bar\\ \\ baz'
```

But there may be a better fix.  This won't escape anything except spaces.

Most functions don't need the path escaping. The few that do include `os.system()`, but it might be safer to replace those calls with `subprocess` and then you avoid the need.  But maybe it's for doing something over SSH where shell escaping is not avoidable. In that case, consider [`shlex.quote()`](https://docs.python.org/3/library/shlex.html#shlex.quote)  
> Return a shell-escaped version of the string s. The returned value is a string that can safely be used as one token in a shell command line, for cases where you cannot use a list.

arc/common.py
```diff
@@ -1565,8 +1565,10 @@ def _check_r_n_p_symbols_between_rmg_and_arc_rxns(arc_reaction: 'ARCReaction',
     """
     result = True
     num_rs, num_ps = len(arc_reaction.r_species), len(arc_reaction.p_species)
-    arc_r_symbols = [atom.element.symbol for atom in chain(*tuple(arc_reaction.r_species[i].mol.atoms for i in range(num_rs)))]
-    arc_p_symbols = [atom.element.symbol for atom in chain(*tuple(arc_reaction.p_species[i].mol.atoms for i in range(num_ps)))]
+    arc_r_symbols = [atom.element.symbol for atom in
+                     chain(*tuple(arc_reaction.r_species[i].mol.atoms for i in range(num_rs)))]
+    arc_p_symbols = [atom.element.symbol for atom in
+                     chain(*tuple(arc_reaction.p_species[i].mol.atoms for i in range(num_ps)))]
```
alongd, line 1571: All the modifications to the file up to this point are style changes. If you think they are necessary, please add them as a different commit. I would recommend removing them, the previous version is "OK", I think

arc/common.py
```diff
@@ -1616,3 +1618,22 @@ def safe_copy_file(source: str,
...
+    Adding backslash before a spaces in the folder name.
+
+    Args:
+        folder_name (str): The string directory we want to modify.
+
+    Returns:
+        new_folder_name (str): modified folder name.
+
+    """
+    folder_name.split(" ")
```
alongd, line 1634: this line has no effect (we don't store the returned value)

arc/common.py
```diff
@@ -1616,3 +1618,22 @@ def safe_copy_file(source: str,
...
+    new_folder_name = ""
+    for i in folder_name.split(" "):
```
alongd, line 1636: minor: I would recommend changing `i` to a different name, `i` is usually an integer iterator. maybe use `split`?

alongd: You can do either `folder_name.split()` or `folder_name.split(" ")`, they are the same (the first one is shorter or "cleaner")

arc/common.py
```diff
@@ -1616,3 +1618,22 @@ def safe_copy_file(source: str,
...
+        new_folder_name += i + "\ "
```
alongd, line 1637: If you'd like, you can achieve this in a single line instead of the loop. Try:
`new_folder_name  = '\ '.join(folder_name.split())`

alongd (commented):
Thanks, please see some comments

arc/common_test.py
```diff
@@ -1370,6 +1370,17 @@ def test_is_xyz_mol_match(self):
         self.assertFalse(common.is_xyz_mol_match(mol1, xyz2))
         self.assertTrue(common.is_xyz_mol_match(mol2, xyz2))
 
+    def test_fill_in_the_blanks(self):
+        """Test the fill_in_the_blanks() function"""
+        ex1 = "michalkfir"
+        ex2 = "michal kfir"
+        ex3 = "mich al kfir"
+        ex4 = "michal  kfir"
+        self.assertTrue(common.fill_in_the_blanks(ex1), "michalkfir")
```
alongd, line 1379: We should test here (and below) for `assertEqual` rather than `assertTrue`

arc/common_test.py
```diff
@@ -1370,6 +1370,17 @@ def test_is_xyz_mol_match(self):
...
+        self.assertTrue(common.fill_in_the_blanks(ex2), "michal\\ kfir")
+        self.assertTrue(common.fill_in_the_blanks(ex3), "mich\\ al\\ kfir")
+        self.assertTrue(common.fill_in_the_blanks(ex4), "michal\\ \\ kfir")
```
alongd, line 1382: why do we have two slashes in the tests?

alongd (commented):
Thanks @MichalKesl, looks good!
I added some comments, please take a look

alongd:
@MichalKesl, when you get a chance, can we finish this PR?

arc/common_test.py
MichalKesl: If we do `michal kfir.replace(" ", "\ ")` the string changes to `michal\\ kfir`, because \ is an operator (like \n).

arc/common_test.py
MichalKesl: Thank you I fixed that

calvinp0:
@MichalKesl is this PR relevant?

alongd:
@calvinp0, I revived it, could you please have a look and merge?

PR #542, opened 2022-08-14
Scissors fix

kfir4444:
Fixed the atoms shuffle in the Molecule object.

### DIFF
--- a/arc/commonTest.py
+++ b/arc/commonTest.py
@@ -8,6 +8,7 @@ This module contains unit tests for ARC's common module
 import copy
 import datetime
 import os
+import random
 import time
 import unittest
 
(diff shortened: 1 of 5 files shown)


arc/commonTest.py
```diff
@@ -1186,6 +1187,23 @@ def test_calc_rmsd(self):
...
+    def test_sort_atoms_in_decending_label_order(self):
+        "tests the sort_atoms_in_decending_label_order function"
+        mol = Molecule(smiles="C1CCCC1")
+        for index, atom in enumerate(mol.atoms):
+            atom.label = str(index)
+        random.shuffle(mol.atoms)
+        common.sort_atoms_in_decending_label_order(mol)
+        for index, atom in enumerate(mol.atoms):
+            self.assertEqual(str(index), atom.label)
+        mol = Molecule(smiles="NC1=NC=NC2=C1N=CN2 Nc1c2ncNc2ncn1")
```
alongd, line 1199: these SMILES have a space. Is this on purpose?

arc/species/species.py
```diff
@@ -1810,21 +1811,26 @@ def scissors(self,
...
-    def _scissors(self, indices: tuple) -> list:
+    def _scissors(self,
+                  indices: tuple,
+                  sort_atom_labels: bool = False) -> list:
         """
         Cut a chemical bond to create two new species from the original one, preserving the 3D geometry.
 
         Args:
             indices (tuple): The atom indices between which to cut (1-indexed, atoms must be bonded).
+            sort_atom_labels(bool, optional): Boolean flag, dettermines whether or not sorting is required.
```
alongd, line 1833: minor: space before `(`

arc/species/species.py
```diff
@@ -1810,21 +1811,26 @@ def scissors(self,
                             atom_indices_reverse = (atom_indices[1], atom_indices[0])
                             if atom_indices not in self.bdes and atom_indices_reverse not in self.bdes:
                                 self.bdes.append(atom_indices)
+        if sort_atom_labels:
```
alongd, line 1814: why do we want it to be False by default?

alongd (commented):
Thanks for this fix, and for working on it so quickly. Much appreciated. Please see some comments, most of them are minor.

arc/commonTest.py
kfir4444: Yes, without the space it is a different species

arc/species/species.py
kfir4444: I thought it would be better if it was not the default, since this overwrites the `atom.label`. On a second thought, changed it to True.

arc/commonTest.py
alongd: I think SMILES cannot have spaces. When I plot this on the RMG website, I just get the first structure (before the space). Can you double-check that this is what you originally meant?

alongd:
Looking good. We just have one remaining comment regarding the space in SMILES, and perhaps we can improve the new function in common?

arc/commonTest.py
kfir4444: Screenshot from 2022-08-15 14-03-30_edit
Screenshot from 2022-08-15 14-03-44_edit

alongd (approved):
Perfect, thanks!!

PR #543, opened 2022-08-24
Ignore IRC status when deciding whether a species has converged

alongd:
Also added a few error exceptions for robustness, and modified non-deterministic Heuristics tests


kfir4444 (approved):
Great fix @alongd!

PR #544, opened 2022-08-30
Allow one conformer to be True

MichalKesl:
Added the option to accept a single optimized conformer to be True.


alongd (approved):
Thanks @MichalKesl !

PR #545, opened 2022-09-01
Fixing TS IRC calcs

alongd:
Here we mainly fix how the scheduler treats species it creates from the IRC runs.
Other minor fixes were added along with tests.


josemvdh (approved):

josemvdh:
Thanks for the help on the issue Alon!

PR #546, opened 2022-09-06
Functional testing

kfir4444:


.github/workflows/cont_int.yml
```diff
@@ -77,3 +77,11 @@ jobs:
         echo $PYTHONPATH
         make test
         codecov --token=f259713a-7f1d-4e9c-b140-bb3bb371d3ef
+
+    - name: Functional Test
+      run: |
+        source ~/.bashrc
+        conda activate arc_env
+        echo $PYTHONPATH
```
alongd, line 85: is this still necessary?

Makefile
```diff
@@ -9,6 +9,9 @@
 test test-unittests:
 	python3 -m unittest discover -p *Test.py -v
 
+test-functional:
+	python3 -m unittest functional/functional_test.py
```
alongd, line 13: our of curiosity, does a `python` command equivalent to the `python3` command here? Do we think that `python` will execute Python 2?

docs/source/installation.rst
```diff
@@ -218,6 +218,13 @@ Tests
     make test
 
   After the tests complete, you may unstash your changes, if relevant (``git stash pop``).
+  
+  In addition, functional tests are helpful in making sure that ARC is installed and functioning correctly. Again, before performing these tests, it is reccommended to first stash your changes (``git stash``).
```
alongd, line 222: Can you add a line break after the `.`?

alongd (commented):
Looking clean and good! Thanks.
Please see some minor questions

Makefile
kfir4444: I do not think that matters, since the only python available in the environment is python 3.

functional/functional_test.py
```diff
@@ -0,0 +1,91 @@
...
+                    'freq': True,
+                    'sp': True,
+                    'rotors': False,
+                    'irc': False,
+                    }
+
+        cls.species_list_1 = [ARCSpecies(label= "2-propanol", smiles="CC(O)C"), ARCSpecies(label= "1-propanol", smiles="CCCO"), ARCSpecies(label= "NN", smiles="NN")]
+
+        cls.arc_object_1 = ARC(project='FunctionalThermoTest',
+                               project_directory=f"{ARC_PATH}/functional/test/thermo",
```
alongd, line 45: instead of using `/` we should always prefer `os.path.join()`. It usually just adds the `/`s, but it behaves more robustly across different OSs.

functional/functional_test.py
```diff
@@ -0,0 +1,91 @@
...
+        summary = self.arc_object_1.summary()
+        for _, ter in summary.items():
+            self.assertTrue(ter)
+
+    def testKinetic(self):
+        """Test kinetics"""
+        self.arc_object_2.execute()
+        summary = self.arc_object_2.summary()
+        for _, ter in summary.items():
+            self.assertTrue(ter)
```
alongd, line 78: can you check whether ARC actually got a rate? E.g., if a kinetic library was generated and it has relevant content, and/or if a rake coefficient is reported in the ARC log file? Same for the `testThermo`

docs/source/installation.rst
alongd: after `functioning correctly. `

alongd (commented):
Excellent, thanks!! I added a few minor commits, feel free to squash the fixes already and we can merge

functional/functional_test.py
```diff
@@ -0,0 +1,106 @@
...
+                    entry += 1
+        self.assertEqual(entry, len(self.species_list_1))
+
+    def testKinetic(self):
+        """Test kinetics"""
+        self.arc_object_2.execute()
+        summary = self.arc_object_2.summary()
+        for _, ter in summary.items():
+            self.assertTrue(ter)
+        self.assertTrue(os.path.exists(os.path.join(f"{ARC_PATH}", "functional", "test", "kinetic", "output", "RMG libraries", "kinetics")))
```
alongd, line 87: I think we should use `os.path.is_dir()`, which specifically checks for a directory, over `os.path.exists()`

functional/functional_test.py
```diff
@@ -0,0 +1,106 @@
...
+        has_content = False
+        with open(file = os.path.join(f"{ARC_PATH}", "functional", "test", "kinetic", "output", "RMG libraries", "kinetics", "reactions.py"), mode='r') as f:
+            for line in f.readlines():
+                if "Arrhenius" in line:
+                    has_content = True
+        self.assertTrue(has_content)
```
alongd, line 93: excellent!

functional/functional_test.py
```diff
@@ -0,0 +1,106 @@
...
+        summary = self.arc_object_1.summary()
+        for _, ter in summary.items():
+            self.assertTrue(ter)
+        self.assertTrue(os.path.exists(os.path.join(f"{ARC_PATH}", "functional", "test", "thermo", "output", "RMG libraries", "thermo")))
+        with open(file=os.path.join(f"{ARC_PATH}", "functional", "test", "thermo", "output", "RMG libraries", "thermo", "FunctionalThermoTest.py"), mode='r') as f:
+            entry = 0
+            for line in f.readlines():
+                if "entry" in line:
+                    entry += 1
+        self.assertEqual(entry, len(self.species_list_1))
```
alongd, line 79: excellent!
(minor: I would rename `entry` as `entries`)

alongd (approved):
Thanks, @kfir4444, for adding functional tests to ARC!

I added three comments, all very minor. Please decide which ones to address, if any, and we can merge this in.

PR #554, opened 2022-10-26
Store xyz in job adapter as_dict() as a string, not as a numpy array

alongd:
`job_dict` is used to "dump" the description of an ARC job in a YAML file, usually to create the restart file. Since YAML doesn't work well with numpy arrays, the xyz must be converted into a string format first. This is the case throughout ARC, but this line seems to be an exception and this PR fixes that.


MichalKesl (dismissed):
Looking good.

MichalKesl (approved):

PR #556, opened 2022-10-28
Process BDEs for cyclic structures

alongd:
The `scissors()` function performs well for cyclic species, but the BDEs processor was not set up properly.
Here we fix that.
Also added a test for `scissors()` for a cyclic species


kfir4444 (approved):
Thanks @alongd for the quick fix!
changes are approved.

PR #558, opened 2022-10-29
Don't use other ESS or other levels for troubleshooting

### LINKED ISSUE #479: Troubleshooting SP job by silently using wb97-x3/def2tzvp
I'm using a fairly modified version of ARC, but I don't think I've tampered with this particular aspect and I think it's a quite dangerous bug. When running CCSD(T)-F12//cc-pvdz-f12 single points that don't converge after enough troubleshooting ARC eventually troubleshoots the single point calculation by using wb97x-d3/def2tzvp. This results in a couple problems, the first is apart from the note in the log that it's running the fact that it does this is entirely silent (the only way I've been able to check this without trying to read the log file is by searching for species that don't have T1 recorded) and the second being that I'm pretty sure it's applying the AECs and BACs for CCSD(T)-F12//cc-pvdz-f12 to the wb97x-d3/def2tzvp calculation resulting in crazy thermochemistries and barriers. I think this troubleshooting method probably made a lot of sense for a opt or scan job, but needs a check to make sure it doesn't do so for an sp job.

alongd:
fixes #479 

We shouldn't silently troubleshoot jobs using different ESS and different levels. It is the users' responsibility to supply an adequate level of theory.

### DIFF
--- a/arc/job/adapter.py
+++ b/arc/job/adapter.py
@@ -999,6 +999,10 @@ class JobAdapter(ABC):
         self.job_status[1]['keywords'] = keywords
         self.job_status[1]['error'] = error
         self.job_status[1]['line'] = line.rstrip()
+        if 'Syntax' in self.job_status[1]['keywords']:
+            logger.error(error)
+            raise JobError(f'Got a syntax error in {self.job_adapter}.\n'
+                           f'Check the level of theory ({self.level}), its compatibility, and any other keywords.')
 
     def rename_output_file(self):
         """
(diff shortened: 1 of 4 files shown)


MichalKesl (commented):
The changes looks right to me, I would like to know why the tests failed after the removal of `available_ess` from trsh_ess_job().

alongd:
@MichalKesl, I think we can merge this once the tests pass

kfir4444:
@alongd @MichalKesl  Could we revive or close this PR?

MichalKesl (approved):

PR #560, opened 2022-11-05
save non standard object to yaml.

kfir4444:
Solves the error:
`can't pickle dict_values objects`
(and hopefully other similar errors).
Also added a unit test to the `to_yaml()` function.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -420,6 +420,9 @@ def to_yaml(py_content: Union[list, dict]) -> str:
     Returns: str
         The corresponding YAML representation.
     """
+    if not isinstance(py_content, (list, dict)):
+        logger.warning(f"Type of output data is not a list or dictionary. Trying to convert {type(py_content)} to list:")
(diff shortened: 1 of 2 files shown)


arc/common.py
```diff
@@ -420,6 +420,9 @@ def to_yaml(py_content: Union[list, dict]) -> str:
     Returns: str
         The corresponding YAML representation.
     """
+    if not isinstance(py_content, (list, dict)):
+        logger.warning(f"Type of output data is not a list or dictionary. Trying to convert {type(py_content)} to list:")
+        py_content = list(py_content)
```
alongd, line 425: Won't this store a (real) dictionary as a list or within a list? Will our restart still work? What's the intended outcome for a normal case (not when fixing the dict.values() error)?

kfir4444:
This is fixed in [#581](https://github.com/ReactionMechanismGenerator/ARC/pull/581)

PR #562, opened 2022-11-12
Troubleshoot job memory issues from server-generated log

alongd:
Here we add the ability to read a server-generated job log file (not the ESS log file).
The specific implementation is based on HTCondor `job.log` and can be extended.
A test was added

### DIFF
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -4,7 +4,7 @@ The ARC troubleshooting ("trsh") module
 
 import math
 import os
-from typing import Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -48,7 +48,9 @@ delete_command, inconsistency_ab, inconsistency_az, maximum_barrier, preserve_pa
 def determine_ess_status(output_path: str,
                          species_label: str,
                          job_type: str,
-                         software: str = None):
+                         job_log: Optional[str] = None,
+                         software: Optional[str] = None,
+                         ) -> Tuple[str, List[str], str, str]:
     """
     Determine the reason that caused an ESS job to crash, assign error keywords for troubleshooting.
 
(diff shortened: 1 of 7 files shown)


arc/job/trsh.py
```diff
@@ -417,6 +424,39 @@ def determine_ess_status(output_path: str,
...
+    """
+    keywords, error, line = list(), '', ''
+    if job_log is not None:
+        if os.path.isfile(job_log):
+            with open(job_log, 'r') as f:
+                lines = f.readlines()
+        else:
+            lines = job_log.splitlines()
+        for line in lines:
+            if 'MEMORY EXCEEDED' in line:
```
kfir4444, line 447: I think we should use lower case letters and use `.lower()`, so it could try to capture additional errors in other servers, in additional to this one.

arc/scheduler.py
```diff
@@ -887,9 +887,22 @@ def end_job(self, job: 'JobAdapter',
...
+            original_mem = job.job_memory_gb
+            if job.job_status[1]['error'] == 'Insufficient job memory.':
+                job.job_memory_gb *= 3
+                logger.warning(f'Job {job.job_name} errored because of insufficient memory. '
+                               f'Was {original_mem} GB, rerunning job with {job.job_memory_gb} GB.')
+                self._run_a_job(job=job, label=label)
+            elif job.job_status[1]['error'] == 'Memory requested is too high.':
+                logger.warning(f'Job {job.job_name} errored because the requested memory is too high. '
+                               f'Was {original_mem} GB, rerunning job with {job.job_memory_gb} GB.')
+                job.job_memory_gb *= 0.5
```
kfir4444, line 900: I think we could try to save the used memory and the requested memory to make a safer guess than just cutting the job memory in half. This could cause an unstable jobs, that alternate too much and not enough memory. Either way, let's prioritize implementing maximum iterations per jobs in light of this change.

kfir4444 (commented):
Thanks @alongd!
please see these couple of comments.

alongd:
Thanks, comments addressed and squashed

kfir4444 (approved):
Great job! Thanks! @alongd

kfir4444 (commented):
Let's wait for the CI to pass, and then merge.

kfir4444 (approved):
Thanks @alongd

PR #566, opened 2022-11-22
Lite intallation

kfir4444:
Added a lite installation feature, managed to drop the arc's size from ~190MB to ~80MB.

### DIFF
--- a/devtools/lite.sh
+++ b/devtools/lite.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+# delete all the Test and unnecessary files
+pushd .
+
+rm -rf functional
+rm -rf ipython
+rm -rf arc/testing
+
+find "$PWD" -path '*Test.py' -type f -delete
+
+popd || exit
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ docs/build/*
 
 # timer.dat
 timer.dat
+
+# .vscode
+.vscode
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,9 @@ install-sella:
 install-xtb:
 	bash devtools/install_xtb.sh
 
+lite:
+	bash devtools/lite.sh
+
 clean:
 	find -type d -name __pycache__ -exec rm -rf {} +
 	rm -rf testing


devtools/lite.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# delete all the Test and unnecessary files
+pushd .
```
alongd, line 3: When using `pushd .` (storing the current path), I think we should end the script with `popd || exit` (returning to the original path)

devtools/lite.sh
```diff
@@ -0,0 +1,9 @@
...
+
+rm -r functional
```
alongd, line 5: Should we add `-f` (so `-rf` actually) to ignore nonexisting files? Might be useful if a user executed this script, stopped it for some reason, and then run it again

devtools/lite.sh
```diff
@@ -0,0 +1,9 @@
...
+rm -r ipython
+rm -r arc/testing
+
+find "$PWD" -path '*Test.py' -type f -delete
```
alongd, line 9: What does this do?

alongd (commented):
Thanks!

devtools/lite.sh
kfir4444: good point!

devtools/lite.sh
kfir4444: This finds all the files of the type `*Test.py` in the current working directory, and delete them.

devtools/lite.sh
```diff
@@ -0,0 +1,11 @@
...
+# delete all the Test and unnecessary files
+pushd .
+
+rm -rf functional
+rm -rf ipython
+rm -rf arc/testing
+
+find "$PWD" -path '*Test.py' -type f -delete
+
+popd || exit
```
alongd, line 11: Minor: please finish scripts with an empty line

alongd (commented):
Looking good, added a minor comment. Would you like to add all the `make` option to the docs?

alongd (approved):
Thanks!

PR #567, opened 2022-11-23
Parse xTB freqs from an auxiliary file if needed

alongd:
Solves an issue where xTB save frequencies in a vibspectrum file instead of the primary output file

### DIFF
--- a/arc/parser.py
+++ b/arc/parser.py
@@ -14,7 +14,7 @@ from rmgpy.exceptions import InputError as RMGInputError
 from arkane.exceptions import LogError
 from arkane.ess import ess_factory, GaussianLog, MolproLog, OrcaLog, QChemLog, TeraChemLog
 
-from arc.common import determine_ess, get_close_tuple, get_logger, is_same_pivot, read_yaml_file
+from arc.common import determine_ess, get_close_tuple, get_logger, is_same_pivot, is_str_float, read_yaml_file
 from arc.exceptions import InputError, ParserError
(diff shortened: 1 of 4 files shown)


arc/parser.py
```diff
@@ -127,7 +127,19 @@ def parse_frequencies(path: str,
...
-
+        if freqs.size == 0:
+            # Could not read freqs from output.out, try reading from vibspectrum
+            vibspectrum_path = os.path.join(os.path.dirname(path), 'vibspectrum')
+            if os.path.isfile(vibspectrum_path):
+                lines = _get_lines_from_file(path=vibspectrum_path)
+                for line in lines:
+                    if '$' not in line and '#' not in line:
+                        splits = line.split()
+                        if len(splits) < 5:
```
Laxzal, line 138: Hey Alon, what if the splits is higher than 5? Or are we expecting the splits to always be lower than 5?

alongd: We expect to have 5 or 6 splits, e.g.:
```
     5                       0.00         0.00000        -       - 
     6        a           4225.72         0.00000       YES     YES
```

This checks whether the num of splits is less than 5 (4 or lower), if it is then we skip the line (`continue`)

Laxzal: Okay thanks, just wanted to double check that we wouldn't lose any information in the if len(splits) check!

Laxzal (approved):
Code looks good. Also tested it from my side and the run was successful and worked as intended.

PR #571, opened 2022-12-02
Tests: Added test_determine_symmetry to common tests

alongd:
The symmetry script (from the RMG qm module) was not tested in ARC. Added some simple tests.

### DIFF
--- a/arc/common_test.py
+++ b/arc/common_test.py
@@ -827,6 +827,42 @@ class TestCommon(unittest.TestCase):
         self.assertTrue(common.check_that_all_entries_are_in_list(
             bonds, [(0, 1), (1, 2), (0, 3), (0, 4), (1, 6), (2, 7), (2, 8), (2, 9), (5, 10)]))  # No (1, 5) bond.
 
+    def test_determine_symmetry(self):
+        """
+        Test the determine_symmetry() function
+        The tests implemented here determine the external symmetries of ethanol (non-symmetric)
+        methane (highly symmetric) and a chiral species.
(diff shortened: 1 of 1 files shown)


arc/commonTest.py
```diff
@@ -827,6 +827,31 @@ def test_get_bonds_from_dmat(self):
         self.assertTrue(common.check_that_all_entries_are_in_list(
             bonds, [(0, 1), (1, 2), (0, 3), (0, 4), (1, 6), (2, 7), (2, 8), (2, 9), (5, 10)]))  # No (1, 5) bond.
 
+    def test_determine_symmetry(self):
+        """Test the determine_symmetry() function"""
```
Laxzal, line 831: changes look good. the only thing I would like added is just a bit more explanation of the test.
inlcuding the phrase
""" Test the determine symmetry function"""
Also include about which the 3 tests in this function are doing and what you are expecting. I just think adding a bit more description will help when this test is run again in the future by someone else.

Laxzal (changes requested):

alongd:
Thanks, @Laxzal. Added and squashed.

alongd:
@Laxzal, I think we can merge this in once tests pass

kfir4444 (changes requested):
Thanks @alongd!
I think the tests misses one case, where there are both symmetry and chirality. Could you add it?
Example: `CC=C=CC`

alongd:
Thanks for the suggestion, @kfir4444, added

kfir4444 (approved):
Thanks for addressing the comments @alongd,
approved and merged

PR #572, opened 2022-12-02
Parity plots should be drawn as squares to emphesize the pair values

alongd:
Setting the width and height dimensions to the same value.


kfir4444 (approved):
Thanks @alongd!

PR #573, opened 2022-12-02
BugFix: Store the scan_path in the restart file

alongd:
The .pivots attribute of Job adapters is a list within a list, check whether the pivots equal the .pivots or the .pivots[0] list and update the scan_path if needed

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -943,7 +943,7 @@ class Scheduler(object):
                         self.species_dict[label].checkfile = check_path
             if job.job_type == 'scan' or job.directed_scan_type == 'ess':
                 for rotors_dict in self.species_dict[label].rotors_dict.values():
-                    if rotors_dict['pivots'] == job.pivots:
+                    if rotors_dict['pivots'] in [job.pivots, job.pivots[0]]:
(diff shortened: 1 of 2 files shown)


kfir4444:
Thanks @alongd!
Could you please enumerate about the bug and the way you fixed it?

alongd:
We store various attribute for a rotor such as the four atom indices in the torsion, e.g., `[1, 2, 3, 7]`.
To define the rotor we need the "torsion" list, but to just identify it and report to the user we use only the two middle atoms on the torsional axis, the "pivots", e.g., `[2, 3]`.
To accommodate several ND rotors per species, torsions and pivots are not lists of ints, but lists of lists of ints.
The way pivots are stored per rotor dictionary (in `ARCSpecies.rotors_dict`) is a list of ints if it is a 1D rotor. But the way they are stored in a Job is more generic to accommodate ND rotors.

When we end a scan (especially when ending a scan that underwent troubleshooting) we don't always store the path to the scan output file correctly. Here we fix that by improving the condition for identifying a rotor by its pivots.

kfir4444 (approved):
Thanks @alongd!

PR #574, opened 2022-12-07
Codecov fixup

kfir4444:


.github/workflows/cont_int.yml
```diff
@@ -83,3 +83,6 @@ jobs:
         conda activate arc_env
         make test-functional
         codecov --token=f259713a-7f1d-4e9c-b140-bb3bb371d3ef
```
alongd, line 85: Can you check whether or not we need to keep this line?

alongd:
After the tests I see 72% coverage:
```
TOTAL                                    25093   5756  10866   1357    72%
```

How come here it says 24.35%?

alongd:
Not all tests were renamed, e.g., `commonTest.py`. Was this on purpose?

functional/functional_test.py
```diff
@@ -92,7 +93,10 @@ def testThermo(self):
 
     def testKinetic(self):
         """Test kinetics"""
-        self.arc_object_2.execute()
+        try:
+            self.arc_object_2.execute()
+        except InputError as e:
+            self.skipTest(f"execution of ARC failed: {e}, skipping this test for now...")
```
alongd, line 99: why do we skip this? Don't we expect this test to pass, and if it doesn't then this means we broke something?

alongd (commented):
Looking great! Thanks for upgrading ARC's tests and making coverage functional again
Added just one question

functional/functional_test.py
kfir4444: No, it's the same error we receive sometimes in this test. I added an [issue](https://github.com/ReactionMechanismGenerator/ARC/issues/578) with additional information and potential fixes.

alongd (approved):
Thanks!

PR #575, opened 2022-12-11
Use self.scan_res in Job adapter instead of redefining it

alongd:
self.scan_res is already set in _initialize_adapter() in common.py


kfir4444 (approved):
Thanks @alongd!
This is a very good step into implementing local settings, in addition to the global settings.

Issue #576, opened 2022-12-11
Error with syntax


naddeu:
Hello, 
when running ARC with basis set "def2-tzvp" received an error:
`Warning: Troubleshooting h job opt_a15284 which failed with status: "errored,"
with keywords: ['Syntax'] in gaussian. 
Troubleshooting opt job in gaussian for h using int=(Acc2E=14)`
and Gaussian started using cbs-qb3 for the optimization job

This is the input.yml file I used
`#AutoTST
project: ax_1010

level_of_theory: B3LYP/6-31G(d,p)//apfd/def2-tzvp

species:
  - label: a
    smiles: C[CH]O
    
  - label: a1
    smiles: '[CH2]CO'

  - label: b
    smiles: C[O]

  - label: b1
    smiles: '[CH2]O'

  - label: c
    smiles: '[CH2]CCO'

  - label: c1
    smiles: C[CH]CO

  - label: d
    smiles: '[CH2]CCC'

  - label: d1
    smiles: CC[CH]C
    
  - label: e
    smiles: CCCC([O])=O
    
  - label: e1
    smiles: '[CH2]CCC(=O)O'

  - label: f
    smiles: CC(=O)O[O]

  - label: f1
    smiles: '[CH2]C(=O)OO'

  - label: g
    smiles: C[CH]C1C=CC=C1

  - label: g1
    smiles: '[CH2]CC1C=CC=C1'

  - label: h
    smiles: '[CH]1C=CC=CCC1'

  - label: h1
    smiles: '[CH]1CC=CC=CC1'
      
  - label: i
    smiles: C=CC1[CH]CC=C1
    
  - label: i1
    smiles: C=C[C]1C=CCC1

  - label: j
    smiles: C=C1[CH]C(C)C=C1

  - label: j1
    smiles: '[CH2]C1C=CC(=C)C1'

reactions:
  - label: a <=> a1
  
  - label: b <=> b1

  - label: c <=> c1
  
  - label: d <=> d1
  
  - label: e <=> e1
  
  - label: f <=> f1

  - label: g <=> g1
  
  - label: h <=> h1
  
  - label: i <=> i1

  - label: j <=> j1

`
I hoped to receive valid reaction rates with optimized TS geometries

kfir4444:
Dear @naddeu,
Thanks for reporting this issue, it is now solved with [this PR](https://github.com/ReactionMechanismGenerator/ARC/pull/577),
[...]

PR #577, opened 2022-12-11
Fixup def2 tzvp

kfir4444:
This PR helps to solve issues such as [issue#576](https://github.com/ReactionMechanismGenerator/ARC/issues/576).
Starting with `def2tzvp` due to it's frequency of occurrence.
In the future, this problem must be solved more robustly, such that other basis sets/methods will be caught early.
Thanks @naddeu for bringing this issue up!

### DIFF
--- a/arc/job/adapters/gaussian.py
+++ b/arc/job/adapters/gaussian.py
@@ -7,6 +7,7 @@ https://gaussian.com/
 import datetime
 import math
 import os
+import re
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
 from mako.template import Template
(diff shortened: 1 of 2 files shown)


kfir4444:
For additional [context](https://gaussian.com/basissets/), it shows that the hyphen is dropped from all `def2` types of basis

arc/job/adapters/gaussian.py
```diff
@@ -193,6 +194,9 @@ def __init__(self,
                             tsg=tsg,
                             xyz=xyz,
                             )
+        if isinstance(self.level, Level):
```
alongd, line 197: can you connect the two conditions in the same line?

arc/job/adapters/gaussian.py
```diff
@@ -193,6 +194,8 @@ def __init__(self,
                             tsg=tsg,
                             xyz=xyz,
                             )
+        if isinstance(self.level, Level) and self.level.basis is not None:
+                self.level.basis = re.sub('def2-', 'def2', self.level.basis.lower())
```
alongd, line 198: I just noticed that this is implemented under `_initialize_adapter` and will be applied to all adapters. I think it must be relocated just to the Gaussian adapter, since Q-Chem, for example, does things just the opposite way... (required the dash)

alongd (approved):
Thanks!!

PR #579, opened 2022-12-18
fixup GHP

kfir4444:


.github/workflows/gh-pages.yml
```diff
@@ -14,23 +14,25 @@ jobs:
         shell: bash -l {0}
 
     steps:
-      - uses: actions/checkout@v2
+    - uses: actions/checkout@v2
```
alongd, line 17: The syntax looks non YAML-like, shouldn't there be a double space there?

kfir4444: This is taken from our CI workflow, so probably not.

alongd (approved):
Thanks!

PR #585, opened 2022-12-26
submit.py renamed to arc_submit.py

### LINKED ISSUE #584: We should rename submit.py
We should rename `submit.py` to, e.g., `arc_submit.py`, see https://github.com/ReactionMechanismGenerator/T3/pull/91
After this change, the docs should be updated and all current users must be notified so they can rename their `.arc/submit.py`.

Rishav1707:
### This PR closes the following issue:
closes #584 

### Changes made:
- Renamed `submit.py` to `arc_submit.py` file.


kfir4444:
Thanks @Rishav1707,
however, this issue requires additional work, for example, changing the references for this file in the code.

alongd:
Hi @Rishav1707, we're happy to see contributions from the community. But we also expect them to be comprehensive. The fix you suggested should include additional modifications in the actual code to correctly import this file after it is renamed. As the feature request suggests, the docs should also be updated appropriately. It's indeed a "good first issue", but I believe that its appropriate for users of this software who would like to contribute, since a bigger context is needed.

PR #587, opened 2023-01-01
Updating devtool installation

Laxzal:
devtool bash scripts search for mamba, and if found, will use mamba as the installation command.

### DIFF
--- a/devtools/install_autotst.sh
+++ b/devtools/install_autotst.sh
@@ -2,15 +2,16 @@
 CONDA_BASE=$(conda info --base)
 source $CONDA_BASE/etc/profile.d/conda.sh
 
-
-if [ $(mamba) ] > /dev/null 2>&1; then 
-	COMMAND_PKG=conda
-	echo "conda found"
-	else
+#Check if mamba/conda is installed
+if [ -x "$(command -v mamba)" ]; then
+	echo "mamba is installed."
 	COMMAND_PKG=mamba
-	echo "mamba found"
+elif [ -x "$(command -v conda)" ]; then
+	echo "conda is installed."
+	COMMAND_PKG=conda
+else
+    echo "mamba and conda are not installed. Please download and install mamba or conda - we strongly recommend mamba"
 fi
-
 
 # temporarily change directory to install software, and move one directory up in the tree
 pushd .
--- a/devtools/install_autotst.sh
+++ b/devtools/install_autotst.sh
@@ -2,15 +2,14 @@
 CONDA_BASE=$(conda info --base)
 source $CONDA_BASE/etc/profile.d/conda.sh
 
-#Check if mamba/conda is installed
-if [ -x "$(command -v mamba)" ]; then
-	echo "mamba is installed."
-	COMMAND_PKG=mamba
-elif [ -x "$(command -v conda)" ]; then
-	echo "conda is installed."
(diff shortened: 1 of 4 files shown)


devtools/install_autotst.sh
```diff
@@ -2,6 +2,16 @@
 CONDA_BASE=$(conda info --base)
 source $CONDA_BASE/etc/profile.d/conda.sh
 
+
+if [ $(mamba) ] > /dev/null 2>&1; then
```
kfir4444, line 6: Can you explain this line to me?

devtools/install_autotst.sh
```diff
@@ -2,6 +2,16 @@
 CONDA_BASE=$(conda info --base)
 source $CONDA_BASE/etc/profile.d/conda.sh
 
+
+if [ $(mamba) ] > /dev/null 2>&1; then 
+	COMMAND_PKG=conda
+	echo "conda found"
+	else
+	COMMAND_PKG=mamba
+	echo "mamba found"
```
kfir4444, line 11: What if the user doesn't have both?

kfir4444 (changes requested):
Thanks @Laxzal!
I have some comments/questions, could you address them?

devtools/install_autotst.sh
Laxzal: It's just a redirection line. 
/dev/null accepts and discards all input;
2>&1 merges output from stream 2 with stream 1

So basically I am telling the system that it doesn't need to report anything in the terminal to the user

devtools/install_autotst.sh
Laxzal: I guess you could echo that they are missing an installation command but I think they would run into this error when it cannot install anyway. 

Further, maybe we should add to the website instructions to just tell them to install anaconda or mambaforge

devtools/install_sella.sh
```diff
@@ -2,18 +2,46 @@
...
+touch ./etc/conda/deactivate.d/env_vars.sh
+
+#This sets up the LD_LIBRARY_PATH to include Sella_Env, but only when the environment is active
+echo 'export OLD_LD_LIBRARY_PATH=${LD_LIBRARY_PATH}' >> $CONDA_BASE/envs/sella_env/etc/conda/activate.d/env_vars.sh
+echo 'export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:'"$CONDA_BASE"'/envs/sella_env/lib' >> $CONDA_BASE/envs/sella_env/etc/conda/activate.d/env_vars.sh
+#This will reset the LD_LIBRARY_PATH back to the original LD_LIBRARY_PATH when the environment is deactivated
+echo 'export LD_LIBRARY_PATH=${OLD_LD_LIBRARY_PATH}' >> $CONDA_BASE/envs/sella_env/etc/conda/deactivate.d/env_vars.sh
+echo 'unset OLD_LD_LIBRARY_PATH' >> $CONDA_BASE/envs/sella_env/etc/conda/deactivate.d/env_vars.sh
+
+
```
kfir4444, line 44: Can you remove some of these empty lines?

kfir4444 (commented):
Thanks @Laxzal!
One quick comment

kfir4444 (approved):
Thanks @Laxzal! Great job!

PR #590, opened 2023-01-08
Allow heuristics TS generation for reactions with isomorphic reactants or products

alongd:
This PR solves a bug where ARC got stuck when a reaction had two identical reactants or two identical products.
The issue was traced to the function `mapping.map_arc_rmg_species()`, where a very simple fix was added

Tests were added to `Reaction.get_reactants_and_products()`, `mapping.map_arc_rmg_species()`, `heuristics.react()`, and `heuristics_for_h_abstraction`

### DIFF
--- a/arc/reaction.py
+++ b/arc/reaction.py
@@ -741,6 +741,7 @@ class ARCReaction(object):
 
     def get_reactants_and_products(self,
                                    arc: bool = True,
+                                   return_copies: bool = True,
                                    ) -> Tuple[List[Union[ARCSpecies, Species]], List[Union[ARCSpecies, Species]]]:
         """
         Get a list of reactant and product species including duplicate species, if any.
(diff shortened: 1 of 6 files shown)


kfir4444:
@alongd can you rebase so this can pass CI?

arc/reaction.py
```diff
@@ -756,16 +758,20 @@ def get_reactants_and_products(self,
         reactants, products = list(), list()
         for r_spc in self.r_species:
             if arc:
-                reactants.extend([r_spc] * self.get_species_count(species=r_spc, well=0))
+                for i in range(self.get_species_count(species=r_spc, well=0)):
+                    reactants.append(r_spc.copy() if return_copies else r_spc)
```
kfir4444, line 762: just for future safety, I would have used deep copy here. This causes issues that are really hard to debug in the future.

kfir4444 (approved):
Thanks @alongd!
I added a single comment, if you want to fix it in a later PR it's fine.

arc/reaction.py
alongd: The objects here are `ARCSpecies`, not RMG `Species`, we do something equivalent to `deep=True` in ARC

PR #591, opened 2023-01-15
Generate H Abstraction TSs for reactions with linear motifs

alongd:
ARC's heuristics approach for generating TS guesses depends on Z Matrices. Currently, we do not consider dummy atoms correctly (required for defining linear motifs in a ZMat). This PR fixes two minor bugs, and adds a bunch of tests.


kfir4444 (approved):
Thanks for the contribution @alongd!

PR #594, opened 2023-01-18
Replace ":" in folder name when creating folders in the scale.py script

alongd:
This creates a folder for a "sub" ARC job to compute freq scale factors for new levels of theory.
When we use solvation the `__str()__` method of the `Level` class outputs `:` chars.
Here we replace them with `..` when we use the level of theory string representation to name folders.
A test was added


kfir4444 (approved):
Thanks @alongd!

kfir4444 (approved):

PR #596, opened 2023-01-28
Adapted the composite scan/freq/irc levels to G4

alongd:
We use G4 more frequently than CBS-QB3
See: https://aip.scitation.org/doi/10.1063/1.2436888


kfir4444 (commented):
Thanks @alongd !
We could use this opportunity to discuss the comments in our code.

alongd:
G4 didn't give promising results, perhaps the BACs aren't optimized, closing this PR for now

Issue #593, opened 2023-01-16
1D scan files are all saved as `[].png`


alongd:
**Describe the bug**
1D scan files are all saved as `[].png`. This is uninformative, plus if there's more than one 1D scan, the images are overwritten and only the last scan remains.

**How to reproduce**
Run `CCO` in ARC, see the the output under the `rotors` folder

**Expected behavior**
A unique and informative file name should be generated per 1D rotor

**Additional context**
In the "Job" PR we made the "torsions" attribute of a rotor a list of lists. Probably this caused a bug in naming these files.

PR #597, opened 2023-01-28
Fix 1D rotor scan file names

alongd:
The `scan` argument in `plot_1d_rotor_scan()` is a list of lists (for 1D scan it's a single list within a list). This modification was introduced in the Job PR, but created a bug that 1D scan image filenames weren't generated correctly. All fig names were `[].png`, and if a species has more than one 1D rotor, the images override each other and only the last instance is visible.
Here we fix this issue.

fixes https://github.com/ReactionMechanismGenerator/ARC/issues/593


kfir4444 (approved):
Thanks @alongd! Approved.

PR #598, opened 2023-02-01
Changed the conversion of MB to Bytes to 1E6

Laxzal:
Originally in the code, the memory for PBS was being converted by multiplying by 1E3. This has been corrected now to 1E6


kfir4444 (approved):
Great job @Laxzal!

PR #601, opened 2023-02-19
Openbabel

kfir4444:
This PR adds a OpenBabel Job adapter for opt/sp jobs


arc/job/adapters/obabel.py
```diff
@@ -0,0 +1,270 @@
...
+                            tsg=tsg,
+                            xyz=xyz,
+                            )
+
+
+    def write_input_file(self, settings) -> None:
+        """
+        Write the input file to execute the job on the server.
+        """
+        if self.level.method is not None:
```
alongd, line 175: Can this adapter be called with `self.level.method` set to `None`? If so, how would ARC know to run this adapter?

arc/job/adapters/obabel.py
```diff
@@ -0,0 +1,270 @@
...
+        Execute a job incore.
+        """
+        self._log_job_execution()
+
+        if "freq" in self.job_type:
+            raise NotImplementedError("Openbabel does not contain a frequency calculation option.\n"
+                                      "Use another ESS from the available options.")
+        
+        if self.level.method is not None:
+            ob_default_settigns["FF"] = self.level.method
```
alongd, line 233: I think this is repetitive with what we have under `write_input_file`

arc/job/adapters/obabel.py
```diff
@@ -0,0 +1,270 @@
...
+
+        if not self.check_settings(ob_default_settigns):
+            logger.error(f'Given wrong settings in input: {ob_default_settigns}\n'
```
alongd, line 236: I think we should raise an error here, not just log an error

arc/job/adapters/scripts/ob_script.py
```diff
@@ -0,0 +1,211 @@
...
+    mol = xyz_to_OBMol(xyz)
+    if job_type == 'sp':
+        sp = run_sp(mol=mol, ff_method=ff_method)
+        save_output_file(path = str(args.yml_path), key="sp", val=sp)
+
+    if job_type in ['opt', 'conformers']:
+        constraints = input_dict["constraints"] if "constraints" in input_dict.keys() else None
+        opt_xyz, sp = constraint_opt(mol=mol, constraints_dict=constraints, ff_method=ff_method)\
+                      if constraints is not None else unconstraint_opt(mol=mol, ff_method=ff_method)
+        save_output_file(path = str(args.yml_path), key="opt_xyz", val=opt_xyz)
```
alongd, line 202: Can you check whether `args.yml_path` is a string or a list? I think it usually returns a list, and we should do something like `args.yml_path[0]` to access the content

alongd, line 202: Can you check whether `args.yml_path` is a string or a list? I think it usually returns a list, and we should do something like `args.yml_path[0]` to access the content

arc/job/adapters/scripts/ob_script.py
```diff
@@ -0,0 +1,211 @@
...
+def save_output_file(path,
+                     key = None,
+                     val = None,
+                     content_dict = None,
+                     ):
+    """
+    Save the output of a job to the YAML output file.
+    Args:
+        key (str, optional): The key for the YAML output file.
+        val (Union[float, dict, np.ndarray], optional): The value to be stored.
```
alongd, line 166: YAML file are infamouse for their treatment of np value types
I suggest to make sure not to pass `np.ndarray` here, or to convert it inside the function

arc/job/adapters/scripts/ob_script.py
```diff
@@ -0,0 +1,211 @@
...
+        content_dict (dict, optional): A dictionary to store.
+    """
+    yml_out_path = os.path.join(path, 'output.yml')
+    content = read_yaml_file(yml_out_path) if os.path.isfile(yml_out_path) else dict()
+    if content_dict is not None:
```
alongd, line 171: Can `content_dict` be `None` after the above line? Did you mean to check whether it's not an empty dict?

arc/job/adapters/scripts/ob_script.py
```diff
@@ -0,0 +1,211 @@
...
+    Returns: Union[dict, list]
+        The content read from the file.
+    """
+    with open(path, 'r') as f:
+        content = yaml.load(stream=f, Loader=yaml.FullLoader)
+    return content
+
+
+def save_yaml_file(path: str,
+                   content: list,
```
alongd, line 143: looks line content is a dict

arc/job/adapters/scripts/ob_script.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def run_sp(mol : openbabel.OBMol, ff_method : str) -> int:
+    ff = openbabel.OBForceField.FindForceField(ff_method)
+    ff.Setup(mol)
+    return ff.Energy()*4.18 # kcal/mol to kj/mol
+
+
+
+def unconstraint_opt(mol: openbabel.OBMol,
```
alongd, line 29: I wonder if the two opt functions can be merged (and called through one directive in `main`) for better maintainability

arc/job/adapters/scripts/ob_script.py
```diff
@@ -0,0 +1,211 @@
...
+        econv (float): the convergence parameter.
+        ff_method (str): the force field method required for the optimization.
+    Returns:
+        optimized xyz string, energy in kj/mol
+    """
+    ff = openbabel.OBForceField.FindForceField(ff_method)
+    ff.Setup(mol)
+    ff.SteepestDescent(steps, econv)
+    ff.GetCoordinates(mol)
+    energy = ff.Energy()*4.18 # kcal/mol to kj/mol
```
alongd, line 47: maybe use 4.184 which is a bit more exact? (although FF energies are not accurate anyway, but why not)

devtools/install_ob.sh
```diff
@@ -0,0 +1,15 @@
...
+if command -v mamba 2>/dev/null; then
+    COMMAND_PKG=mamba
+elif command -v conda >/dev/null 2>&1; then
+    COMMAND_PKG=conda
+else
+    echo "Error: mamba or conda is not installed. Please download and install mamba or conda - we strongly recommend mamba"
+    exit 1
+fi
+$COMMAND_PKG clean -a -y
+$COMMAND_PKG env create -n ob_env conda-forge::openbabel
```
alongd, line 15: we should add YAML to this env

arc/job/adapters/obabel_test.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def test_write_input_file(self):
+        """Test the write_input_file function"""
+        self.job_3.write_input_file(settings = ob_default_settigns)
+        self.assertTrue(os.path.isfile(os.path.join(self.job_3.local_path, "input.yml")))
+        content = read_yaml_file(os.path.join(self.job_3.local_path, "input.yml"))
+        expected = {'FF': 'ghemical',
+ 'job_type': 'opt',
+ 'opt_gradient_settings': {'econv': 1e-06, 'steps': 2000},
+ 'xyz': '9\n'
```
alongd, line 118: You could use a multiline string `"""` to make sections like that more readable

arc/job/adapters/__init__.py
```diff
@@ -2,7 +2,7 @@
 import arc.job.adapters.gaussian
 import arc.job.adapters.molpro
 import arc.job.adapters.orca
-import arc.job.adapters.openbable
+import arc.job.adapters.obabel
```
alongd, line 5: can you squash this commit into the original change?

alongd (commented):
Perfect, thanks for this addition!!
Please see some comments

arc/job/adapters/obabel.py
```diff
@@ -0,0 +1,270 @@
...
+from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+import subprocess
+
+from arc.common import ARC_PATH, get_logger, save_yaml_file, read_yaml_file
+from arc.imports import settings
+from arc.job.adapter import JobAdapter
+from arc.job.adapters.common import _initialize_adapter
+from arc.job.factory import register_job_adapter
+from arc.level import Level
+from arc.settings.settings import ob_default_settigns
```
Laxzal, line 17: This needs to be fixed from `ob_default_settigns` to `ob_default_settings`

arc/job/adapters/scripts/ob_script.py
Laxzal: I believe it will be a string due to how Kfir has set the function `parse_command_line_arguments()`

arc/job/adapters/scripts/ob_script.py
Laxzal: Small point but I think we should also set the parameters in the function to be the type as described in the function doc string.

Laxzal (changes requested):

arc/job/adapters/obabel.py
kfir4444: with API maybe?

arc/job/adapters/scripts/ob_script.py
kfir4444: This comes from here:
https://github.com/ReactionMechanismGenerator/ARC/blob/2a980e511f9b422fe867865c617809727be0cdd4/arc/job/adapter.py#L1317
I try to avoid using type hinting in the scripts because they sometime requires additional imports, and are not as important in this case.

devtools/install_ob.sh
kfir4444: No, it's distributed with ob probably

arc/job/adapters/scripts/ob_script.py
alongd: I recommend to consolidate the functions into a single call. Then if constraint opt is requested, call another function to initialize the params, then continue either way with a similar workflow

devtools/install_ob.sh
alongd: Could you check?

alongd (commented):
Looking good, added two minor comments

Laxzal (approved):
Looks good to me

PR #602, opened 2023-02-23
Solve problems in arc's directed scans.

kfir4444:
A few functions returned a `np.float64`, which caused errors when dumped into yaml files, and several other small corrections.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -996,10 +996,11 @@ class ARC(object):
                     default_flag = ' (default)'
                 else:
                     default_flag = ''
-                self.opt_level = Level(repr=self.opt_level)
                 logger.info(f'Geometry optimization:{default_flag} {self.opt_level}')
             else:
                 logger.warning("Not performing geometry optimization, since it was not requested by the user.")
+            if self.opt_level:
+                self.opt_level = Level(repr=self.opt_level)
 
             if self.job_types['freq']:
                 if not self.freq_level:
@@ -1011,10 +1012,11 @@ class ARC(object):
                         info = ' (default)'
                 else:
                     info = ''
-                self.freq_level = Level(repr=self.freq_level)
                 logger.info(f'Frequencies:{info} {self.freq_level}')
             else:
                 logger.warning("Not performing frequency calculation, since it was not requested by the user.")
+            if self.freq_level:
+                self.freq_level = Level(repr=self.freq_level)
 
             if self.job_types['sp']:
                 if not self.sp_level:
@@ -1022,10 +1024,11 @@ class ARC(object):
                     default_flag = ' (default)'
                 else:
                     default_flag = ''
-                self.sp_level = Level(repr=self.sp_level)
                 logger.info(f'Energy:{default_flag} {self.sp_level}')
             else:
                 logger.warning("Not performing single point energy calculation, since it was not requested by the user.")
+            if self.sp_level:
+                self.sp_level = Level(repr=self.sp_level)
 
             if self.job_types['rotors']:
                 if not self.scan_level:
@@ -1037,13 +1040,14 @@ class ARC(object):
                         info = ' (default)'
                 else:
                     info = ''
-                self.scan_level = Level(repr=self.scan_level)
                 logger.info(f'Rotor scans:{info} {self.scan_level}')
             else:
                 logger.warning("Not performing rotor scans, since it was not requested by the user. This might "
                                "compromise finding the best conformer, as dihedral angles won't be corrected. "
                                "Also, the calculated thermodynamic properties and rate coefficients "
                                "will be less accurate.")
+            if self.scan_level:
+                self.scan_level = Level(repr=self.scan_level)
 
             if self.job_types['irc']:
                 if not self.irc_level:
@@ -1051,10 +1055,11 @@ class ARC(object):
                     default_flag = ' (default)'
                 else:
                     default_flag = ''
-                self.irc_level = Level(repr=self.irc_level)
                 logger.info(f'IRC:{default_flag} {self.irc_level}')
             else:
                 logger.warning("Not running IRC computations, since it was not requested by the user.")
+            if self.irc_level:
+                self.irc_level = Level(repr=self.irc_level)
 
             if self.job_types['orbitals']:
                 if not self.orbitals_level:
@@ -1062,10 +1067,11 @@ class ARC(object):
                     default_flag = ' (default)'
                 else:
                     default_flag = ''
-                self.orbitals_level = Level(repr=self.orbitals_level)
                 logger.info(f'Orbitals:{default_flag} {self.orbitals_level}')
             else:
                 logger.debug("Not running molecular orbitals visualization, since it was not requested by the user.")
+            if self.orbitals_level:
+                self.orbitals_level = Level(repr=self.orbitals_level)
 
         if not self.job_types['fine'] and (self.opt_level is not None and self.opt_level.method_type == 'dft'
                                            or self.composite_method is not None):
--- a/arc/parser.py
+++ b/arc/parser.py
@@ -233,6 +233,8 @@ def parse_geometry(path: str) -> Optional[Dict[str, tuple]]:
         content = read_yaml_file(path)
         if isinstance(content, dict) and 'xyz' in content.keys():
             return content['xyz']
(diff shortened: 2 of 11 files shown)


arc/species/converter_test.py
```diff
@@ -4673,6 +4673,26 @@ def test_modify_coords(self):
                                           fragments=fragments,
                                           )
         self.assertAlmostEqual(calculate_dihedral_angle(coords=new_xyz, torsion=indices, index=1), 200, places=3)
+        coords={'coords': ((-0.7862825353221515, -0.28824023055636216, 0.4782944637692894),
```
alongd, line 4676: If this is a new “sub test”, add an empty line bofore it for readability

arc/species/zmat.py
```diff
@@ -1113,7 +1113,7 @@ def _add_nth_atom_to_coords(zmat: dict,
                       cd_length * math.sin(bcd_angle) * math.sin(abcd_dihedral)])
         d = m.dot(d)  # Rotate the coordinate system into the reference frame of orientation defined by A, B, C.
         # Add the coordinates of atom C to the resulting atom D:
-        coords.append((d[0] + coords[c_index][0], d[1] + coords[c_index][1], d[2] + coords[c_index][2]))
+        coords.append((float(d[0] + coords[c_index][0]), float(d[1] + coords[c_index][1]), float(d[2] + coords[c_index][2])))
```
alongd, line 1116: Please fix the commit message to something that reflects the modification

arc/parser.py
```diff
@@ -233,6 +233,8 @@ def parse_geometry(path: str) -> Optional[Dict[str, tuple]]:
         content = read_yaml_file(path)
         if isinstance(content, dict) and 'xyz' in content.keys():
             return content['xyz']
+        if isinstance(content, dict) and 'opt_xyz' in content.keys():
```
alongd, line 236: Perhaps change the tani script to store the xyz output in an xyz key instead of the opt_xyz key. We wouldn’t want additional future adapters to add their own custom keys for the geometry (unless there’s good reason to) and have to accommodate for them in this (and other?) functions

arc/plotter.py
```diff
@@ -1145,7 +1145,15 @@ def plot_1d_rotor_scan(angles: Optional[Union[list, tuple, np.array]] = None,
     plt.xticks(np.arange(min_angle, min_angle + 361, step=60))
     plt.ylabel('Electronic energy (kJ/mol)')
     if label is not None:
-        original_dihedral_str = f' from {original_dihedral:.1f}$^o$' if original_dihedral is not None else ''
+        if isinstance(original_dihedral, str):
```
alongd, line 1148: Please implement this here:

https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/common.py#L1033

arc/species/converter.py
```diff
@@ -1901,7 +1901,7 @@ def get_center_of_mass(xyz):
     cm_x /= sum(masses)
     cm_y /= sum(masses)
     cm_z /= sum(masses)
-    return cm_x, cm_y, cm_z
+    return float(cm_x), float(cm_y), float(cm_z)
```
alongd, line 1904: Can you add a test?

alongd (commented):
Looking good!
Please see a few comments 
Also, it would significantly strengthen this PR if you can add a directed scan test using tani

arc/parser.py
kfir4444: There is an important distinction between `xyz` and `opt_xyz`, how would we differentiate them without using different keys?
I don't think it's just the torchani adapter that relates differently to these properties..

kfir4444:
> Looking good! Please see a few comments Also, it would significantly strengthen this PR if you can add a directed scan test using tani

Thanks for the commants @alongd!
I was planning to add a functional scans test in the future, for now this PR sets the ground for additional work that I need to do in order to get the directed scans in operational order, especially with the new openbabel and torchani adapters :smiley:
I will add an issue on the functional scans test and mention this PR.

arc/parser.py
alongd: @kfir4444, let me know when this is resolved, and let's merge this in

kfir4444: @alongd I will probably add more commits to this PR before

arc/species/converter_test.py
```diff
@@ -4008,6 +4008,10 @@ def test_get_center_of_mass(self):
         self.assertAlmostEqual(cm_x, 0.7201, 3)
         self.assertAlmostEqual(cm_y, -0.4880, 3)
         self.assertAlmostEqual(cm_z, -0.1603, 3)
+        
+        self.assertTrue(isinstance(cm_x, float))
```
alongd, line 4012: `assertIsInstance`?

arc/main.py
```diff
@@ -1067,8 +1067,9 @@ def set_levels_of_theory(self):
             else:
                 logger.debug("Not running molecular orbitals visualization, since it was not requested by the user.")
 
-        if not self.job_types['fine'] and (self.opt_level is not None and self.opt_level.method_type == 'dft'
-                                           or self.composite_method is not None):
+        if not self.job_types['fine'] and (self.opt_level is not None and \
+            (self.opt_level['method_type'] if isinstance(self.opt_level, dict) else self.opt_level.method_type) == 'dft'
```
alongd, line 1071: Why would `opt_level` be a dict? And can we catch it early and turn it into a `Level` instance? It might cause more issues downstream if we let an attribute be of two types

arc/scheduler.py
```diff
@@ -2197,6 +2198,8 @@ def parse_opt_geo(self,
         logger.debug(f'parsing opt geo for {job.job_name}')
         if job.job_status[1]['status'] == 'done':
             opt_xyz = parser.parse_xyz_from_file(path=job.local_path_to_xyz or job.local_path_to_output_file)
+            if isinstance(opt_xyz, str):
+                opt_xyz = str_to_xyz(opt_xyz)
```
alongd, line 2205: The signature of `parse_xyz_from_file` is `Optional[Dict[str, tuple]]`. It should return a dict, not an str. Can you check what's going on? We should used consistent types whenever possible.

arc/scheduler.py
```diff
@@ -1157,7 +1159,8 @@ def run_sp_job(self,
             level (Level): An alternative level of theory to run at. If ``None``, self.sp_level will be used.
         """
         level = level or self.sp_level
-
+        if isinstance(level, dict):
```
alongd, line 1162: This goes against the signature of the method, `level: Optional[Level] = None,`
Can you check why it's inconsistent?

(this should be a different commit, no?)

alongd (commented):
Thanks, I added a few comments.
There are at least three cases where we use inconsistent types. It's an invitation for bugs down the road. Can you check?

arc/scheduler.py
kfir4444: Another edge case, if `level = None and self.sp_level is not None` but `job_type['sp'] = False`, then self.sp_level remains a dictionary

kfir4444: This occurs when you wish to run a rotors job type and had an input of `xyz`, at:
https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/scheduler.py#L421
Another possible solution is to use `scan_level` in:
https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/scheduler.py#L427

arc/main.py
alongd: Thanks for pointing to this. Interesting. I agree we should be robust and address these edge cases. Please see my additional comment above that was not addressed:  "can we catch it early and turn it into a `Level` instance?"

While we should treat these cases, the correct solution is to make each of our attributes of a single type (always? as much as possible?) at the outset of the program. In general, there could be additional silent bugs like these, we should root them.

If there are other cases where we treat attributes as possibly two types, they should get the same treatment.

arc/parser.py
alongd: Was this addressed? Are we leaving it both 'xyz' and 'opt_xyz'?

alongd (commented):
Thanks, I added two comments

alongd (dismissed):
Looking good! Let's wait for the tests and we can merge

Laxzal (approved):
looks good to me

PR #603, opened 2023-03-05
Rename the output file of a recent opt job before post_sp_actions

alongd:
Job output files are automatically renamed to `output.out` in ARC, unless they are YAML files. Recently I had several instances where I ran ops ans sp jobs at the same DFT level, making ARC to practically skip the redundant sp job, extracting the electronic energy straight from the converged opt job. The opt jobs indeed converged, but their output file wasn't renamed, causing ARC to crash since it cannot find the expected file.

To be on the safe side, I'm adding a call to the `rename_output_file()` Job method before calling `post_sp_actions`.


kfir4444 (approved):
Great job @alongd !

PR #605, opened 2023-03-05
Call submit_job() recursively if the max num of job limit was reached

alongd:
If the max number of jobs submitted to a server (either via local or ssh) was reached, sleep for 5 minutes and try again recursively. Currently implemented for Slurm.


kfir4444 (approved):
Thanks @alongd!

PR #607, opened 2023-03-06
Make parse_1d_scan_coords() more robust

alongd:
The `parse_1d_scan_coords()` crashed for a specific TS scan output file for two independent reasons.
Both bugs were fixed now, and a test was added.


kfir4444 (changes requested):
Thanks @alongd,
I this the added test file is too large (>17 MB), could you replace it with a smaller output file, or remove unnecessary lines?

alongd:
Thanks @kfir4444, the file size was reduced

kfir4444 (approved):
Thanks @alongd! approved

PR #609, opened 2023-03-12
Fix Reaction.get_element_mass()

alongd:
The `get_element_mass` method in Reaction iterates through the elements of the reactants and returns the mass of the respective elements. However, it doesn't take into account that when a reaction has two identical reactant (e.g., `OH + OH <=> H2O2`), ARC only stores one instance of the reactant species to avoid running duplicate identical QM jobs.
This method was fixed and it uses now the `get_reactants_and_products` method that considers identical reactants/products.
A test was added.


kfir4444 (approved):
Great find and fix @alongd !

PR #612, opened 2023-03-16
Only add a species to the thermo library if requested

alongd:
Here we add a new attribute to ARCSpecies, `include_in_thermo_lib`
We already have an attribute called `compute_thermo`, which will cause ARC not to process the species via statmech. The new attribute will make ARC compute thermo data, but it won't be included in the final RMG thermo library output.
The rational is to facilitate a new T3 feature where species that participate in reactions we compute will not end up in the thermo library if their thermo data is already known to a satisfactory precision. However, for post-analysis purposes, it could still be beneficial to have the thermo data already process by ARC/Arkane.


kfir4444 (approved):
Thanks @alongd!

PR #613, opened 2023-03-16
Modify tani env

kfir4444:


alongd (commented):
Thanks! Can you explain the rationale?

kfir4444:
> Thanks! Can you explain the rationale?

Sure @alongd!
The old tani_env was indeterministic in package version. When specifying:
`  - pytorch::pytorch`
without a version, most likely mamba/conda takes a version that "works" with all package requirements and is the latest.
The issue I solve here stems from the fact that pytorch deprecated a function: `torch.symeig` in favor of `torch.linalg.eigh`. However, TorchANI uses `torch.symeig`, so we no longer have the privilege of not specifying which exact version of pytorch. Yet, when specifying an exact pytorch version, there are conflicts with other packages, that seem to last a small infinity.
So, to solve this issue, @Laxzal and I used an existing `tani_env` with the latest pytorch version that allowed usage of torchani, and froze it with all the specific packages, so we always know exactly which packages to install for torchani, and we know that it always works.

alongd (approved):
Thanks!
Let's keep an eye for when tani updates and uses `torch.linalg.eigh`

PR #614, opened 2023-03-19
Corrected Ketoenol family label in job adapters common

alongd:


alongd:
We could/should use lowercase in the future in ARC to avoid issues like this. Currently ARC takes the family name literally and searches RMG for it

PR #615, opened 2023-03-22
Capture `ConverterError`s in molecules_from_xyz()

alongd:
The `set_multiplicity()` function can raise a ConverterError which should also be captured instead of crashing ARC


kfir4444:
Let's wait for CI and merge.

Issue #618, opened 2023-03-24
Replacing if level_of_theory is not None with if isinstance(level_of_theory, Level)


Laxzal:
**Describe the bug**
Running an input.yml file produced this error

```python

Traceback (most recent call last):
  File "/Local/ce_dana/Code/ARC//ARC.py", line 69, in <module>
    main()
  File "/Local/ce_dana/Code/ARC//ARC.py", line 65, in main
    arc_object.execute()
  File "/Local/ce_dana/Code/ARC/arc/main.py", line 583, in execute
    fine_only=self.fine_only,
  File "/Local/ce_dana/Code/ARC/arc/scheduler.py", line 484, in __init__
    self.schedule_jobs()
  File "/Local/ce_dana/Code/ARC/arc/scheduler.py", line 537, in schedule_jobs
    self.determine_most_stable_conformer(label)  # also checks isomorphism
  File "/Local/ce_dana/Code/ARC/arc/scheduler.py", line 2001, in determine_most_stable_conformer
    self.troubleshoot_conformer_isomorphism(label=label)
  File "/Local/ce_dana/Code/ARC/arc/scheduler.py", line 3348, in troubleshoot_conformer_isomorphism
    conformer=conformer,
  File "/Local/ce_dana/Code/ARC/arc/scheduler.py", line 770, in run_job
    if level_of_theory is not None and level_of_theory.args is not None:
AttributeError: 'str' object has no attribute 'args
```

**How to reproduce**
I believe it was this reaction in my input.yml

```yml
project: 'arc_ll_hab'

level_of_theory: wb97xd/def2tzvp

compute_thermo: False

species:
 - label: 'r_82_[O-][N+](=O)O'
   smiles: '[O-][N+](=O)O'
 - label: 'r_82_[H]'
   smiles: '[H]'
 - label: 'r_82_[H][H]'
   smiles: '[H][H]'
 - label: 'r_82_[O-][N+](=O)[O]'
   smiles: '[O-][N+](=O)[O]'

reactions:
 - label: 'r_82_[O-][N+](=O)O + r_82_[H] <=> r_82_[H][H] + r_82_[O-][N+](=O)[O]'
```

**Expected behavior**

change in the code:

`if level theory is not None` to `if isinstance(level_of_theory, Level)`

Issue #610, opened 2023-03-12
ARC accept multiple types for `level`.


kfir4444:
**Describe the bug**
ARC object accepts two input types of `level`, which causes issues later.

**Possible solution**
I suggest writing a function in `main.py`, that reads the `input_dict` and turns all `level` keys that are not None into `Level` objects.

**Additional context**
[PR ](https://github.com/ReactionMechanismGenerator/ARC/pull/602),

kfir4444:
solved with [602](https://github.com/ReactionMechanismGenerator/ARC/pull/602)

PR #619, opened 2023-03-24
Changed if isinstance(level_of_theory, Level)

Laxzal:
Adjusted line 770 to now check if the instance is a Level rather than if level_of_theory is None

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -767,7 +767,7 @@ class Scheduler(object):
             args['shift'] = shift
         if scan_trsh:
             args['keyword']['scan_trsh'] = scan_trsh
-        if level_of_theory is not None and level_of_theory.args is not None:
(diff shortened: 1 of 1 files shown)


kfir4444 (changes requested):
Thanks @Laxzal,
First, could you squash the commits?
In addition, as per your [issue](https://github.com/ReactionMechanismGenerator/ARC/issues/618), let's find out why a string still passes all the way to the scheduler, if we recently closed another [issue](https://github.com/ReactionMechanismGenerator/ARC/issues/610) on that topic, and try to find the source for it.

alongd:
Looks like we took care of all specific levels, but not of `level_of_theory`:
https://github.com/ReactionMechanismGenerator/ARC/pull/602/commits/1e30fbf86ded2b8c03cd40d0a729c88308be3b36

@Laxzal, please squash

alongd (approved):
Thanks!

PR #621, opened 2023-03-24
A few patches to the restart

alongd:
ARC's restart should be remade one day. Here are a few patches to avoid running undesired jobs when restarting ARC


kfir4444 (approved):
Great! @alongd

PR #626, opened 2023-03-26
Added 'is not None' to if conditions

Laxzal:
In the original issue, when a `self.variable =0`, then the `if self.variable:` was not met in the`as_dict()` function found in the script: https://github.com/ReactionMechanismGenerator/ARC/blob/ba4f5b9aa8cb1b3ab8d6135e0d1e25ee10f5cc76/arc/job/adapter.py#L824

### DIFF
--- a/arc/job/adapter.py
+++ b/arc/job/adapter.py
@@ -792,7 +792,7 @@ class JobAdapter(ABC):
         if self.dihedrals is not None:
             job_dict['dihedrals'] = self.dihedrals
         job_dict['ess_settings'] = self.ess_settings
-        if self.ess_trsh_methods:
+        if isinstance(self.ess_trsh_methods, list) and self.ess_trsh_methods:
             job_dict['ess_trsh_methods'] = self.ess_trsh_methods
         job_dict['execution_type'] = self.execution_type
         if self.fine:
@@ -815,15 +815,15 @@ class JobAdapter(ABC):
             job_dict['rotor_index'] = self.rotor_index
         if self.server is not None:
             job_dict['server'] = self.server
-        if self.server_nodes:
+        if isinstance(self.server_nodes, dict) and self.server_nodes:
             job_dict['server_nodes'] = self.server_nodes
         if self.species is not None:
             job_dict['species_labels'] = [species.label for species in self.species]
-        if self.torsions:
+        if self.torsions is not None:
             job_dict['torsions'] = self.torsions
-        if self.tsg:
+        if self.tsg is not None:
             job_dict['tsg'] = self.tsg
-        if self.xyz:
+        if self.xyz is not None:
             job_dict['xyz'] = xyz_to_str(self.xyz)
         return job_dict
 


arc/job/adapter.py
```diff
@@ -792,10 +792,10 @@ def as_dict(self) -> dict:
         if self.dihedrals is not None:
             job_dict['dihedrals'] = self.dihedrals
         job_dict['ess_settings'] = self.ess_settings
-        if self.ess_trsh_methods:
+        if self.ess_trsh_methods  is not None:
```
alongd, line 795: but we also don't want to save just an empty dict/list
maybe `if isinstance(self.ess_trsh_methods, list) and len(self.ess_trsh_methods):`

arc/job/adapter.py
```diff
@@ -792,10 +792,10 @@ def as_dict(self) -> dict:
...
             job_dict['ess_trsh_methods'] = self.ess_trsh_methods
         job_dict['execution_type'] = self.execution_type
-        if self.fine:
+        if self.fine is not None:
```
alongd, line 798: `fine` is never `None`, it's only `True` or `False`

arc/job/adapter.py
```diff
@@ -815,15 +815,15 @@ def as_dict(self) -> dict:
             job_dict['rotor_index'] = self.rotor_index
         if self.server is not None:
             job_dict['server'] = self.server
-        if self.server_nodes:
+        if self.server_nodes is not None:
```
alongd, line 818: suggestion:
`if isinstance(self.server_nodes , dict) and len(self.server_nodes.keys()):`

alongd (commented):
Thanks! Added a few comments

arc/job/adapter.py
Laxzal: How about:
`if isinstance(self.server_nodes, dict) and self.server_nodes:`
Since len(self.server_nodes.keys()) will evaluate to false if empty

arc/job/adapter.py
Laxzal: `if isinstance(self.ess_trsh_methods, list) and self.ess_trsh_methods:`
 Same thought applies, if the list is empty, it would be false

arc/job/adapter.py
alongd: Sure, I think their equal (the former is perhaps more "formal"?) Same for the above comment

Laxzal: I think my change is meant to be more `pythonic` - less words or something. Eitherway, I just wonder if there are situations where `len(self.server_nodes.keys())` are better suited for certain unexpected cases

arc/job/adapter.py
```diff
@@ -792,10 +792,10 @@ def as_dict(self) -> dict:
         if self.dihedrals is not None:
             job_dict['dihedrals'] = self.dihedrals
         job_dict['ess_settings'] = self.ess_settings
-        if self.ess_trsh_methods  is not None:
+        if isinstance(self.ess_trsh_methods, list) and self.ess_trsh_methods:
             job_dict['ess_trsh_methods'] = self.ess_trsh_methods
         job_dict['execution_type'] = self.execution_type
-        if self.fine is not None:
+        if self.fine:
```
alongd, line 798: I'm not sure about this one.
Is there a default value for `self.fine`? If the default is `True` then we must save in the restart dict the `False` case, but this condition won't do that
Perhaps to be safe, always save it (unless it's `None` like in the original code)?

alongd:
I added one comment, please rebase and squash

arc/job/adapter.py
Laxzal: the original code shows me:

```python
        if self.fine:
            job_dict['fine'] = self.fine
```

So just remove the if condition and say
`job_dict['fine'] = self.fine`?

alongd (approved):
Thanks!

PR #628, opened 2023-03-28
Extracted sorting indices and sort the plot by them

kfir4444:
This is to avoid plotting errors, such as:

That after the patch becomes:
image


MichalKesl (commented):
Looks great

MichalKesl (approved):
Looks great

PR #633, opened 2023-04-03
PBS Memory Conversion

Laxzal:
Update the memory conversion for PBS in the adapter.py to accurately request gb of memory instead of mb


kfir4444 (approved):
Terrific, let's wait for the CI and merge

kfir4444 (approved):

PR #634, opened 2023-04-03
Commented out the reset of the job_dict conformers

calvinp0:
The current workaround is to remove the reset of the self.job_dict[label]['conformers']

@alongd was there a specific reason for this line of code originally?


alongd:
I think we assume that if we're in `process_conformers`, then all conformers have finished running. When a specific job finished, we remove it from the `job_dict[label][<job id>]` dict, but conformers have another level in this dict of `job_dict[label][conformers][<job id>]`, so we reset the entire thing.

Does commenting this out solve the reset issues? I'd like to understand it better, but if so, please remove the line instead of commenting it out.

PR #641, opened 2023-04-16
auto tst command fix

calvinp0:
The subprocess command for auto tst would fail for some reactions due… to character issues


kfir4444 (approved):
:+1:

PR #642, opened 2023-04-16
Refactoring H_Abstraction Code

calvinp0:
get an xyz guess if both reactants are the same


alongd:
@calvinp0, can you remove the second "merge" commit?

alongd (approved):

alongd (approved):
Thanks! You can merge after it passes

PR #650, opened 2023-05-02
Make sure that parse_xyz returns a dict xyz

kfir4444:


arc/parser.py
```diff
@@ -231,10 +231,10 @@ def parse_geometry(path: str) -> Optional[Dict[str, tuple]]:
         raise InputError(f'Could not find file {path}')
     if path.endswith('.yml'):
         content = read_yaml_file(path)
-        if isinstance(content, dict) and 'xyz' in content.keys():
-            return content['xyz']
-        if isinstance(content, dict) and 'opt_xyz' in content.keys():
-            return content['opt_xyz']
+        if 'xyz' in content.keys():
```
alongd, line 234: why removing the safety check `if isinstance(content, dict)`? If we're reading a YAML what doesn't hold a dict, this like will crash

arc/parser.py
```diff
@@ -231,10 +231,10 @@ def parse_geometry(path: str) -> Optional[Dict[str, tuple]]:
...
+            return content['xyz'] if isinstance(content['xyz'], dict) else str_to_xyz(content['xyz'])
```
alongd, line 235: 👍

arc/parser.py
```diff
@@ -231,10 +231,10 @@ def parse_geometry(path: str) -> Optional[Dict[str, tuple]]:
         raise InputError(f'Could not find file {path}')
     if path.endswith('.yml'):
         content = read_yaml_file(path)
-        if isinstance(content, dict) and 'xyz' in content.keys():
```
alongd, line 234: Can you please modify the commit message to "Make **sure**...."

alongd (commented):
Thanks! Added minor comments

arc/parser.py
alongd: I think this wasn't addressed

arc/parser.py
alongd: I think this wasn't addressed

arc/parser.py
Lilachn91, line 234: Why did you remove isinstance statement here, and not keep it like below?

alongd (approved):

Lilachn91 (commented):
Good job :)

PR #651, opened 2023-05-07
Fix the molpro input file memory

alongd:
Molpro's input file memory is per cpu core, this PR returns this consideration into the calculation.
The conversion from mW to GB was done using [this website](https://deviceanalytics.com/words-to-bytes-converter/) specifying a 64-bit architecture.

A test was added

I snack in two minor TS-related commits into this small PR:
1. Added a shortcut to specify `xyz` when initializing an `ARCReaction` object (in addition to `ts_xyz_guess` which is less intuitive)
2. Initialize user TS guesses with the `success` flag (relates to the method used to generate the guess and whether it gave any guesses, which is always true for a user guess - it has nothing to do with whether the guess is actually reasonable or not)

### DIFF
--- a/arc/job/adapters/molpro.py
+++ b/arc/job/adapters/molpro.py
@@ -319,13 +319,8 @@ ${self.species[0].occ}wf,spin=${input_dict['spin']},charge=${input_dict['charge'
         """
         Set the input_file_memory attribute.
         """
-        # Molpro's memory is per cpu core and in MW (mega word; 1 MW ~= 8 MB; 1 GB = 128 MW)
-        # https://www.molpro.net/pipermail/molpro-user/2010-April/003723.html
-        # In the link, they describe the conversion of 100,000,000 Words (100Mword) is equivalent to
-        # 800,000,000 bytes (800 mb). 
-        # Formula - (100,000,000 [Words]/( 800,000,000 [Bytes] / (job mem in gb * 1000,000,000 [Bytes])))/ 1000,000 [Words -> MegaWords] 
-        # The division by 1E6 is for converting into MWords
-        self.input_file_memory = math.ceil((1E8/(8E8 /(self.job_memory_gb * 1E9)))/1E6)
+        # Molpro's memory is per cpu core and in MW (mega word; 1000 MW = 7.45 GB on a 64-bit machine)
+        self.input_file_memory = math.ceil(self.job_memory_gb / (7.45e-3 * self.cpu_cores))
         
     def execute_incore(self):
         """
--- a/arc/job/adapters/molpro.py
+++ b/arc/job/adapters/molpro.py
@@ -320,6 +320,15 @@ ${self.species[0].occ}wf,spin=${input_dict['spin']},charge=${input_dict['charge'
         Set the input_file_memory attribute.
         """
         # Molpro's memory is per cpu core and in MW (mega word; 1000 MW = 7.45 GB on a 64-bit machine)
+        # The conversion from mW to GB was done using this (https://deviceanalytics.com/words-to-bytes-converter/)
+        # specifying a 64-bit architecture.
+        #
+        # See also:
+        # https://www.molpro.net/pipermail/molpro-user/2010-April/003723.html
+        # In the link, they describe the conversion of 100,000,000 Words (100Mword) is equivalent to
+        # 800,000,000 bytes (800 mb).
+        # Formula - (100,000,000 [Words]/( 800,000,000 [Bytes] / (job mem in gb * 1000,000,000 [Bytes])))/ 1000,000 [Words -> MegaWords]
+        # The division by 1E6 is for converting into MWords
         self.input_file_memory = math.ceil(self.job_memory_gb / (7.45e-3 * self.cpu_cores))
         
     def execute_incore(self):
--- a/arc/job/adapters/molpro.py
+++ b/arc/job/adapters/molpro.py
@@ -8,6 +8,7 @@ import datetime
 import math
 import os
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+import socket
 
 from mako.template import Template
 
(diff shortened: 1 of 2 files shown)


arc/job/adapters/molpro.py
```diff
@@ -319,13 +319,8 @@ def set_input_file_memory(self) -> None:
...
         Set the input_file_memory attribute.
         """
-        # Molpro's memory is per cpu core and in MW (mega word; 1 MW ~= 8 MB; 1 GB = 128 MW)
-        # https://www.molpro.net/pipermail/molpro-user/2010-April/003723.html
-        # In the link, they describe the conversion of 100,000,000 Words (100Mword) is equivalent to
-        # 800,000,000 bytes (800 mb). 
-        # Formula - (100,000,000 [Words]/( 800,000,000 [Bytes] / (job mem in gb * 1000,000,000 [Bytes])))/ 1000,000 [Words -> MegaWords] 
-        # The division by 1E6 is for converting into MWords
-        self.input_file_memory = math.ceil((1E8/(8E8 /(self.job_memory_gb * 1E9)))/1E6)
+        # Molpro's memory is per cpu core and in MW (mega word; 1000 MW = 7.45 GB on a 64-bit machine)
```
calvinp0, line 322: I still think the formula is handy to have, so that whoever reads this part of the code will have an understanding how MegaWords is formulated. Or at the very least, the link to Molpro where they discuss the formula they specifcally use

arc/job/adapters/molpro.py
```diff
@@ -319,13 +319,8 @@ def set_input_file_memory(self) -> None:
...
+        self.input_file_memory = math.ceil(self.job_memory_gb / (7.45e-3 * self.cpu_cores))
```
calvinp0, line 323: I think we need to test this assertion of the CPU cores being recognised by Molpro. We will have to look into the file generated from Molpro when it uses the input file, or at the very least, check our input file that we specify the cpu cores available to the software. I say this because the one test we use, where we assert the memory is 40 (changed from 1750) may be problematic from previous runs I have done.

arc/job/adapters/molpro.py
alongd: I'll keep it. I originally removed it since it's slightly inaccurate, the old comment said `1 MW ~= 8 MB` and the formula that followed strengthened it, but the new formula says `1000 MW = 7.45 GB`
I thought it would be confusing to keep both

arc/job/adapters/molpro.py
alongd: Here are input and output files from an ARC spawned Molpro job using this branch:
[input.txt](https://github.com/ReactionMechanismGenerator/ARC/files/11472154/input.txt)
[output.txt](https://github.com/ReactionMechanismGenerator/ARC/files/11472155/output.txt)

My experience is that jobs tend to crash on the current main branch

calvinp0: `Both the -m and -G options are by default given in megawords (m) but unit gigaword (g) can also be used (e.g. -m1000 is equivalent to -m1000m and to -m1g). The total memory M per node allocated by molpro amounts to (n⋅m+G)/N, where n is the total number of processes (-n option), m is the stack memory per process (-m option), G the GA memory (-G option), and N the number of nodes. In addition, at least 200 MW per process should be added for the program itself. In total, a calculation needs about 8⋅[n⋅(m+0.3)+G]/N GB (gigabytes) of memory (n,m,G in gw), and this should not exceed the physical memory of the machine(s) used. We do note that in many calculations leaving some memory for the system to buffer I/O operations may improve the performance significantly. `

I think we have to set `-n` here as the number of process?

calvinp0: I will get back to you on this when I re-run an old job that had an issue with the molpro original memory prior to my change

alongd: You mean besides [this](https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/settings/submit.py#L710) `-n`?
We should also actually do `-n <cpus> -t 1` [see here](https://github.com/DanaResearchGroup/DRGScripts/blob/main/Servers/Atlas/.arc/submit.py#L100), which could also be a cause for why we might get different behaviors

calvinp0: Yes, it may have got lost with everyone probably having slightly different submit.py files for their environments in Zeus. I also want to double check if molpro will register `-n` if we provide it memory in the input file because I know that even if we change the memory in the command line, it will ignore it if we still provide it memory in the input file

calvinp0: Hey @alongd 

So, I think I know why the current MolPro set up on ARC doesn't run smoothly for you. The reason being is that for some reason, when on Atlas, MolPro correctly identifies the number of NProcs.

This is in your output file for it:

```
 Nodes     nprocs
 tech-wn179    9
```

So I guess it was originally doing 4000 * 9


Whereas, when I try it on Zeus, and attempt to do `-n 16` for the command line, the Nprocs still show as 1

```
 Nodes     nprocs
 n024         1
```

So I am not sure if it is a configuration difference between the 2 servers

calvinp0:
So until I get real confirmation, I have a theory on the molpro issue and @kfir4444 came to the same conclusion

With Atlas, when we specify -n {cpu_cores}, in the submit file, Molpro actually recognises it and takes it into consideration - thus the total MW we define are multiplied by the number of cores

However, Zeus appears to be different when submitting through PBS...Even if we define -n 16 in the submit file with the command line for Zeus, it will still only be 1 node
```
Nodes     nprocs
 n027         1
```
Thus for 4000MW  on Zeus it is 4000 MW total whereas on Atlas on it would be 64000MW total due it being 4000 * 16

HOWEVER, if I am on Zeus terminal and do the molpro -n 16 [input.in](http://input.in/) it will show this
``` 
Nodes              nprocs
 zeus   16

So until I can figure out how to define the number of cpus nodes  for zeus, we will have to update the PR to check which server we are using if Zeus, dont divide by cpu cors, and if Atlas, divide by cpu cores

calvinp0:
I have submitted a ticket to Zeus about this issue we are experiencing

alongd:
@calvinp0, shall I review, or are we waiting on the Zeus support team?

calvinp0:
> @calvinp0, shall I review, or are we waiting on the Zeus support team?

I'd say review. The Zeus team have yet to respond to our other issues regarding the submission of jobs. So it may take some time for them to get back to us

calvinp0 (approved):

alongd:
Looks overall good. We have two commits to squash (a fixup and the last two), please squash and rebase

calvinp0 (approved):

PR #652, opened 2023-05-09
Fix env

kfir4444:
This PR contains two features:

1. An update to the ARC environment, specifically numpy and cantera version update, and removal of nosetests, which is deprecated
2. replacing the `@wip` decorator with a one that is not dependent on nosetests.

### DIFF
--- a/arc/settings/settings.py
+++ b/arc/settings/settings.py
@@ -335,13 +335,9 @@ for autotst_pypath in [autotst_pypath_1, autotst_pypath_2, autotst_pypath_3, aut
 paths = list()
 paths.append(os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(sys.executable))),
                             'xtb_env', 'bin', 'xtb'))
-paths.append(os.path.join(home, 'anaconda3', 'envs', 'arc_env', 'bin', 'xtb'))
 paths.append(os.path.join(home, 'anaconda3', 'envs', 'xtb_env', 'bin', 'xtb'))
-paths.append(os.path.join(home, 'miniconda3', 'envs', 'arc_env', 'bin', 'xtb'))
 paths.append(os.path.join(home, 'miniconda3', 'envs', 'xtb_env', 'bin', 'xtb'))
-paths.append(os.path.join(home, '.conda', 'envs', 'arc_env', 'bin', 'xtb'))
 paths.append(os.path.join(home, '.conda', 'envs', 'xtb_env', 'bin', 'xtb'))
-paths.append(os.path.join('/Local/ce_dana', 'anaconda3', 'envs', 'arc_env', 'bin', 'xtb'))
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -3,10 +3,11 @@ channels:
   - defaults
   - rmg
   - conda-forge
+  - cantera
 dependencies:
   - cairo
   - cairocffi
-  - rmg::cantera >=2.3.0
+  - cantera::cantera=2.6
```
alongd, line 10: maybe `>=`?

kfir4444:
> Do you think we should bump numpy more? See #517
Probably not possible for now, we would need to update python:
  ├─ numpy 1.22.0**  is installable with the potential options
  │  ├─ numpy 1.22.0 would require
  │  │  └─ python >=3.9,<3.10.0a0 , which can be installed;
  │  ├─ numpy 1.22.0 would require
  │  │  └─ python >=3.10,<3.11.0a0 , which can be installed;
  │  └─ numpy 1.22.0 would require
  │     └─ python >=3.8,<3.9.0a0 , which can be installed;
  └─ python 3.7**  is uninstallable because there are no viable options
     ├─ python [3.7.0|3.7.1|...|3.7.9] conflicts with any installable versions previously reported;
     └─ python 3.7.1 would require
        └─ openssl >=1.1.1,<1.1.2.0a0 , which does not exist (perhaps a missing channel).

arc/settings/settings.py
```diff
@@ -334,7 +334,7 @@
 
 paths = list()
 paths.append(os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(sys.executable))),
-                            'arc_env', 'bin', 'xtb'))
+                            'xtb_env', 'bin', 'xtb'))
 paths.append(os.path.join(home, 'anaconda3', 'envs', 'arc_env', 'bin', 'xtb'))
```
calvinp0, line 338: With the removal of xtb from arc_env, I think we need to remove the path appends that include the arc_env

calvinp0 (changes requested):
Everything looks good to me with the exception of the minor pedantic removal of certain paths when looking for xtb

arc/settings/settings.py
```diff
@@ -349,7 +349,7 @@
         break
 
 arc_pypath_1 = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(sys.executable))),
-                            'arc_env', 'bin', 'python')
+                            'xtb_env', 'bin', 'python')
```
alongd, line 352: why was this  changed ?

calvinp0: That was my mistake. It was during the xTB issue. Changing it back

alongd (dismissed):
Looking good!
I don't support locking dependable versions using a `=`, I do think we should allow `>=`
Other than that it's good to go

alongd (approved):

alongd:
Are we clear to merge?

calvinp0 (approved):

PR #656, opened 2023-05-14
Added support for additional methods

alongd:
Matching additional methods with our supported ESS


kfir4444 (commented):
Thanks @alongd,
I think you are missing an empty line at the end of the file, is that intentional?

alongd:
Yes, there were two empty lines at the end, I removed one (there's still one empty line left)

kfir4444 (approved):
Thanks! @alongd

PR #662, opened 2023-05-23
Changing MOPAC channel

calvinp0:
RMG team has removed mopac from the rmg channel, thus we need to change the channel of our mopac installation from rmg to conda-forge


kfir4444 (approved):
Nice catch!

PR #663, opened 2023-05-25
Read a mol repr from an Arkane YAML file if given

alongd:
Save a mol representation in an Arkane YAML file under a `mol` key
and read the mol object properly if given.

A twin PR of https://github.com/ReactionMechanismGenerator/RMG-Py/pull/2447


kfir4444 (approved):
LGTM

PR #664, opened 2023-05-26
Fixes TS E0 checks

alongd:
This PR primarily improves ARC E0 checks for a reaction.
Sometimes E0 values are known (e.g., from Arkane YAML files) and we don't have the e_elect values.
ARC should use the E0 values instead when comparing relative reactants - TS - products well heights

Other minor additions in this PR:
- Initialize TSGuess object derived from user guesses with `success=True` (important for facilitating user TS guesses)
- Added a shortcut argument `xyz` to `ARCReaction` object (more intuitive in input files)
- Save the species `adjlist` if given, so that ARC/RMG won't regenerate it for libraries, and perhaps get it wrong from xyz perception


kfir4444 (approved):
Thanks @alongd! LGTM.

alongd:
Should be merged after https://github.com/ReactionMechanismGenerator/ARC/pull/663

alongd:
@kfir4444, I think this PR is good to go, please take a look

kfir4444 (approved):
LGTM

PR #666, opened 2023-06-01
Added solvation option to Orca job adapter

alongd:
Added SMD and PCM solvation types
Note that the keywords here are for Orca 4 and not Orca 5, this should still be fixed

Tests added

### DIFF
--- a/arc/job/adapters/orca.py
+++ b/arc/job/adapters/orca.py
@@ -1,5 +1,5 @@
 """
-An adapter for executing Orca jobs
+An adapter for executing Orca 4 jobs
 
 https://orcaforum.kofo.mpg.de/app.php/portal
 """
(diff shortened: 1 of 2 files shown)


Lilachn91:
@alongd I fixed the tests, should be fine. Note it's a temporary fix until I figure out parallelization in Atlas. 
After I'll edit the adapter to work also with solvation on Orca 5 and test it, we can merge.

Thanks :)

arc/job/adapters/orca.py
```diff
@@ -316,7 +316,7 @@ def write_input_file(self) -> None:
             if self.level.solvation_method.lower() == 'smd':
                 self.add_to_args(val=f"""
 %cpcm SMD true
-      SMDsolvent "{self.level.solvent}"
```
alongd: I'd like this commit to be split with the orca.py changes eventually squashed into the original orca.py commit in this PR, and the orca_test.py changes becoming a new commit with a clearer commit message

alongd (commented):
Thanks! I added one comment, I guess we're waiting to figure out the cpu/memory handling?

PR #667, opened 2023-06-03
Adjusted CI step to clean as much of the ubunutu image as possible

calvinp0:
Currently our CI fails due to not enough space on the runner image. The adjustment of the Clean Ubuntu Image step cleans up aproximately 30GB of unrequired files


kfir4444 (approved):
lgtm

PR #670, opened 2023-06-06
Gaussian Troubleshoot Combinations Improvement

calvinp0:
The current version of ARC cannot fully troubleshoot Gaussian due to lack of combination troubleshooting. This PR attempts to rectify this issue.

For example, if a job receives a first error of L401/301, and so we remove the chkfile by changing `guess=read` to `guess=mix` but then it its next failed run it gives another error where we need to add `scf=(nosymm)`, then this modification will create an `input.gjf` file that uses BOTH troubleshooting methods.

### DIFF
--- a/arc/job/adapters/common.py
+++ b/arc/job/adapters/common.py
@@ -7,6 +7,7 @@ import datetime
 import os
 import shutil
 import sys
+import re
 
 from pprint import pformat
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
@@ -220,8 +221,13 @@ def _initialize_adapter(obj: 'JobAdapter',
     if obj.execution_type != 'incore' and obj.job_adapter in obj.ess_settings.keys():
         obj.determine_job_array_parameters()
 
-    obj.scan_res = obj.args['trsh']['scan_res'] \
-        if obj.args and 'trsh' in obj.args.keys() and 'scan_res' in obj.args['trsh'].keys() else rotor_scan_resolution
+    # Set scan_res if required by trsh
+    if obj.args and 'trsh' in obj.args.keys() and 'scan_res' in obj.args['trsh'].keys():
+        obj.scan_res = obj.args['trsh']['scan_res']
+        # Remove it from the args dict
+        obj.args['trsh'].pop('scan_res')
+    else:
+        obj.scan_res = rotor_scan_resolution
 
     obj.set_files()
     check_argument_consistency(obj)
@@ -280,29 +286,130 @@ def update_input_dict_with_args(args: dict,
     """
     Update the job input_dict attribute with keywords and blocks from the args attribute.
 
+    The function iterates over each key in the 'args' dictionary. Depending on the key type 
+    ('block', 'keyword', 'trsh'), it updates the 'input_dict' accordingly. For 'block' and 
+    'keyword' types, it appends the corresponding values to the 'input_dict'. For 'trsh', 
+    it handles both list and non-list values.
+    
     Args:
-        args (dict): The job arguments.
-        input_dict (dict): The job input dict.
+        args (dict): The job arguments with the following structure:
+            args = {
+                'block': {key1: value1, key2: value2, ...},
+                'keyword': {key1: value1, key2: value2, ...},
+                'trsh': {key1: value1, key2: value2, ... or key: [value1, value2, ...]},
+                ...
+            }
+
+        input_dict (dict): The job input dict with a structure that may be updated:
+            input_dict = {
+                'block': <block string>,
+                'keywords': <keywords string>,
+                'scan_trsh': <scan_trsh string>,
+                'trsh': <trsh string>,
+                ...
+            }
 
     Returns:
-        dict: The updated input_dict.
+        dict: The updated input_dict with appended or modified values based on the args.
+
+    Example:
+        Input args structure:
+            args = {
+                'block': {'1': 'block text 1', '2': 'block text 2'},
+                'keyword': {'opt': 'keyword opt', 'freq': 'keyword freq'},
+                'trsh': ['trsh value 1', 'trsh value 2']
+            }
+
+        Input input_dict structure:
+            input_dict = {
+                'block': 'existing block text',
+                'keywords': 'existing keywords',
+                'scan_trsh': 'existing scan_trsh',
+                'trsh': 'existing trsh'
+            }
+
+        Output input_dict structure:
+            input_dict = {
+                'block': 'existing block text\n\nblock text 1\nblock text 2\n',
+                'keywords': 'existing keywords keyword opt keyword freq ',
+                'scan_trsh': 'existing scan_trsh',
+                'trsh': 'existing trsh trsh value 1 trsh value 2 '
+            }
+
     """
     for arg_type, arg_dict in args.items():
         if arg_type == 'block' and arg_dict:
             input_dict['block'] = '\n\n' if not input_dict['block'] else input_dict['block']
             for block in arg_dict.values():
-                input_dict['block'] += f'{block}\n'
+                # Chek if input_dict['block'] already contains a value and that it ends with a newline
+                if input_dict['block'] and not input_dict['block'].endswith('\n'):
+                    input_dict['block'] += '\n'
+                input_dict['block'] += f'{block}'
         elif arg_type == 'keyword' and arg_dict:
             for key, value in arg_dict.items():
                 if key == 'scan_trsh':
                     if 'scan_trsh' not in input_dict.keys():
                         input_dict['scan_trsh'] = ''
-                    input_dict['scan_trsh'] += f'{value} '
+                    # Check if input_dict['scan_trsh'] already contains a value
+                    if input_dict['scan_trsh']:
+                        input_dict['scan_trsh'] += f' {value}'
+                    else:
+                        input_dict['scan_trsh'] += f'{value}'
                 else:
                     if 'keywords' not in input_dict.keys():
                         input_dict['keywords'] = ''
-                    input_dict['keywords'] += f'{value} '
+                # Check if input_dict['keywords'] already contains a value
+                    if input_dict['keywords']:
+                        input_dict['keywords'] += f' {value}'
+                    else:
+                        input_dict['keywords'] += f'{value}'
+        elif arg_type == 'trsh':
+            if isinstance(arg_dict, list):
+                # arg_dict is a list, iterate through its elements
+                for val in arg_dict:
+                    if 'trsh' not in input_dict.keys():
+                        input_dict['trsh'] = ''
+                    # Check if input_dict['trsh'] already contains a value
+                    if input_dict['trsh']:
+                        input_dict['trsh'] += f' {val}'
+                    else:
+                        input_dict['trsh'] += f'{val}'
+            else:
+                # arg_dict is a dictionary, proceed as before
+                for key, value in arg_dict.items():
+                    # Check if value is a list
+                    if isinstance(value, list):
+                        for val in value:
+                            if key not in input_dict.keys():
+                                input_dict[key] = ''
+                            if input_dict[key]:
+                                input_dict[key] += f' {val}'
+                            else:
+                                input_dict[key] += f'{val}'
+                    else:
+                        if key not in input_dict.keys():
+                            input_dict[key] = ''
+                        if input_dict[key]:
+                            input_dict[key] += f' {value}'
+                        else:
+                            input_dict[key] += f'{value}'
+
     return input_dict
+
+def input_dict_strip(input_dict: dict) -> dict:
+    """
+    Strip all values in the input dict of leading and trailing whitespace.
+    """
+    stripped_dict = {
+        key: (
+            val.rstrip() if isinstance(val, str) and val.startswith('\n') and not val.endswith('\n') else
+            val.lstrip() if isinstance(val, str) and not val.startswith('\n') and val.endswith('\n') else
+            val.strip() if isinstance(val, str) and not val.startswith('\n') and not val.endswith('\n') else val
+        )
+        for key, val in input_dict.items() if val is not None
+    }
+
+    return stripped_dict
 
 
 def set_job_args(args: Optional[dict],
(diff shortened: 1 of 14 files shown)


kfir4444 (commented):
Thanks @calvinp0!
a quick comment, other than that it looks great!

arc/job/trsh.py
```diff
@@ -851,48 +851,64 @@ def trsh_ess_job(label: str,
...
             trsh_keyword = 'scf=(NDump=30)'
+            if 'CheckFile' in job_status['keywords'] or 'checkfile=None' in ess_trsh_methods:
+                remove_checkfile = True
         elif 'SCF' in job_status['keywords'] and 'scf=NoDIIS' not in ess_trsh_methods:
             # Switching off Pulay's Direct Inversion
             logger.info(f'Troubleshooting {job_type} job in {software} for {label} using scf=NoDIIS')
             ess_trsh_methods.append('scf=NoDIIS')
             trsh_keyword = 'scf=NoDIIS'
+            if 'CheckFile' in job_status['keywords'] or 'checkfile=None' in ess_trsh_methods:
+                remove_checkfile = True
```
alongd, line 931: I wonder if we can (A) put these lines somewhere where they'll be always called but just written once, and (B) generalize them for any previous trsh keyword and not just checkfile

Another thought: what about creating a small class of `Trsh` that stores the used trsh keywords and specifies which keyword should be kept until explicitly removed, and which should only be tried/used once. This class could be an attribute of a `Species`, and passed around to the Job. It'll will create a "memory" of trsh methods used. trsh here should manipulate this object (add/remove keywords). I think that ultimately this is more sustainable and the implementation is keyword/ess independent.

arc/species/species.py
```diff
@@ -575,9 +575,8 @@ def number_of_heavy_atoms(self) -> int:
                 self._number_of_heavy_atoms = len([symbol for symbol in self.get_xyz()['symbols'] if symbol != 'H'])
             elif self.is_ts:
                 for ts_guess in self.ts_guesses:
-                    if ts_guess.xyz is not None:
-                        self._number_of_heavy_atoms = \
-                            len([line for line in ts_guess.xyz.splitlines() if line.split()[0] != 'H'])
+                    if ts_guess.get_xyz() is not None:
```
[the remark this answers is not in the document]
alongd: This is the `get_xyz()` method of `TSGuess`, not of `ARCSpecies`, so it doesn't have a `generate` arg

arc/job/adapters/gaussian.py
```diff
@@ -48,7 +48,7 @@
 %%mem=${memory}mb
 %%NProcShared=${cpus}
 
-#P ${job_type_1} ${restricted}${method}${slash_1}${basis}${slash_2}${auxiliary_basis} ${job_type_2} ${fine} IOp(2/9=2000) ${keywords} ${dispersion}
+#P ${job_type_1} ${restricted}${method}${slash_1}${basis}${slash_2}${auxiliary_basis} ${job_type_2} ${fine} IOp(2/9=2000) ${keywords} ${dispersion} ${trsh}
```
alongd, line 51: OK to add `${trsh}`, but looks like `${keywords}` was meant to do the same thing, consider appending to `${keywords}` instead?

arc/job/adapters/gaussian.py
```diff
@@ -230,7 +230,13 @@ def write_input_file(self) -> None:
         input_dict['method'] = self.level.method
         input_dict['multiplicity'] = self.multiplicity
         input_dict['scan_trsh'] = self.args['keyword']['scan_trsh'] if 'scan_trsh' in self.args['keyword'] else ''
+        input_dict['trsh'] = self.args['trsh']['trsh'] if 'trsh' in self.args['trsh'] else ''
```
alongd, line 235: We have many instances where we try to retrieve a value from `self.args`, but we must first check that it contains the keys.
At some point we should implement a simple function that gets `key_1` and `key_2` and does all the checks for `args` in one place, and returns the requested value if it exists. It'll make the rest of the code more elegant. You don't have to implement it as part of this PR, just a thought

arc/job/adapters/gaussian_test.py
```diff
@@ -27,110 +27,153 @@ def setUpClass(cls):
...
-                                    job_type='composite',
-                                    level=Level(method='cbs-qb3-paraskevas'),
-                                    project='test',
-                                    project_directory=os.path.join(ARC_PATH, 'arc', 'testing', 'test_GaussianAdapter'),
-                                    species=[ARCSpecies(label='spc1', xyz=['O 0 0 1'])],
-                                    testing=True,
-                                    args={'keyword': {'general': 'IOp(1/12=5,3/44=0)'}},
-                                    )
-        cls.job_9 = GaussianAdapter(execution_type='local',
+        # cls.job_1 = GaussianAdapter(execution_type='incore',
```
alongd: is this commented out block a debug leftover?

arc/job/trsh_test.py
```diff
@@ -283,7 +283,7 @@ def test_trsh_ess_job(self):
         self.assertTrue(remove_checkfile)
         self.assertEqual(software, 'gaussian')
         self.assertEqual(memory, 16)
-        self.assertFalse(couldnt_trsh)
+        self.assertFalse(couldnt_trsh)
```
alongd: remove extra white space?

arc/job/trsh.py
```diff
@@ -851,66 +851,110 @@ def trsh_ess_job(label: str,
         couldnt_trsh = True
 
     elif software == 'gaussian':
+        trsh_keyword = [] # TODO: Current making this a list to allow for multiple keywords to be tried. Will need another approach
```
alongd, line 854: maybe the `Trsh` class approach could assist here?
BTW, I prefer not to commit `TODO` comments (sorry for the hypocrisy, I know I have done so myself more than a few times)

arc/job/adapters/gaussian_test.py
```diff
@@ -27,109 +27,109 @@ def setUpClass(cls):
...
-        # cls.job_9 = GaussianAdapter(execution_type='local',
-        #                     job_type='optfreq',
-        #                     level=Level(method='wb97xd',
-        #                                 basis='def2-TZVP'),
-        #                     project='test',
-        #                     project_directory=os.path.join(ARC_PATH, 'arc', 'testing', 'test_GaussianAdapter'),
-        #                     species=[ARCSpecies(label='anion', xyz=['O 0 0 1'], charge=-1, is_ts=False)],
-        #                     testing=True,
-        #                     )
+        cls.job_1 = GaussianAdapter(execution_type='incore',
```
alongd: this uncomments the commenting out made in a different commit, please organize the commits

alongd (commented):
Thanks @calvinp0 for taking the time to fix and improve out troubleshooting!
I made some suggestions in the comments.

@Lilachn91 says that we have an issue with the `nosym` keyword in Gaussian and that is is misplaced when creating input files. Perhaps she can contribute a commit to this PR to fix that?

arc/job/trsh.py
calvinp0: Yes, I think a class of Trsh would be more manageable long term, and much easier to handle when it comes to writing input files for the ESS. However, I think it would be a bit more work as it would be impacting a couple of parts of ARC. So maybe for another PR we could talk about doing that?

arc/job/adapters/gaussian.py
calvinp0: I am not 100% certain but I think 'trsh' came from our usage of args['trsh']. I just don't think it was ever used. So I have a suspicion we never fully implemented the troubleshooting - only for the chkfile removal. Basically, whenever we have run ARC, we have only done chkfile removal and that's it - never anything else.

Therefore, I think we should keep {'trsh'} because trsh_keywords eventually ends up in there

calvinp0: https://github.com/ReactionMechanismGenerator/ARC/blob/33ebef206edfc34c6a5be293df92d65b311cceeb/arc/scheduler.py#L777

arc/job/trsh.py
calvinp0: Yeah, i think this is a good example of trsh class

I can remove the TODO - it was more a note to you and kfir if you guys had another idea for this current PR

arc/job/adapters/gaussian.py
alongd: Updating the input file with trsh keywords should happen here in a centralized function:
https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/job/adapters/common.py#L276
The trsh directives from trsh.py should turn into keywords or blocks to be added to the respective input file, passed to the Job adapter as `args`, and added to the input file using the above centralized function.
I see what you added above from scheduler, that makes sense.

arc/job/trsh.py
alongd: Would you like to put all of these (basically identical) likes in one place?

alongd (commented):
Thanks, I added some comments

arc/job/trsh.py
calvinp0: Sorry, I don't understand. Do you mean the SCF stuff into one place?

arc/job/adapters/gaussian.py
calvinp0: Oh, I see what you mean now. I will try adjust to use that code to see if I can make it work

arc/job/trsh.py
alongd: I meant that the line `if 'CheckFile' in job_status['keywords'] or 'checkfile=None' in ess_trsh_methods:` repeats itself many time throughout this block. We should probably have a small function that does this (with a test?), and call that function wherever needed.

calvinp0: I have attempted to turn all these if statements (when possible) into functions
https://github.com/ReactionMechanismGenerator/ARC/blob/66b5d1e3c2e5eba9706ce7b47babe53439c10bf0/arc/job/trsh.py#L1540

arc/job/adapters/common.py
```diff
@@ -220,8 +221,13 @@ def _initialize_adapter(obj: 'JobAdapter',
     if obj.execution_type != 'incore' and obj.job_adapter in obj.ess_settings.keys():
         obj.determine_job_array_parameters()
 
-    obj.scan_res = obj.args['trsh']['scan_res'] \
-        if obj.args and 'trsh' in obj.args.keys() and 'scan_res' in obj.args['trsh'].keys() else rotor_scan_resolution
+    # Set scan_res if required by trsh
```
alongd, line 224: This part looks identical in function to the two lines deleted above. Was there a reason for this change?

arc/job/adapters/common.py
```diff
@@ -302,6 +308,18 @@ def update_input_dict_with_args(args: dict,
...
                     input_dict['keywords'] += f'{value} '
+        elif arg_type == 'trsh' and arg_dict:
+            for key, value in arg_dict.items():
+                # value may be a list, e.g., for the 'trsh' key
+                if isinstance(value, list) and key == 'trsh':
+                    for val in value:
+                        if key not in input_dict.keys():
+                            input_dict[key] = ''
+                        input_dict[key] += f'{val} '
+                elif key == 'trsh':
```
alongd, line 319: Thanks, this is an important addition. Could you find some place in a docstring where it makes sense, and document the desired structure of args, and secifically the bahavior of trsh in it to help maintain it?
See other instances for example, [here](https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/scheduler.py#L81) and [here](https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/species/species.py#L75).

arc/job/adapters/common.py
```diff
@@ -380,3 +398,20 @@ def which(command: Union[str, list],
         return bool(ans)
     else:
         return ans
+
+def combine_parameters(input_dict: dict, terms: list):
+
```
alongd, line 403: Please add a formal docstring like other functions here, with the purpose of the func. inc. an Args and Returns description (also need to add the returned type hint to the func. declaration)

arc/job/adapters/common.py
```diff
@@ -380,3 +398,20 @@ def which(command: Union[str, list],
         return bool(ans)
     else:
         return ans
+
+def combine_parameters(input_dict: dict, terms: list):
```
alongd, line 402: please add tests

arc/job/adapters/gaussian.py
```diff
@@ -218,8 +219,10 @@ def write_input_file(self) -> None:
                     'scan',
                     'slash_1',
                     'slash_2',
+                    'trsh'
                     ]:
             input_dict[key] = ''
+        update_input_dict_with_args(args=self.args, input_dict=input_dict)
```
alongd, line 225: previously, this was how we passed trsh directives into the input file. Now we do it in `input_dict['trsh']`. Just making sure there's no duplicity?

arc/job/adapters/gaussian.py
```diff
@@ -322,7 +327,18 @@ def write_input_file(self) -> None:
...
                 else ' guess=mix'
 
-        input_dict = update_input_dict_with_args(args=self.args, input_dict=input_dict)
+        #Fix SCF
+        terms = ['scf=\((.*?)\)', 'scf=(\w+)']
+        input_dict, parameters = combine_parameters(input_dict, terms)
+        if parameters:
+            input_dict['trsh'] += f" scf=({','.join(parameters)})"
+
+        input_dict =  { key.strip(): (val.rstrip() if isinstance(val, str) and val.startswith('\n') and not val.endswith('\n') else
```
alongd, line 336: this part:
```
(val.rstrip() if isinstance(val, str) and val.startswith('\n') and not val.endswith('\n') else
                  val.lstrip() if isinstance(val, str) and not val.startswith('\n') and val.endswith('\n') else
                  val.strip() if isinstance(val, str) and not val.startswith('\n') and not val.endswith('\n') else val)
```
has some level of complexity. I recommend to extract it into a function in job.common with some tests, and use ti here inside the loop/generator

arc/job/adapters/gaussian_test.py
```diff
@@ -319,6 +462,163 @@ def test_set_files_for_pipe(self):
     def test_gaussian_def2tzvp(self):
         """Test a Gaussian job using def2-tzvp"""
         self.assertEqual(self.job_9.level.basis.lower(), 'def2tzvp')
+    
+    def test_trsh_write_input_file(self):
```
alongd, line 466: Excellent, thank you!

arc/job/trsh.py
```diff
@@ -849,60 +849,61 @@ def trsh_ess_job(label: str,
                  or 'is not appropriate for the this chemistry' in job_status['error']):
         output_errors.append(f'Error: Could not recognize basis set {job_status["error"].split()[-1]} in {software}; ')
         couldnt_trsh = True
-
+
```
alongd, line 852: remove these empty lines above?

arc/job/trsh.py
```diff
@@ -1536,3 +1537,72 @@ def scan_quality_check(label: str,
             return invalidate, invalidation_reason, message, actions
 
     return invalidate, invalidation_reason, message, actions
+ 
+def trsh_keyword_checkfile(job_status, ess_trsh_methods, couldnt_trsh) -> bool:
```
alongd, line 1541: The return types here and below are tuples of bool and other things, something like: `Tuple[bool, list, bool]`

arc/job/trsh.py
```diff
@@ -1536,3 +1537,72 @@ def scan_quality_check(label: str,
...
+    """
+    Check if the job requires removal of checkfile
+    """
+    if 'CheckFile' in job_status['keywords'] or 'checkfile=None' in ess_trsh_methods:
+        ess_trsh_methods.append('checkfile=None')
+        couldnt_trsh = False
+        return True, ess_trsh_methods, couldnt_trsh
+    return False, ess_trsh_methods, couldnt_trsh
+
+def trsh_keyword_intaccuracy(ess_trsh_methods, trsh_keyword, couldnt_trsh) -> list:
```
alongd, line 1551: thanks for restructuring these!

alongd (commented):
Thanks for this addition! Please see some comments below

arc/job/adapters/common.py
calvinp0: It is not identical.

arc/job/adapters/common.py
calvinp0: I don't know the full expectation of the desired structure of arguments. This was just to rectifiy an issue that I came across month ago.

arc/job/adapters/common.py
```diff
@@ -380,3 +451,39 @@ def which(command: Union[str, list],
...
```
alongd, line 455: Minor: Please add the returned type: `-> Tuple[dict, list]`

alongd (approved):
Thanks!

PR #672, opened 2023-06-13
Atom Mapping Fixup

kfir4444:
This PR solves a problem with the atom mapping algorithm in ARC, in which multiple scissions are required on the same species. A new method is introduced and tested. In addition, the make_bond_change function can now add a double bond between atoms that one of them does not have radicals. Additional fix ups are introduced.


calvinp0 (approved):
Changes look great!

PR #677, opened 2023-06-15
Change CI cleaning

kfir4444:
This PR fixes 2 things,
The first is that the free-space action sometimes caused unnecessary failures, which is fixed by forking the action and patching it, and a PR to the main action repo is opened.
The second one is a Sella issue, which is (probably) caused by a syntax error in their new release (from 15.6). This issue is fixed by using the previous release, and informing the Sella developers on this issue.


calvinp0 (approved):
Looks good mate. Good catch on the Sella issue

PR #678, opened 2023-06-19
Load running jobs' xyz as dict from restart

alongd:
Otherwise, ARC fails when it tries to compute dihedral angles after a restarted scan job has finished

### DIFF
--- a/arc/job/adapter.py
+++ b/arc/job/adapter.py
@@ -824,7 +824,7 @@ class JobAdapter(ABC):
         if self.tsg is not None:
             job_dict['tsg'] = self.tsg
         if self.xyz is not None:
-            job_dict['xyz'] = xyz_to_str(self.xyz)
+            job_dict['xyz'] = self.xyz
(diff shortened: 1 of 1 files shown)


arc/main.py
```diff
@@ -285,6 +286,10 @@ def __init__(self,
         self.output = output
         self.standardize_output_paths()  # depends on self.project_directory
         self.running_jobs = running_jobs or dict()
+        for jobs in self.running_jobs.values():
+            for job in jobs:
+                if 'xyz' in job.keys():
+                    job['xyz'] = str_to_xyz(job['xyz'])
```
kfir4444, line 292: Question, how does the xyz gets to `self.running_jobs` as a string? I think we should fix the source of the problem.

kfir4444 (commented):
Thanks @alongd,
please see my question

NellyMitnik (approved):
looks good

kfir4444 (approved):

PR #679, opened 2023-06-22
Created cheat_sheet function to check if smiles is in cheat sheets

MichalKesl:
I have updated the cheat sheets to include C2H3 and added a new function called "cheat sheet." The test aims to confirm the accurate representation of the C2H3 species when providing the SMILES notation in both directions (C=[CH] / [CH]=C).

### DIFF
--- a/arc/species/conformers.py
+++ b/arc/species/conformers.py
@@ -41,7 +41,7 @@ import logging
 import sys
 import time
 from itertools import product
-from typing import List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from openbabel import openbabel as ob
 from openbabel import pybel as pyb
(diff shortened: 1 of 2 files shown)


arc/species/conformers.py
```diff
@@ -126,6 +126,22 @@
...
+    Check if the species is in the cheat sheet, and return its correct xyz if it is.
+
+    Args:
+        mol_list (Union[List[Molecule], Molecule]): Molecule objects to consider (or Molecule, resonance structures will be generated).
+    
+    Returns: Optional[lis[dict]]
+    """
+    for smiles in CHEAT_SHEET.keys():
+        cheat_mol = Molecule(smiles=smiles)
+        for mol in mol_list:
```
alongd, line 140: If we allow `mol_list` to also be a single Molecule and not always a list, then this line may raise an error since we can't loop through a Molecule. I suggest adding in the beginning of the function:
`mol_list = [mol_list] if not isinstance(mol_list, list) else mol_list`

arc/species/conformers.py
```diff
@@ -126,6 +126,22 @@
...
+            if cheat_mol == mol.copy(deep=True):
```
alongd, line 141: we should compare Molecule objects using isomorphism: `if cheat_mol.is_isomorphic(mol.copy(deep=True)`

alongd (commented):
Thanks for the contribution! Please see a few comments

arc/species/conformers.py
kfir4444: Actually, the `==` contains `.is_isomorphic` and even more. See: https://github.com/ReactionMechanismGenerator/RMG-Py/blob/ad51418fecf4fd8651fbef4cc09f21ecf6634f41/rmgpy/molecule/molecule.py#L1002
However, we may also need to deepcopy the other mol object to avoid scrambling.

arc/species/conformers.py
```diff
@@ -126,6 +126,23 @@
...
+    Args:
+        mol_list (Union[List[Molecule], Molecule]): Molecule objects to consider (or Molecule, resonance structures will be generated).
+    
+    Returns: Optional[lis[Dict]]
+    """
+    mol_list = [mol_list] if not isinstance(mol_list, list) else mol_list
+    for smiles in CHEAT_SHEET.keys():
+        cheat_mol = Molecule(smiles=smiles)
+        for mol in mol_list:
+            if cheat_mol.copy(deep=True) == mol.copy(deep=True):
```
alongd, line 142: I think that the comparison should be `if cheat_mol.is_isomorphic(mol.copy(deep=True))`
`is_isomorphic()` is the preferred way to compare molecular structures. Also, no need to use a copy of the cheat mol.

alongd (approved):
Thanks!!

PR #680, opened 2023-06-24
Another round of atom mapping fixes

kfir4444:
Finally, no more reactions in work in progress

### DIFF
--- a/arc/mapping/driver.py
+++ b/arc/mapping/driver.py
@@ -29,7 +29,7 @@ from arc.mapping.engine import (assign_labels_to_products,
                                 RESERVED_FINGERPRINT_KEYS,)
 from arc.common import logger
 
-from rmgpy.exceptions import ActionError
+from rmgpy.exceptions import ActionError, AtomTypeError
 
 if TYPE_CHECKING:
     from rmgpy.data.rmg import RMGDatabase
@@ -248,14 +248,16 @@ def map_rxn(rxn: 'ARCReaction',
 
     try:
         make_bond_changes(rxn, r_cuts, r_label_dict)
-    except (ValueError, IndexError, ActionError):
-        return None
+    except (ValueError, IndexError, ActionError, AtomTypeError) as e:
+        logger.warning(e)
 
     r_cuts, p_cuts = update_xyz(r_cuts), update_xyz(p_cuts)
 
     #step 4:
     pairs_of_reactant_and_products = pairing_reactants_and_products_for_mapping(r_cuts, p_cuts)
-
+    if len(p_cuts):
+        logger.error(f"Could not find isomorphism for scissored species: {[cut.mol.smiles for cut in p_cuts]}")
+        return None
     # step 5:
     maps = map_pairs(pairs_of_reactant_and_products)
 
--- a/arc/mapping/engine.py
+++ b/arc/mapping/engine.py
@@ -1122,13 +1122,15 @@ def make_bond_changes(rxn: 'ARCReaction',
                             atom1 = atom
                         elif int(atom.label) == indicies[1]:
                             atom2 = atom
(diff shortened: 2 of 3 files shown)


arc/mapping/driver.py
```diff
@@ -248,14 +248,16 @@ def map_rxn(rxn: 'ARCReaction',
...
+    except (ValueError, IndexError, ActionError, AtomTypeError) as e:
+        logger.warning(e)
 
     r_cuts, p_cuts = update_xyz(r_cuts), update_xyz(p_cuts)
 
     #step 4:
     pairs_of_reactant_and_products = pairing_reactants_and_products_for_mapping(r_cuts, p_cuts)
-
+    if len(p_cuts):
+        logger.error(f"Could not find isomorphism for scissored species: {[cut.mol.smiles for cut in p_cuts]}")
```
alongd, line 259: I don't think an `ARCSpecies` has a `.smiles` attribute

arc/mapping/driver.py
```diff
@@ -248,14 +248,16 @@ def map_rxn(rxn: 'ARCReaction',
...
-        return None
+    except (ValueError, IndexError, ActionError, AtomTypeError) as e:
+        logger.warning(e)
 
     r_cuts, p_cuts = update_xyz(r_cuts), update_xyz(p_cuts)
 
     #step 4:
     pairs_of_reactant_and_products = pairing_reactants_and_products_for_mapping(r_cuts, p_cuts)
-
+    if len(p_cuts):
```
alongd, line 258: should this be the opposite, `if not len(p_cuts)`?

arc/reaction_test.py
```diff
@@ -1325,13 +1326,6 @@ def test_get_atom_map(self):
         self.assertIn(tuple(rxn.atom_map[5:7]+rxn.atom_map[8:]), permutations([2, 3, 4, 5]))
         self.assertTrue(check_atom_map(rxn))
 
-    @work_in_progress
```
alongd, line 1328: hooray!

arc/mapping/driver.py
```diff
@@ -29,7 +29,7 @@
                                 RESERVED_FINGERPRINT_KEYS,)
 from arc.common import logger
 
-from rmgpy.exceptions import ActionError
+from rmgpy.exceptions import ActionError, AtomTypeError
```
alongd, line 32: squash with the "Improved logging of atom mapping" commit?

arc/mapping/engine.py
```diff
@@ -1122,13 +1122,15 @@ def make_bond_changes(rxn: 'ARCReaction',
                             atom1 = atom
                         elif int(atom.label) == indicies[1]:
                             atom2 = atom
-                    if atom1.radical_electrons == 0:
+                    if atom1.radical_electrons == 0 and atom2.radical_electrons == 0: # Both atoms doesn't have any radicals, but their bond is changing. There probably is resonance, so this will not effect the isomorphism check.
```
alongd, line 1125: Very minor typos: `don't` instead of `doesn't`, `affect` instead of `effect`

alongd (commented):
Thanks, looking good!
I guess the last commit addresses the issue we saw with partial charges?

arc/mapping/driver.py
kfir4444: No, after `pairs_of_reactant_and_products` `p_cuts` should be an empty list

arc/mapping/driver.py
kfir4444: Do you mean in: `[cut.mol.smiles for cut in p_cuts]`? because the `.mol` object does have `.smiles` property.

alongd (approved):
Thanks!

PR #682, opened 2023-07-02
Don't consider conformers for which RDKit could not compute energy

alongd:
Don't consider conformers for which RDKit could not compute energy, since we don't know how to rank the conformers by energy. This could potentially be problematic if one of these conformers has a lower E, but we can't return too much confs to the scheduler to be DFT'ed. We think that ARC's DFT scans may compensate if indeed a lower conformer exists.
In addition, if all conformers have no FF energies, we return them all (a hypothetical case).


NellyMitnik (approved):
I ran the specie: `[O]OOC=CC#CCO`, it **didn't crash** and RDkit computed energy for 10 conformers.

PR #683, opened 2023-07-02
Consider the project directory when reading ARCSpecies or TSGuess xyz

alongd:
Allow users to simply write

```yaml
reactions:
  - label: <some label...>
    xyz:
    - TS1.gjf
    - TS2.gjf
```

and populate the reaction user guesses correctly from the given files considering the project directory as needed.
Tests added.


calvinp0 (approved):
Looks good!

PR #684, opened 2023-07-05
Keep the provided species label when loading from YAML

alongd:
Keep the label provided in an ARC input file for a species even if the YAML file has a different label for it


MichalKesl (approved):
looks good to me

PR #686, opened 2023-07-27
Reset species output paths when deleting all species jobs

alongd:
When we `delete_all_species_jobs`, for example after switching a TS guess to something better and we no longer need the sp, freq, etc jobs of the previous TS, we must also reset the `output` dict of the `Scheduler` object, so that all paths are empty. Note that for `irc` we have two paths, forward and reverse, so we treat it differently (see [here ](https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/scheduler.py#L101)the structure of the output dict)


Lilachn91 (approved):
Looks good :)
Thanks!

PR #689, opened 2023-07-30
Azure server

calvinp0:
Due to acquiring a subscription to Azure, we need to update ARC to allow for the usage of Azure as well as ESS like QCHEM. 
Please see the list below of all the changes. NOTE: Tests still need to be fixed or written for these new changes

1. Added 'rapidfuzz' to ARC environment
2. basis set dataframe for QChem
3. check_normal_mode_displacement - Allow for a try and except case
4. Adapter: Submit Script Format updated with using user provided username {un}, also convert memory to an 'int', and also provide the option of {server_nodes} if required
5. Adapter: Total Submit Memory adjusted to now ensure that when troubleshooting a job, it never attempts to go OVER the maximum memory of the allowed submission memory of the node/server
6. Adapter: SLURM Submit Memory - Using `#SBATCH --mem` as the parameter now as it defines the TOTAL memory of the submission
7. Adapter: SSH File Download - We do not expect to always download or upload certain files depending on the scheduler via SSH. This change allows for recognising if certain files will be uploaded or downloaded depending on the user's scheduler choice
8. Adapter: Null Bytes can appear in files, or rather more specifically, if QChem has an error, it can produce Null Bytes in the out.txt/err.txt files and thus requires a different type of reading of the file contents. This is not 100% full proof though and may need extra work
9. Adapters: In #390 branch, SSH had improvements but were not merged. I have brought forth an improvement from this branch were Remote Files are removed once they are download to the local computer
10. ARC can now recognise that IRC is also supported by QChem
11. common.py: Slight reworking of the trouble shooting of the scan resolution so that it once the obj.scan_res has been defined with the new rotor scan resolution, the scan res is removed from the obj.args['trsh']['scan_res'] properly
12. SSH: Updated decorator to use the correct connect function
13. SSH: If the user provides a server that is not in servers.keys() and server is also not None, then an error is raised to informat the user that they need to fix up the server keys
14. SSH: An error that can occur is when a submission to a scheduler includes an incorrect memory specification, then there is warning to the user that the requested memory is not supported and needs to be checked. May need to make this a ValueError instead of a logger warning
15. SSH: Slight adjustment to checking if there is an stdout after submission attempt
16. SSH: Some servers require private keys. Originally the code was incorrectly adding the private key to the SSH paramiko function. It has now been changed so that system keys are loaded and then if the user provides a private key, it is included in the connect function
17. SSH: Updated default arguments to `get_last_modified_time`
18. SSH: Changed the lowering of the cluster soft
19. SSH: Added a function to remove the directory on the remote server
20. SSH: Azure SLURM has an extra status called 'CF' which means configuring (for the node). This can take 10-15 mins or so before the node is online. We now ensure to caputre this. HOWEVER, a node can get stuck in 'CF' status. Now we check this via checking the current time the node has been active, splitting the time up correctly (different formats of time are possible), and then if it is above 15 minutes, we run the command `scontrol show node {node_id}`. If the stdout includes the phrase 'NOT_RESPONDING' then we return 'errored'
21. XYZ to Smiles: Warning Update to Possible Valence
22. Vectors: Reading Coords that are in string format using regex and putting into a tuple
23. Species: The original code did not functional correctly - nor was never used hence why it was passed into production. It has now been changed to properly return the actual number of heavy atoms.
24. [WIP] Getting the Internal Coordinates for QChem - Still not operational
25. submit.py & settings.py: Updated for SLURM and AZURE 
26. Scheduler: Now imports JobError
27. Scheduler: Fixed adding trsh to the args
28. Scheduler: Added JobError exception for determining job status
29. Scheduler: Now removing remote jobs at the end of the scheduler - !!!MAY NEED TO SEE IF THIS HAS AN EFFECT ON NON SSH JOBS!!!
30. Scheduler: Getting the recent opt job name via properly checking if the opt job was considered done (This was not done before)
31. Scheduler: TODO - We attempt to trouble shoot a frequency we deem not okay. Yet, there is no specific troubleshoot method, so why do we do this?
32. Scheduler: Properly troubleshoot an job
33. Scheduler: Fix conformer troubleshoot if it was a TS conformer
34. parser: TODO - Why do we set raise error as true for normal mode displacement parsing? It has an effect on the function of raising a not implemented error even though it is implememnt
35. parser: Now can parse the normal mode displacement of QCHEM
36. parser: Now can parse the 1d scan coords of QCHEM
37. parser: Can now parse trajectory of QCHEM
38. parser: Can now parse arguments in the scan input file QCHEM
39. parser: NEED to fix parse_ic_info for QCHEM
40. QChem Adapter: Import - Pandas, ARC_PATH, rapidfuzz
41. QChem Adapter: Input Template now supports IRC and {trsh} args and ensures IQMOL_FCHK is false (This can be set to true BUT be aware this .fchk file can be rather large)
42. QChem Adapter: write_input_file - basis set is now matched via the software_input_matching function
43. QChem Adapter: write_input_file - QChem now supports D3 method. We should look at working with other DFT_D methods in the future. More specifically there are other types of D3 methods
44. QChem Adapter: write_input_file - Correctly pass in troubleshooting arguments into the input file
45. QChem Adapter: write_input_file - Capitalised TRUE/FALSE in UNRESTRICTED parameter
46. QChem Adapter: write_input_file - Removed the scan job type and moved it to another section of the input file writing
47. QChem Adapter: write_input_file - If scan is set, the job_type is PES_SCAN. We also set the fine to be XC_GRID 3. However, we may need to look into changing the tolerances later
48. QChem Adapter: write_input_file - We now write correctly the torsional scans for the input file for a scan
49. QChem Adapter: write_input_file - IRC is now supported, however this input file means we run two jobs from the one input file - A  FREQ job and then IRC. This currently works but will need improvements when used more by users
50. QChem Adapter: write_input_file - Ensuring that the SCF CONVERGENCE is 10^-8 for certain job types
51. QChem Adapter: [NEWFUNCTION] generate_scan_angles - to support PES SCAN jobs, we have a function that will look at what the required angle we want to scan, and the step size, and then return a start and end angle between -180, 180 that will ensure we scan the require angle during the stepping
52. QChem Adapter: [NEWFUCNTION] software_input_matching - Since QCHEM has different formatting for basis sets, this function will try take the users format of the basis set and match it against a dataframe (which should always be updated if its missing a format). This uses the new package in the ARC environment called rapidfuzz
53. TrshQChem: Fixed error checking in QChem output files. It would originally mistakenly think SCF failed was the error due to what errors it would look for in the lines
54. TrshQChem: FlexNet Licensing Error - If the license server is not working this will cause ARC to stop
55. TrshQChem: Max Optimisation Cycles is probably checked for in the output file
56. TrshQChem: Max Iteration Cycles is identified now if there is a failure during SCF convergence
57. TrshMolpro: Molpro reports memory errors that need to be properly troubleshooted differently than what we did originally. Now, we will look for how much memory needs to be increased in order for molpro to run successfully. This is done through regex pattern matching. We also check for triples memory increase if required
58. Trsh: determine_job_log_memory_issue - Sometimes the job log can have null bytes in them, usually a QCHEM issue, and so this means we need to open the file to read differently
59. TrshQChem: trsh_ess_job - QCHEMs trsh has been reworked so now that it will combine troubleshoot attempts if they were attempted previously. For example, if we troubleshooted the max opt cycle but now need to turn on SYM IGNORE, it will include both of these statements in the troubleshooting input file
60. TrshQMolpro: trsh_ess_job - Molpro required a chnage in how we troubleshoot the memory. If we get an error for the memory it is because either the MWords per process is not enough, even though we have provided an adequate about of memory to the submit script OR the MWords per process is enough but the TOTAL MEMORY (MWords * CPUS) > Max Node Memory, therefore CORES has to be reduced.
61. TrshSSH:trsh_job_on_server - Fixed it as a 'with' statement so the client is closed when exiting the 'with' statement
62. Molpro Adapter: Molpro needs a different touch to troubleshooting its memory. Here in setting the input file memory we determine if the MWords was enough per process but the total memory was too high. If that's the case, we reduce the processes req. while maintaining the memory per process


arc/checks/ts.py
```diff
@@ -67,7 +67,11 @@ def check_ts(reaction: 'ARCReaction',
         check_ts_energy(reaction=reaction, verbose=verbose)
 
     if 'freq' in checks or (not reaction.ts_species.ts_checks['normal_mode_displacement'] and job is not None):
-        check_normal_mode_displacement(reaction, job=job)
+        try:
+            check_normal_mode_displacement(reaction, job=job)
+        except ValueError as e:
+            logger.error(f'Could not check normal mode displacement, got: \n{e}')
+            reaction.ts_species.ts_checks['normal_mode_displacement'] = True
```
Lilachn91, line 74: Can you explain why do you add that line? 
Thanks!

calvinp0: I believe I pulled this from the ts2 branch
https://github.com/ReactionMechanismGenerator/ARC/blob/5277e5a1b8cef95c01b1ae1f8c77df189b5050cc/arc/checks/ts.py

@alongd Should I remove this change as it should only be for that specific branch?

arc/job/adapters/qchem.py
```diff
@@ -280,11 +361,130 @@ def write_input_file(self) -> None:
...
 
         input_dict = update_input_dict_with_args(args=self.args, input_dict=input_dict)
 
         with open(os.path.join(self.local_path, input_filenames[self.job_adapter]), 'w') as f:
             f.write(Template(input_template).render(**input_dict))
+    def generate_qchem_scan_angles(self,start_angle: int, step: int) -> (int, int, int, int):
+            """
+            Generates the angles for a Q-Chem scan. The scan is split into two parts, one from start_angle to 180, and one from -180 to end_angle.
+
+            Parameters
```
alongd, line 385: this is a different format than what we use throughout ARC, please modify to the `Args:` `Returns:` format for consistency

arc/job/adapters/qchem.py
```diff
@@ -280,11 +361,130 @@ def write_input_file(self) -> None:
...
+                The starting angle for the second part of the scan
+            scan2_end : int
+                The ending angle for the second part of the scan
+            """
+
+            # First, we need to check that the start_angle is within the range of -180 to 180, and if not, convert it to be within that range
+            if start_angle > 180:
+                start_angle = start_angle - 360
+
+
```
alongd, line 408: please remove one line break

arc/job/adapters/qchem.py
```diff
@@ -280,11 +361,130 @@ def write_input_file(self) -> None:
...
+                The ending angle for the first part of the scan
+            scan2_start : int
+                The starting angle for the second part of the scan
+            scan2_end : int
+                The ending angle for the second part of the scan
+            """
+
+            # First, we need to check that the start_angle is within the range of -180 to 180, and if not, convert it to be within that range
+            if start_angle > 180:
+                start_angle = start_angle - 360
```
alongd, line 406: please use https://github.com/ReactionMechanismGenerator/ARC/blob/e9fe17edaba0507e15093dbd4cb2f11305a5b858/arc/common.py#L1274 instead
Can also be applied below for the other params

arc/job/adapters/qchem.py
```diff
@@ -280,11 +361,130 @@ def write_input_file(self) -> None:
...
+            if start_angle == 0 and end_angle == 0:
+                scan1_start = start_angle
+                scan1_end = 180
+                scan2_start = -180
+                scan2_end = end_angle
+            elif start_angle == 180:
+                # This is a special case because the scan will be from 180 to 180
+                # This is not allowed in Q-Chem so we split it into two scans
+                # Arguably this could be done in one scan but it is easier to do it this way
+                # We will need to find the starting angle that when added by the step size will be 180
```
alongd, line 431: why not set `start_angle` to -180 instead and do it in one scan?

arc/job/adapters/qchem.py
```diff
@@ -328,7 +528,7 @@ def set_files(self) -> None:
             self.files_to_download.append(self.get_file_property_dictionary(
                 file_name=output_filenames[self.job_adapter]))
             # 2.3. checkfile
-            self.files_to_download.append(self.get_file_property_dictionary(file_name='orbitals.fchk'))
+            #self.files_to_download.append(self.get_file_property_dictionary(file_name='input.fchk'))
```
alongd, line 531: I see that the file name is change, and the line is commented out. Is this meant to be merged?

arc/job/adapters/qchem.py
```diff
@@ -362,6 +562,52 @@ def execute_queue(self):
...
         """
         self.legacy_queue_execution()
+    
+    def software_input_matching(self, basis):
+        """
+        Check if the user specified software is compatible with the level of theory. If not, try to match the software with
+        similar methods. If no match is found, raise an error.
+
+        Matching is done by comparing the user specified software with the software in the basis_sets.csv file.
+
```
alongd, line 572: please add `Args:` and `Returns:` to the docstring

arc/job/adapters/qchem.py
```diff
@@ -362,6 +562,52 @@ def execute_queue(self):
         Execute a job to the server's queue.
         """
         self.legacy_queue_execution()
+    
+    def software_input_matching(self, basis):
```
alongd, line 566: Looks like this function too fits better in the adapter.py file

arc/level_test.py
```diff
@@ -209,7 +209,7 @@ def test_determine_compatible_ess(self):
         self.assertIsNone(level_2.compatible_ess)
         level_2.determine_compatible_ess()
         self.assertEqual(sorted(level_2.compatible_ess), sorted(['gaussian', 'qchem', 'terachem']))
-
+
```
alongd, line 212: the commit message says "Will require some additional tests"
What do you have in mind? Are you still working on this file?

arc/job/adapters/qchem_test.py
```diff
@@ -165,7 +165,7 @@ def test_write_input_file(self):
    METHOD        b3lyp
    UNRESTRICTED  TRUE
    BASIS         def2-TZVP
-   IQMOL_FCHK    TRUE
+   IQMOL_FCHK    FALSE
```
alongd, line 168: please squash this commit into the original QChem test commit you made

arc/job/adapter.py
```diff
@@ -754,13 +754,13 @@ def set_cpu_and_mem(self):
                            f'exceeds {100 * job_max_server_node_memory_allocation}% of the the maximum node memory on '
                            f'{self.server}. Setting it to {job_max_server_node_memory_allocation * max_mem:.2f} GB.')
             self.job_memory_gb = job_max_server_node_memory_allocation * max_mem
-            total_submit_script_memory = self.job_memory_gb * 1024 * 1.05 if (self.job_memory_gb * 1024 * 1.05) <= max_mem else max_mem * 1000 # MB
+            total_submit_script_memory = self.job_memory_gb * 1024 * 1.05 if (self.job_memory_gb * 1024 * 1.05) <= (max_mem * 1024) else max_mem * 1024 # MB
```
alongd, line 757: Can you squash? I commented about this on the original commit

alongd (commented):
Thanks, @calvinp0, for investing the time and effort into this very needed extension!
Please see my comments below

.gitignore
```diff
@@ -36,6 +36,7 @@ scratch/
 
 # csv database
 *.csv
+!basis_sets.csv
```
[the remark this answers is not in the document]
calvinp0: So that whenever someone `gits add .` it will make sure to add the `basis_sets.csv`. Basically it is excluded from the rule `*.csv`

arc/species/species.py
```diff
@@ -579,9 +579,8 @@ def number_of_heavy_atoms(self) -> int:
                 self._number_of_heavy_atoms = len([symbol for symbol in self.get_xyz()['symbols'] if symbol != 'H'])
             elif self.is_ts:
                 for ts_guess in self.ts_guesses:
-                    if ts_guess.xyz is not None:
-                        self._number_of_heavy_atoms = \
-                            len([line for line in ts_guess.xyz.splitlines() if line.split()[0] != 'H'])
+                    if ts_guess.get_xyz() is not None:
+                        self._number_of_heavy_atoms = len([symbol for symbol in ts_guess.get_xyz()['symbols'] if symbol != 'H'])
```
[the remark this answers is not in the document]
alongd: I would consider hydrogen isotopes as "light" atoms like H

arc/job/adapter.py
```diff
@@ -753,10 +754,10 @@ def set_cpu_and_mem(self):
                            f'exceeds {100 * job_max_server_node_memory_allocation}% of the the maximum node memory on '
                            f'{self.server}. Setting it to {job_max_server_node_memory_allocation * max_mem:.2f} GB.')
             self.job_memory_gb = job_max_server_node_memory_allocation * max_mem
-            total_submit_script_memory = self.job_memory_gb * 1024 * 1.05  # MB
+            total_submit_script_memory = self.job_memory_gb * 1024 * 1.05 if (self.job_memory_gb * 1024 * 1.05) <= max_mem else max_mem * 1000 # MB
```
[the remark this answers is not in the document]
MichalKesl: According to ```set_cpu_and_mem(self)``` function, max_mem is in GB

https://github.com/ReactionMechanismGenerator/ARC/blob/5d5fc3a0c739c6a43e713abf613e3742931f96ae/arc/job/adapter.py#L749

arc/job/adapter.py
```diff
@@ -1312,6 +1331,14 @@ def troubleshoot_server(self):
         if run_job:
             # resubmit job
             self.execute()
+    
+    def remove_remote_files(self):
```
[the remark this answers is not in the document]
MichalKesl: Here

    https://github.com/ReactionMechanismGenerator/ARC/blob/b7cdbb5fe6ebe1c7404afdd2b9d744a54277cce0/arc/scheduler.py#L977

PR #690, opened 2023-08-03
Improved TS energy checks

alongd:
Calling the checks from a centralized function
Refactoring ts_check functions
Tests added

### DIFF
--- a/arc/checks/ts.py
+++ b/arc/checks/ts.py
@@ -37,10 +37,16 @@ LOWEST_MAJOR_TS_FREQ, HIGHEST_MAJOR_TS_FREQ = settings['LOWEST_MAJOR_TS_FREQ'], 
 
 
 def check_ts(reaction: 'ARCReaction',
-             verbose: bool = True,
              job: Optional['JobAdapter'] = None,
              checks: Optional[List[str]] = None,
              rxn_zone_atom_indices: Optional[List[int]] = None,
+             species_dict: Optional[dict] = None,
+             project_directory: Optional[str] = None,
+             kinetics_adapter: Optional[str] = None,
+             output: Optional[dict] = None,
+             sp_level: Optional['Level'] = None,
+             freq_scale_factor: float = 1.0,
+             verbose: bool = True,
              ):
     """
     Check the TS in terms of energy, normal mode displacement, and IRC.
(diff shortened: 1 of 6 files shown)


arc/checks/ts.py
```diff
@@ -110,45 +110,29 @@ def check_ts_energy(reaction: 'ARCReaction',
...
+                                   energy_type='electronic energy')
+    if all([val is not None for val in [r_ee, p_ee, ts_ee]]):
+        if ts_ee > r_ee + 1.0 and ts_ee > p_ee + 1.0:
+            reaction.ts_species.ts_checks['e_elect'] = True
+            return
+        if verbose:
+            logger.error(f'TS of reaction {reaction.label} has a lower electronic energy value than expected.')
+            reaction.ts_species.ts_checks['e_elect'] = False
+            return
     if verbose:
```
kfir4444, line 156: Note that this is the third `if verbose` in this function. Could you please unify them?

kfir4444 (changes requested):
Hey @alongd thanks for this PR!
I have added some comments.

arc/scheduler.py
```diff
@@ -2431,20 +2429,22 @@ def check_rxn_e0_by_spc(self, label: str):
         """
         for rxn in self.rxn_list:
             labels = rxn.reactants + rxn.products + [rxn.ts_label]
-            if label in labels and not rxn.ts_species.ts_checks['E0'] \
+            if label in labels and rxn.ts_species.ts_checks['E0'] is None \
```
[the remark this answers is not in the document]
alongd: Right, this is just semantics, not logic. I think the current version is more explicit.

alongd:
Thanks, @kfir4444, I addressed the comments with fixup commits, let me know if you have additional comments or whether I can go ahead and squash

arc/checks/ts.py
kfir4444: I agree, if there is a return between them then it will exit the function, and there is no need for 3 of them, right?

arc/checks/ts.py
```diff
@@ -110,60 +133,45 @@ def check_ts_energy(reaction: 'ARCReaction',
...
+                                   ts_label=reaction.ts_label, chosen_ts=reaction.ts_species.chosen_ts,
+                                   energy_type='electronic energy')
+    if all([val is not None for val in [r_ee, p_ee, ts_ee]]):
+        if ts_ee > r_ee + 1.0 and ts_ee > p_ee + 1.0:
+            reaction.ts_species.ts_checks['e_elect'] = True
+            return
+        if verbose:
+            logger.error(f'TS of reaction {reaction.label} has a lower electronic energy value than expected.')
+            reaction.ts_species.ts_checks['e_elect'] = False
+            return
```
alongd, line 155: return

kfir4444 (approved):

PR #695, opened 2023-08-21
Added a `Linear` TS search job adapter

alongd:
Adds a new Linear TS search adapter that builds TS guesses from atom-mapped reactants and products via Z-matrix chimera construction with Hammond-biased weighting. The adapter is incore-only, plugs into ARC's scheduler like heuristics/autotst_ts, and delegates heavy geometry work to a new linear_utils/ subpackage (5 modules). Currently implemented for isomerization/unimolecular reactions.

The PR also carries supporting additions to arc/species/zmat.py (anchors, smart-anchor detection, zmat re-indexing helpers), arc/species/converter.py (atom-map reordering), arc/reaction/reaction.py (is_unimolecular, refined is_isomerization), and a biradical-preservation fix in arc/species/species.py. CI was hardened: pinned action, `-n 4 --dist worksteal` for stability, and obabel test made self-contained.


arc/job/adapters/ts/linear_utils/addition.py
```diff
@@ -0,0 +1,1524 @@
...
+                node = queue.popleft()
+                if node in visited:
+                    continue
+                visited.add(node)
+                component.add(node)
+                queue.extend(local_adj[node] - visited)
+            frags.append(component)
+        return frags
+
+    def _heavy_formulas(formulas: List[Dict[str, int]]) -> List[Dict[str, int]]:
```
calvinp0: I personally am not the biggest fan of nested functions - are we able to pull these out? I feel it helps with readability cause I got a bit confused at like 92/93 thinking these were two floating variables between functions before realising this was all in one function

arc/job/adapters/ts/linear_utils/addition.py
```diff
@@ -0,0 +1,1247 @@
...
+        if existing_edges:
+            BondClass = type(existing_edges[0])
+
+    for a_idx, b_idx in (cross_bonds or []):
+        if BondClass is None:
+            break
+        try:
+            mol_copy.get_bond(copy_atoms[a_idx], copy_atoms[b_idx])
+            continue  # bond already exists
+        except (ValueError, KeyError):
```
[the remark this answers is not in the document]
calvinp0: Yeah, I think we should not do bare excepts

.github/workflows/ci.yml
```diff
@@ -23,21 +23,21 @@ jobs:
...
-          swap-storage: true
+          swap-storage: false
 
       - name: Set up micromamba (arc_env)
         uses: mamba-org/setup-micromamba@v3
         with:
           environment-name: arc_env
           environment-file: ARC/environment.yml
-          cache-environment: true
+          cache-environment: false
```
calvinp0: is there a reason we are not caching the environment?

alongd: To keep the nightly cron (line 9 of ci) catch env drift issues, I'll implement:
`cache-environment: ${{ github.event_name != 'schedule' }}` to make PR's CI faster

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+    if not all(0 <= idx < n_atoms for idx in (migrating, origin, target)):
+        return None
+
+    coords = np.array(r_xyz['coords'], dtype=float).copy()
+    atom_to_idx = {atom: idx for idx, atom in enumerate(r_mol.atoms)}
+
+    # Calibrated TS target distances for [1,3]-sigmatropic shifts.
+    sbl_break = get_single_bond_length(symbols[origin], symbols[migrating]) or 1.5
+    sbl_form = get_single_bond_length(symbols[target], symbols[migrating]) or 1.5
+    _SIGMA_BREAK_STRETCH = 0.77
```
calvinp0: I think these (incl. _SIGMA_FORM_STRETCH) should become module level

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+    for atom in r_mol.atoms:
+        ia = atom_to_idx[atom]
+        if symbols[ia] == 'H':
+            continue
+        for nbr in atom.bonds.keys():
+            ib = atom_to_idx[nbr]
+            if symbols[ib] == 'H':
+                continue
+            d = float(np.linalg.norm(coords[ia] - coords[ib]))
+            if d < 0.9:
```
calvinp0: Is there also a situation where it can be wildly above?

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+                             ref_pos: np.ndarray,
+                             fallback_perp: np.ndarray,
+                             ) -> np.ndarray | None:
+    """
+    Place a point at the intersection of two spheres around two centers.
+
+    Given two spheres of radii ``r1`` (around ``center1``) and ``r2``
+    (around ``center2``), this returns the intersection point that lies
+    on the same side of the inter-center axis as ``ref_pos``. When the
+    spheres do not overlap, the point is placed at distance ``r1`` from
```
calvinp0: Maybe its just worth mentionig in the doctstring but if there is a case when th circle is inside the other circle completely. I guess it depends if you consider that as a 'no-intersection' and if that situation is every possible.

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+        if nbr_idx == o_other_side:
+            continue
+        if symbols[nbr_idx] != 'O':
+            continue
+        h_on_o = [k for k in adj[nbr_idx] if symbols[k] == 'H']
+        if h_on_o:
+            o_hydroxyl = nbr_idx
+            break
+
+    _BV_CO_SHORTEN_TARGET = 1.28  # Å, calibrated from DFT TS (1.279)
```
calvinp0: Module level

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+        mig_candidates.append(nbr_idx)
+    if not mig_candidates:
+        return None
+
+    # Calibrated TS target distances (from DFT-curated BV step-2 reference).
+    sbl_oo = get_single_bond_length('O', 'O') or 1.48
+    sbl_cc = get_single_bond_length('C', 'C') or 1.54
+    sbl_co = get_single_bond_length('C', 'O') or 1.43
+
+    _BV_OO_STRETCH = 0.27    # Å per side (total ≈ 2 × 0.27 above sbl)
```
calvinp0: Module level

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+    _BV_CC_STRETCH = 0.76    # Å above sbl for migrating-group departure
```
calvinp0: Module level

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+    _BV_CO_STRETCH = 0.73    # Å above sbl(C,O) for migrating-group approach
```
calvinp0: Module level

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+        if o_hydroxyl is not None:
+            frag_oh = bfs_fragment(adj, o_hydroxyl, block={c_parent})
+            coords = _set_bond_distance(
+                coords, fixed=c_parent, mobile=o_hydroxyl,
+                target_dist=_BV_CO_SHORTEN_TARGET,
+                mobile_frag=frag_oh)
+
+        # 4d. Concerted H migration from O_hydroxyl to the carbonyl O
+        # (~1.40 Å from each, calibrated).
+        _BV_H_TRANSFER_TARGET = 1.40
```
calvinp0: Module level

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+                key = (min(c_carbonyl, nbr_idx), max(c_carbonyl, nbr_idx))
+                if bond_orders.get(key, 1.0) >= 1.5:
+                    o_carbonyl_dbl = nbr_idx
+                    break
+            h_on_oh = [k for k in adj[o_hydroxyl] if symbols[k] == 'H']
+            if o_carbonyl_dbl is not None and h_on_oh:
+                h_idx = h_on_oh[0]
+                p1 = coords[o_hydroxyl]
+                p2 = coords[o_carbonyl_dbl]
+                ax = p2 - p1
```
calvinp0: Is this not something already kinda done in _two_sphere_intersection function?

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+                                    mobile=o_other_side,
+                                    target_dist=d_oo_target,
+                                    mobile_frag=frag_other)
+
+        # 4b. Triangulate C_migrating between C_parent and O_other simultaneously
+        # (a sequential stretch-then-contract would undo itself).
+        frag_mig = bfs_fragment(adj, c_migrating, block={c_parent})
+        p1 = coords[c_parent]
+        p2 = coords[o_other_side]
+        axis = p2 - p1
```
calvinp0: Is this not something similar in _two_sphere_intersection?

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+
+    _BV_CO_SHORTEN_TARGET = 1.28  # Å, calibrated from DFT TS (1.279)
+
+    best_ts: dict | None = None
+    best_score = float('inf')
+
+    for c_migrating in mig_candidates:
+        coords = np.array(uni_xyz['coords'], dtype=float).copy()
+
+        # 4a. Stretch the O-O peroxide bond.
```
calvinp0: It's under the comment of step 5 but then here its 4a (then 4b etc.)

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+                  'isotopes': uni_xyz.get('isotopes', tuple(0 for _ in range(n_atoms))),
+                  'coords': tuple(tuple(float(c) for c in row) for row in coords)}
+        if colliding_atoms(ts_xyz):
+            continue
+
+        # Score = sum of squared deviations from calibrated targets.
+        d_oo = float(np.linalg.norm(coords[oo_bond[0]] - coords[oo_bond[1]]))
+        d_cc = float(np.linalg.norm(coords[c_parent] - coords[c_migrating]))
+        d_co = float(np.linalg.norm(coords[c_migrating] - coords[o_other_side]))
+        score = ((d_oo - d_oo_target) ** 2
```
calvinp0, line 725: if the function also does C_parent–O_hydroxyl shortening and H transfer, the score ignores those by the looks of it. Maybe do add optionl score terms when those atoms exist?
Cause if two candidates tie on the 3 main distances, one could, I guess, have a muc better or worse H migration geom

arc/job/adapters/ts/linear_utils/families.py
```diff
@@ -0,0 +1,1596 @@
...
+        ia = atom_to_idx[atom]
+        for nbr, bond in atom.bonds.items():
+            ib = atom_to_idx[nbr]
+            key = (min(ia, ib), max(ia, ib))
+            bond_orders[key] = bond.order
+
+    # Step 1: Find the O-O peroxide bond.
+    oo_bond = None
+    for a, b in split_bonds:
+        if symbols[a] == 'O' and symbols[b] == 'O':
```
calvinp0: will there ever be a chance when a/b can be not valid indexes?

arc/family/family.py
```diff
@@ -40,10 +41,39 @@ def get_rmg_db_subpath(*parts: str, must_exist: bool = False) -> str:
...
+        return tuple(f.readlines())
+
+
 class ReactionFamily(object):
     """
     A class for representing a reaction family.
 
+    Instances are cached per ``(label, consider_arc_families)`` so the
+    family ``groups.py`` file is read and parsed at most once per process.
+    The cached object is treated as immutable; do not mutate its public
```
calvinp0: Just to double check, are any of `self.reactants`, `self.entries`, or `self.actions` ever mutated downstream? If so, then the cache can leak state between calls

arc/family/family.py
```diff
@@ -40,10 +41,39 @@ def get_rmg_db_subpath(*parts: str, must_exist: bool = False) -> str:
     return candidates[0]
 
 
+@functools.lru_cache(maxsize=None)
+def _read_groups_file_lines(label: str, consider_arc_families: bool) -> tuple[str, ...]:
```
calvinp0: So this feeds into `get_groups_file_as_lines` right, which method says def get_groups_file_as_lines(...) -> list[str]:

calvinp0: Also since we have this function, it seems redundant to then have a self function inside the class thats sole purpose is to call this function? But okay with leaving it as is, just thought to point it out

arc/family/family.py
```diff
@@ -40,10 +41,39 @@ def get_rmg_db_subpath(*parts: str, must_exist: bool = False) -> str:
...
+    if not os.path.isfile(groups_path):
+        if consider_arc_families:
+            groups_path = os.path.join(ARC_FAMILIES_PATH, f'{label}.py')
+        if not os.path.isfile(groups_path):
+            raise FileNotFoundError(f'Could not find the groups file for family {label}')
+    with open(groups_path, 'r') as f:
+        return tuple(f.readlines())
+
+
 class ReactionFamily(object):
```
calvinp0, line 68: Since ReactionFamily is now cached, there is a line that appears in nested loops
```
Group().from_adjacency_list(get_group_adjlist(...))
```

We should consider caching it too?

```
self.groups_by_label = {
    label: Group().from_adjacency_list(adjlist)
    for label, adjlist in self.entries.items()
}
```

calvinp0:
Not in this PR, but it does touch this file significantly `family.py`
https://github.com/ReactionMechanismGenerator/ARC/blob/58770524bab03e3cfe649875ec5b2a37a91c6968/arc/family/family.py#L713
I believe that two lines above, we just initialised `rmg_families, arc_families = list(), list()`, so it is always `False`

.github/workflows/ci.yml
```diff
@@ -23,21 +23,21 @@ jobs:
...
       - name: Clean Ubuntu Image
-        uses: jlumbroso/free-disk-space@main
+        uses: jlumbroso/free-disk-space@v1.3.1
         with:
           tool-cache: true
           android: true
           dotnet: true
           haskell: true
           large-packages: true
-          swap-storage: true
```
calvinp0, line 33: Were we getting OOM when this was true?

arc/job/adapters/ts/linear_utils/families.py
alongd: In principle, yes. Currently we catch collisions, I'll also add a detection for very large distances. Thanks!

arc/job/adapters/ts/linear_utils/families.py
alongd: I agree. Currently it fails if one circle is within the other. I'll add a fallback to collinear placement

arc/job/adapters/ts/linear_utils/families.py
alongd: you're right, the math is duplicated. also, I see it in other families.py locations and in local_geometry, addition, and postprocess. I'll consolidate

arc/job/adapters/ts/linear_utils/families.py
alongd: the concern is legit, but the impact might be nuanced in my opinion.
I added two optional terms to the score, gated on `o_hydroxyl is not None` and `o_carbonyl_dbl is not None`

arc/job/adapters/ts/linear_utils/families.py
alongd: Thanks, I appreciate the question. Here, `a` and `b` are always valid indices
Of course, if we have nonsense input the function could break, but this is also true for other params in this function, and to many others. I don't think we should guard against all possibilities, the current state is realistic, I'll leave as-is

arc/family/family.py
alongd: no, there's no modification downstream, so no leak risk in the current code. But we're not guarded against future code changes that might introduce mutations. If we want, we could make these tuples instead of lists. I'll leave it as is, I think the risk is low

arc/family/family.py
alongd: Thanks, I'll make the helper public and drop the wrapper

arc/family/family.py
alongd: Good catch! applied

alongd:
@calvinp0, comments addressed, ready for the next round. Thanks!

arc/job/adapters/ts/linear_utils/math_zmat.py
```diff
@@ -0,0 +1,388 @@
...
+        r_e = r_ee
+        p_e = p_ee
+
+    delta_e = sum(p_e) - sum(r_e)
+    if abs(delta_e) < 1e-3:
+        return 0.5
+
+    if reorg_energy is None:
+        lam_endo = delta_e_sat / (2.0 * (w_max - 0.5)) if (w_max - 0.5) > 0 else float('inf')
+        lam_exo  = delta_e_sat / (2.0 * (0.5 - w_min)) if (0.5 - w_min) > 0 else float('inf')
```
calvinp0, line 323: double spacing for `lam_exo  =`

arc/job/adapters/ts/linear_utils/math_zmat.py
```diff
@@ -0,0 +1,388 @@
...
+    constraints = list()
+    atom_occurrences: dict[int, int] = dict()
+    for bond in expected_breaking_bonds + expected_forming_bonds:
+        for atom in bond:
+            atom_occurrences[atom] = atom_occurrences.get(atom, 0) + 1
+    atoms_sorted_by_frequency = [k for k, _ in sorted(atom_occurrences.items(),
+                                                       key=lambda item: (-item[1], item[0]))]
+    for i, atom in enumerate(atoms_sorted_by_frequency):
+        for bond in expected_breaking_bonds + expected_forming_bonds:
+            if atom in bond and all(a not in atoms_sorted_by_frequency[:i] for a in bond):
```
calvinp0, line 127: I want to double-check the intended behavior here. `get_r_constraints()` does not include all changing bonds. For each atom, it picks the first eligible bond and then `break`s. Also, because of the `all(a not in atoms_sorted_by_frequency[:i] for a in bond)` condition, bonds involving an earlier higher-frequency atom are skipped for later atoms.

So for changing bonds like `[(1, 2), (1, 3)]`, this would only keep one constraint involving atom 1, not both. Is the goal to produce a minimal canonical subset of R constraints, or should every expected forming/breaking bond appear in `R_atom`?

arc/job/adapters/ts/linear_utils/path_spec.py
```diff
@@ -0,0 +1,1461 @@
...
+def has_recipe_channel_mismatch(path_spec: ReactionPathSpec,
+                                xyz: dict,
+                                ) -> tuple[bool, str]:
+    """
+    Return ``(True, reason)`` if *xyz* violates the recipe channel.
+
+    Conservative checks:
+
+    1. **Failed to break** — any bond in ``path_spec.breaking_bonds`` whose
+       distance is below 1.30 Å.
```
calvinp0: Is this true for all bonds? Or would we possible scale it using the single bond length as the base

arc/job/adapters/ts/linear_utils/isomerization.py
```diff
@@ -0,0 +1,1248 @@
...
+    if len(migrants) != 2:
+        return None
+
+    c1, c2 = centers
+    m1, m2 = migrants
+    symbols = r_xyz['symbols']
+    coords = np.array(r_xyz['coords'], dtype=float)
+    atom_to_idx = atom_index_map(r_mol)
+
+    scale = 2.0 * (1.0 - weight)
```
calvinp0, line 1083: I just want to double check, so if weight is 1.0 then scale is 0 and then we expect the PAULING_DELTA to have no impact?

arc/job/adapters/ts/linear_utils/local_geometry.py
```diff
@@ -0,0 +1,1157 @@
...
+        return xyz
+    return xyz_with_coords(xyz, coords)
+
+
+# ---------------------------------------------------------------------------
+# Helper 3 — migrating-H local cleanup (donor–acceptor triangulation)
+# ---------------------------------------------------------------------------
+
+def clean_migrating_h(xyz: dict,
+                      mol: Molecule,
```
calvinp0: is `mol` used in this func? I don't see it being called

arc/job/adapters/ts/linear_utils/postprocess.py
```diff
@@ -0,0 +1,2181 @@
...
+            vec = coords[b] - coords[a]
+            direction = vec / d_cur
+            half_push = (d_target - d_cur) * 0.5
+            coords[a] -= direction * half_push
+            coords[b] += direction * half_push
+
+    return xyz_with_coords(xyz, coords)
+
+
+def orient_h_on_reactive_centers(xyz: dict,
```
calvinp0, line 184: Is this not kinda the same as https://github.com/ReactionMechanismGenerator/ARC/blob/ea9392510f2b1a14508b4716eb9a043cccba809b/arc/job/adapters/ts/linear_utils/local_geometry.py#L187

I think a lot of funcs here are similar in local_geometry.py
Is there a way to consolidate these functions in a file?

arc/job/adapters/ts/linear_utils/postprocess.py
```diff
@@ -0,0 +1,2181 @@
...
+        if symbols[a] == 'H' or symbols[b] == 'H':
+            continue
+        d_cur = float(np.linalg.norm(coords[a] - coords[b]))
+        if d_cur < 1e-6:
+            continue
+        sbl = get_single_bond_length(symbols[a], symbols[b]) or 1.5
+        d_target = sbl + PAULING_DELTA
+        # Only act when interpolation failed to stretch at all; do not touch
+        # already-stretched bonds from well-interpolated TSs.
+        if d_cur < sbl:
```
calvinp0, line 159: adjust_reactive_bond_distances() docstring says breaking heavy-atom bonds shorter than their TS estimate are stretched, but the code only stretches when the current distance is shorter than the single-bond length
The inline comment seems to explain this, but maybe should be in the docstring too

arc/job/adapters/ts/linear.py
```diff
@@ -0,0 +1,4030 @@
...
+        """
+        self.execute_incore()
+
+
+# ---------------------------------------------------------------------------
+# Top-level dispatchers
+# ---------------------------------------------------------------------------
+
+def _frag_composition_key(uni_mol: Molecule,
+                          atom_to_idx: dict,
```
calvinp0: I don't see `atom_to_idx` used in this helper

arc/job/adapters/ts/linear.py
```diff
@@ -0,0 +1,4030 @@
...
+                h_idx_loc = int(mig_rec['h_idx'])
+                d_loc = int(mig_rec['donor'])
+                a_loc = int(mig_rec['acceptor'])
+                ts_xyz = orient_h_away_from_axis(
+                    ts_xyz, uni_mol, d_loc, a_loc, exclude_h={h_idx_loc})
+                frag_centers.add(d_loc)
+                frag_centers.add(a_loc)
+                frag_exempt_hs.add(h_idx_loc)
+            if frag_centers:
+                ts_xyz = apply_reactive_center_cleanup(
```
calvinp0: Refer to prev comment

arc/job/adapters/ts/linear.py
```diff
@@ -0,0 +1,4030 @@
...
+                    # its triangulated position is preserved.
+                    template_centers: set[int] = set()
+                    template_exempt_hs: set[int] = set()
+                    for mig_rec in migration_records:
+                        template_centers.add(int(mig_rec['donor']))
+                        template_centers.add(int(mig_rec['acceptor']))
+                        template_exempt_hs.add(int(mig_rec['h_idx']))
+                    if template_centers:
+                        for mig_rec in migration_records:
+                            ts_xyz = orient_h_away_from_axis(
```
calvinp0: This post-migration cleanup pattern appears in both here and then a roughl 400-500 lines down

In both cases, the migrating H has already been placed upstream by `migrate_verified_atoms()` / `migrate_h_between_fragments()`, so we intentionally call:

```python
apply_reactive_center_cleanup(..., migrations=None, exempt_h_indices=...)
```
to avoid calling clean_migrating_h() a second time.

Could we wrap this sequence into a helper, maybe _cleanup_after_existing_h_migration(...)? That would make the “migrating H is already placed; only orient/regularize/symmetrise the donor-acceptor shells” behaviour explicit and avoid the two copies drifting apart later.

arc/species/converter.py
```diff
@@ -2470,3 +2470,67 @@ def kabsch(xyz1: dict, xyz2: dict) -> float:
     coords1, coords2 = np.array(xyz1['coords']), np.array(xyz2['coords'])
     _, score = Rotation.align_vectors(coords1, coords2)
     return score
+
+
+def order_xyz_by_atom_map(xyz: dict,
```
calvinp0, line 2475: See comment  on order_mol_by_atom_map()

arc/species/converter.py
```diff
@@ -2470,3 +2470,67 @@ def kabsch(xyz1: dict, xyz2: dict) -> float:
...
+    Returns:
+        dict: The ordered xyz coordinates.
+    """
+    symbols = [xyz['symbols'][i] for i in atom_map]
+    isotopes = [xyz['isotopes'][i] for i in atom_map] if 'isotopes' in xyz else None
+    coords = [xyz['coords'][i] for i in atom_map]
+    return xyz_from_data(coords=coords, symbols=symbols, isotopes=isotopes)
+
+
+def order_mol_by_atom_map(mol: Molecule,
```
calvinp0, line 2520: The two implementations seem internally consistent, but I think the docstring describes the opposite map direction.

The code uses `atom_map` as `new[i] = old[atom_map[i]]`, so `atom_map[new_i] = old_i`. But the docstring says `atom_map[reactant_i] = product_i`, which sounds like old -> new.

Could we clarify the docstring or invert the map if old -> new is the intended convention?

arc/species/converter.py
```diff
@@ -2470,3 +2470,67 @@ def kabsch(xyz1: dict, xyz2: dict) -> float:
...
+
+    Raises:
+        ValueError: If ``atom_map`` length does not match the number of atoms in ``mol``,
+            or if any mapped index is out of range.
+
+    Returns:
+        Molecule: A deep copy of ``mol`` with atoms reordered according to ``atom_map``.
+    """
+    n = len(mol.atoms)
+    if len(atom_map) != n:
```
calvinp0, line 2554: I think also we should check duplication.

So, for example, [0, 0, 1] passes range/length checks and silently duplicates one atom while dropping another.

arc/species/species.py
calvinp0, line 38: Is there a way we can avoid this cyclic import. I want to avoid more cyclic check failures from CodeQL

arc/species/zmat.py
```diff
@@ -1324,6 +1353,152 @@ def get_connectivity(mol: Molecule) -> dict[int, list[int]]:
     return connectivity
 
 
+def find_smart_anchors(mol: Molecule,
```
calvinp0, line 1356: This helper is mostly clear, but the docstring promises “1–3 unique, contiguous atom indices.” The code only excludes the immediately previous atom when picking the next anchor. Could we make that invariant explicit by tracking `used = {anchor_0, anchor_1}` when selecting `anchor_2`, or by validating before return?

arc/species/zmat.py
```diff
@@ -1324,6 +1353,152 @@ def get_connectivity(mol: Molecule) -> dict[int, list[int]]:
...
+
+    # Step 1: define the reactive core.
+    reactive_core: set = set()
+    for bond in breaking_bond_set | forming_bond_set:
+        reactive_core.update(bond)
+
+    spectator_atoms: set = set(range(num_atoms)) - reactive_core
+    connectivity: dict[int, list[int]] = get_connectivity(mol)
+
+    def _atomic_number(atom_index: int) -> int:
```
calvinp0, line 1422: I see we have nested helper functions here. I will leave it up to you if you want to keep them inside as they are relatively small so not too bad to notice and follow

arc/species/zmat.py
```diff
@@ -2338,3 +2513,80 @@ def map_index_to_int(index: int | str) -> int:
...
+    """
+    zmat_out = {
+        'symbols': zmat['symbols'],
+        'coords': zmat['coords'],
+        'vars': dict(zmat['vars']),
+        'map': zmat['map'],
+    }
+
+    coords = xyz['coords'] if isinstance(xyz, dict) and 'coords' in xyz else xyz
+    n_atoms_xyz = len(coords)
```
calvinp0, line 2573: `update_zmat_by_xyz()` normalizes `coords` from `xyz`, but then passes `xyz` rather than `coords` into `vectors.calculate_param()`.

arc/common_test.py
```diff
@@ -30,11 +30,33 @@ class TestCommon(unittest.TestCase):
...
+        a previous run cannot leave the next run inheriting the prior
+        ``restart_paths_globalized.yml``.
+        """
+        globalized_restart_path = os.path.join(
+            common.ARC_TESTING_PATH, 'restart', '4_globalized_paths',
+            'restart_paths_globalized.yml')
+        if os.path.isfile(globalized_restart_path):
+            try:
+                os.remove(path=globalized_restart_path)
+            except OSError:
```
calvinp0: Should probably log something here if we hit the except statement

functional/functional_test.py
```diff
@@ -22,14 +22,37 @@ class TestFunctional(unittest.TestCase):
...
+        """
+        for folder in ['thermo', 'kinetic']:
+            shutil.rmtree(os.path.join(ARC_PATH, 'functional', 'test', folder),
+                          ignore_errors=True)
+        for stray in ('nul', 'run.out'):
+            stray_path = os.path.join(ARC_PATH, 'functional', stray)
+            if os.path.isfile(stray_path):
+                try:
+                    os.remove(stray_path)
+                except OSError:
```
calvinp0: Ditto here for the except

devtools/install_rmg.sh
```diff
@@ -9,34 +9,31 @@ echo ">>> Starting RMG-Py installer..."
...
     case "$1" in
         -p|--pip)    MODE=source; SOURCE_MODE=pip; shift ;;
         -c|--conda)  MODE=source; SOURCE_MODE=conda; shift ;;
-        -r|--rms)    INSTALL_RMS=true; shift ;;
         -s|--ssh)    USE_SSH=true; shift ;;
         -S|--source) MODE=source; shift ;;
         -h|--help)
 cat <<EOF
-Usage: $0 [--source] [--pip] [--conda] [--rms] [--ssh]
+Usage: $0 [--source] [--pip] [--conda] [--ssh]
```
calvinp0, line 27: Since we have removed RMG, we should mention in the docs for user to search out the proper installation of RMG

arc/job/adapters/ts/linear_utils/math_zmat.py
alongd: this is "on purpose" for comparability of the two lines, so it's easier to see `(w_max - 0.5)` directly above `(0.5 - w_min)`

.github/workflows/ci.yml
alongd: We weren't getting OOMs, but moving from 2 to 6 workers in CI risked OOM under peak load

arc/job/adapters/ts/linear_utils/math_zmat.py
alongd: Intentional, it's a minimal canonical R-constraint set: one constraint per reactive atom, not one per bond. An atom appearing in multiple bonds (e.g. a migrating H in both a break and a form) contributes a single R, which avoids over-constraining the optimizer. Added a docstring note to clarify.

arc/job/adapters/ts/linear_utils/path_spec.py
alongd: Thanks, I agree it'll be more chemically correct to scale it. Adding a scale using `common.get_single_bond_length()`

arc/job/adapters/ts/linear_utils/isomerization.py
alongd: correct. weight=1.0 means "geometry is fully product-like"

arc/job/adapters/ts/linear_utils/local_geometry.py
alongd: removing `mol` and `weight`. thanks

arc/job/adapters/ts/linear_utils/postprocess.py
alongd: these two have different underlying logic. I'll check if there are additional funcs to consolidate (already did a pass in the last iteration, I'll check this file specifically now)

arc/job/adapters/ts/linear_utils/postprocess.py
alongd: I agree, fixed

arc/job/adapters/ts/linear.py
alongd: fixed

arc/job/adapters/ts/linear.py
alongd: agree, consolidating

arc/species/converter.py
alongd: I agree. we check it now with `if len(set(atom_map)) != n`

arc/species/species.py
alongd: converter only imports from species.py within `TYPE_CHECKING`. CodeQL thinks it's a real import. There's no real issue. We can think of options to change the code to silence CodeQL, but I think this is a CodeQL issue, they need to recognize that there's no cyclic import in such cases

arc/species/zmat.py
alongd: Done. `_pick_next` now takes `used: set[int]` instead of `previous_atom: int`, so every already-chosen anchor is passed in and excluded by construction (rather than relying on graph properties + argument ordering)

arc/species/zmat.py
alongd: `calculate_param` accepts both forms, but it does seem inconsistent. changes now to `coords`

arc/job/adapters/ts/linear_utils/addition.py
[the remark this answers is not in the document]
alongd: agree, fixed as suggested

arc/species/converter.py
alongd: I extended the docstring to explain this better

LeenFahoum (commented):
I checked out the branch and tested the Linear TS Adapter using the showcase notebook with several unimolecular reactions. Overall, this is a very impressive feature, and the generated TS guesses worked well for most of the cases I tested.
I noticed one problematic constrained ring/puckering case in the Intra Diels-Alder monocyclic family:
`C=C1C=CC=C1 ⇌ C1=CC2CC2=C1`
I sent Alon a few related isomerization cases offline to help test and improve this behavior. My general impression is that the adapter works well overall, with this specific constrained ring case needing some additional attention.

alongd:
Thanks, @LeenFahoum!
I fixed this family, please take a look
image

alongd:
@calvinp0 Thanks for reviewing
@LeenFahoum, thanks for playing around and testing this branch

Please let me know if you have additional comments, and let's advance toward merging if possible

LeenFahoum:
Thanks, Alon.
I don’t have any additional comments from my side. I tested the branch, and the results looked correct to me. The adapter seems to perform well for generating TSGs.

LeenFahoum (dismissed):

LeenFahoum (approved):

PR #696, opened 2023-08-28
Make sure to update, upgrate and autoremove before latex installation

kfir4444:
This is due to the Failed to fetch error, which (according to [this](https://askubuntu.com/questions/1241009/failed-to-fetch-installation-errors)) is caused by outdated repository information.


alongd (approved):
Thanks!

PR #701, opened 2023-09-21
Fixed: allow mutiple scissions in the atom mapping process.

kfir4444:
This PR presents a new method of performing the scission in atom mapping, based on the atom indices, which are required for atom mapping anyway. This method also simplify the process.
Tests were also added.


arc/reaction_test.py
```diff
@@ -35,7 +36,15 @@ def setUpClass(cls):
         """
         cls.maxDiff = None
         cls.rmgdb = rmgdb.make_rmg_database_object()
-        rmgdb.load_families_only(cls.rmgdb)
+        cls.rmgdb.load(path=rmg_settings['database.directory'],
```
alongd, line 39: We should try to load through ARC's rmgdb module, so we guarantee that during an actual run ARC's also able to load all families, not only during the tests.

arc/reaction_test.py
```diff
@@ -1654,6 +1654,23 @@ def test_get_atom_map(self):
                 self.assertIn(atom_map[int(atom.label)], c_symmetry_h_2 if atom.symbol == "C" else h_symmetry2)
         self.assertTrue(check_atom_map(rxn=rxn))
 
+        # XY_elimination_hydroxyl
```
alongd, line 1657: I have another PR that organizes these tests. We should separate this large test into smaller unit tests, so we know which ones fail, and we can run them individually when debugging.
Please put this added test in a new unit test function

arc/mapping/engine.py
```diff
@@ -1219,3 +1219,39 @@ def cuts_on_cycle_of_labeled_mol(spc: 'ARCSpecies')-> bool:
             if bde[0]-1 in cycle and bde[1]-1 in cycle:
                 return True
     return False
+
+
+def determine_bdes_indices_based_on_atom_labels(spc, bde):
+    """
+    A function for determining whether or not the species in question containt the bond specified by the BDE. Also, assigns the BDE if the user requests.
```
alongd, line 1226: Since BDE really stands for bond dissociation energy, and we have taken it out of context and don't use it for energy anymore, how about we rename the variables to somethign else, e.g., BDI (bond dissociation indices)?

arc/mapping/engine.py
```diff
@@ -1219,3 +1219,39 @@ def cuts_on_cycle_of_labeled_mol(spc: 'ARCSpecies')-> bool:
...
+    Args:
+        spc (ARCSpecies): The species in question, with labels atom indices.
+        bde (Tuple(int, int)): The bde in question.
+        add_bdes (bool): Whether or not to add the bde to the species.
+    Returns:
+        bool: Whether or not the bde is based on the atom labels.
```
alongd, line 1232: From the function name I understand that it should determine the indices, but from this returns description I understand that it just returns True or False. Should we rename the function to reflect what it does?

arc/mapping/engine.py
```diff
@@ -1219,3 +1219,39 @@ def cuts_on_cycle_of_labeled_mol(spc: 'ARCSpecies')-> bool:
...
+    Examples:
+        >>> spc = ARCSpecies(label="ethane", smiles="CC")
```
alongd, line 1234: the example isn't very clear. Especially the second case. Is `bde` determined after it is passed to the function?
and in general, how do we know the atom indices (in 3D) from the SMILES?

arc/mapping/engine.py
```diff
@@ -1219,3 +1219,39 @@ def cuts_on_cycle_of_labeled_mol(spc: 'ARCSpecies')-> bool:
...
+        >>> bde = (1, 2)
+        >>> label_species_atoms([spc])
+        >>> determine_bdes_indices_based_on_atom_labels(spc, bde)
+        True
+        >>> determine_bdes_indices_based_on_atom_labels(spc, bde, add_bdes=True)
+        >>> bde = (2, 3)
+        False
+    """
+    index1, index2 = bde[0]-1, bde[1]-1
```
alongd, line 1243: we have a function in common that translates 1-indexed lists to 0-indexed lists, let's use it here

arc/mapping/engine.py
```diff
@@ -1219,3 +1219,39 @@ def cuts_on_cycle_of_labeled_mol(spc: 'ARCSpecies')-> bool:
...
+    new_bde = list()
+    atoms = list()
+    for index, atom in enumerate(spc.mol.atoms):
+        if atom.label == str(index1) or atom.label ==  str(index2):
```
alongd, line 1247: minor: extra white space before `str(index2)`

arc/mapping/engine.py
```diff
@@ -1219,3 +1219,39 @@ def cuts_on_cycle_of_labeled_mol(spc: 'ARCSpecies')-> bool:
...
+            new_bde.append(index+1)
+            atoms.append(atom)
+        if len(new_bde) == 2:
+            break
+    
+    if len(new_bde) == 2:
+        spc.bdes = [tuple(new_bde)]
+        return atoms[1] in atoms[0].bonds.keys()
```
alongd, line 1255: If this live evaluates to `False`, then we have already modified the species BDEs. Is this a desired/expected behavior?

arc/mapping/engine.py
```diff
@@ -1255,3 +1255,39 @@ def determine_bdes_indices_based_on_atom_labels(spc, bde):
         return atoms[1] in atoms[0].bonds.keys()
     else:
         return False
+
+
+def cut_species_based_on_atom_indices(species: List["ARCSpecies"], bdes: List[Tuple[int, int]]) -> List["ARCSpecies"]:
```
alongd, line 1260: No need for the quotation marks around `ARCSpecies`, it is [imported properly](https://github.com/ReactionMechanismGenerator/ARC/blob/main/arc/mapping/engine.py#L23) (unless changed here?)

arc/mapping/engine.py
```diff
@@ -1255,3 +1255,39 @@ def determine_bdes_indices_based_on_atom_labels(spc, bde):
...
+    Returns:
+        List[ARCSpecies]: The species after scission.
+    """
+    if not bdes:
+        return species
+    
+    for bde in bdes:
+        for index, spc in enumerate(species):
+            if determine_bdes_indices_based_on_atom_labels(spc, bde):
+                candidate = species.pop(index)
```
alongd, line 1275: index increases like `0, 1, 2, ...`
so first we pop species `0` from `[s0, s1, s2, ...]`, so we get `s0`, and the list becomes `[s1, s2, ...]`
then index increments to `1`, so now we pop `s2` from `[s1, s2, ...]`
Is this the intended behavior? Looks like we're skipping species

arc/mapping/engine.py
```diff
@@ -1255,3 +1255,39 @@ def determine_bdes_indices_based_on_atom_labels(spc, bde):
...
+
+
+def cut_species_based_on_atom_indices(species: List["ARCSpecies"], bdes: List[Tuple[int, int]]) -> List["ARCSpecies"]:
+    """
+    A function for scissoring species based on their atom indices.
+    Args:
+        species (List[ARCSpecies]): The species list that requires scission.
+        bdes (List[Tuple[int, int]]): A list of the atoms between which the bond should be scissored. The atoms are described using the atom labels, and not the actuall atom positions.
+    Returns:
+        List[ARCSpecies]: The species after scission.
```
alongd, line 1267: should be `Optional[List[ARCSpecies]]`, also above in the type hint

arc/mapping/engine.py
```diff
@@ -1255,3 +1255,39 @@ def determine_bdes_indices_based_on_atom_labels(spc, bde):
...
+    
+    for bde in bdes:
+        for index, spc in enumerate(species):
+            if determine_bdes_indices_based_on_atom_labels(spc, bde):
+                candidate = species.pop(index)
+                candidate.final_xyz = candidate.get_xyz()
+                if candidate.mol.copy(deep=True).smiles == "[H][H]":
+                    labels = [atom.label for atom in candidate.mol.copy(deep=True).atoms]
+                    try:
+                        H1 = candidate.scissors()[0]
```
alongd, line 1280: can you change to lowercase, e.g., `h1` (PEP8)

arc/mapping/engine.py
```diff
@@ -1255,3 +1255,39 @@ def determine_bdes_indices_based_on_atom_labels(spc, bde):
...
+                    except SpeciesError:
+                        return None
+                    H2 = H1.copy()
+                    H2.mol.atoms[0].label = labels[0] if H1.mol.atoms[0].label != labels[0] else labels[1]
+                    species += [H1, H2]
+                else:
+                    try:
+                        species += candidate.scissors()
```
alongd, line 1288: The docsting says we return the species after scission, but looks like we ass to the species list, and the original species is also returned (also above when treating H2). Is this on purpose? If so, need to update the docstring

arc/mapping/engine.py
```diff
@@ -1255,3 +1255,39 @@ def determine_bdes_indices_based_on_atom_labels(spc, bde):
...
+                    except SpeciesError as e:
+                        print(e)
```
alongd, line 1290: is it printing on purpose instead of logging?

arc/mapping/engine.py
```diff
@@ -1306,3 +1306,20 @@ def copy_species_list_for_mapping(species: List["ARCSpecies"]) -> List["ARCSpeci
         for atom1, atom2 in zip(copy.mol.atoms, spc.mol.atoms):
             atom1.label = atom2.label
     return copies
+
+
+def find_all_bdes(rxn: "ARCReaction", label_dict: dict, is_reactants: bool) -> List[Tuple[int, int]]:
```
alongd, line 1311: nice improvement!

arc/mapping/engine.py
```diff
@@ -1306,3 +1306,20 @@ def copy_species_list_for_mapping(species: List["ARCSpecies"]) -> List["ARCSpeci
...
+        rxn (ARCReaction): The reaction in question.
+        label_dict (dict): A dictionary of the atom indices to the atom labels.
+        is_reactants (bool): Whether or not the species list represents reactants or products.
+    Returns:
+        List[Tuple[int, int]]: A list of tuples of the form (atom_index1, atom_index2) for each broken bond. Note that these represent the atom indicies to be cut, and not final BDEs.
+    """
+    bdes = list()
+    for action in rxn.family.forward_recipe.actions:
+        if action[0].lower() == ("break_bond" if is_reactants else "form_bond"):
+            bdes.append((label_dict[action[1]]+1, label_dict[action[3]]+1))
```
alongd, line 1324: add white space before and after operators (PEP8)

alongd (commented):

Can you relocate the added tests to right after the respective commit that added the function being tested? This way it is easier to review a function and its related tests and see the relevance of the new tests

arc/reaction_test.py
kfir4444: There is an issue with that correctly. Looking for replaying on that [issue](https://github.com/ReactionMechanismGenerator/ARC/issues/702) so I could proceed with fixing it. This is a temporary patch.

arc/mapping/engine_test.py
```diff
@@ -1469,5 +1469,17 @@ def test_find_all_bdes(self):
         bdes = find_all_bdes(rxn, self.r_label_dict_rxn_1, True)
         self.assertEqual(bdes, [(2, 4)])
 
+    def test_copy_species_list_for_mapping(self):
+        """tests the copy_species_list_for_mapping function"""
+        species = [ARCSpecies(label="test_1", smiles = "BrC(F)Cl"), ARCSpecies(label="test_2", smiles = "OOC(F)CCCONNO")]
+        label_species_atoms(species)
+        species_copy = copy_species_list_for_mapping(species)
+        for s1, s2 in zip(species, species_copy):
```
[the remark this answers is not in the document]
kfir4444: This is the test itself. We go over each species in the copy, and verify that the species are not the same object (i.e., not saved in the same memory slot), but the species are isomorphic and the atoms are labeled the same.

arc/mapping/engine_test.py
```diff
@@ -1481,5 +1481,12 @@ def test_copy_species_list_for_mapping(self):
                 self.assertIsNot(atom1, atom2)
                 self.assertEqual(atom1.label, atom2.label)
 
+    def test_determine_bdes_indices_based_on_atom_labels(self):
+        """tests the determine_bdes_indices_based_on_atom_labels function"""
+        spc = ARCSpecies(label="ethane", smiles="CC")
+        label_species_atoms([spc])
+        self.assertTrue(determine_bdes_indices_based_on_atom_labels(spc, (1, 2)))
+        self.assertFalse(determine_bdes_indices_based_on_atom_labels(spc, (2, 3)))
```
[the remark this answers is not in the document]
kfir4444: No. If (starting from 1, due to bde convention) atom number 3 (hydrogen atom) was connected to atom number 2 (the second carbon atom) then this would have been true.

arc/mapping/engine.py
kfir4444: We do not generally know the atom order from the SMILES. However, in this case (2 heavy atoms, the rest are hydrogen atoms), we, in fact, do know the order, from convention. If you think it doesn't help and may confuse the users, I will remove the example all together.

arc/mapping/engine.py
kfir4444: It is the intended behavior.

arc/mapping/engine.py
kfir4444: The species are returned after all required scissoring is done. If none were requested, then no scissoring are done. It does return them after scission, so I do not understand the confusion.

arc/reaction_test.py
kfir4444: Could you point me to the PR?
Also, we should think about performing some of these randomly, they are getting out of hand in terms of computation time..

arc/mapping/engine.py
kfir4444: Tried it, it caused issues, returned to the original. Maybe we can try it in a separate PR, since it is true for many places in this file.

arc/mapping/engine_test.py
```diff
@@ -1463,5 +1463,11 @@ def test_add_adjacent_hydrogen_atoms_to_map_based_on_a_specific_torsion(self):
         for key in range(4):
             self.assertEqual(out_dict[key], key)
 
+    def test_find_all_bdes(self):
+        rxn = ARCReaction(r_species=[self.r_1, self.r_2], p_species=[self.p_1, self.p_2])
```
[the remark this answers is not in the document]
kfir4444: Yes. I added `.copy()` in the fixup to avoid changing the original object.

alongd:
I tried to map the following rxn on this branch
`C=CC([O])[CH]C=CCC <=> CCC=C[CH]CC=C[O]`

and got an error:
```
IndexError                                Traceback (most recent call last)
/tmp/ipykernel_225830/1280115450.py in <module>
----> 1 print(rxn1.atom_map)

~/Code/ARC/arc/reaction.py in atom_map(self)
    156                 and all(species.get_xyz(generate=False) is not None for species in self.r_species + self.p_species):
    157             for backend in ["ARC", "QCElemental"]:
--> 158                 _atom_map = map_reaction(rxn=self, backend=backend)
    159                 if _atom_map is not None:
    160                     self._atom_map = _atom_map

~/Code/ARC/arc/mapping/driver.py in map_reaction(rxn, backend, db, flip)
     71             return _map if _map is not None else map_reaction(rxn, backend=backend, db=db, flip=True)
     72         try:
---> 73             _map = map_rxn(rxn, backend=backend, db=db)
     74         except ValueError as e:
     75             return map_reaction(rxn, backend=backend, db=db, flip=True)

~/Code/ARC/arc/mapping/driver.py in map_rxn(rxn, backend, db)
    237     rmg_reactions = get_rmg_reactions_from_arc_reaction(arc_reaction=rxn, backend=backend)
    238     r_label_dict, p_label_dict = get_atom_indices_of_labeled_atoms_in_an_rmg_reaction(arc_reaction=rxn,
--> 239                                                                                       rmg_reaction=rmg_reactions[0])
    240 
    241     # step 2:

IndexError: list index out of range
```

Can you take a look?

alongd:
BTW, I think this is the correct mapping:
image

kfir4444:
> `C=CC([O])[CH]C=CCC <=> CCC=C[CH]CC=C[O]`

Did it previously work on the main branch? I think I know the issue, and I did not change the function that returns a wrong value.
Also, I this this issue probably stem from the fact that `get_rmg_reactions_from_arc_reaction(arc_reaction=rxn, backend=backend)` for this reaction returns an empty list.
Also, could you share the full script, and maybe also check that the reaction family is found correctly?

alongd (approved):
Thanks!

PR #708, opened 2023-10-23
Report electronic energy in an output file

JintaoWu98:
1. Add the "report_e_elect" as the argument/attribute, and method "save_e_elect" to decide whether electronic energy will be in the output or not, after the single-point energy calculation.
2. A test case is also added for this purpose to check if e_elect_summary.yml is created and have the electronic energy info


alongd (approved):
Looking good, thanks!

PR #709, opened 2023-10-25
Orca 5 sol

Lilachn91:
Added the option to write Orca v5 input files with solvation methods, plus fixed the error message when sp file is missing for rate calculations.


alongd (approved):
Thanks!

PR #711, opened 2023-11-12
Sella env fix

calvinp0:
Sella environment had issues installing due to the restriction of python being 3.7 which in turn cause supporting packages required for Sella not be installed. This then created issues for our testing


alongd (approved):
Thanks! Can you squash the 1st and last commits?

PR #713, opened 2023-11-23
Gauss Trsh Fixes

calvinp0:
@Lilachn91 pointed out that `nosymm` should not be part of the opt or scf combination of parameters but be its own parameter. This has been implemented. Furthermore, at times we end up creating an input file that may have `opt=([param1])` and `opt=([param2])` when rather they should be combined into one `opt=([param1,param2])`

### DIFF
--- a/arc/job/adapters/common.py
+++ b/arc/job/adapters/common.py
@@ -522,4 +522,15 @@ def combine_parameters(input_dict: dict, terms: list) -> Tuple[dict, List]:
                         value = re.sub(term, '', value)
             input_dict_copy[key] = value
 
+    # Parameters may appear as one word, so need to split them via comma
+    parameters = [param.split(',') for param in parameters]
+    # Flatten the list of lists
+    parameters = [item for sublist in parameters for item in sublist]
+    # Remove empty spaces from the beginning and end of each parameter
+    parameters = [param.strip() for param in parameters]
+    # Parameters may appear multiple times in the input_dict, so remove duplicates
+    parameters = list(set(parameters))
+    # Sort the list
+    parameters.sort()
+
     return input_dict_copy, parameters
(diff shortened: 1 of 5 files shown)


arc/job/adapters/common.py
```diff
@@ -522,4 +522,15 @@ def combine_parameters(input_dict: dict, terms: list) -> Tuple[dict, List]:
                         value = re.sub(term, '', value)
             input_dict_copy[key] = value
 
+    # Parameters may appear as one word, so need to split them via comma
```
alongd, line 525: Nice addition. Can you add some tests?

alongd (commented):
Looks good, thanks!
I'm missing a test generating a Gaussian input file with the nosymm keyword properly placed

calvinp0:
> Gaussian

https://github.com/ReactionMechanismGenerator/ARC/blob/e407531b7aa4e3b2a0571da4bd8f4d11b661285c/arc/job/adapters/gaussian_test.py#L525

I believe this covers the nosymm test with the correct placement

calvinp0:
> Looks good, thanks!
> 
> I'm missing a test generating a Gaussian input file with the nosymm keyword properly placed

https://github.com/ReactionMechanismGenerator/ARC/blob/e407531b7aa4e3b2a0571da4bd8f4d11b661285c/arc/job/adapters/gaussian_test.py#L525

@alongd 
I believe this covers the nosymm test with correct placement

arc/job/adapters/common.py
calvinp0: In the previous PR that was merged, we implemented tests for the function

https://github.com/ReactionMechanismGenerator/ARC/blob/26b200ba111fe603d2e7b890820f0ee6bed13258/arc/job/adapters/common_test.py#L172C1-L206C58

alongd (approved):
Thanks for the fix!

PR #715, opened 2023-11-26
Various TS check improvements

alongd:
This PR contains various commit from the infamous `ts2` branch.
Also, added an option for users to specify `skip_nmd: True` in the input file to skip the TS normal mode displacement check if it is problematic.


calvinp0 (approved):
LGTM!

JintaoWu98:
Looks great!

PR #716, opened 2023-11-26
Gauss trsh fix

calvinp0:
NDamp was misspelled (appears to have been for a while but never noticed)
`L502/L1002，Inaccurate quadrature in CalDSu.` - This error can appear has multiple ways to attempt to solve (but not always successful). Have attempted to provide guess=INDO now as an option if this error continues to occur.
https://github.com/liyuanhe211/Solution_for_Every_Gaussian_Error_Message#arrow_forward-l502l1002inaccurate-quadrature-in-caldsu-hourglass_flowing_sand
https://wongzit.github.io/gaussian-common-errors-and-solutions/#inaccurate-quadrature-in-caldsu


arc/job/trsh.py
```diff
@@ -904,8 +906,15 @@ def trsh_ess_job(label: str,
                 logger.info(f'Troubleshooting {job_type} job in {software} for {label} using more memory: {memory} GB '
                             f'instead of {memory_gb} GB')
                 ess_trsh_methods.append('memory')
-        # Log information
-        if logger_info:
+
+        if attempted_ess_trsh_methods:
+            if attempted_ess_trsh_methods == ess_trsh_methods:
+                logger.info(f'{logger_phrase} has already attempted all possible troubleshooting methods. Will not troubleshoot again.')
```
alongd, line 911: This will look like:
`Troubleshooting {job_type} job in {software} for {label} has already attempted...`
Maybe change to:
`{logger_phrase} wan unsuccessful. Already attempted these troubleshooting methods. Will not troubleshoot again.`
Or anything else that explains the situation.

arc/job/trsh.py
```diff
@@ -858,7 +858,8 @@ def trsh_ess_job(label: str,
         logger_info = []
         couldnt_trsh = True
         fine = False
-
+        if ess_trsh_methods:
```
alongd: I think that the first part of the commit message (`Incl. guess=INDO trsh`) doesn't imply on the content of this commit. Also, this commit has both code change and test change, which is OK, but maybe better to reorganize the commits and squash the test added here to the 1st commit in this PR, and rename this commit

arc/job/adapters/gaussian.py
```diff
@@ -364,8 +364,15 @@ def write_input_file(self) -> None:
             input_dict['job_type_1'] += f' SCRF=({self.level.solvation_method}, Solvent={self.level.solvent})'
 
         if self.species[0].number_of_atoms > 1:
-            input_dict['job_type_1'] += ' guess=read' if self.checkfile is not None and os.path.isfile(self.checkfile) \
-                else ' guess=mix'
+            if input_dict['job_type_1']:
+                input_dict['job_type_1'] += ' '
+            if 'guess=INDO' in input_dict['trsh']:
+                input_dict['job_type_1'] += 'guess=INDO'
+                # Remove guess=INDO from trsh
```
alongd, line 371: no need for this comment, the code is self-explanatory

arc/job/adapters/gaussian.py
```diff
@@ -364,8 +364,15 @@ def write_input_file(self) -> None:
             input_dict['job_type_1'] += f' SCRF=({self.level.solvation_method}, Solvent={self.level.solvent})'
 
         if self.species[0].number_of_atoms > 1:
-            input_dict['job_type_1'] += ' guess=read' if self.checkfile is not None and os.path.isfile(self.checkfile) \
-                else ' guess=mix'
+            if input_dict['job_type_1']:
+                input_dict['job_type_1'] += ' '
```
alongd, line 368: Why is this needed? Can we instead add the space to `' guess=INDO'` like we do below for `read` and `mix`?

arc/job/trsh.py
```diff
@@ -858,8 +858,7 @@ def trsh_ess_job(label: str,
         logger_info = []
         couldnt_trsh = True
         fine = False
-        if ess_trsh_methods:
-            attempted_ess_trsh_methods = ess_trsh_methods.copy()
+        attempted_ess_trsh_methods = ess_trsh_methods.copy() if ess_trsh_methods else None
```
alongd, line 861: Awesome. Please squash with the previous commit

alongd (commented):
Looking good!
I added mostly minor comments, also the commits should be organized a bit

arc/job/adapters/gaussian.py
calvinp0: Because I am not a fan of hard coding empty space if there is no requirement for it and maybe in input_dict['job_type_1'] there is an empty space already. If that makes sense

alongd (approved):
Thanks!

PR #718, opened 2023-12-05
Make Reaction atom_map a list of atom maps

alongd:


alongd:
Deprecated. @kfir4444, see if any commits from here are useful for you, if they are still relevant

PR #719, opened 2023-12-05
Queue change - PBS

calvinp0:
A new troubleshoot function that will recognise, on PBS, if the job failed due to exceeding the walltime. If this is the case, it will then first check to see if the user has provided multiple queue names and times to use. If so, it will then check if the provided times of these queues are sufficient. If not, it will then query PBS to determine if there are queues available that provide the user sufficient time


arc/job/trsh_test.py
```diff
@@ -28,253 +28,312 @@ def setUpClass(cls):
...
         status, keywords, error, line = trsh.determine_ess_status(
-            output_path=path, species_label='Zr2O4H', job_type='opt')
-        self.assertEqual(status, 'errored')
-        self.assertEqual(keywords, ['CheckFile'])
-        self.assertEqual(error, 'No data on chk file.')
-        self.assertIn('Error termination via Lnk1e', line)
-        self.assertIn('g09/l301.exe', line)
-
-        path = os.path.join(self.base_path['gaussian'], 'l301.out')
+            output_path=path, species_label="Zr2O4H", job_type="opt"
```
alongd, line 69: There are many style changes here of just changing `'` into `"`. Are they on purpose?

arc/job/trsh_test.py
```diff
@@ -634,20 +698,111 @@ def test_scan_quality_check(self):
...
 
-
-if __name__ == '__main__':
+    @patch(
+        "arc.common.get_logger"
+    )
+    @patch(
+        "arc.job.local.execute_command"
+    )
+    @patch(
```
alongd, line 725: Can you explain what this decorator does?

arc/job/local.py
```diff
@@ -256,7 +256,7 @@ def submit_job(path: str,
                        submit_filename=submit_filename,
                        recursion=True,
                        )
-        if cluster_soft.lower() == 'pbs' and  any('qsub: would exceed' in err_line for err_line in stderr):
+        if cluster_soft.lower() == 'pbs' and  (any('qsub: would exceed' in err_line for err_line in stderr ) or any('qsub: Maximum number of jobs' in err_line for err_line in stderr)):
```
alongd, line 259: Do we need to add a similar change in `ssh.py`?

arc/job/adapter.py
```diff
@@ -1313,6 +1338,27 @@ def troubleshoot_server(self):
...
+        """
+        queues, run_job = trsh_job_queue(job_name=self.job_name,
+                                            server=self.server,
+                                            max_time=self.max_job_time,
+                                            attempted_queues = self.attempted_queues,
+                                            )
+        
+        if queues is not None:
+            # We use self.max_job_time to determine which queues to troubleshoot.
+            filtered_queues = {queue_name: walltime for queue_name, walltime in queues.items() if convert_to_hours(walltime) >= self.max_job_time}
```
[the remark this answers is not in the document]
alongd: I think it's more intuitive for users to just give the walltime of a queue as an int in hours in settings.py
Do we plan on letting users add this in settings, or to only automatically read from the server?

alongd (commented):
Thanks for this addition! I added some comments, overall looks very good

arc/job/adapter.py
```diff
@@ -493,12 +493,23 @@ def write_submit_script(self) -> None:
             # If your server has different node architectures, implement something similar here.
             architecture = '\n#$ -l harpertown' if self.cpu_cores <= 8 else '\n#$ -l magnycours'
 
+        # Extract the 'queue' dictionary from servers[self.server], defaulting to an empty dictionary if 'queue' is not a key
+        settings_queues = servers[self.server].get('queue', {})
+
+        # Get the first item from the 'queue' dictionary if it is not empty, else default to None
```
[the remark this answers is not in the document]
calvinp0: The assumption is the user will define the queues dict with priority in mind.
I guess we could do `default_queue` but I do not particularly want to start creating more keys for the settings.py

arc/job/adapter.py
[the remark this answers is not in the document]
calvinp0: the server, at least for pbs, returns walltime in the format HH:MM:SS - so i figured that it would be easier to request the user to do the same if there were to define `queues` dict in the settings.py
and i have already added a comment to the settings.py to show the user what is expected of them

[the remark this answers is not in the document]
calvinp0: https://github.com/ReactionMechanismGenerator/ARC/blob/b0f23c8940460673abc2bacf2edb5a27f3551b05/arc/settings/settings.py#L61

arc/job/trsh_test.py
calvinp0: The way I understand it is that when the function is called, for example the `execute_command` function in arc.job.local, this patch will take 'take over it' and will replace the output with what the user has defined in the @patch

arc/job/adapter.py
```diff
@@ -1313,6 +1338,27 @@ def troubleshoot_server(self):
...
+                                            server=self.server,
+                                            max_time=self.max_job_time,
+                                            attempted_queues = self.attempted_queues,
+                                            )
+        
+        if queues is not None:
+            # We use self.max_job_time to determine which queues to troubleshoot.
+            filtered_queues = {queue_name: walltime for queue_name, walltime in queues.items() if convert_to_hours(walltime) >= self.max_job_time}
+
+            # Now we sort the queues by walltime, and choose the one with the longest walltime.
```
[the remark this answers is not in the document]
calvinp0: I have changed it so that it will work its way from shortest to longest

arc/job/adapter.py
```diff
@@ -493,21 +493,36 @@ def write_submit_script(self) -> None:
             # If your server has different node architectures, implement something similar here.
             architecture = '\n#$ -l harpertown' if self.cpu_cores <= 8 else '\n#$ -l magnycours'
 
+        # Extract the 'queue' dictionary from servers[self.server], defaulting to an empty dictionary if 'queue' is not a key
+        default_queue, _ = next(iter(servers[self.server].get('queues', {}).items()), (None, None))
```
alongd, line 497: Will the iterator remember where it was last time and calling `next` will actually give the 2nd, 3rd, and so on elements?
Looks to me like it'll always return the 1st element since `iter` is defined anew each time this function is being called

arc/job/local.py
alongd: Was this addressed?

arc/job/trsh.py
```diff
@@ -18,6 +18,7 @@
                         is_same_pivot,
                         is_same_sequence_sublist,
                         is_str_float,
+                        convert_to_hours
```
alongd, line 21: minor annoying comment: please sort these alphabetically

arc/job/trsh.py
```diff
@@ -1114,6 +1115,156 @@ def trsh_conformer_isomorphism(software: str,
...
+                for queue in excluded_queues:
+                    if queue in queues:
+                        queues.pop(queue, None)
+            if len(queues) == 0:
+                logger.error(f'Could not troubleshoot {job_name} on {server} as no queues were found.')
+                return None, False
+            else:
+                return queues, True
+    else:
+        logger.error(f'Could not troubleshoot {job_name} on {server} as server is not PBS.')
```
alongd, line 1266: maybe "since the server runs {cluster_soft} and not PBS"

arc/job/trsh_test.py
alongd: was this addressed?

alongd (commented):
Thanks! Only some minor comments

arc/job/adapter.py
calvinp0: I think I had an idea like you mentioned but then changed at somepoint. Now it seems it just grabs the `1st` in the queue as default queue and then if self.queue is not defined it will use the default queue

So essentially I am using this to grab the first queue in the settings.py and then if self.queue wasn't defined (as it wasnt troubleshooted) it will use the default queue

arc/job/local.py
calvinp0: I do not know. I haven't attempted to use the SSH part of ARC into Zeus (or a PBS server) so for now I will not implement it there. Also the SSH script needs updating so that will be done in another PR

arc/job/trsh_test.py
calvinp0: It's a style change that was done automatically I believe... I think one of the extensions I use (Trunk maybe?) decided to format the strings to a certain Python style

alongd (approved):
Thanks!

PR #721, opened 2023-12-12
Updated args['trsh'] if condition

calvinp0:
When a scan job was being troubleshooted, it was not removing the scan resolution troubleshoot correctly prior to writing the input gaussian file. This change in the code ensures that it will be removed correctly later in the code.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -784,7 +784,16 @@ class Scheduler(object):
             self.deduce_job_adapter(level=Level(repr=level_of_theory), job_type=job_type)
         args = {'keyword': {}, 'block': {}}
         if trsh:
-            args['trsh'] = {'trsh': trsh}
+            if isinstance(trsh, (str, list)):
+                args['trsh'] = {'trsh': trsh}
+            elif isinstance(trsh, dict) and 'trsh' in args:
+                for key, value in trsh.items():
+                    if isinstance(args['trsh'][key], list) and isinstance(value, list) and key in args['trsh']:
+                        args['trsh'][key].extend(value)
+                    else:
+                        args['trsh'][key] = value
+            else:
+                args['trsh'] = trsh
         if shift:
             args['shift'] = shift
         if scan_trsh:


arc/scheduler.py
```diff
@@ -784,7 +784,21 @@ def run_job(self,
             self.deduce_job_adapter(level=Level(repr=level_of_theory), job_type=job_type)
         args = {'keyword': {}, 'block': {}}
         if trsh:
-            args['trsh'] = {'trsh': trsh}
+            if isinstance(trsh, (str, list)):
+                args['trsh'] = {'trsh': trsh}
+            # if trsh is a dictionary
+            elif isinstance(trsh, dict):
```
alongd, line 790: Can you combine two consecutive `if`s (if no `else` is needed for just one of them)? E.g.:
```python
elif isinstance(trsh, dict) and 'trsh' in args:
```
Also, I think there's no need for the comment about the type, the code is self-explanatory

arc/scheduler.py
```diff
@@ -784,7 +784,21 @@ def run_job(self,
...
+                if 'trsh' in args:
+                    for key, value in trsh.items():
+                        if key in args['trsh']:
+                            if isinstance(args['trsh'][key], list) and isinstance(value, list):
+                                args['trsh'][key].extend(value)
+                            else:
+                                args['trsh'][key] = value
+                        else:
```
alongd, line 798: Both `else` lead to the same action, please consider combining the `if` statements and keep just one `else`

arc/scheduler.py
```diff
@@ -784,7 +784,21 @@ def run_job(self,
...
+                            args['trsh'][key] = value
+                else:
+                    args['trsh'] = trsh
```
alongd, line 801: Do we think this is OK, since `trsh` is a dict?

alongd (commented):
Thanks for the quick response for this issue.
LGTM. I added some comments, mostly on style.

arc/scheduler.py
calvinp0: Yes, it becomes a dict within a dict
For example
args['trsh]['trsh']
or
args['trsh]['scan_res']

alongd (approved):
Thanks for addressing this quickly!

PR #722, opened 2023-12-13
trsh logic change in run_job func call

calvinp0:
this is to ensure that the dictionary in args['trsh'] is passed through the function correctly and to stop ['', '', 'nosymm'] syntax error


arc/scheduler.py
```diff
@@ -1012,7 +1012,7 @@ def _run_a_job(self,
                      max_job_time=job.max_job_time,
                      rotor_index=job.rotor_index,
                      reactions=job.reactions,
-                     trsh=list(job.args['trsh'].values()) if 'trsh' in job.args else '',
+                     trsh=job.args['trsh'] if 'trsh' in job.args and 'trsh' else [],
```
alongd, line 1015: Take a look at ` and 'trsh'`, looks like it'll always evaluate to True

calvinp0: Thanks, fixed it.

alongd (approved):
LGTM

PR #723, opened 2023-12-14
Special chars creating submit issues

calvinp0:
issue was stemming from local.py, the var called cmd
cmd = f'cd {path}; {submit_cmd} {submit_filename}'

here we defined path but never put it in quotations so special chars like `[` and `]` would be misinterpreted by the shell

so in python would change to the direct of /rxn_120_CCN/ instead of /rxn_120_[CH2]CN/ (assuming out intention was to go to  /rxn_120_[CH2]CN/)

so just needed to change it to

cmd = f'cd "{path}"; {submit_cmd} {submit_filename}'


JintaoWu98:
LGTM!

JintaoWu98 (approved):
LGTM!

PR #724, opened 2023-12-18
Multi species

JintaoWu98:
This branch allows us to create a cluster species including multi-individual species. To be more specific, the current arc creates different input Gaussian input files if it receives multispecies. With this new feature, we can combine all the species in one Gaussian input file, followed by the relevant post-processing steps.

### DIFF
--- a/arc/job/adapter.py
+++ b/arc/job/adapter.py
@@ -375,7 +375,7 @@ class JobAdapter(ABC):
         ARC will allocate, e.g., 8 workers, to simultaneously get processes (one by one) from the HDF5 bank
         and execute them. On average, each worker in this example executes 125 jobs.
         """
-        if self.execution_type == 'incore':
+        if self.execution_type == 'incore' or self.run_multi_species:
             return None
         if len(self.job_types) > 1:
             self.iterate_by.append('job_types')
@@ -533,7 +533,10 @@ class JobAdapter(ABC):
         Set local and remote job file paths.
         """
         folder_name = 'TS_guesses' if self.reactions is not None else 'TSs' if self.species[0].is_ts else 'Species'
-        self.local_path = os.path.join(self.project_directory, 'calcs', folder_name, self.species_label, self.job_name)
+        if self.run_multi_species == False:
+            self.local_path = os.path.join(self.project_directory, 'calcs', folder_name, self.species_label, self.job_name)
+        else:
+            self.local_path = os.path.join(self.project_directory, 'calcs', folder_name, self.species[0].multi_species, self.job_name)
         self.local_path_to_output_file = os.path.join(self.local_path, 'output.out')
         self.local_path_to_orbitals_file = os.path.join(self.local_path, 'orbitals.fchk')
         self.local_path_to_check_file = os.path.join(self.local_path, 'check.chk')
@@ -545,7 +548,8 @@ class JobAdapter(ABC):
 
         if self.server is not None:
             # Parentheses don't play well in folder names:
-            species_name_remote = self.species_label.replace('(', '_').replace(')', '_')
+            species_name_remote = self.species_label if isinstance(self.species_label, str) else self.species[0].multi_species
+            species_name_remote = species_name_remote.replace('(', '_').replace(')', '_')
             path = servers[self.server].get('path', '').lower()
             path = os.path.join(path, servers[self.server]['un']) if path else ''
             self.remote_path = os.path.join(path, 'runs', 'ARC_Projects', self.project,
--- a/arc/job/adapters/common.py
+++ b/arc/job/adapters/common.py
@@ -105,6 +105,7 @@ def _initialize_adapter(obj: 'JobAdapter',
                         job_status: Optional[List[Union[dict, str]]] = None,
                         level: Optional[Level] = None,
                         max_job_time: Optional[float] = None,
+                        run_multi_species: bool = False,
                         reactions: Optional[List['ARCReaction']] = None,
                         rotor_index: Optional[int] = None,
                         server: Optional[str] = None,
@@ -164,6 +165,7 @@ def _initialize_adapter(obj: 'JobAdapter',
     # When restarting ARC and re-setting the jobs, ``level`` is a string, convert it to a Level object instance
     obj.level = Level(repr=level) if not isinstance(level, Level) and level is not None else level
     obj.max_job_time = max_job_time or default_job_settings.get('job_time_limit_hrs', 120)
+    obj.run_multi_species = run_multi_species
     obj.number_of_processes = 0
     obj.reactions = [reactions] if reactions is not None and not isinstance(reactions, list) else reactions
     obj.remote_path = None
--- a/arc/job/adapters/common.py
+++ b/arc/job/adapters/common.py
@@ -115,7 +115,7 @@ def _initialize_adapter(obj: 'JobAdapter',
                         times_rerun: int = 0,
                         torsions: Optional[List[List[int]]] = None,
                         tsg: Optional[int] = None,
-                        xyz: Optional[List[dict]] = None,
+                        xyz: Optional[Union[dict,List[dict]]] = None,
                         ):
     """
     A common Job adapter initializer function.
@@ -251,25 +251,51 @@ def _initialize_adapter(obj: 'JobAdapter',
     check_argument_consistency(obj)
 
 
-def is_restricted(obj) -> bool:
+def is_restricted(obj: 'JobAdapter') -> Union[bool, List[bool]]:
     """
     Check whether a Job Adapter should be executed as restricted or unrestricted.
+    If the job adapter contains a list of species, return True or False per species.
 
     Args:
         obj: The job adapter object.
 
     Returns:
+        Union[bool, List[bool]]: Whether to run as restricted (``True``) or not (``False``).
+    """
+    if not obj.run_multi_species:
+        return is_species_restricted(obj)
+    else:
+        return [is_species_restricted(obj, species) for species in obj.species]
+
+
+def is_species_restricted(obj: 'JobAdapter',
+                          species: Optional['ARCSpecies'] = None,
+                          ) -> bool:
+    """
+    Check whether a species should be executed as restricted or unrestricted.
+
+    Args:
+        obj: The job adapter object.
+        species (ARCSpecies, optional): The species to check.
+
+    Returns:
         bool: Whether to run as restricted (``True``) or not (``False``).
     """
-    if (obj.multiplicity > 1 and obj.level.method_type != 'composite') \
-            or (obj.species[0].number_of_radicals is not None and obj.species[0].number_of_radicals > 1):
+
+    if obj.level.method_type in ['force_field','composite','semiempirical']:
+        return True
+    
+    multiplicity = obj.multiplicity if species is None else species.multiplicity
+    number_of_radicals = obj.species[0].number_of_radicals if species is None else species.number_of_radicals
+    species_label = obj.species[0].label if species is None else species.label
+    if multiplicity > 1 or (number_of_radicals is not None and number_of_radicals > 1):
         # run an unrestricted electronic structure calculation if the spin multiplicity is greater than one,
         # or if it is one but the number of radicals is greater than one (e.g., bi-rad singlet)
         # don't run unrestricted for composite methods such as CBS-QB3, it'll be done automatically if the
         # multiplicity is greater than one, but do specify uCBS-QB3 for example for bi-rad singlets.
-        if obj.species[0].number_of_radicals is not None and obj.species[0].number_of_radicals > 1:
-            logger.info(f'Using an unrestricted method for species {obj.species_label} which has '
-                        f'{obj.species[0].number_of_radicals} radicals and multiplicity {obj.multiplicity}.')
+        if number_of_radicals is not None and number_of_radicals > 1:
+            logger.info(f'Using an unrestricted method for species {species_label} which has '
+                        f'{number_of_radicals} radicals and multiplicity {multiplicity}.')
         return False
     return True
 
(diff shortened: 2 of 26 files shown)


arc/job/adapters/common.py
```diff
@@ -182,18 +182,34 @@ def _initialize_adapter(obj: 'JobAdapter',
     obj.pivots = [[tor[1] + 1, tor[2] + 1] for tor in obj.torsions] if obj.torsions is not None else None
     obj.tsg = tsg
     obj.workers = None
-    obj.xyz = obj.species[0].get_xyz() if obj.species is not None and xyz is None else xyz
+    if obj.run_multi_species == False:
+        obj.xyz = obj.species[0].get_xyz() if obj.species is not None and xyz is None else xyz
+    else:
+        obj.xyz = list()
+        for index in range(len(obj.species)):
```
alongd, line 189: Sine we only use `index` in the context of `obj.species[index]` below, better to simply loop with
```python
for spc in obj.species:
```

arc/job/adapters/common.py
```diff
@@ -182,18 +182,34 @@ def _initialize_adapter(obj: 'JobAdapter',
...
+            obj.xyz.append(obj.species[index].get_xyz() if obj.species is not None and xyz is None else xyz)
```
alongd, line 190: If we got here, then we can assume that `obj.species` indeed isn't None (otherwise, the loop would have crashed)
If we care about this check, then it should come before the loop

arc/job/adapters/common.py
```diff
@@ -182,18 +182,34 @@ def _initialize_adapter(obj: 'JobAdapter',
...
         obj._set_job_number()
 
     if obj.species is not None:
-        obj.charge = obj.species[0].charge
-        obj.multiplicity = obj.species[0].multiplicity
-        obj.is_ts = obj.species[0].is_ts
-        obj.species_label = obj.species[0].label
-        if len(obj.species) > 1:
-            obj.species_label += f'_and_{len(obj.species) - 1}_others'
+        if obj.run_multi_species == False:
```
alongd, line 197: use `not` instead of `=== False`

arc/job/adapters/common.py
```diff
@@ -182,18 +182,34 @@ def _initialize_adapter(obj: 'JobAdapter',
...
+            obj.is_ts = obj.species[0].is_ts
+            obj.species_label = obj.species[0].label
+            if len(obj.species) > 1:
+                obj.species_label += f'_and_{len(obj.species) - 1}_others'
+        else:
+            obj.charge = list()
+            obj.multiplicity = list()
+            obj.is_ts = obj.species[0].is_ts
+            obj.species_label = list()
+            for index in range(len(obj.species)):
```
alongd, line 209: better to loop with `for spc in obj.species:`

arc/job/adapters/molpro.py
```diff
@@ -95,6 +95,7 @@ class MolproAdapter(JobAdapter):
         job_status (list, optional): The job's server and ESS statuses.
         level (Level, optional): The level of theory to use.
         max_job_time (float, optional): The maximal allowed job time on the server in hours (can be fractional).
+        run_multi_species (bool, optional): Whether to run a job for multiple species in the same input file.
```
alongd, line 98: If it's not too much trouble, we should relocate the cfour modification from the gaussian commit into this commit (not super important though)

arc/parser.py
```diff
@@ -815,6 +816,83 @@ def parse_xyz_from_file(path: str) -> Optional[Dict[str, tuple]]:
     return xyz
 
 
+def parse_multi_species_xyz_from_file(path: str, species_name_list: list) -> Optional[Dict[str, tuple]]:
```
alongd: This should be a commit of its own, not inside `Modify ts adapters to include rum_multi_species `

arc/job/adapters/gaussian.py
```diff
@@ -301,7 +301,7 @@ def write_input_file(self) -> None:
                     keywords.extend(['tight', 'maxstep=5'])
                 else:
                     keywords.extend(['tight', 'maxstep=5'])
-            input_dict['job_type_1'] = f"opt=({', '.join(key for key in keywords)})"
+            input_dict['job_type_1'] = "opt" if self.level.method == 'uff' else f"opt=({', '.join(key for key in keywords)})"
```
alongd, line 304: This should be a different commit than the tests below.

Also, we should generalize and not hard-code for `uff`:
``` python
if self.level.method_type not in ['dft', 'composite', 'wavefunction']
```

arc/job/adapters/common.py
```diff
@@ -251,25 +251,52 @@ def _initialize_adapter(obj: 'JobAdapter',
...
-    if (obj.multiplicity > 1 and obj.level.method_type != 'composite') \
-            or (obj.species[0].number_of_radicals is not None and obj.species[0].number_of_radicals > 1):
+
+    if obj.level.method_type in ['force_field','composite','semiempirical']:
+        return True
+    
+    multiplicity = obj.multiplicity if species is None else species.multiplicity
+    number_of_radicals = obj.species[0].number_of_radicals if species is None else species.number_of_radicals
+    species_label = obj.species[0].label if species is None else species.label
+    if (multiplicity > 1 and obj.level.method_type != 'composite') \
```
alongd, line 291: We no longer need `and obj.level.method_type != 'composite'`, this is being checked above we can assume here that it is not composite

arc/scheduler.py
```diff
@@ -13,6 +13,7 @@
 import numpy as np
 from IPython.display import display
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+from pathlib import Path
```
alongd, line 16: The import should be added in the commit where `Path` is being used (unless we change to using `os.path.dirname()`)

arc/scheduler.py
```diff
@@ -597,10 +598,15 @@ def schedule_jobs(self):
                         if not (job.job_id in self.server_job_ids and job.job_id not in self.completed_incore_jobs):
                             successful_server_termination = self.end_job(job=job, label=label, job_name=job_name)
                             if successful_server_termination:
+                                multi_species = any(spc.multi_species == label for spc in self.species_list)
+                                if multi_species:
+                                    self.multi_species_path_dict = self.make_multi_species_output_file(label=label, job=job)
```
alongd, line 603: Maybe we should call `make_multi_species_output_file()` inside `parse_opt_geo()`?
THis place seems too generic or too high-level to be calling these functions

arc/scheduler.py
```diff
@@ -605,8 +605,8 @@ def schedule_jobs(self):
                                 if success:
                                     self.parse_opt_e_elect(label=label, job=job)
                                     self.spawn_post_opt_jobs(label=label, job_name=job_name)
-                                # if multi_species:
-                                #     self.delete_multi_species_output_file(label=label, job=job)
+                                if multi_species:
```
alongd, line 608: should be squshed with the previous commit where it is added as a comment

Also, we need to consider where to call this. Maybe inside `parse_opt_e_elect()` as well?

arc/parser.py
```diff
@@ -345,58 +345,58 @@ def parse_e_elect(path: str,
...
-                    # output_file.writelines(sections)
-                    for inner_list in sections:
-                        line = ' '.join(map(str, inner_list))
-                        output_file.write(software + '\n' + line + '\n')
-                e_elect_multi[species_name_list[i]] = parse_e_elect(path=output_file_path)
-                os.remove(output_file_path)    
-    except (LogError, NotImplementedError):
-        logger.warning(f'Could not read e_elect from {path}')
-    return e_elect_multi
+# def parse_multi_species_e_elect(path: str,
```
alongd, line 348: this isn't final, right?

alongd (commented):
@JintaoWu98, thanks for this first major contribution to ARC!
The PR looks good and well thought of. I added comments, most of them on improving writing style, some with questions.

calvinp0:
@JintaoWu98 @alongd 

Hi guys, I just wanted to ask, what happens if one species in the multi gauss file failes due to a convergence issue etc. How do we troubleshoot it? Do we rerun all the species again, or will ARC be able to recognise that, for example, 1 of the 3 species in the input.gjf failed and it will only troubleshoot that particular species

JintaoWu98:
> @JintaoWu98 @alongd
> 
> Hi guys, I just wanted to ask, what happens if one species in the multi gauss file failes due to a convergence issue etc. How do we troubleshoot it? Do we rerun all the species again, or will ARC be able to recognise that, for example, 1 of the 3 species in the input.gjf failed and it will only troubleshoot that particular species

alongd:
I agree, troubleshooting specific species in the future is the way to go

arc/job/adapter.py
```diff
@@ -375,7 +375,8 @@ def determine_job_array_parameters(self):
         ARC will allocate, e.g., 8 workers, to simultaneously get processes (one by one) from the HDF5 bank
         and execute them. On average, each worker in this example executes 125 jobs.
         """
-        if self.execution_type == 'incore':
+        if self.execution_type == 'incore' \
+                or self.species is not None and all([spc.multi_species is not None for spc in self.species]):
```
alongd, line 379: Can we use `self.run_multi_species` here instead of `self.species is not None and all([spc.multi_species is not None for spc in self.species])`?

arc/job/adapter.py
```diff
@@ -545,7 +549,7 @@ def set_file_paths(self):
 
         if self.server is not None:
             # Parentheses don't play well in folder names:
-            species_name_remote = self.species_label.replace('(', '_').replace(')', '_')
+            species_name_remote = self.species_label.replace('(', '_').replace(')', '_') if isinstance(self.species_label, str) else ''
```
alongd, line 552: This will make `species_name_remote` an empty sting in case we have a multispecies. Is that desirable?

arc/job/adapters/common.py
```diff
@@ -233,25 +251,51 @@ def _initialize_adapter(obj: 'JobAdapter',
     check_argument_consistency(obj)
 
 
-def is_restricted(obj) -> bool:
+def is_restricted(obj: 'JobAdapter') -> Union[bool, List[bool]]:
```
alongd, line 254: Can you add tests for these two functions in common?

arc/job/adapters/gaussian.py
```diff
@@ -245,8 +248,8 @@ def write_input_file(self) -> None:
         if self.level.method_type in ['semiempirical', 'force_field']:
             self.checkfile = None
 
-        if not is_restricted(self):
-            input_dict['restricted'] = 'u'
+        # if self.run_multi_species is False and not is_restricted(self):
```
alongd, line 251: why did this commented-out code remain here? Can we remove it?

arc/job/adapters/gaussian.py
```diff
@@ -263,8 +271,10 @@ def write_input_file(self) -> None:
             self.level.method = 'cbs-qb3'
 
         # Job type specific options
+        # max_cycles 500
+        max_c = self.args['trsh'].split()[1] if 'max_cycles' in self.args['trsh'] else 100
```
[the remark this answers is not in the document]
alongd: Was this comment addressed?

arc/job/adapters/gaussian.py
```diff
@@ -263,8 +266,10 @@ def write_input_file(self) -> None:
...
         if self.job_type in ['opt', 'conformers', 'optfreq', 'composite']:
-            keywords = ['ts', 'calcfc', 'noeigentest', 'maxcycles=100'] if self.is_ts else ['calcfc']
+            keywords = ['ts', 'calcfc', 'noeigentest', f'maxcycles={max_c}' + '}'] if self.is_ts else ['calcfc']
```
alongd, line 272: Why do we need the extra `+ '}'`?

arc/scheduler.py
alongd: Was this comment addressed? If so, you can mark it as "resolved"

arc/scheduler.py
alongd: Was this addressed?

arc/scheduler.py
```diff
@@ -767,13 +773,18 @@ def run_job(self,
...
         """
         max_job_time = max_job_time or self.max_job_time  # if it's None, set to default
         ess_trsh_methods = ess_trsh_methods if ess_trsh_methods is not None else list()
-        species = self.species_dict[label] if label is not None else None
+        species = None
+        if isinstance(label, str) and label in self.output:
+            species = self.species_dict[label]
+        elif isinstance(label, list):
+            species = [spc for spc in self.species_list if spc.label in label]
+        run_multi_species = all([spc.multi_species is not None for spc in species]) if isinstance(species, list) else False
```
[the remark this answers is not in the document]
alongd: Was this addressed?

arc/scheduler.py
```diff
@@ -3503,6 +3615,74 @@ def save_restart_dict(self):
             logger.debug(f'Dumping restart dictionary:\n{self.restart_dict}')
             save_yaml_file(path=self.restart_path, content=self.restart_dict)
 
+    def make_multi_species_output_file(self,
```
alongd, line 3618: Can you try relocating this function from Scheduler to plotter?
Also please add a test for it

alongd (commented):
Thanks @JintaoWu98 !

I added another round of comments, please take a look

JintaoWu98 (commented):
Please let me know if there are any other comments. Thanks!

alongd (commented):
Thanks for this contribution, @JintaoWu98 !
Looking good. Please go ahead and squash the fixup commits

alongd:
@JintaoWu98, is this PR ready to be merged? Can you run a multispecies project using the updated brunch to confirm it works as expected?

alongd (approved):
Thanks for this contribution!

PR #726, opened 2024-01-30
Updating Dockerfile

calvinp0:
Required updates for the Dockerile and provided different entrypoints (for interactive and non-interactive)


JintaoWu98 (approved):
Thank you, LGTM!

PR #732, opened 2024-02-12
Dockerfile Update - Julia

calvinp0:
RMG-Py environment has now been updated to restrict the Julia version. Since this has been done, we do not need to patch the environment file ourselves.


Dockerfile
```diff
@@ -118,7 +116,9 @@ RUN echo "alias rmge='micromamba activate rmg_env'" >> ~/.bashrc \
     && echo "alias rmgdb='cd \$rmgdb_path'" >> ~/.bashrc \
     && echo "alias arcode='cd /home/rmguser/Code/ARC'" >> ~/.bashrc \
     && echo "alias conda='micromamba'" >> ~/.bashrc \
-    && echo "alias mamba='micromamba'" >> ~/.bashrc
+    && echo "alias mamba='micromamba'" >> ~/.bashrc \
```
MichalKesl, line 119: shouldn't this be  >> ~/.bashrc (without \ at the end)?

calvinp0: This is just how Dockerfile needs to be written if you are having multiple RUN statements put together

calvinp0: `\` will always be followed by `&&` to denote that there is another statement that will occur in that RUN layer (statement)

MichalKesl (approved):
Looks great

PR #733, opened 2024-02-13
Docker CI Skip No Push Step if Merge

calvinp0:
Runs the step if the event is a pull request, regardless of the branch. Runs the step if the event is a push, but only if the push is not to the main branch. Ensures that the step is skipped for any pushes directly to the main branch


JintaoWu98 (approved):
LGTM! Thanks!

NellyMitnik (approved):

PR #735, opened 2024-02-14
Define output dict for multi species in main.py

JintaoWu98:
We have problem when we launch `restart.yml` saying unexpected keyword argument 'output_multi_spc'. It turns out previously we defined `Scheduler.output_multi_spc` parallel to `Scheduler.output` in `Scheduler` `__init__`, however, we forgot to define it in the `ARC` class in `main.py`. So now we add the relevant terms in `main.py`.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -154,6 +154,8 @@ class ARC(object):
         trsh_ess_jobs (bool, optional): Whether to attempt troubleshooting failed ESS jobs. Default is ``True``.
         output (dict, optional): Output dictionary with status and final QM file paths for all species.
                                  Only used for restarting.
+        output_multi_spc (dict, optional): Output dictionary with status and final QM file paths for the multi species. 
+                                           Only used for restarting.
         running_jobs (dict, optional): A dictionary of jobs submitted in a precious ARC instance, used for restarting.
         ts_adapters (list, optional): Entries represent different TS adapters.
         report_e_elect (bool, optional): Whether to report electronic energy. Default is ``False``.
@@ -180,6 +182,8 @@ class ARC(object):
             Job types not defined in adaptive levels will have non-adaptive (regular) levels.
         output (dict): Output dictionary with status and final QM file paths for all species. Only used for restarting,
                        the actual object used is in the Scheduler class.
+        output_multi_spc (dict): Output dictionary with status and final QM file paths for the multi species. 
+                                 Only used for restarting, the actual object used is in the Scheduler class.
         bac_type (str): The bond additivity correction type. 'p' for Petersson- or 'm' for Melius-type BAC.
                         ``None`` to not use BAC.
         arkane_level_of_theory (Level): The Arkane level of theory to use for AEC and BAC.
@@ -258,6 +262,7 @@ class ARC(object):
                  opt_level: Optional[Union[str, dict, Level]] = None,
                  orbitals_level: Optional[Union[str, dict, Level]] = None,
                  output: Optional[dict] = None,
+                 output_multi_spc: Optional[dict] = None,
                  project: Optional[str] = None,
                  project_directory: Optional[str] = None,
                  reactions: Optional[List[Union[ARCReaction, Reaction]]] = None,
@@ -291,6 +296,7 @@ class ARC(object):
         if not os.path.exists(self.project_directory):
             os.makedirs(self.project_directory)
         self.output = output
+        self.output_multi_spc = output_multi_spc
         self.standardize_output_paths()  # depends on self.project_directory
         self.running_jobs = running_jobs or dict()
         for jobs in self.running_jobs.values():
@@ -515,6 +521,7 @@ class ARC(object):
             restart_dict['orbitals_level'] = self.orbitals_level.as_dict() \
                 if not isinstance(self.orbitals_level, (dict, str)) else self.orbitals_level
         restart_dict['output'] = self.output
+        restart_dict['output_multi_spc'] = self.output_multi_spc if self.output_multi_spc else dict()
         restart_dict['project'] = self.project
         restart_dict['reactions'] = [rxn.as_dict() for rxn in self.reactions]
         restart_dict['running_jobs'] = self.running_jobs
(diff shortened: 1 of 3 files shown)


arc/main.py
```diff
@@ -180,6 +180,8 @@ class ARC(object):
             Job types not defined in adaptive levels will have non-adaptive (regular) levels.
         output (dict): Output dictionary with status and final QM file paths for all species. Only used for restarting,
                        the actual object used is in the Scheduler class.
+        output_multi_spc (dict): Output dictionary with status and final QM file paths for the multi species.
```
alongd, line 183: please also add above under `Args:`

arc/main.py
```diff
@@ -291,6 +294,7 @@ def __init__(self,
         if not os.path.exists(self.project_directory):
             os.makedirs(self.project_directory)
         self.output = output
+        self.output_multi_spc = output_multi_spc
```
alongd, line 297: we store it as an attribute, but we don't do anything with it. I think we should transmit it to Scheduler (and also add it as an argument in Scheduler like here)

alongd (commented):
Thanks! Can you please see the two comments below?

arc/main.py
JintaoWu98: We actually transmitted it to Scheduler as an attribute in the previous PR already (like what had been done to the output dict), I think we just forgot to define it in the main.py.

calvinp0: Also, do we need it part of this function for the dumping of the yaml? @alongd 
https://github.com/ReactionMechanismGenerator/ARC/blob/688f9602e7d41252cc0e22a27cd30a4d59eafcf5/arc/main.py#L464C1-L467C12

JintaoWu98: > I could be misunderstanding but I think @alongd meant here
> 
> https://github.com/ReactionMechanismGenerator/ARC/blob/688f9602e7d41252cc0e22a27cd30a4d59eafcf5/arc/main.py#L591

Thank you for pointing that out. Indeed, we didn't transmit the output dictionary to the Scheduler. I have a question, though. I thought `output_multi_spc` was the multi-species counterpart to `output`. However, since we didn't transmit `output` here, why is there a need to transmit `output_multi_spc`? Am I missing something?

calvinp0: Ah maybe you're right. I'm not really familiar with the restart function. Do you know with these changes you've made that if a restart of a multi species works? As in ARC picks up that a multi species is/was run(ning)

JintaoWu98: I'm not that familiar with the restart function either. Currently, I only have `restart.yml` from completed projects (both single and multi-species), which have already finished running. With these changes, I launched `restart.yml` for single and multi-species projects individually and compared them. They do yield similar folders, such as `log_and_restart_archive`, and no errors were reported in the `err.txt`. However, I'm not sure how this will turn out in the future with an unfinished project.

calvinp0: Areyou able to do with it unfinished. What I mean is start the job with the input.yml and then as soon as you see it submits its first job to the server, cancel ARC `qdel <job id>` and then restart it

JintaoWu98: Now I get the following error, this from when I terminate the ARC submitted job
```
  File "/home/jintaowu/Code/ARC/ARC.py", line 69, in <module>
    main()
  File "/home/jintaowu/Code/ARC/ARC.py", line 65, in main
    arc_object.execute()
  File "/home/jintaowu/Code/ARC/arc/main.py", line 620, in execute
    skip_nmd=self.skip_nmd,
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 508, in __init__
    self.schedule_jobs()
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 595, in schedule_jobs
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 993, in end_job
    self._run_a_job(job=job, label=label)
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 1060, in _run_a_job
    xyz=job.xyz,
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 801, in run_job
    checkfile = self.species_dict[label].checkfile if isinstance(label, str) else None
KeyError: 'multi_spc1'
```
And this is the error when I restart the job
```
  File "/home/jintaowu/Code/ARC/ARC.py", line 69, in <module>
    main()
  File "/home/jintaowu/Code/ARC/ARC.py", line 65, in main
    arc_object.execute()
  File "/home/jintaowu/Code/ARC/arc/main.py", line 620, in execute
    skip_nmd=self.skip_nmd,
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 483, in __init__
    self.run_opt_job(species.label, fine=self.fine_only)
  File "/home/jintaowu/Code/ARC/arc/scheduler.py", line 1188, in run_opt_job
    if self.output_multi_spc[self.species_dict[label].multi_species].get(key, False):
KeyError: 'multi_spc1'
```

alongd:
I confirm that ARC's restart does not crash on this branch. @calvinp0, do you have any additional comments, or can we merge?

arc/scheduler.py
```diff
@@ -1185,16 +1198,21 @@
...
+            label_single_spc = label
             label = [species.label for species in self.species_list
                      if species.multi_species == self.species_dict[label].multi_species]
         self.run_job(label=label, 
                      xyz=self.species_dict[label].initial_xyz if isinstance(label, str) else None,
                      level_of_theory=self.opt_level,
                      job_type='opt', 
                      fine=fine)
+        if self.species_dict[label].multi_species:
+            self.output_multi_spc[self.species_dict[label_single_spc].multi_species][key] = True
```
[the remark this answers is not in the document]
calvinp0: Hey @JintaoWu98, these code scan issues appear to be still in the latest push. It looks like they will be issues if certain conditions are met (as the Check Failure says, local var may be used before initialisation)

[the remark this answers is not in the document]
calvinp0: @JintaoWu98 Same as above comment

calvinp0:
> I confirm that ARC's restart does not crash on this branch. @calvinp0, do you have any additional comments, or can we merge?

The code scanning issues need to be resolved I think

JintaoWu98:
> > I confirm that ARC's restart does not crash on this branch. @calvinp0, do you have any additional comments, or can we merge?
> 
> The code scanning issues need to be resolved I think

@calvinp0, thanks for pointing it out. The updated branch is now more straightforward and robust.

calvinp0 (approved):
Thanks @JintaoWu98!

PR #736, opened 2024-03-04
Modified Cl2 xyz

kfir4444:
The bond distance suggested in earlier version was 1.1 ångström, while based on experimental results, it is 1.988 ångström (based on https://cms.gutow.uwosh.edu/gutow/Chem371_F19/KGTS/Cl2/Cl2.html). This caused issues with convergence for the Molpro software when trying to optimize the species geometry.


alongd:
Looks good.
If you want, you can take a `CCSD(T)=FULL/daug-cc-pVTZ` value from CCCBDB of `2.0110` A

calvinp0 (approved):

PR #740, opened 2024-05-06
Small atom mapping debugging

kfir4444:
Two small fixes for the atom mapping module.

### DIFF
--- a/arc/mapping/engine.py
+++ b/arc/mapping/engine.py
@@ -23,7 +23,7 @@ from arc.exceptions import SpeciesError
 from arc.species import ARCSpecies
 from arc.species.conformers import determine_chirality
 from arc.species.converter import compare_confs, sort_xyz_using_indices, translate_xyz, xyz_from_data, xyz_to_str
-from arc.species.vectors import calculate_angle, calculate_dihedral_angle, calculate_distance, get_delta_angle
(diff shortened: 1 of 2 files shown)


arc/mapping/engine.py
```diff
@@ -1117,8 +1117,12 @@ def r_cut_p_cut_isomorphic(reactant, product):
         bool: ``True`` if they are isomorphic, ``False`` otherwise.
     """
     res1 = generate_resonance_structures(object_=reactant.mol.copy(deep=True), save_order = True)
```
alongd, line 1131: I would recommend to modify the `generate_resonance_structures` func so it always returns at least the original Molecule. Please see where else this func is used and whether this change will be appropriate

arc/mapping/engine.py
```diff
@@ -856,17 +856,23 @@ def map_hydrogens(spc_1: ARCSpecies,
...
+                                    torsion=rotor_dict['torsion'],
+                                    atom_map=atom_map,
+                                    find_torsion_end_to_map=True,
+                                )
+                                success = True
+                                break
+                            except VectorsError:
+                                # Trying to map a linear torsion
+                                # but the torsion is undefined, since v1 x v2 = 0 if v1 || v2
+                                # using other internal coordinates to map the hydrogens.
```
alongd, line 874: does this fix solve the issue or skips it? We could consider using a different atom instead of the last atom in the torsion

alongd (commented):
Thanks. Please see two comments

PR #743, opened 2024-05-21
Report energy in `conformers_before_optimization.txt`

JintaoWu98:
The previous `conformers_before_optimization.txt` doesn't include energy information for the conformer before the optimization, so the updated one will include this.

### DIFF
--- a/arc/plotter.py
+++ b/arc/plotter.py
@@ -900,6 +900,7 @@ def save_conformers_file(project_directory: str,
                          ts_methods: Optional[List[str]] = None,
                          im_freqs: Optional[List[List[float]]] = None,
                          log_content: bool = False,
+                         before_optimization: bool = True,
                          ):
     """
     Save the conformers before or after optimization.
@@ -918,6 +919,7 @@ def save_conformers_file(project_directory: str,
         ts_methods (list, optional): Entries are method names used to generate the TS guess.
         im_freqs (list, optional): Entries lists of imaginary frequencies.
         log_content (bool): Whether to log the content of the conformers file. ``True`` to log, default is ``False``.
+        before_optimization (bool): Whether the conformers are before DFT optimization. ``True`` for before, default is ``True``.
     """
     spc_dir = 'rxns' if is_ts else 'Species'
     geo_dir = os.path.join(project_directory, 'output', spc_dir, label, 'geometry', 'conformers')
--- a/arc/plotter.py
+++ b/arc/plotter.py
@@ -925,16 +925,16 @@ def save_conformers_file(project_directory: str,
     geo_dir = os.path.join(project_directory, 'output', spc_dir, label, 'geometry', 'conformers')
     if not os.path.exists(geo_dir):
         os.makedirs(geo_dir)
-    if energies is not None and any(e is not None for e in energies):
-        optimized = True
-        min_e = extremum_list(energies, return_min=True)
-        conf_path = os.path.join(geo_dir, 'conformers_after_optimization.txt')
-    else:
-        optimized = False
+    if before_optimization:
         conf_path = os.path.join(geo_dir, 'conformers_before_optimization.txt')
+    else:
+        conf_path = os.path.join(geo_dir, 'conformers_after_optimization.txt')
+    min_e = extremum_list(energies, return_min=True)
     with open(conf_path, 'w') as f:
         content = ''
-        if optimized:
+        if before_optimization:
+            content += f'Conformers for {label}, computed using a force field:\n\n'
+        else:
             level_of_theory = level_of_theory.simple() if isinstance(level_of_theory, Level) else level_of_theory
             content += f'Conformers for {label}, optimized at the {level_of_theory} level:\n\n'
         for i, xyz in enumerate(xyzs):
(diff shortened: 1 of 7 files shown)


arc/plotter.py
```diff
@@ -923,18 +924,22 @@
...
-    else:
-        optimized = False
-        conf_path = os.path.join(geo_dir, 'conformers_before_optimization.txt')
     with open(conf_path, 'w') as f:
         content = ''
         if optimized:
             level_of_theory = level_of_theory.simple() if isinstance(level_of_theory, Level) else level_of_theory
             content += f'Conformers for {label}, optimized at the {level_of_theory} level:\n\n'
+        if before_optimized:
+            level_of_theory = level_of_theory.simple() if isinstance(level_of_theory, Level) else level_of_theory
```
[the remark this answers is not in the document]
alongd: `level_of_theory ` isn't used

arc/plotter.py
```diff
@@ -923,18 +924,22 @@ def save_conformers_file(project_directory: str,
...
+            content += f'Conformers for {label}, generated from RDKit:\n\n'
```
alongd, line 942: You can pass the string level of theory into this function (e.g., MMFF94s or UFF) and print it here instead of writing RDKit

arc/scheduler.py
```diff
@@ -1845,6 +1845,8 @@ def process_conformers(self, label):
                                      multiplicity=self.species_dict[label].multiplicity,
                                      charge=self.species_dict[label].charge,
                                      is_ts=False,
+                                     energies=self.species_dict[label].conformer_energies,
+                                     before_optimized=True,
```
alongd, line 1849: Please rename to "before_optimization"

alongd (commented):
Thanks! I added some minor comments

arc/plotter.py
```diff
@@ -923,18 +924,21 @@ def save_conformers_file(project_directory: str,
     geo_dir = os.path.join(project_directory, 'output', spc_dir, label, 'geometry', 'conformers')
     if not os.path.exists(geo_dir):
         os.makedirs(geo_dir)
-    if energies is not None and any(e is not None for e in energies):
+    if before_optimization or energies is None or not any(e is not None for e in energies):
```
alongd, line 927: We should make the conditions much simpler and differentiate the case simply using the `before_optimization` arg instead of looking at the energies. something like:
Also, now we don't need to set `optimized`, we can just use `before_optimization`.
Can you go over the logic and make this function simpler and robust?

arc/plotter.py
```diff
@@ -900,6 +900,7 @@ def save_conformers_file(project_directory: str,
                          ts_methods: Optional[List[str]] = None,
                          im_freqs: Optional[List[List[float]]] = None,
                          log_content: bool = False,
+                         before_optimization: Optional[bool] = None,
```
alongd, line 903: I think this shouldn't be Optional (cannot be `None`) because we're always either before or after opt.
We should make sure to send a proper value in all the calls to this functions, including tests

alongd (commented):
Thanks, please see some comments

arc/scheduler.py
```diff
@@ -1139,6 +1139,7 @@ def run_ts_conformer_jobs(self, label: str):
                         f'{tsg.method_direction if tsg.method_direction is not None else ""} '
                         f'{tsg.method_index if tsg.method_index is not None else ""} '
                         for tsg in self.species_dict[label].ts_guesses],
+            before_optimization=True,
```
alongd, line 1142: please fix the alignment of this line

arc/species/species.py
```diff
@@ -1098,7 +1098,7 @@ def generate_conformers(self,
                                                       )
         if len(lowest_confs):
             self.conformers.extend([conf['xyz'] for conf in lowest_confs])
-            self.conformer_energies.extend([None] * len(lowest_confs))
+            self.conformer_energies.extend([conf['FF energy'] for conf in lowest_confs])
```
alongd, line 1101: Can you explain what this does?

alongd (commented):
Thanks! Please see two minor comments

arc/scheduler.py
JintaoWu98: The full block looks as the following, so I think `before_optimization=True` aligns with all the other parameters and inputs.
```
plotter.save_conformers_file(
    project_directory=self.project_directory,
    label=label,
    xyzs=[tsg.initial_xyz for tsg in self.species_dict[label].ts_guesses],
    level_of_theory=self.ts_guess_level,
    multiplicity=self.species_dict[label].multiplicity,
    charge=self.species_dict[label].charge,
    is_ts=True,
    ts_methods=[f'{tsg.method} '
                f'{tsg.method_direction if tsg.method_direction is not None else ""} '
                f'{tsg.method_index if tsg.method_index is not None else ""} '
                for tsg in self.species_dict[label].ts_guesses],
    before_optimization=True,
)
```

arc/species/species.py
JintaoWu98: https://github.com/ReactionMechanismGenerator/ARC/blob/98b443c4feec46cc2823cda21ece5998cf3c783e/arc/scheduler.py#L1842-L1851

Line 1849 `self.species_dict[label].conformer_energies` requires attribute `conformer_energies` to be well defined, previous code will lead to a list of `None`.

PR #744, opened 2024-06-17
Don't execute jobs on a server if the maximum job limit has been reached

alongd:
Limit the overall number of simultaneous jobs on a server if the user defined `max_simultaneous_jobs` in the servers dict in settings.
This feature was not tested yet.


calvinp0 (approved):
LGTM

PR #745, opened 2024-06-20
Use the WELL_GAP constant in conformers

alongd:


JintaoWu98 (approved):
I think it works as we expected. Thank you very much!

alongd (commented):
Thanks @JintaoWu98 !

PR #748, opened 2024-06-23
Fixed the sella env

kfir4444:
Numpy came up with version 2.0.0, which modified the API. Sella still works with <2.0.0, so locked in the last version.


alongd (approved):
LGTM

PR #749, opened 2024-06-23
Don't run freq for IRC species

alongd:
An IRC job is supposed to give us the minimal energy path that connects a TS to its adjacent wells (reactants and products).
ARC optimizes the two resulting IRC output to get the wells. However, it also unnecessarily spawns a freq job for the IRC species. This is a bit of an overkill, since we just want to see that the connectivity is correct, we don't use the IRC results per se in the calculations, so we don't care about their frequency (or whether there's an imaginary frequency). Plus, this computation is relatively expensive. sp and scan jobs are also not needed.
Here we tell ARC not to spawn freq jobs for species defined from IRC jobs

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -441,7 +441,8 @@ class Scheduler(object):
                             self.run_sp_job(label=species.label)
                         if self.job_types['onedmin']:
                             self.run_onedmin_job(species.label)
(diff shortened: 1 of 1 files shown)


JintaoWu98 (approved):
LGTM

PR #751, opened 2024-06-28
Reduce clutter in restart dict

alongd:
Don't save default values of `ARC`, `ARCSpecies`, and `ARCReaction` objects as dictionaries
Tests modified accordingly


JintaoWu98 (approved):
LGTM!

PR #752, opened 2024-06-30
Don't consider atoms in a ring as distinct groups for symmetry

alongd:
When generating conformers and looking for torsion symmetry, avoid considering atoms that are in the same ring.


JintaoWu98 (approved):
LGTM!

PR #758, opened 2024-08-06
Gaussian unconverge trsh

calvinp0:
There are times when we do not get convergence of the optimisation of a molecule and we only receive an output of `l9999.exe`. Currently, we do not make any drastic changes then to the input file for Gaussian for troubleshooting and may end up determining we cannot converge the molecule and therefore give up.

However, further up the file we may find the number of steps is a limit (the example output file shows 27 steps is the max for opt). Therefore, this aims to troubleshoot it by increasing the cycles for opt. Removed increase of scf maxcycle as DM/QC (512) have their own default values compared to default algorithm (128). If non-convergence is still not occuring, then checking if the scf is shaking manually should be done. https://wongzit.github.io/method-to-solve-the-scf-not-converged/

Edit: I have now included more methods for dealing with other errors, such as L502 - `Inaccurate quadrature in CalDSu`

I have attempted to provide comments against each trsh_keyword to explain what they do

### DIFF
--- a/arc/job/adapters/gaussian_test.py
+++ b/arc/job/adapters/gaussian_test.py
@@ -218,7 +218,7 @@ class TestGaussianAdapter(unittest.TestCase):
                             )
         
         # Gaussian: Checkfile error and SCF error
-        # First SCF error - qc,nosymm
+        # First SCF error - maxcycle=128,nosymm
         job_status = {'keywords': ['SCF', 'NoSymm']}
         output_errors, ess_trsh_methods, remove_checkfile, level_of_theory, software, job_type, fine, trsh_keyword, \
             memory, shift, cpu_cores, couldnt_trsh = trsh.trsh_ess_job(label, level_of_theory, server, job_status,
@@ -240,7 +240,7 @@ class TestGaussianAdapter(unittest.TestCase):
                                             )
         
         # Gaussian: Additional SCF error
-        # Second SCF error - Includes previous SCF error and NDamp=30
+        # Second SCF error - Includes previous SCF error and xqc
         job_status = {'keywords': ['SCF']}
         output_errors, ess_trsh_methods, remove_checkfile, level_of_theory, software, job_type, fine, trsh_keyword, \
             memory, shift, cpu_cores, couldnt_trsh = trsh.trsh_ess_job(label, level_of_theory, server, job_status,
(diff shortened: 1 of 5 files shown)


JintaoWu98 (dismissed):
LGTM!

arc/job/trsh.py
```diff
@@ -878,6 +887,8 @@ def trsh_ess_job(label: str,
 
         # Check if SCF is in the keyword
         ess_trsh_methods, trsh_keyword, couldnt_trsh = trsh_keyword_scf(job_status, ess_trsh_methods, trsh_keyword, couldnt_trsh)
+        if 'scf=(maxcycle=512)' in ess_trsh_methods:
```
alongd, line 890: Why 512?

alongd (commented):
Thanks, I added one small comment

arc/job/adapters/gaussian_test.py
```diff
@@ -651,7 +675,7 @@ def test_trsh_write_input_file(self):
 %mem=14336mb
 %NProcShared=8
 
-#P opt=(calcfc,maxstep=5,tight)  guess=mix wb97xd  integral=(grid=ultrafine, Acc2E=14) IOp(2/9=2000)     nosymm scf=(direct,tight,xqc)
+#P opt=(calcfc,maxstep=5,tight)  guess=mix wb97xd  integral=(grid=ultrafine, Acc2E=14) IOp(2/9=2000)     nosymm scf=(direct,maxcycle=512,tight)
```
Lilachn91, line 678: Does is also work when the `maxcycles` is in the `scf ` block and not in the `opt` block? 
Also, I'm familiar with it as `maxcycles` (or `maxcyc`) and not `maxcycle`.

arc/job/trsh.py
calvinp0: It was just a number that was suggested. There is no rhyme or reason why that value specifically. As I mentioned in the description, this troubleshoot addition is a last resort brute force.

arc/job/adapters/gaussian_test.py
calvinp0: https://wongzit.github.io/gaussian-common-errors-and-solutions/#inaccurate-quadrature-in-caldsu - Here they use maxcycle as the word
https://chemistry.stackexchange.com/questions/58305/gaussian-calculation-problem-maxcycle-values-for-opt-and-scf - Here they use both scf and opt maxcycle
https://www.researchgate.net/post/How-to-increase-the-number-of-maximum-cycles-in-gaussian-03-software-windows-mode-for-excited-state-optimisation - Here they use maxcycle

In the end, I don't think it matters if it is `maxcycle` `maxcyc` or `maxcycles`

arc/job/trsh.py
calvinp0: I have removed the scf cycle increase as it would have consequences on certain aspects if this is being done automatedly.

JintaoWu98 (approved):
LGTM!

PR #761, opened 2024-08-11
Increasing opt maxcycles for fine opt

calvinp0:
Commonly, our fine opt fails due to number of steps exceeded. Whilst we may troubleshoot this after an initial failure, I believe it is best to set the maxcycles at 100 to begin with so that we increase the chance of convergence prior to troubleshooting. This can reduce our creation of opt jobs.

### DIFF
--- a/arc/job/adapters/gaussian.py
+++ b/arc/job/adapters/gaussian.py
@@ -274,6 +274,9 @@ class GaussianAdapter(JobAdapter):
                 # There are no analytical 2nd derivatives (FC) for this method.
                 keywords = ['ts', 'noeigentest', 'maxcycles=100'] if self.is_ts else []
             if self.fine:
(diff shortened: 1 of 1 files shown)


alongd:
I agree it's good to begin with a larger number than we have today.
Maybe also change the Gaussian input file template we have?

calvinp0:
> I agree it's good to begin with a larger number than we have today.
> 
> Maybe also change the Gaussian input file template we have?

I agree we should but I was thinking it would require a bigger overhaul - whereby we define in input_dict keys that are related to variables 
So for example if we have scf
Input_dict['scf']=['maxcycle=512','xqc']
And then in the template we set the section for {scf}

alongd:
@calvinp0, can you rebase?

PR #763, opened 2024-08-18
Avoid crushing when can't trsh scan

alongd:
In trsh `scan_quality_check()`, ARC checks broken bond and any lowest conformation, excluding those with broken bonds. Ths is step 1.2 in the function.
Here we make a minor fix to a KeyError in the line `actions = {'change conformer': conf_xyzs[min_index]}`, and now we only call `conf_xyzs[min_index]` if the length of the list and the index value makes sense.

### DIFF
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -4,11 +4,11 @@ The ARC troubleshooting ("trsh") module
 
 import math
 import os
+import re
 from typing import List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
-import re
 
 from arc.common import (check_torsion_change,
                         convert_to_hours,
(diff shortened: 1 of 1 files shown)


NellyMitnik (approved):
I ran a species that caused ARC to crush for the corresponding error. With this fix, it converged successfully.
```- adjlist: |
    multiplicity 2
    1  O u0 p2 c0 {5,S} {12,S}
    2  O u1 p2 c0 {6,S}
    3  C u0 p0 c0 {4,S} {6,D} {8,S}
    4  C u0 p0 c0 {3,S} {7,D} {9,S}
    5  C u0 p0 c0 {1,S} {7,D} {10,S}
    6  C u0 p0 c0 {2,S} {3,D} {11,S}
    7  C u0 p0 c0 {4,D} {5,D}
    8  H u0 p0 c0 {3,S}
    9  H u0 p0 c0 {4,S}
    10 H u0 p0 c0 {5,S}
    11 H u0 p0 c0 {6,S}
    12 H u0 p0 c0 {1,S}
  label: C5H5O2(505)
```

PR #765, opened 2024-08-20
Optimize conformer comparison algorithm

JintaoWu98:
The previous conformer comparison algorithm required a distance matrix (d-matrix) for operations. In this update, we introduce a new variable, `fl_distance`, which represents the distance between the first and last elements. This method provides a less costly preliminary comparison before invoking the d-matrix comparison. Now, the d-matrix comparison will only be executed when necessary, optimizing the overall efficiency of the algorithm.

### DIFF
--- a/arc/species/conformers.py
+++ b/arc/species/conformers.py
@@ -519,15 +519,23 @@ def conformers_combinations_by_lowest_conformer(label, mol, base_xyz, multiple_t
             newest_conformers_dict[tor] = list()  # Keys are torsions for plotting.
             for xyz, energy, dihedral in zip(xyzs, energies, sampling_points):
                 exists = False
-                if any([converter.compare_confs(xyz, conf['xyz']) for conf in new_conformers + newest_conformer_list]):
-                    exists = True
+                dmat1, fl_distance1 = None, None
+                for conf in new_conformers + newest_conformer_list:
+                    fl_distance1, dmat1, conf, similar = converter.compare_confs_fl(xyz,conf)
+                    if not similar:
+                        break
+                    if converter.compare_confs(xyz, conf['xyz'], skip_conversion=True, dmat1=dmat1, dmat2=conf['dmat']):
+                        exists = True
+                        break
                 if xyz is not None and energy is not None:
                     conformer = {'index': len_conformers + len(new_conformers) + len(newest_conformer_list),
                                  'xyz': xyz,
                                  'FF energy': round(energy, 3),
                                  'source': f'Changing dihedrals on most stable conformer, iteration {i}',
                                  'torsion': tor,
-                                 'dihedral': round(dihedral, 2)}
+                                 'dihedral': round(dihedral, 2),
+                                 'dmat': dmat1,
+                                 'fl_distance': fl_distance1}
                     newest_conformers_dict[tor].append(conformer)
                     if not exists:
                         newest_conformer_list.append(conformer)
@@ -541,7 +549,9 @@ def conformers_combinations_by_lowest_conformer(label, mol, base_xyz, multiple_t
                          'FF energy': None,
                          'source': f'Changing dihedrals on most stable conformer, iteration {i}, but FF energy is None',
                          'torsion': tor,
-                         'dihedral': round(dihedral, 2)})
+                         'dihedral': round(dihedral, 2),
+                         'dmat': dmat1,
+                         'fl_distance': fl_distance1})
         new_conformers.extend(newest_conformer_list)
         if not newest_conformer_list:
             newest_conformer_list = [lowest_conf_i]
@@ -1113,9 +1123,15 @@ def get_lowest_confs(label: str,
     for index in range(len(conformer_list)):
         if (e is not None and conformer_list[index][energy] > min_e + e) or (n is not None and len(lowest_confs) >= n):
             break
-        if index > 0 and not any([converter.compare_confs(lowest_conf['xyz'], conformer_list[index]['xyz'])
-                                  for lowest_conf in lowest_confs]):
-            lowest_confs.append(conformer_list[index])
+        if index > 0:
+            for lowest_conf in lowest_confs:
+                _, dmat1, lowest_conf, similar = converter.compare_confs_fl(conformer_list[index]['xyz'],lowest_conf)
+                if not similar or not converter.compare_confs(conformer_list[index]['xyz'], lowest_conf['xyz'],
+                                                              skip_conversion=True,
+                                                              dmat1=dmat1,dmat2=lowest_conf['dmat']):
+                    lowest_confs.append(conformer_list[index]) if lowest_conf==lowest_confs[-1] else None
+                else:
+                    break
     return lowest_confs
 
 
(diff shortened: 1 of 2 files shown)


arc/species/conformers.py
```diff
@@ -519,15 +519,31 @@ def conformers_combinations_by_lowest_conformer(label, mol, base_xyz, multiple_t
...
+                dmat1, fl_distance1 = None, None
+                for conf in new_conformers + newest_conformer_list:
+                    conf['dmat'] = conf.get('dmat')
+                    conf['fl_distance'] = conf.get('fl_distance')
+                    fl_distance1, fl_distance2, dmat1, dmat2 = converter.compare_confs_fl(xyz,conf['xyz'],
+                                                                                          dmat2=conf['dmat'],
+                                                                                          fl_distance2=conf['fl_distance'])
+                    conf['fl_distance'] = fl_distance2 if conf['fl_distance'] is None else conf['fl_distance']
+                    conf['dmat'] = dmat2 if conf['dmat'] is None else conf['dmat']
+                    if dmat1 is None:
```
alongd, line 531: I understand that if dmat1 is None, this implies that the fl_distance was too large. But this code is not self-explanatory, therefore hard to maintain and understand.
I recommend to pass the two conformer dictionaries to `compare_confs_fl`, then have `compare_confs_fl` return three things: whether the conformers are similar or not and the two updated conformer dictionaries. The we update our confs with the returned dictionaries

alongd (commented):
Thanks! I added my initial review.

arc/species/conformers.py
```diff
@@ -519,15 +519,31 @@ def conformers_combinations_by_lowest_conformer(label, mol, base_xyz, multiple_t
             newest_conformers_dict[tor] = list()  # Keys are torsions for plotting.
             for xyz, energy, dihedral in zip(xyzs, energies, sampling_points):
                 exists = False
-                if any([converter.compare_confs(xyz, conf['xyz']) for conf in new_conformers + newest_conformer_list]):
-                    exists = True
+                dmat1, fl_distance1 = None, None
+                for conf in new_conformers + newest_conformer_list:
+                    conf['dmat'] = conf.get('dmat')
```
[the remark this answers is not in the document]
JintaoWu98: I think otherwise we may get into a case where `conf` dictionary doesn't have `dmat` at the beginning.

arc/species/conformers.py
JintaoWu98: Thanks, I agree. I modified `compare_confs_fl` with your advice. However, it's a bit inconvenient to pass both conf dicts under `conformers_combinations_by_lowest_conformer`, since we are looping `xyz`, as shown [here](https://github.com/ReactionMechanismGenerator/ARC/blob/89ce5f58ccdec85d1c0f428b0b11b58a45918d34/arc/species/conformers.py#L524). But if you would like me to modify it further, please let me know.

arc/species/converter.py
```diff
@@ -2011,6 +2011,43 @@ def compare_zmats(z1, z2, r_tol=0.01, a_tol=2, d_tol=2, verbose=False, symmetric
...
+    """
+    simliar = False
+    conf2['fl_distance'] = conf2.get('fl_distance')
+    conf2['dmat'] = conf2.get('dmat')
+    xyz1, xyz2 = check_xyz_dict(xyz1), check_xyz_dict(conf2['xyz'])
+    dmat1 = None
+    fl_distance1 = np.linalg.norm(np.array(xyz1['coords'][0]) - np.array(xyz1['coords'][-1]))
+    if conf2['fl_distance'] is None:
+        conf2['fl_distance'] = np.linalg.norm(np.array(xyz2['coords'][0]) - np.array(xyz2['coords'][-1]))
+    if abs(fl_distance1-conf2['fl_distance']) / fl_distance1 > rtol:
```
alongd, line 2042: You can use `np.isclose` here

arc/species/converter.py
```diff
@@ -2011,6 +2011,43 @@ def compare_zmats(z1, z2, r_tol=0.01, a_tol=2, d_tol=2, verbose=False, symmetric
...
+        return fl_distance1, dmat1, conf2, simliar
+    simliar = True
+    dmat1 = xyz_to_dmat(xyz1)
```
alongd, line 2045: why do we compute dmat1 in this function?
This function also doesn't use `conf2['dmat']`, maybe we can simplify it?

arc/species/conformers.py
```diff
@@ -519,15 +519,23 @@ def conformers_combinations_by_lowest_conformer(label, mol, base_xyz, multiple_t
...
+                        break
                 if xyz is not None and energy is not None:
                     conformer = {'index': len_conformers + len(new_conformers) + len(newest_conformer_list),
                                  'xyz': xyz,
                                  'FF energy': round(energy, 3),
                                  'source': f'Changing dihedrals on most stable conformer, iteration {i}',
                                  'torsion': tor,
-                                 'dihedral': round(dihedral, 2)}
+                                 'dihedral': round(dihedral, 2),
+                                 'dmat': dmat1 if dmat1 is not None else None,
```
alongd, line 537: If `dmat1` either has a value or can be `None`, and you want its value or else `None`, then considering just keeping `'dmat': dmat1` instead of the entire line. Same for the line below

arc/species/conformers.py
```diff
@@ -1113,9 +1123,17 @@ def get_lowest_confs(label: str,
     for index in range(len(conformer_list)):
         if (e is not None and conformer_list[index][energy] > min_e + e) or (n is not None and len(lowest_confs) >= n):
             break
-        if index > 0 and not any([converter.compare_confs(lowest_conf['xyz'], conformer_list[index]['xyz'])
-                                  for lowest_conf in lowest_confs]):
-            lowest_confs.append(conformer_list[index])
+        if index>0:
```
alongd, line 1126: add a space before and after operators

alongd (commented):

arc/species/converter.py
JintaoWu98: The purpose of this function `compare_confs_fl`, is as following:
1. To compare two Cartesian coordinates representing conformers by examining the distances between the first and last atoms.
2. Distance matrices are computed and returned only if these distances are identical.

Considering potential modifications to the second purpose of this process, here are three options:
1. Rename the function to `compare_confs_pre` to indicate that it performs a preliminary, quicker comparison before the more resource-intensive compare_confs calculation.
2. Create a new function to do this calculation.
3. Merge this part to `compare_confs`, though feasible, I do not recommend this due to the complexity it would add, especially in returning the variable `conf2`, which could complicate the primary function.

arc/species/conformers.py
```diff
@@ -1113,9 +1123,15 @@ def get_lowest_confs(label: str,
     for index in range(len(conformer_list)):
         if (e is not None and conformer_list[index][energy] > min_e + e) or (n is not None and len(lowest_confs) >= n):
             break
-        if index > 0 and not any([converter.compare_confs(lowest_conf['xyz'], conformer_list[index]['xyz'])
-                                  for lowest_conf in lowest_confs]):
-            lowest_confs.append(conformer_list[index])
+        if index > 0:
```
JintaoWu98, line 1126: I found a mistake in the previous version, could you please review this part again? Thx!

arc/species/converter.py
alongd: OK

alongd (approved):
Looking good. Thanks!

PR #767, opened 2024-09-29
Define `set_rdkit_ring_dihedrals` and make test

JintaoWu98:
This PR introduces a new helper function, `set_rdkit_ring_dihedrals`, to the `converter.py` module. This function allows for setting dihedral angles in a ring structure using RDKit, facilitating the manipulation of molecular conformations.


alongd:
@JintaoWu98, what's the status of this PR? I don't see the new function being used, would you like to wait with it?

PR #777, opened 2024-12-06
Adapt parse_nd_scan_energies() to a slightly different Gaussian 2D scan format

alongd:
A test was added with a trimmed log file


JintaoWu98 (approved):
LGTM!

PR #784, opened 2025-04-08
Allow ARC to use mockter as a TS guess adapter

alongd:
will be used for functional tests of ARC and T3
Tests added

### DIFF
--- a/arc/job/adapters/mockter.py
+++ b/arc/job/adapters/mockter.py
@@ -122,8 +122,8 @@ class MockAdapter(JobAdapter):
         self.command = 'mockter'
         self.url = ''
 
-        if species is None:
-            raise ValueError('Cannot execute Mockter without an ARCSpecies object.')
+        if species is None and reactions is None:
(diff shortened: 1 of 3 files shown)


arc/job/adapters/mockter.py
```diff
@@ -238,7 +238,21 @@ def set_input_file_memory(self) -> None:
...
+        if self.xyz is not None:
+            return self.xyz
+        if self.species is not None and len(self.species):
+            return self.species[0].get_xyz()
+        if self.reactions is not None and len(self.reactions):
+            ts_xyz_str = ''
+            for r in self.reactions[0].r_species:
+                ts_xyz_str += xyz_to_str(r.get_xyz()) + '\n'
+            return str_to_xyz(ts_xyz_str)
+
```
OfirSimhi1612, line 255: We should probably have a fallback return here, or just return `None` or raise an error. Either way need to make sure we handle this case of no `xyz` returned

OfirSimhi1612 (approved):
small comment, otherwise LGTM!

PR #785, opened 2025-04-11
Don't gen xyz when checking if a species is monoatomic

alongd:
To avoid inf. recursions (is_monoatomic → get_xyz → get_cheap_conformer → is_monoatomic), don't generate a cheap xyz when checking if a species is monoatomic

### DIFF
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -1020,7 +1020,9 @@ class ARCSpecies(object):
         """
         if self.mol is not None and len(self.mol.atoms):
             return len(self.mol.atoms) == 1
-        xyz = self.get_xyz()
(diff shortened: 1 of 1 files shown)


arc/job/trsh.py
```diff
@@ -93,7 +94,7 @@ def determine_ess_status(output_path: str,
                     if 'l9999.exe' in line or 'link 9999' in line:
                         cycle_issue = False
                         neg_eigenvalues = False
-                        for j in range(i,len(reverse_lines)):
+                        for j in range(i, len(reverse_lines)):
```
OfirSimhi1612, line 97: nit: we can use `len_reversed_lines`

OfirSimhi1612 (approved):
LGTM!

PR #786, opened 2025-04-11
Added a feature that allows not to trsh rotors

alongd:
1D rotor scans for large systems fail too often. In some applications we'd like to compute rotors, but if they fail that means that more sophisticated methods should be applied. Currently ARC tries to troubleshoot the rotor runs, taking too long, and not always succeeding. This PR keeps the above default behavior, but adds an option to avoid trshing rotor runs


samcogan21:
LGTM!

samcogan21 (commented):
LGTM!

kfir4444 (approved):
LGTM, maybe some tests are needed?

PR #789, opened 2025-05-09
Use the molecule repo instead of RMG for objects

alongd:


alongd:
Comment to reviewers: the xTB test passes locally for me, but fails online. Let me know if you have any suggestions.

arc/species/converter.py
```diff
@@ -1588,13 +1588,13 @@ def set_radicals_by_map(mol, radical_map):
         atom.radical_electrons = radical_map.atoms[i].radical_electrons
 
 
-def order_atoms_in_mol_list(ref_mol, mol_list) -> bool:
+def order_atoms_in_mol_list(ref_mol: Molecule, mol_list: List[Molecule] | None) -> bool:
```
kfir4444, line 1458: again, let's try to be consistant with the typing

devtools/delete.sh
```diff
@@ -0,0 +1,43 @@
...
+for env in "${ENVS_TO_REMOVE[@]}"; do
+    if conda env list | grep -qw "$env"; then
+        echo "🗑️ Removing conda environment: $env"
+        conda remove -n "$env" --all -y
+    else
+        echo "ℹ️ Conda environment '$env' not found."
+    fi
+done
+
+echo "✅ Clean completed."
```
kfir4444, line 43: Why do we need that script?

arc/species/converter.py
```diff
@@ -1091,7 +1090,7 @@ def get_zmat_param_value(coords: Dict[str, tuple],
                          indices: List[int],
                          mol: Molecule,
                          index: int = 0,
-                         ) -> float:
+                         ) -> float | None:
```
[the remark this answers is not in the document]
calvinp0: I think it's a stylistic choice that came in with recent Python versions. I have seen it more and more use. Even linting extensions I use are suggesting it now

I also checked on PyLint and saw this

https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/consider-alternative-union-syntax.html

Seems to do with reducing imports I guess

arc/scripts/save_arkane_thermo.py
```diff
@@ -0,0 +1,52 @@
...
+"""
+
+import os
+
+from common import save_yaml_file
+
+from rmgpy.data.thermo import ThermoLibrary
+from rmgpy.thermo import NASAPolynomial, NASA, ThermoData, Wilhoit
+
+RT = 298.15
```
kfir4444, line 15: May be confusing (RT as in PV=nRT)...

arc/species/vectors.py
```diff
@@ -76,20 +76,45 @@ def get_dihedral(v1: List[float],
...
+    if np.linalg.norm(v1) == 0.0 or np.linalg.norm(v2) == 0.0 or np.linalg.norm(v3) == 0.0:
+        raise VectorsError('Input vectors must be non-zero.')
+
+    # 3) compute normals
     v2_x_v1 = np.cross(v2, v1)
-    v2_x_v1 /= float(np.linalg.norm(v2_x_v1))
+    norm1 = np.linalg.norm(v2_x_v1)
+    if norm1 < 1e-8:
+        # colinear → zero dihedral (but inputs were non-zero so no error)
+        return 0.0
```
kfir4444, line 96: That can cause a lot of problems for mapping, etc. This should throw an error, we don't have a perfect solution for that yet, and it should be handled by the caller for this function. Also, if anything, this should approach infinity, not 0.0

arc/species/converter.py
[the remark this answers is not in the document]
alongd: Thanks for the discussion guys. It was done consciously. When I wrote new funcs in existing modules I used the `Optional[]` style. Since this file is a new module altogether, I chose the modern style, and stayed consistent within this module

arc/species/converter.py
alongd: same module (converter.py), so I chose for it to be self-consistent

arc/species/vectors.py
```diff
@@ -91,15 +92,14 @@ def get_dihedral(v1: List[float],
     # 3) compute normals
     v2_x_v1 = np.cross(v2, v1)
     norm1 = np.linalg.norm(v2_x_v1)
-    if norm1 < 1e-8:
-        # colinear → zero dihedral (but inputs were non-zero so no error)
-        return 0.0
+    if norm1 < tol:
```
kfir4444, line 95: I think this should also take into account norm2 also, because if they are both small then it's fine.

arc/molecule/vf2.pyx
```diff
@@ -0,0 +1,399 @@
+"""
+This module contains graph ismorphism functions that implement the VF2
+algorithm of Vento and Foggia.  https://doi.org/10.1109/TPAMI.2004.75
+"""
+
+from arc.exceptions import VF2Error
+from arc.molecule.graph cimport Graph
+
+
+cdef class VF2:
```
kfir4444, line 10: There exist some very nice and optimized implementation of this:
https://networkx.org/documentation/stable/reference/algorithms/isomorphism.html
With the (apparently) better VF2++, I think it should be easier than to implement another module. Also, this could be quite nice if we implement networkx graphs in ARCSpecies, this could really help with ML applications, etc.

arc/molecule/adjlist.py
```diff
@@ -0,0 +1,754 @@
...
+
+    @staticmethod
+    def check_partial_charge(atom):
+        """
+        Checks whether the partial charge attribute of the atom checks out with
+        the theoretical one:
+
+        """
+        if atom.symbol in ['X','Li']:
+            return  # because we can't check it.
```
kfir4444, line 69: I think we should rethink the return statement. If the function doesn't return anything we should be consistent.

arc/molecule/vf2.pyx
alongd: We could think about it, and implement if needed. For now, I'm following the RMG module with only minimal modifications, so if we choose to upgrade the `molecule` module as RMG grows by incorporating their modifications, then it's relatively easy. But I'm open to modifying VF2 in the future. I'll avoid adding another modification (relative to RMG) into this PR.

arc/molecule/adjlist.py
```diff
@@ -0,0 +1,754 @@
...
+                                                'radicals {1}.'.format(multiplicity, n_rad))
+        elif n_rad == 2:
+            if not int(multiplicity) in [1, 3]:
+                raise InvalidAdjacencyListError('Multiplicity {0} not in agreement with total number of '
+                                                'radicals {1}.'.format(multiplicity, n_rad))
+        elif n_rad == 3:
+            if not int(multiplicity) in [4, 2]:
+                raise InvalidAdjacencyListError('Multiplicity {0} not in agreement with total number of '
+                                                'radicals {1}.'.format(multiplicity, n_rad))
+        elif n_rad == 4:
```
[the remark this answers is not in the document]
alongd: This is again taken directly from RMG. Would you like to suggest something better?

arc/constants.pxd
```diff
@@ -0,0 +1 @@
+cdef double pi, Na, kB, R, h, hbar, c, e, m_e, m_p, m_n, amu, a0, E_h, F, E_h_kJmol, bohr_to_angstrom
```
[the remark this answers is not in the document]
alongd: I think the reason has to do with the cythinozed modules that import these constants. Here i mimicked the [RMG implementaiton](https://github.com/ReactionMechanismGenerator/RMG-Py/blob/main/rmgpy/constants.pxd)

arc/common_test.py
```diff
@@ -867,6 +867,21 @@ def test_determine_symmetry(self):
...
+                self.molecule = molecule
+        arc_species = ARCSpecies(label='test_species', smiles='C')
+        mock_species = MockSpecies(label='test_species', molecule=[arc_species.mol])
+        self.assertFalse(common.is_obj_of_rmg_species_type(arc_species))
+        self.assertTrue(common.is_obj_of_rmg_species_type(mock_species))
+        self.assertFalse(common.is_obj_of_rmg_species_type('not a species'))
+        self.assertFalse(common.is_obj_of_rmg_species_type(12345))
+        self.assertFalse(common.is_obj_of_rmg_species_type(None))
+
+
```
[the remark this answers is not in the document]
alongd: I agree, I think it was fixes in another commit on this branch

arc/parser/parser.py
```diff
@@ -0,0 +1,516 @@
...
+                line_index += 1
+            while len(lines) and line_index < len(lines) and 'conformer' not in lines[line_index]:
+                if 'relative energy:' in lines[line_index].lower():
+                    energy = float(lines[line_index].split()[2])
+                line_index += 1
+            xyzs.append(str_to_xyz(xyz))
+            energies.append(energy)
+        else:
+            line_index += 1
+    return xyzs, energies
```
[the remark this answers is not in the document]
alongd: In my version parser.py ends with an empty line. Maybe GitHub doesn't show it

arc/processor.py
```diff
@@ -230,13 +246,25 @@ def compare_thermo(species_for_thermo_lib: list,
...
+    stdout, stderr = execute_command(command=commands, no_fail=True)
+    if len(stderr):
+        logger.error(f'Error while running RMG thermo script: {stderr}')
     species_list = read_yaml_file(path=species_thermo_path)
     for original_spc, rmg_spc in zip(species_for_thermo_lib, species_list):
         h298, s298, comment = rmg_spc.get('h298', None), rmg_spc.get('s298', None), rmg_spc.get('comment', None)
         if h298 is not None and s298 is not None:
-            original_spc.rmg_thermo = {'H298': h298, 'S298': s298, 'comment': comment}
+            original_spc.rmg_thermo.H298 = h298
+            original_spc.rmg_thermo.S298 = s298
```
kfir4444, line 266: Minor: upper cases S

arc/species/perceive.py
```diff
@@ -0,0 +1,932 @@
...
+        multiplicity (int, optional): Spin multiplicity (2S+1). If None, inferred.
+        n_radicals (int, optional): Number of radical centers. If None, defaults to multiplicity-1.
+        n_fragments (int, optional): Expected number of fragments. Defaults to 1.
+        single_bond_tolerance (float): Tolerance for single bond length perception, default is 1.20.
+        is_fragment (bool): If True, treat the input as a fragment rather than a full molecule.
+
+    Returns:
+        Molecule | None: The perceived Molecule, or None if perception failed.
+    """
+    charge = charge or 0
```
kfir4444, line 53: Why not just add 0 as default?

arc/species/vectors.py
```diff
@@ -76,20 +76,45 @@ def get_dihedral(v1: List[float],
...
-        dihedral = 2 * np.pi - dihedral
-    if np.isnan(dihedral):
-        raise VectorsError(f'Could not calculate a dihedral angle for\nv1: {v1}, v2: {v2}, v3: {v3}.')
-    conversion = 180 / math.pi if 'degs' in units else 1
-    return float(dihedral * conversion)
+        dihedral = 2 * math.pi - dihedral
+
+    # 6) unit conversion
+    if 'degs' in units:
+        dihedral = math.degrees(dihedral)
```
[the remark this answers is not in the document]
kfir4444: I agree, but I also think the logic of the units should stay the same for now too

arc/species/perceive.py
```diff
@@ -0,0 +1,932 @@
...
+            if symbols in [('N', 'H'), ('H', 'N')]:
+                return 3
+            if symbols in [('C', 'H'), ('H', 'C')]:
+                return 2
+        if len(symbols) == 3:
+            # hard code for some common triatomic species
+            if symbols in [('C', 'H', 'H'), ('H', 'C', 'H'), ('H', 'H', 'C')]:
+                return 3
+    total_e = sum(NUMBER_BY_SYMBOL[s] for s in symbols) - total_charge
+    return 1 if total_e % 2 == 0 else 2
```
kfir4444, line 372: Wait, but for multiplicity 3 and 5 we also have an even number of electrons, why is this returning 1 always?

arc/constants.pxd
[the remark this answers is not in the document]
alongd: I think it is compiled. We have in setup.py: `Extension('arc.constants', ['arc/constants.py'], include_dirs=['.']),`

arc/processor.py
alongd: The attribute of `rmg_thermo` is S298, which the parameter in this code is `s298`, I think it's defined correctly

arc/species/perceive.py
alongd: currently not used anywhere else, so not critical (also good that it's defined in proximity to the test)

kfir4444 (approved):
LGTM, great work!

calvinp0 (approved):
LGTM. Great work!

PR #790, opened 2025-05-24
Run CCS/VTZ for [H] instead of more complex WF methods

alongd:
Methods such as `DLPNO-CCSD(T)` which is frequenctly used for kinetics due to it's efficiency and relatively good accuracy cannot be applied for `[H]`. This method relies on electron pairs for defining correlation spaces, and the H radical has only one electron.
Therefore, reactions such as `RH + H <=> R + H2` cannot be computed using this method. However, since the energy of H is relatively simple, it is "known" percisely or can be computed using another Coupled-Cluster method without the DLPNO (Domain-based Local Pair Natural Orbital) assumption. This PR runs a simple and quick CC job for H if a DLPNO method was requested for it.

### DIFF
--- a/arc/job/ssh.py
+++ b/arc/job/ssh.py
@@ -578,8 +578,7 @@ def check_job_status_in_stdout(job_id: int,
             return 'errored'
     elif servers[server]['cluster_soft'].lower() == 'htcondor':
         return 'running'
-    else:
-        raise ValueError(f'Unknown cluster software {servers[server]["cluster_soft"]}')
+    raise ValueError(f'Unknown cluster software {servers[server]["cluster_soft"]}')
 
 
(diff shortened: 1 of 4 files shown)


arc/job/ssh.py
```diff
@@ -539,7 +539,7 @@ def _create_dir(self, remote_path: str) -> None:
 def check_job_status_in_stdout(job_id: int,
                                stdout: Union[list, str],
                                server: str,
-                               ) -> str:
+                               ) -> Optional[str]:
```
OfirSimhi1612, line 542: nit: need to update the function doc string return type

OfirSimhi1612 (approved):
LGTM!

OfirSimhi1612 (approved):
LGTM

PR #800, opened 2025-09-27
Dockerfile Update 3.12

calvinp0:
Refactor Dockerfile to be updated for RMG-Py 3.9 and ARC 3.12
Refactored scripts for user transition from rmguser to mambauser
Added tests (smoke tests) that are small test to check that the Docker image has built properly with ARC and RMG


JintaoWu98 (approved):
LGTM!

PR #806, opened 2025-11-04
Updated the alias command for RMG

calvinp0:
Removed the `-jl` from the RMG command that is used in the Docker. RMG does not work the old julia connection anymore.


JintaoWu98:
Thanks, LGTM!

JintaoWu98 (approved):

PR #808, opened 2025-12-07
Added a .index attr to ARCSpecies

alongd:
Used by T3 for migrating away from RMG `Species()` and into `ARCSpecies()`


ranbenayoun (approved):
looks good

PR #817, opened 2026-01-07
Mutated conformer jobs issue

calvinp0:
> **Standalone PR, base `main`.** This was the bottom of stack #942; the stack has been dissolved.
> #940's audit is folded in here — the two covered the same defect class in the same files, so the
> boundary between them only created duplicated hunks. #941 and #943 are now independent PRs off
> `main`: they were written on top of this work but do not depend on it.

---

---

## What went wrong

`TSGuess.index` is a TS guess's **identity**: it is assigned when the guess is appended to the
species, it is what `chosen_ts` and `chosen_ts_list` refer to, and it is how a solved TS is
attributed to the adapter that produced it. It is *not* a position in a list. Five sites broke
that contract.

### 1. conf_opt result ingestion attributed results to the wrong guess

Every successful guess gets a conformer optimization job, numbered by its position in
`successful_tsgs`. Clustering shrinks and reorders that list relative to the full `ts_guesses`
list, but the `enumerate` position over the former was used to subscript the latter when writing
the optimized energy and geometry back:

```python
self.species_dict[label].ts_guesses[i].energy = energy   # i is a position in a *different* list
self.species_dict[label].ts_guesses[i].index = i         # ...and the identity is overwritten
```

So the optimized result landed on an unrelated guess, and the second line destroyed that guess's
identity as well. This is the failure originally reported: after dedup collapsed several similar
TS conformers, ARC wrote conf_opt results back using the pre-dedup integers and went out of range.

Guesses are now matched on `conformer_index`, and a missing match warns instead of mis-assigning.

### 2. The frequency check could accept a TS it never verified

`chosen_ts` stores a `TSGuess.index`, but `check_negative_freq()` matched on `conformer_index` —
the position among the *successful* guesses. It also relied on the loop variable leaking out of
the `for` loop, so a no-match left `tsg` bound to the last guess rather than to `None`. Because
`check_imaginary_frequencies(None)` returns `True` by design — a guess whose frequencies have not
been computed must not be disqualified during pre-screening — **a TS with zero, two or three
imaginary frequencies could be accepted as converged while its real frequencies were never
recorded.**

Matching is now by identity, through a new `get_chosen_tsg()` helper. A `chosen_ts` that matches
no guess is treated as an unverified frequency check: it logs the label, `chosen_ts` and the
available indices, and looks for a different TS conformer rather than silently accepting the
geometry.

When *no* guess is selected at all — which a repaired restart file can produce, see below — the
check is likewise left unverified, but the geometry is kept and no guess switch is made. Switching
there would discard a TS that may already have passed every check, which is the failure mode #941
exists to prevent.

### 3. Pipe ts_opt ingestion clobbered the identity

The same overwrite of `TSGuess.index` with the conformer index, on the pipe path.

### 4. The identity was allocated from a list length

Raised in review. Every allocation site spelled the next identity as
`len(ts_species.ts_guesses)`, which is the next free identity only while that list never shrinks.
`cluster_tsgs()` does shrink it, keeping the survivors' original — now sparse — indices:

```
before:          [0, 1, 2, 3]
after cluster:   [0, 2]                 len() == 2, and 2 is still in use
next allocation: index = len() == 2     ← duplicate identity
```

Reproduced against the real clustering path, including the user-xyz guess that carries no index
at all:

```
before:         [None, 0, 1, 2, 3]
after cluster:  [None, 2]
after append:   [None, 2, 2]
```

With two guesses holding index 2, `chosen_ts` silently resolves to whichever appears first, and
the `chosen_ts_list` filter that stops an exhausted guess being re-selected excludes both at once.

### 5. Three more sites used a position where an identity was meant

Folded in from #940. None of these changes which TS guess is selected.

| site | what it did |
|---|---|
| `Scheduler.troubleshoot_ess` | subscripted `ts_guesses` with the number parsed out of a `tsg<i>` job name. That number identifies the TS-search **adapter** dispatched, not a guess — an adapter may contribute several guesses or none — so it annotated an unrelated guess, and raised `IndexError`, aborting the run, whenever the adapter number exceeded the guesses generated so far. Now matched on the adapter that produced each guess. |
| `Scheduler.run_ts_conformer_jobs` | reported `ts_guesses[0]` as the provenance of the lone successful guess while optimizing `successful_tsgs[0]`'s coordinates. Several adapters append `success=False` guesses, so those are not the same object, and `chosen_ts_method` could name a method that produced no geometry. |
| `output._get_ts_imag_freq` | subscripted `ts_guesses` with `chosen_ts`, reporting a different guess's imaginary frequency whenever the two numberings diverged. |

## The fix

| site | before | after |
|---|---|---|
| conf_opt ingestion | `ts_guesses[i]`, then `.index = i` | look up by `conformer_index`; warn on no match |
| frequency check | match on `conformer_index`, leaked loop variable | `get_chosen_tsg()`, match on identity |
| pipe ts_opt ingestion | `.index = conformer_index` | identity left alone |
| identity allocation | `len(ts_species.ts_guesses)` | `ARCSpecies.next_ts_guess_index()` |

`next_ts_guess_index()` returns one past the largest index in use, so an identity is never reused
no matter how clustering has reshaped the list. It replaces the length-based allocation at all
fourteen sites: ten across nine TS adapters, one in `scheduler.py`, and three in `species.py`.
Three further holes are closed with it:

- a TS guess read from a queue job's YAML no longer keeps an index that is already taken;
- guesses built from user xyz input, which were assigned no index at all and were then reached by
  list position, are allocated one too. Without this, a user-supplied TS geometry is unselectable:
  `chosen_ts` is set from `tsg.index`, so it stays `None` and `determine_most_likely_ts_conformer()`
  reports "Could not determine a likely TS conformer";
- a restart file holding missing or duplicated indices is repaired on load. Only the ambiguous
  guesses are re-indexed, so references to unambiguous identities stay valid. Every reference to a
  duplicated index is then dropped, since none of them can be resolved — `chosen_ts` is reset so a
  TS conformer is selected again, and the duplicated indices are stripped from `chosen_ts_list` so
  that neither the guess that kept the index nor the guess that was re-indexed is barred from
  selection.

### 6. Frequencies that belong to no identifiable guess

Found by adversarial review of the first fix for #2, and reproduced independently twice. Leaving
the check "unverified" when no guess is selected **strands the species**: for a TS,
`check_freq_job()` neither troubleshoots nor respawns, `output['job_types']['freq']` stays `False`,
and `chosen_ts` is only ever assigned by `determine_most_likely_ts_conformer()` — which that path
declines to reach. Two routes hit it, neither exotic:

- a standalone TS given two or more xyz guesses, where `main` matched `None == None`, bound the
  first guess and returned `True` — so a TS that converges today would never converge;
- any repaired restart file, and the duplicated indices that trigger the repair were produced by
  the very code this PR removes, so **that is the common case for existing restarts**, not an edge
  case.

The frequencies still describe the geometry that was optimized, so the check is now decided on them
directly, without attributing them to a guess. A bad result still does not switch guesses.

### 7. A guess cannot be selected without an optimized geometry

`determine_most_likely_ts_conformer()` assigns `initial_xyz` from `TSGuess.opt_xyz`. The lone
successful guess is sent straight to a geometry optimization and now carries a relative energy of
0.0, which made it selectable on a later `switch_ts` with `opt_xyz` still `None` — blanking the
geometry and burning a full opt/freq/sp cycle on the raw guess. Selection now requires `opt_xyz`.

## Why it matters

The first three sites move a result, or a verdict, onto a guess it does not belong to. The fourth
makes the identity itself ambiguous, which is what the other three rely on being unique. Together
they mean a TS can be reported as converged on the strength of another guess's frequencies, and
the per-adapter attribution the benchmark is built from can silently point at the wrong method.

## Evidence

Regression tests accompany each changed site, written so that reverting the fix fails them:

| test | what it pins |
|---|---|
| `test_parse_conformer_attributes_ts_results_by_conformer_index` | list position, `index` and `conformer_index` all differ; results must follow `conformer_index`, identities must survive |
| `test_parse_conformer_warns_when_no_ts_guess_matches` | an unmatched result touches nothing and is logged |
| `test_check_negative_freq_*` (3) | frequencies are attributed by identity; two imaginary frequencies are rejected; an unmatched `chosen_ts` is logged, not accepted |
| `TestIngestTsOpt` (3) | pipe ingestion with `index=7, conformer_index=0` preserves the identity and updates only the matching guess |
| `test_next_ts_guess_index_survives_clustering` | the allocator never returns an index still in use after clustering |
| `test_process_completed_tsg_queue_jobs_keeps_indices_unique` | a YAML-supplied index that is already taken is reallocated |
| `test_renumber_ambiguous_ts_guesses_on_restart` | duplicated indices are repaired on load, an ambiguous `chosen_ts` is reset, and the ambiguous entries are stripped from `chosen_ts_list` while unambiguous ones survive |
| `test_process_xyz_assigns_ts_guess_indices` | user guesses get an identity, continuing past the indices in use (`0, 4` → `5`) |
| `test_check_negative_freq_no_chosen_ts_leaves_the_check_unverified` | freqs arriving with no selection leave the check unverified and do **not** switch guesses |

**374 tests pass** (`species_test`, `scheduler_test`, `output_test`, `job/pipe`), run serially — this suite is
flaky under `xdist` because its workers collide on shared `Projects/` directories.

### DIFF
--- a/arc/job/adapters/ts/autotst_ts.py
+++ b/arc/job/adapters/ts/autotst_ts.py
@@ -277,9 +277,8 @@ class AutoTSTAdapter(JobAdapter):
                                                        execution_time=tok,
                                                        xyz=xyz,
                                                        success=True,
-                                                       index=len(rxn.ts_species.ts_guesses),
                                                        )
-                                    rxn.ts_species.ts_guesses.append(ts_guess)
+                                    rxn.ts_species.append_ts_guess(ts_guess)
                                     save_geo(xyz=xyz,
                                              path=self.local_path,
                                              filename=f'AutoTST {direction}',
@@ -294,9 +293,8 @@ class AutoTSTAdapter(JobAdapter):
                                                t0=tic,
                                                execution_time=tok,
                                                success=False,
-                                               index=len(rxn.ts_species.ts_guesses),
                                                )
-                            rxn.ts_species.ts_guesses.append(ts_guess)
+                            rxn.ts_species.append_ts_guess(ts_guess)
                             i += 1
 
             if len(self.reactions) < 5:
--- a/arc/job/adapters/ts/gcn_ts.py
+++ b/arc/job/adapters/ts/gcn_ts.py
@@ -378,7 +378,6 @@ def run_subprocess_locally(direction: str,
     ts_xyz = None
     tsg = TSGuess(method='GCN',
                   method_direction=direction,
-                  index=len(ts_species.ts_guesses),
                   )
     tsg.tic()
     # Routed via run_in_conda_env so arc_env's activation vars don't
@@ -434,16 +433,18 @@ def process_tsg(direction: str,
         if unique:
             tsg.success = True
             tsg.process_xyz(ts_xyz)
-            save_geo(xyz=ts_xyz,
-                     path=local_path,
-                     filename=f'GCN {direction} {tsg.index}',
-                     format_='xyz',
-                     comment=f'GCN {direction}',
-                     )
     else:
         tsg.success = False
     if unique and tsg.success:
-        ts_species.ts_guesses.append(tsg)
+        # Appended before the geometry is saved, so the file is named with the identity the guess
+        # actually ends up holding rather than one predicted before the append.
+        ts_species.append_ts_guess(tsg)
+        save_geo(xyz=ts_xyz,
+                 path=local_path,
+                 filename=f'GCN {direction} {tsg.index}',
+                 format_='xyz',
+                 comment=f'GCN {direction}',
+                 )
 
 
 register_job_adapter('gcn', GCNAdapter)
--- a/arc/job/adapters/ts/goflow_ts.py
+++ b/arc/job/adapters/ts/goflow_ts.py
@@ -623,11 +623,10 @@ def process_goflow_tsg(tsg_dict: dict,
     tsg = TSGuess(method='GoFlow',
                   method_direction=tsg_dict.get('method_direction', 'F'),
                   method_index=method_index,
-                  index=len(ts_species.ts_guesses),
                   success=True,
                   )
     tsg.process_xyz(ts_xyz)
-    ts_species.ts_guesses.append(tsg)
+    ts_species.append_ts_guess(tsg)
     save_geo(xyz=ts_xyz,
              path=local_path,
              filename=f'GoFlow {method_index}',
--- a/arc/job/adapters/ts/heuristics.py
+++ b/arc/job/adapters/ts/heuristics.py
@@ -288,7 +288,6 @@ class HeuristicsAdapter(JobAdapter):
                         break
                 if unique:
                     ts_guess = TSGuess(method='Heuristics',
-                                       index=len(rxn.ts_species.ts_guesses),
                                        method_index=method_index,
                                        t0=tsg.t0,
                                        execution_time=tsg.execution_time,
@@ -296,7 +295,7 @@ class HeuristicsAdapter(JobAdapter):
                                        family=rxn.family if families is None else families[method_index],
                                        xyz=xyz,
                                        )
-                    rxn.ts_species.ts_guesses.append(ts_guess)
+                    rxn.ts_species.append_ts_guess(ts_guess)
                     save_geo(xyz=xyz,
                              path=self.local_path,
                              filename=f'Heuristics_{method_index}',
--- a/arc/job/adapters/ts/kinbot_ts.py
+++ b/arc/job/adapters/ts/kinbot_ts.py
@@ -380,7 +380,6 @@ class KinBotAdapter(JobAdapter):
             ts_guess = TSGuess(method=method,
                                method_direction=method_direction,
                                method_index=method_index,
-                               index=len(rxn.ts_species.ts_guesses),
                                execution_time=result.get('execution_time'),
                                )
             success = bool(result.get('success')) and result.get('coords') is not None
@@ -409,7 +408,7 @@ class KinBotAdapter(JobAdapter):
             if not success:
                 ts_guess.success = False
             if unique:
-                rxn.ts_species.ts_guesses.append(ts_guess)
+                rxn.ts_species.append_ts_guess(ts_guess)
                 method_index += 1
 
     def execute_queue(self):
--- a/arc/job/adapters/ts/linear.py
+++ b/arc/job/adapters/ts/linear.py
@@ -926,7 +926,6 @@ class LinearAdapter(JobAdapter):
                         if unique:
                             method = f'Linear (w={w:.2f}, {xyz_i})'
                             ts_guess = TSGuess(method=method,
-                                               index=len(rxn.ts_species.ts_guesses),
                                                method_index=w_i,
                                                t0=t0,
                                                execution_time=t_ex,
@@ -934,7 +933,7 @@ class LinearAdapter(JobAdapter):
                                                family=rxn.family,
                                                xyz=xyz,
                                                )
-                            rxn.ts_species.ts_guesses.append(ts_guess)
+                            rxn.ts_species.append_ts_guess(ts_guess)
 
                             save_geo(xyz=xyz,
                                      path=self.local_path,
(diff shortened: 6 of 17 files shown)


alongd (commented):
thanks @calvinp0 , LGTM!
please squash and remove the last "merge" commit?

alongd:
Review of `2ffffdd`. All three review threads are genuinely addressed in code — no `tsg.index = …` clobber remains in either ingestion path, and `next_ts_guess_index()` replaces all 11 length-based allocation sites with `renumber_ambiguous_ts_guesses()` repairing duplicates on load. Three things to fix before this goes in; (1) is the blocker.

### 1. The last push reverted the user-xyz identity allocation (blocker)

`2ffffdd` restored main's version of the user-guess block in `process_xyz` (`arc/species/species.py:1926`), dropping `index=self.next_ts_guess_index()`, and deleted `test_user_guesses_are_assigned_an_identity`. It also reinstated the redundant `self.ts_guesses[tsg_index].success = True` immediately after `success=True` was passed to the constructor. This looks like collateral from resolving the squash against main rather than a decision.

`renumber_ambiguous_ts_guesses()` cannot cover for it: `from_dict` calls it at `species.py:904`, *before* `process_xyz` at 905 creates those guesses. Measured at head, both construction paths:

```
direct ctor      indices: [None, None]
from input dict  indices: [None, None]
```

Driving `determine_most_likely_ts_conformer()` on two user guesses with energies set:

| commit | result |
|---|---|
| `1f71aa8` (previous push) | `indices: [0, 1]` → `chosen_ts: 0`, `initial_xyz` set |
| `2ffffdd` (head) | `indices: [None, None]` → `Could not determine a likely TS conformer for TS0`, `chosen_ts: None` |

because `selected_i = tsg.index` at `arc/scheduler.py:2410` is `None`. To be fair to the change: main behaves identically, so this is a pre-existing break of the user-supplied-TS-xyz path, not a regression this PR introduces — but the commit message still claims *"Guesses built from user xyz input … are allocated one too"*, and after the revert it doesn't. CI is green only because the test that covered it went out in the same push.

**Fix:** restore `index=self.next_ts_guess_index()` and the deleted test; drop the redundant `success = True` line while you're there.

### 2. `renumber_ambiguous_ts_guesses()` leaves the duplicated index in `chosen_ts_list`

The repair resets `chosen_ts` but not `chosen_ts_list` (the new test asserts `chosen_ts_list == [2]`). Re-selection filters on `tsg.index not in chosen_ts_list` (`arc/scheduler.py:2408`), so the guess that *keeps* index 2 — as likely as any to be the one actually chosen and optimized — is permanently barred, and only the re-indexed duplicate stays eligible. The repair moves the ambiguity rather than removing it. Strip the repaired indices from `chosen_ts_list` as well.

### 3. `get_chosen_tsg()`'s `chosen_ts is None` fallback can discard a validated TS

The fallback returns the sole successful guess (good), but returns `None` when 0 or ≥2 guesses succeeded, which now routes into the new warn-and-`switch_ts` branch. Combined with (2), a restart-repaired TS — `chosen_ts` reset to `None`, several successful guesses — whose freq job lands afterwards will discard a validated geometry and switch guesses, which is the failure mode #941 exists to prevent. Make the `chosen_ts is None` case leave the frequency check unverified (skip) rather than switch.

### Non-blocking follow-ups

- `arc/output.py:537` (`chosen < len(spc.ts_guesses)`) and `arc/scheduler.py:3799` (`ts_guesses[get_i_from_job_name(job.job_name)]`) still use a position where an identity is meant. The first gets worse here: `next_ts_guess_index()` and the renumbering can produce indices ≥ `len(ts_guesses)`, so its guard silently returns `None` instead of the TS imaginary frequency.
- The no-match branch calls `switch_ts()` but `post_freq_actions` returns `(False, False)`, whose second element is documented as "whether a different TS guess was selected as a result". Latent only — `scheduler.py:2715` discards it and `pipe_coordinator.py:378` ignores the return.
- `renumber_ambiguous_ts_guesses` has no return annotation (repo standard: type hints on all signatures, params and return).
- `pipe_run_test.TestIngestTsOpt` mocks a path that `scheduler_test.test_parse_conformer_attributes_ts_results_by_conformer_index` drives with real log files; the repo standard prefers real objects, and the real-object route is demonstrated three files over.
- `index=<spc>.next_ts_guess_index()` … `ts_guesses.append(tsg)` now appears at ten sites across nine adapters plus the scheduler. An `ARCSpecies.append_ts_guess(tsg)` that assigns the index on append would collapse them and stop a future adapter reintroducing `len(...)`.

alongd (approved):

PR #835, opened 2026-03-16
Increase ARC's code coverage statistics.

kfir4444:
This PR related to the low (~67%) code coverage with the tests, since the inclusion of the molecule branch.
The lowered coverage is caused by the molecule module being unreachable to codecov due to the compilation. This is evident in codecov, where it seems that tests passes 100% but the modules related to the tests appear at 0.0%. This is solved by the using the `CYTHON_TRACE` macro. Though, due to it lowereing performence by ~10x, it is enabled only during CI via costum environment flag.
In addition, tests were added to the statmech module (specifically ARKANE), species, utils and job modules.

### DIFF
--- a/arc/job/factory_test.py
+++ b/arc/job/factory_test.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+# encoding: utf-8
+
+"""
+This module contains unit tests for ARC's factories
+"""
+
+import os
+import shutil
+import unittest
+from unittest.mock import patch
+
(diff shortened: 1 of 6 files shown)


alongd:
Can you explain the `lowering performance by ~10x` part?

alongd (commented):
Thanks! I left some comments.
I think we should avoid mock tests whenever possible, and use our real object, unless there's an operational reason to do otherwise

PR #839, opened 2026-03-26
Several more atom mapping improvements

kfir4444:
This PR improves several aspects of atom mappings.
two key fixes:
1) Stopping conformer generation in `ARCSpecies._scissors`. This is a major issue that improves both accuracy and performence:
    1) Performence - less RDKit calls
    2) Accuracy - Generating conformers distorts the XYZ of the scissored products.
    Moreover, this is critical for cyclic species, since debugging showed we did not include the original XYZ in the ARCSpecies, which caused errors.
2)  the `identify_superimposable_candidates`, which was being called $\mathcal{O}(N^2)$ times and reduces to $\mathcal{O}(N)$. More calls are not required, and does not increase accuracy. (There is some more math behind it, and it can be discussed offline @alongd).

Other fixes added, and more generalization of AM testing (manually verified)

### DIFF
--- a/arc/family/family.py
+++ b/arc/family/family.py
@@ -24,7 +24,7 @@ ARC_FAMILIES_PATH = settings['ARC_FAMILIES_PATH']
 logger = get_logger()
 
 
-REACTION_FAMILY_CACHE: Dict[Tuple[str, bool], 'ReactionFamily'] = {}
+REACTION_FAMILY_CACHE: dict[tuple[str, bool], 'ReactionFamily'] = {}
 
 # Pre-compiled regex patterns
 ENTRY_PATTERN = re.compile(r'entry\((.*?)\)', re.DOTALL)
@@ -139,11 +139,12 @@ class ReactionFamily(object):
         self.groups_as_lines = read_groups_file_lines(label, consider_arc_families)
         self.reversible = is_reversible(self.groups_as_lines)
         self.own_reverse = is_own_reverse(self.groups_as_lines)
+        
         reactant_labels = get_initial_reactant_labels_from_template(self.groups_as_lines)
-        temp_entries = get_entries(self.groups_as_lines, entry_labels=reactant_labels, recursive=True)
-        self.reactants = get_reactant_groups_from_template(self.groups_as_lines, entries=temp_entries)
-        leaf_labels = [label for group in self.reactants for label in group]
-        self.entries = {label: temp_entries[label] for label in leaf_labels if label in temp_entries}
+        all_necessary_entries = get_entries(self.groups_as_lines, entry_labels=reactant_labels, recursive=True)
+        self.reactants = get_reactant_groups_from_template(self.groups_as_lines, entries=all_necessary_entries)
+        self.entries = all_necessary_entries
+        
         self.groups = {}
         for reactant_group in self.reactants:
             for label in reactant_group:
@@ -196,6 +197,9 @@ class ReactionFamily(object):
                 for group_label in group_labels:
                     group = self.groups_by_label[group_label]
                     for mol in reactant.mol_list or [reactant.mol]:
+                        if not any(a.atomtype for a in mol.atoms):
+                            # Update atomtypes if they are missing (e.g., from SMILES)
+                            mol.update_atomtypes(log_species=False, raise_exception=False)
                         splits = group.split()
                         if mol.is_subgraph_isomorphic(other=group, save_order=True) \
                                 or len(splits) > 1 and any(mol.is_subgraph_isomorphic(other=g, save_order=True) for g in splits):
@@ -337,9 +341,15 @@ class ReactionFamily(object):
                         group_2 = self.groups_by_label[reactant_to_group_map_2['subgroup']]
                         isomorphic_subgraphs_1 = mol_1.find_subgraph_isomorphisms(other=group_1, save_order=True)
                         isomorphic_subgraphs_2 = mol_2.find_subgraph_isomorphisms(other=group_2, save_order=True)
+                        
                         if len(isomorphic_subgraphs_1) and len(isomorphic_subgraphs_2):
                             for isomorphic_subgraph_1 in isomorphic_subgraphs_1:
                                 for isomorphic_subgraph_2 in isomorphic_subgraphs_2:
+                                    # Create the combined isomorphic subgraph.
+                                    # Note: get_isomorphic_subgraph needs to know which subgraph corresponds to which template index.
+                                    # It assumes mol_1 corresponds to the first group match and mol_2 to the second.
+                                    # The labels are already inside the group_atom.label.
+                                    
                                     isomorphic_subgraph_dicts.append(
                                         {'mols': [mol_1, mol_2],
                                          'subgroups': (reactant_to_group_map_1['subgroup'],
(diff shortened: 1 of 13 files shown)


kfir4444:
The previous error in CI was caused by `get_entries` incorrectly including intermediate Logical OR groups (e.g., `OR{X_H, Xrad_H}`) in the result dictionary when recursing through the RMG database hierarchy. Fixed now.

arc/mapping/engine.py
```diff
@@ -1194,12 +1189,19 @@ def pairing_reactants_and_products_for_mapping(r_cuts: list[ARCSpecies],
...
+    pairs: List[Tuple[ARCSpecies, ARCSpecies]] = list()
+    r_res = [generate_resonance_structures_safely(react.mol, save_order=True) for react in r_cuts]
+    for i, react in enumerate(r_cuts):
+        res1 = r_res[i]
         for idx, prod in enumerate(p_cuts):
-            if r_cut_p_cut_isomorphic(react, prod):
-                pairs.append((react, prod))
-                p_cuts.pop(idx)
+            found = False
+            for res in res1:
```
alongd, line 1233: can res1 be `None`? Shall we protect against it?

arc/mapping/engine.py
```diff
@@ -1194,12 +1189,19 @@ def pairing_reactants_and_products_for_mapping(r_cuts: list[ARCSpecies],
     Returns:
         list[tuple[ARCSpecies,ARCSpecies]]: A list of paired reactant and products, to be sent to map_two_species.
     """
-    pairs: list[tuple[ARCSpecies, ARCSpecies]] = list()
-    for react in r_cuts:
+    pairs: List[Tuple[ARCSpecies, ARCSpecies]] = list()
+    r_res = [generate_resonance_structures_safely(react.mol, save_order=True) for react in r_cuts]
+    for i, react in enumerate(r_cuts):
+        res1 = r_res[i]
```
alongd, line 1230: `generate_resonance_structures_safely()` return None on errors, need to see that this is safe

arc/family/family.py
```diff
@@ -125,9 +167,11 @@ def generate_products(self,
         for reactant_idx, reactant in enumerate(reactants):
             for groups_idx, group_labels in enumerate(self.reactants):
                 for group_label in group_labels:
-                    group = Group().from_adjacency_list(
-                        get_group_adjlist(self.groups_as_lines, entry_label=group_label))
+                    group = self.groups[group_label]
                     for mol in reactant.mol_list or [reactant.mol]:
+                        if not any(a.atomtype for atom in (reactant.mol_list or [reactant.mol]) for a in atom.atoms):
```
alongd: will this `any()` always return True (so skipping the action below because of the `not any`) after the first pass?

arc/family/family.py
```diff
@@ -63,13 +96,22 @@ def __init__(self,
...
+        self.reactants = get_reactant_groups_from_template(self.groups_as_lines, entries=all_necessary_entries)
+        
+        # self.entries should contain all entries that were traversed to find the leaves
+        self.entries = all_necessary_entries
+        
+        self.groups = {}
+        for reactant_group in self.reactants:
+            for label in reactant_group:
+                if label not in self.groups and label in self.entries:
+                    self.groups[label] = Group().from_adjacency_list(self.entries[label])
```
alongd, line 152: should we first check that `label` is in `self.groups` to be safe? (same for other additions in this file)

arc/species/species.py
```diff
@@ -1992,13 +1992,23 @@ def _scissors(self,
                 sort_atoms_in_descending_label_order(split)
 
         if len(mol_splits) == 1:  # If cutting leads to only one split, then the split is cyclic.
+            mol1 = mol_splits[0]
+            for atom in mol1.atoms:
+                theoretical_charge = elements.PeriodicSystem.valence_electrons[atom.symbol] \
```
alongd: this looks very familiar to me, I'm almost certain we have this code elsewhere in ARC. Probably worth extracting in all places into a common tested function

arc/mapping/engine.py
```diff
@@ -1194,12 +1189,19 @@ def pairing_reactants_and_products_for_mapping(r_cuts: list[ARCSpecies],
     Returns:
         list[tuple[ARCSpecies,ARCSpecies]]: A list of paired reactant and products, to be sent to map_two_species.
     """
-    pairs: list[tuple[ARCSpecies, ARCSpecies]] = list()
-    for react in r_cuts:
+    pairs: List[Tuple[ARCSpecies, ARCSpecies]] = list()
```
alongd: Are `List` and `Tuple` imported? We transitions into Python 3.14, these should simply be `list` and `tuple` (and nor imported now)
Please check other occurrences in this PR and adapt

arc/family/family.py
```diff
@@ -2,7 +2,9 @@
 A module for working with RMG reaction families.
 """
 
-from typing import TYPE_CHECKING
+from __future__ import annotations
+
+from typing import Dict, List, Optional, Tuple, TYPE_CHECKING
```
alongd: We don't need those in Python 3.14. remove (just keep `TYPE_CHECKING` as in the original) and rebuild the env?

arc/family/family.py
```diff
@@ -63,13 +96,22 @@ def __init__(self,
...
         self.reversible = is_reversible(self.groups_as_lines)
         self.own_reverse = is_own_reverse(self.groups_as_lines)
-        self.reactants = get_reactant_groups_from_template(self.groups_as_lines)
+        
+        reactant_labels = get_initial_reactant_labels_from_template(self.groups_as_lines)
+        # Use a temporary dict to load all necessary entries recursively
+        all_necessary_entries = get_entries(self.groups_as_lines, entry_labels=reactant_labels, recursive=True)
+        self.reactants = get_reactant_groups_from_template(self.groups_as_lines, entries=all_necessary_entries)
+        
+        # self.entries should contain all entries that were traversed to find the leaves
```
alongd: remove comments anywhere the code is self explanatory (better)

alongd (commented):

arc/family/family.py
kfir4444: I need to read more about the new way python deals with typing. For now, I'm reverting these.

arc/species/species_test.py
```diff
@@ -2907,6 +2907,49 @@
...
+        mol = Molecule().from_smiles('C1CC1')
+        
+        # Find a C-C bond to remove to simulate a ring opening
+        for bond in mol.get_all_edges():
+            if bond.atom1.is_carbon() and bond.atom2.is_carbon():
+                c1, c2 = bond.atom1, bond.atom2
+                mol.remove_bond(bond)
+                break
+        
+        self.assertEqual(c1.radical_electrons, 0)
```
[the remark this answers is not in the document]
kfir4444: If c2 or c1 are none, I want this test to break.

arc/family/family.py
```diff
@@ -23,6 +25,37 @@
 logger = get_logger()
 
 
+REACTION_FAMILY_CACHE: Dict[Tuple[str, bool], 'ReactionFamily'] = {}
```
alongd: please change `Dict` and `Tuple` to `dict` and `tuple`, and remove from __future__ import annotations

please check that all type annotations in this PR align with Py 3.14

arc/family/family.py
```diff
@@ -868,45 +922,79 @@ def get_recipe_actions(groups_as_lines: list[str]) -> list[list[str]]:
...
-    for i, entry in enumerate(entries):
-        label_match = re.search(r'label = "(.*?)"', entry)
-        group_match = re.search(r'group =(.*?)(?=\w+ =)', entry, re.DOTALL)
-        if label_match is not None and group_match is not None and label_match.group(1) in entry_labels:
-            specific_entries[label_match.group(1)] = clean_text(group_match.group(1))
-        if i > 2000:
-            break
-    return specific_entries
+    groups_str = "\n" + "".join(groups_as_lines)
+    # Split by entry( but keep the delimiter-ish part
```
alongd: see the comment, seems broken? "("

arc/mapping/engine.py
alongd: I think this comment wasn't addressed. take a look?

arc/family/family.py
kfir4444: No, that's the regex expression (L942)

arc/mapping/engine.py
kfir4444: I think it was see L#1193

alongd (dismissed):
Thanks!

alongd (approved):

PR #855, opened 2026-04-01
Refactor ND scan subsystem with adaptive sparse 2D sampling, surface validation, and coupling classification

alongd:
- New module arc/species/nd_scan.py: Extracts ND scan logic from scheduler.py into a dedicated subsystem with clean interfaces for grid generation, continuous scan state, adaptive sparse 2D sampling, surface validation, coupling metrics, and rotor classification
- Adaptive sparse 2D scans: For eligible brute_force_sp/brute_force_opt 2D rotors, ARC can now run batched adaptive sampling instead of the full Cartesian grid, controlled by a sampling_policy field on the rotor — no new directed_scan_type values are introduced 
- Surface validation: Neighbor-edge continuity checks, periodic boundary consistency, and branch-jump detection diagnose PES discontinuities without blocking finalization
- Coupling analysis and classification: Separable-fit residual analysis classifies each adaptive 2D rotor as separable, coupled, or unreliable with an explicit recommended action
- Sparse-aware 2D plotting: plot_2d_rotor_scan dispatches to an interpolation-based sparse contour path for adaptive scans, with sampled/failed/invalid point overlays
- xTB directed_scan fix: The xTB adapter now correctly runs --opt for brute_force_opt directed scan jobs (was silently running SP)
- Bug fix: Corrected parse_e_elect(path=...) → parse_e_elect(log_file_path=...) kwarg in check_directed_scan_job

### DIFF
--- a/arc/job/adapters/xtb_adapter.py
+++ b/arc/job/adapters/xtb_adapter.py
@@ -218,7 +218,9 @@ class xTBAdapter(JobAdapter):
         directives, block = '', ''
         uhf = self.species[0].number_of_radicals or self.multiplicity - 1
 
-        if self.job_type in ['opt', 'conf_opt', 'scan']:
+        if self.job_type in ['opt', 'conf_opt', 'scan'] \
+                or (self.job_type == 'directed_scan' and self.directed_scan_type is not None
+                    and 'opt' in self.directed_scan_type):
             directives += ' --opt'
             directives += self.add_accuracy()
             if self.constraints and self.job_type != 'scan':
--- a/arc/plotter.py
+++ b/arc/plotter.py
@@ -1296,6 +1296,12 @@ def plot_2d_rotor_scan(results: dict,
     if len(results['scans']) != 2:
         raise InputError(f'results must represent a 2D rotor, got {len(results["scans"])}D')
 
+    # Dispatch to sparse plotting for adaptive scans
+    if is_sparse_2d_scan(results):
+        _plot_2d_rotor_scan_sparse(results, path=path, label=label, cmap=cmap,
+                                   resolution=resolution, original_dihedrals=original_dihedrals)
+        return
+
     results['directed_scan'] = clean_scan_results(results['directed_scan'])
 
     # phis0 and phis1 correspond to columns and rows in energies, respectively
@@ -1360,9 +1366,9 @@ def plot_2d_rotor_scan(results: dict,
     label = ' for ' + label if label else ''
     plt.title(f'2D scan energies (kJ/mol){label}')
     min_x = min_y = -180
-    plt.xlim = (min_x, min_x + 360)
+    plt.gca().set_xlim(min_x, min_x + 360)
     plt.xticks(np.arange(min_x, min_x + 361, step=60))
-    plt.ylim = (min_y, min_y + 360)
+    plt.gca().set_ylim(min_y, min_y + 360)
     plt.yticks(np.arange(min_y, min_y + 361, step=60))
 
     if mark_lowest_conformations:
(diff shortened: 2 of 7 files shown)


alongd:
I rebased this branch onto current `main` (263 commits of drift) as `nd_scans_rebased`, pushed it to the canonical repo, and opened #930 from it — deliberately *not* touching `nd_scans` here, since that branch is checked out in a live working session with uncommitted local patches and shouldn't be force-pushed or rebased in place. Below are the review findings from doing that rebase and a follow-up read-through.

### Conflict resolution note (for the author)

At `arc/scheduler.py`, in `check_directed_scan_job`, neither side of the merge conflict was correct wholesale, so I kept both:

- Current `main` has a TS isomorphism short-circuit (`if is_ts: is_isomorphic = True`) that this branch had dropped. Without it, the new invalid-point routing would send every point of a TS directed scan to `mark_scan_point_invalid` and discard the surface.
- This branch has the correct `parse_e_elect(log_file_path=...)` kwarg that `main` lacks (see bug #1 below).

The resolution in `nd_scans_rebased` keeps both the TS guard and the corrected kwarg.

### Two latent bugs on `main` itself (not introduced by this PR)

1. `arc/scheduler.py` calls `parser.parse_e_elect(path=...)`, but `make_parser` (`arc/parser/parser.py`) builds a callable whose parameter is `log_file_path` — `path` is only the `{path}` key in the error-message template. Any completed directed-scan job raises `TypeError` on `main` today. This is masked in the test suite by a `MagicMock` standing in for the parser. This branch already has the correct kwarg, which is why it surfaced as a merge conflict rather than a silent bug.
2. `Scheduler.__init__` calls `run_scan_jobs()` from its per-species restart loop before `self.pipe_planner` is assigned, and `run_scan_jobs` dereferences it via `try_pipe_rotor_scans_1d()`. Restarting a project with pending 1D rotor scans raises `AttributeError`. Fixed in `nd_scans_rebased` by constructing the coordinator/planner before the loop (commit "Fix pipe_planner initialization order in Scheduler.__init__").

### Fixed here — finalization crash

`finalize_directed_scan_results` did `parse_nd_scan_energies_func(...)[0]`, but that parser returns the results dict itself — `make_parser` types it `Optional[Dict]`, and every ESS adapter declares the same. This raised `KeyError: 0` at finalization, i.e. after a multi-hour scan completed and before the rotor YAML was written, losing the run's output. The existing unit test's mock returned `[mock_results]`, so it was written to match the bug rather than the real contract; I fixed the call site and added a regression test.

### Design findings for discussion (independent review pass — flagging these to confirm, not asserting they're wrong)

- `sampling_policy` doesn't appear to have an `input.yml` surface. Its only writer (`initialize_adaptive_scan_state`) sits downstream of its only reader (`is_adaptive_enabled`), so the adaptive path looks unreachable from user input as written — the PR's own test sets the dict key by hand rather than going through any config path. Worth confirming whether there's a wiring path I'm missing.
- At ARC's default `rotor_scan_resolution = 8.0`, seed points (~285) already exceed `ADAPTIVE_DEFAULT_MAX_POINTS = 200`, so it looks like the adaptive phase never actually runs at default settings, and truncation in `a0`-major order would leave a large gap in one dimension. I also didn't see a convergence-based stopping criterion — only `max_points_reached` / `grid_exhausted` / `full_coverage`.
- The separability metric seemed to invert on adaptively-sampled data in a quick check: on a surface separable by construction it reported `nonseparability ~0.26` / `coupled` / `confidence 1.0`, while the same surface on a dense grid scored `0.000000`. The likely cause is estimating 1D marginals by averaging over points sharing an angle key, which degenerates when most angles are singletons (as they are under adaptive/sparse sampling). All current tests of this metric use complete factorial grids, so this gap wouldn't show up in CI as written.
- `calculate_cross_term_strength` returns `calculate_nonseparability_score(...)` directly, so in `nonsep > 0.15 or cross_term > 0.10`, `COUPLING_NONSEP_THRESHOLD = 0.15` never ends up being the operative cut (both terms are the same value).
- Non-isomorphic points are routed to `mark_scan_point_invalid` but also still written to `directed_scan` via `record_directed_scan_point`. Downstream consumers read `directed_scan` and filter only on `energy is not None`, so an invalid point could shift the energy zero for the whole rotor rather than being excluded.
- Batch acquisition sorts candidates once without updating the sampled set between picks within a batch, so points within a single batch can collapse onto the same region.

Happy to discuss any of these — flagging them now since I had the branch open for the rebase anyway. Detailed PR for the rebase + fixes is #930.

PR #859, opened 2026-04-08
Added the RitS TS search adapter

alongd:
An adapter for executing RitS (Right into the Saddle) TS-guess jobs.

RitS is a flow-matching ML model that generates 3D transition-state geometries
directly from atom-mapped reactant + product structures, without requiring an
initial guess. Unlike GCN (which is restricted to isomerizations), RitS can
handle bimolecular reactions and supports charged species, so it covers a
strictly larger reaction space.

Code source     : https://github.com/isayevlab/RitS
Paper           : [10.26434/chemrxiv.15001681/v1](https://chemrxiv.org/doi/full/10.26434/chemrxiv.15001681/v1)
Pretrained ckpt : https://doi.org/10.5281/zenodo.19474153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,3 +124,63 @@ jobs:
           name: codecov-umbrella
           verbose: true
           fail_ci_if_error: false
+
+  rits-install:
+    name: RitS install (CPU lane)
+    runs-on: ubuntu-latest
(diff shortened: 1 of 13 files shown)


alongd:
integrated into https://github.com/ReactionMechanismGenerator/ARC/pull/887

PR #862, opened 2026-04-09
TS Edge Case & Pipe QUEUE bug

calvinp0:
# Two changes:
---
## Change 1:
In the Scheduler, only break after conformer troubleshooting if new jobs (conf_opt or conf_sp) are actually running for that species. This ensures the scheduler correctly falls through to the "all conformers done" check if troubleshooting was attempted but failed to launch new tasks.

Essentially, this was a race to completion bug. If there were, for example, 3 TS guesses being troubleshooted and if the 2 out the 2 finished troubleshoot, then the last one, if it failed an exhausted all attempts would cause a bug where ARC would declare there was no TS that converged

## Change 2:

There is a bug in ARC when Pipe is active. The issue is that when a batch job is submited, and let's say it has 20 jobs in it - so 20 workers needed, and then only 10 of those were picked up by workers but the other 10 are in Q mode, ARC misunderstand this and attempts to resubmit those 10 as it thinks they were not provided workers properly. And ARC will continue to do this ad infinitum

### DIFF
--- a/arc/job/pipe/pipe_run.py
+++ b/arc/job/pipe/pipe_run.py
@@ -363,20 +363,11 @@ class PipeRun:
                     logger.debug(f'Could not promote task {task_id} to FAILED_TERMINAL '
                                  f'(lock contention or concurrent state change): {e}')
 
-        # Only flag resubmission for genuinely retried tasks (attempt_index > 0).
-        # Fresh PENDING tasks (attempt_index == 0) are waiting for the initial
-        # submission's workers to start — don't resubmit for those.
-        # After a resubmission, allow a grace period for workers to start before
-        # flagging again (prevents duplicate submissions).
-        active_after_retry = counts[TaskState.CLAIMED.value] + counts[TaskState.RUNNING.value]
-        resubmit_grace = 120  # seconds
-        time_since_submit = (now - self.submitted_at) if self.submitted_at else float('inf')
-        if retried_pending > 0 and active_after_retry == 0 and time_since_submit > resubmit_grace:
-            self._needs_resubmission = True
-            logger.info(f'Pipe run {self.run_id}: {retried_pending} retried tasks '
-                        f'need workers. Resubmission needed.')
-        else:
-            self._needs_resubmission = False
+        # Never resubmit a new scheduler job for retried tasks.
+        # Workers still in the scheduler queue (PBS Q state) will claim
+        # retried PENDING tasks when they start.  If the scheduler job
+        # was killed, that is a manual intervention issue.
+        self._needs_resubmission = False
 
         terminal = (counts[TaskState.COMPLETED.value]
                     + counts[TaskState.FAILED_ESS.value]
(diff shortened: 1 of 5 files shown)


arc/scheduler.py
```diff
@@ -630,7 +630,15 @@ def schedule_jobs(self):
...
-                                    break
+                                    # Only break if other conformer jobs are still in flight.
+                                    # When the last conformer exhausts troubleshooting without
+                                    # converging, we must fall through to the "all done" check
+                                    # below so it can call determine_most_likely_ts_conformer
+                                    # on the conformers that already succeeded — otherwise ARC
+                                    # mistakenly concludes no TS guess converged.
+                                    if any(is_conformer_job(j)
+                                           for j in self.running_jobs.get(label, [])):
+                                        break
```
alongd, line 641: Is the unconditional break in the loop prevent the "all done" check from running?

arc/job/pipe/pipe_run.py
```diff
@@ -366,16 +370,29 @@ def reconcile(self) -> Dict[str, int]:
...
+        # (PBS Q state) and will claim retried tasks when they start.
         # After a resubmission, allow a grace period for workers to start before
         # flagging again (prevents duplicate submissions).
+        # However, fresh_pending must not suppress resubmission indefinitely —
+        # if the scheduler job was killed/cancelled, those workers will never
+        # start.  After fresh_stale_timeout we assume the scheduler job is dead.
         active_after_retry = counts[TaskState.CLAIMED.value] + counts[TaskState.RUNNING.value]
-        resubmit_grace = 120  # seconds
+        resubmit_grace = 120  # seconds — minimum wait after any submission
+        fresh_stale_timeout = 10800  # seconds (3 hours) — max time to trust fresh tasks' scheduler job
```
alongd: This is hardcoded. On long queues it might be too aggressive, and on fast-Qs it might be too conservative

alongd (commented):
Thanks! Please see the couple of comments below

arc/scheduler.py
calvinp0: The break is conditional by `if any(is_conformer_job(j) for j in self.running_jobs.get(label, [])):`

arc/job/pipe/pipe_run.py
calvinp0: Yes it is hard coded cause right now ARC has trouble understanding what job in a batch is Q and what is a worker issue (why it could be stuck in pending). I think this is a limitation of Pipe and that it only reads the json file ont he hard disk and not actually checks the queue itself

calvinp0: I am going to have to change the logic altogether here because of this limitation. We will have to assume that PENDING in the task json indicates to us that the job in the batch is in Q. There appears to be no other way unless we poll the actual scheduler itself. Else, ARC will assume a PENDING job (that is actually queued) needs resubmission. And it will do this forever - or at least until they are running on queue

alongd (approved):
Thanks!

PR #864, opened 2026-04-10
Decouple TS Scheduler Check

calvinp0:
## The Bug

  When the last conformer job for a TS exhausted troubleshooting, determine_most_likely_ts_conformer was never called — 47 successful conformers were ignored and TS0 was marked "did not converge."

### Structural fix — Two-Phase Loop:
  - Phase 1 (simplified): Process individual job completions — end_job, parse, troubleshoot, break. No aggregate pipeline decisions. The fragile inline for/else checks and the conditional troubleshooting break are gone.
  - Phase 2 (new): After all job events, unconditionally check stage transitions for every species. No break can skip it.

 ### Three new methods:
  1. _check_conformer_stage_complete — detects when all conf_opt/conf_sp jobs finished, selects best conformer, spawns opt. Works for both pipe-ejected and regular scheduler jobs.
  2. _check_tsg_stage_complete — detects when all TSG jobs finished, spawns conformer jobs.
  3. _check_species_complete — detects when all jobs for a species finished, calls check_all_done.

 ### Edge cases found: 
  - All conformers fail (no energy): Now logs a clear error and sets ts_guesses_exhausted = True so the species properly reports failure instead of silently stalling.
  - Isomorphism troubleshooting exhausted: Now marks conf_opt = True to prevent infinite re-firing of the check every 30 seconds.

 All guards are idempotent — Phase 2 runs every iteration but each check fires at most once per species per stage.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -622,55 +622,21 @@ class Scheduler(object):
                         job = self.job_dict[label]['conf_opt'][i] if 'conf_opt' in job_name \
                             else self.job_dict[label]['conf_sp'][i]
                         if not (job.job_id in self.server_job_ids and job.job_id not in self.completed_incore_jobs):
-                            # this is a completed conformer job
                             successful_server_termination = self.end_job(job=job, label=label, job_name=job_name)
                             if successful_server_termination:
                                 troubleshooting_conformer = self.parse_conformer(job=job, label=label, i=i)
                                 if 'conf_opt' in job_name and self.job_types['conf_sp'] and not troubleshooting_conformer:
                                     # Accumulate for deferred pipe batching of conf_sp.
                                     self._pending_pipe_conf_sp.setdefault(label, set()).add(i)
-                                if troubleshooting_conformer:
-                                    break
-                            # Just terminated a conformer job.
-                            # Are there additional conformer jobs currently running for this species?
-                            # Note: end_job already removed the current job from running_jobs,
-                            # so we don't need to exclude job_name.
-                            for spec_jobs in job_list:
-                                if 'conf_opt' in spec_jobs or 'conf_sp' in spec_jobs:
-                                    break
-                            else:
-                                # All conformer jobs terminated.
-                                # Check isomorphism and run opt on most stable conformer geometry.
-                                logger.info(f'\nConformer jobs for {label} successfully terminated.\n')
-                                if self.species_dict[label].is_ts:
-                                    self.determine_most_likely_ts_conformer(label)
-                                else:
-                                    self.determine_most_stable_conformer(label, sp_flag=True if self.job_types['conf_sp'] else False)  # also checks isomorphism
-                                if self.species_dict[label].initial_xyz is not None:
-                                    # if initial_xyz is None, then we're probably troubleshooting conformers, don't opt
-                                    if not self.composite_method:
-                                        self.run_opt_job(label, fine=self.fine_only)
-                                    else:
-                                        self.run_composite_job(label)
                             self.timer = False
                             break
                     if 'tsg' in job_name:
                         job = self.job_dict[label]['tsg'][get_i_from_job_name(job_name)]
                         if not (job.job_id in self.server_job_ids and job.job_id not in self.completed_incore_jobs):
-                            # This is a successfully completed tsg job. It may have resulted in several TSGuesses.
                             self.end_job(job=job, label=label, job_name=job_name)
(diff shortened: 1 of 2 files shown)


arc/scheduler.py
```diff
@@ -840,6 +798,114 @@ def schedule_jobs(self):
...
+    def _check_tsg_stage_complete(self, label: str) -> None:
+        """
+        Check whether all TS guess jobs for a species have finished.
+        If so, spawn conformer jobs for the TS.
+        """
+        if 'tsg' not in self.job_dict.get(label, {}):
+            return
+        if any('tsg' in j for j in self.running_jobs.get(label, [])):
+            return
+        if not self.species_dict[label].is_ts:
```
alongd, line 865: intuitively, I'd start with this check (will return from this func more often than the other checks)

arc/scheduler.py
```diff
@@ -803,20 +769,12 @@ def schedule_jobs(self):
...
-                               for p in self.active_pipes.values())
-                    )
-                    if not has_pending_pipe_work:
-                        self.check_all_done(label)
-                    if not self.running_jobs[label]:
-                        # Delete the label only if it represents an empty entry.
-                        del self.running_jobs[label]
+            for label in list(self.unique_species_labels):
+                if label in self.output and self.output[label]['convergence'] is False:
+                    continue
```
alongd, line 774: This skips failed species but still runs all three checks for already-converged species. Maybe instead, `continue` if convergence is not None?

arc/scheduler.py
```diff
@@ -840,6 +798,114 @@ def schedule_jobs(self):
...
+        if 'conf_opt' not in self.job_dict.get(label, {}):
+            return
+        if any('conf_opt' in j or 'conf_sp' in j
+               for j in self.running_jobs.get(label, [])):
+            return
+        if label in self._pending_pipe_conf_sp:
+            return
+        if any(label in {t.owner_key for t in p.tasks}
+               for p in self.active_pipes.values()
+               if any(t.task_family in ('conf_opt', 'conf_sp', 'ts_opt') for t in p.tasks)):
```
alongd, line 818: This checks ts_opt , but the function is about conformers. If a species has active ts_opt jobs but its conformers are all done, this guard would incorrectly block the conformer stage from completing. Restrict to ('conf_opt', 'conf_sp') only?

arc/scheduler.py
```diff
@@ -840,6 +798,114 @@ def schedule_jobs(self):
...
+            self.output[label]['job_types']['conf_opt'] = True
+
+    def _check_tsg_stage_complete(self, label: str) -> None:
+        """
+        Check whether all TS guess jobs for a species have finished.
+        If so, spawn conformer jobs for the TS.
+        """
+        if 'tsg' not in self.job_dict.get(label, {}):
+            return
+        if any('tsg' in j for j in self.running_jobs.get(label, [])):
```
alongd, line 863: If TSG jobs are submitted through pipe mode, the method only checks self.running_jobs for 'tsg' entries but doesn't check self.active_pipes

arc/scheduler.py
```diff
@@ -840,6 +798,114 @@ def schedule_jobs(self):
...
+
+        if self.species_dict[label].is_ts:
+            has_successful_conformer = any(
+                tsg.energy is not None for tsg in self.species_dict[label].ts_guesses)
+        else:
+            has_successful_conformer = any(
+                e is not None for e in self.species_dict[label].conformer_energies)
+
+        if not has_successful_conformer:
+            logger.error(f'All conformer jobs for {label} failed. '
```
alongd, line 835: should we also mark a failed conformer state in the species or `.output` in scheduler? (I don't remember but we should have a f;lag for this that should be turned on here)

arc/scheduler_test.py
```diff
@@ -757,6 +758,135 @@ def test_add_label_to_unique_species_labels(self):
...
+            ts_guess_level=Level(repr=default_levels_of_theory['ts_guesses']),
+            project_directory=os.path.join(ARC_PATH, 'Projects', 'arc_project_for_testing_delete_after_usage6'),
+            testing=True,
+            job_types=self.job_types1,
+            orbitals_level=default_levels_of_theory['orbitals'],
+            adaptive_levels=None,
+        )
+        return sched, spc.label
+
+    def test_check_conformer_stage_complete_spawns_opt_for_ts(self):
```
alongd, line 790: The test covers the TS path. Add a test for non-TS species?

alongd (commented):
Thanks @calvinp0! Please see some comments

PR #865, opened 2026-04-10
Invalid orbital range bug

calvinp0:
# Bugs: 
Guard 1 — Prevention (scheduler.py:1440): Case bug. 'DLPNO' in level.method but Level.__init__ normalizes to lowercase. Dead code since the day it was written — never fired once.                                                                                                                                       
                                                                                                                                                                                                                                                                                                                           
Guard 2 — Troubleshooting (trsh.py:1070): Structurally unreachable. This one actually used lowercase 'dlpno' correctly, but it was an elif after the Memory branch. The error flow made it impossible to reach:                                                                                                          
                                                                         
  1. ORCA crashes with INVALID ORBITAL RANGE in err.txt                                                                                                                                                                                                                                                                    
  2. determine_ess_status reads the log file, finds "ORCA finished by error termination in MDCI", scans for "Please increase MaxCore" or "parallel calculation exceeds number of pairs" — finds neither
  3. Falls through the for-else to: "MDCI error in Orca. Assuming memory allocation error." → keywords = ['MDCI', 'Memory']                                                                                                                                                                                                
  4. trsh_ess_job sees 'Memory' in keywords → enters Memory branch → increases memory → resubmits                                                                                                                                                                                                                          
  5. Same crash → step 2 → infinite loop                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                           
  The DLPNO check at step 4 was behind elif, so it could never fire when Memory was in the keywords. Two bugs compounding — the first one prevents the problem, the second one should have caught it but couldn't due to the control flow.   



---

Following on for why ARC did the trsh ad infinitum:

  1. ORCA fails → determine_ess_status sees "ORCA finished by error termination in MDCI", doesn't find "Please increase MaxCore" or "parallel calculation exceeds number of pairs" in the log → falls through to else: keywords = ['MDCI', 'Memory']                                                                       
  2. trsh_ess_job enters Orca Memory branch → 'memory' not in ess_trsh_methods → appends 'memory' → calculates new memory via estimate_orca_mem_cpu_requirement(num_heavy_atoms=0) → couldnt_trsh stays False                                                                                                              
  3. Scheduler resubmits with new memory → same ORCA crash                                                                                                                                                                                                                                                                 
  4. trsh_ess_job enters Orca Memory branch again → 'memory' already in list (not re-added) → calculates same memory estimate → couldnt_trsh stays False                                                                                                                                                                   
  5. Repeat step 3-4 forever

### DIFF
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -838,6 +838,7 @@ def trsh_ess_job(label: str,
                  cpu_cores: int,
                  ess_trsh_methods: list,
                  is_h: bool = False,
+                 is_monoatomic: bool = False,
                  ) -> tuple:
     """
     Troubleshoot issues related to the electronic structure software, such as convergence.
@@ -856,6 +857,7 @@ def trsh_ess_job(label: str,
         cpu_cores (int): The total number of cpu cores requested for a job.
         ess_trsh_methods (list): The troubleshooting methods tried for this job.
         is_h (bool): Whether the species is a hydrogen atom (or its isotope). e.g., H, D, T.
+        is_monoatomic (bool): Whether the species is monoatomic (single atom).
 
     Todo:
         - Change server to one that has the same ESS if running out of disk space.
@@ -1016,7 +1018,10 @@ def trsh_ess_job(label: str,
             couldnt_trsh = True
 
     elif 'orca' in software:
-        if 'Memory' in job_status['keywords']:
+        if 'dlpno' in level_of_theory.method and (is_monoatomic or is_h):
+            raise TrshError(f'DLPNO methods are incompatible with monoatomic species {label} in Orca. '
+                            f'This should have been caught by the Scheduler before job submission.')
+        elif 'Memory' in job_status['keywords']:
             # Increase memory allocation.
             # job_status will be for example
             # `Error  (ORCA_SCF): Not enough memory available! Please increase MaxCore to more than: 289 MB`.
@@ -1067,9 +1072,6 @@ def trsh_ess_job(label: str,
             logger.info(f'Troubleshooting {job_type} job in {software} for {label} using {cpu_cores} cpu cores.')
             if 'cpu' not in ess_trsh_methods:
                 ess_trsh_methods.append('cpu')
-        elif 'dlpno' in level_of_theory.method and is_h:
-            logger.error('DLPNO method is not supported for H atom (or its isotope D or T) in Orca.')
-            couldnt_trsh = True
         else:
             couldnt_trsh = True
 
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -67,9 +67,9 @@ if TYPE_CHECKING:
 logger = get_logger()
 
 LOWEST_MAJOR_TS_FREQ, HIGHEST_MAJOR_TS_FREQ, default_job_settings, \
-    default_job_types, default_ts_adapters, max_rotor_trsh, rotor_scan_resolution, servers_dict = \
+    default_job_types, default_ts_adapters, max_ess_trsh, max_rotor_trsh, rotor_scan_resolution, servers_dict = \
     settings['LOWEST_MAJOR_TS_FREQ'], settings['HIGHEST_MAJOR_TS_FREQ'], settings['default_job_settings'], \
-    settings['default_job_types'], settings['ts_adapters'], settings['max_rotor_trsh'], \
+    settings['default_job_types'], settings['ts_adapters'], settings['max_ess_trsh'], settings['max_rotor_trsh'], \
     settings['rotor_scan_resolution'], settings['servers']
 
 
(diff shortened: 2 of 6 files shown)


alongd:
We must add a trsh counter (or do we already have one?), so we don't do anything infinitely

arc/scheduler.py
```diff
@@ -1436,10 +1436,16 @@ def run_sp_job(self,
                              level_of_theory='ccsd/cc-pvdz',
                              job_type='sp')
                 return
-        mol = self.species_dict[label].mol
-        if mol is not None and len(mol.atoms) == 1 and mol.atoms[0].element.symbol == 'H' and 'DLPNO' in level.method:
-            # Run only CCSD for an H atom instead of DLPNO-CCSD(T) / etc.
-            level = Level(repr='ccsd/vtz', software=level.software, args=level.args)
+        if self.species_dict[label].is_monoatomic() and 'dlpno' in level.method:
+            species = self.species_dict[label]
+            if species.number_of_atoms == 1 and species.mol.atoms[0].element.symbol in ('H', 'D', 'T'):
```
alongd: no need for `species.number_of_atoms == 1`, we already know it is_monoatomic

arc/scheduler.py
```diff
@@ -1436,10 +1436,16 @@ def run_sp_job(self,
...
+                logger.warning(f'Using HF/{level.basis} for {label} (single electron, no correlation).')
```
alongd: it's not a warning per se...
Info?
same below

arc/scheduler.py
```diff
@@ -1436,10 +1436,16 @@ def run_sp_job(self,
...
+                level = Level(method='hf', basis=level.basis, software=level.software)
+            else:
+                canonical_method = level.method.replace('dlpno-', '')
```
alongd, line 1453: 👍

arc/scheduler.py
```diff
@@ -3572,6 +3573,7 @@ def troubleshoot_ess(self,
                          server=job.server,
                          job_status=job.job_status[1],
                          is_h=is_h,
+                         is_monoatomic=bool(self.species_dict[label].is_monoatomic()),
```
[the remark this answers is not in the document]
alongd: I tend to agree

arc/job/trsh.py
```diff
@@ -1016,7 +1018,11 @@ def trsh_ess_job(label: str,
             couldnt_trsh = True
 
     elif 'orca' in software:
-        if 'Memory' in job_status['keywords']:
+        if 'dlpno' in level_of_theory.method and (is_monoatomic or is_h):
```
alongd, line 1021: this shouldn't happen, if it does, it means Scheduler is buggy. I almost think we could raise an error here (so devs know)

alongd (commented):
Thanks!! Added some comments

calvinp0:
> We must add a trsh counter (or do we already have one?), so we don't do anything infinitely

Depends on what we are troubleshooting - liek for TS guess, we eventually try everything and then declare all methods attempted

arc/job/trsh.py
calvinp0: Fair, changed it

calvinp0:
> We must add a trsh counter (or do we already have one?), so we don't do anything infinitely

I added a counter now, and defaulted it to 10 in the settings.py

alongd:
I think the trsh counter could be per trsh method (not to try the same one too many times), Maybe we can add a trsh_counter dict to Species?

calvinp0:
> I think the trsh counter could be per trsh method (not to try the same one too many times), Maybe we can add a trsh_counter dict to Species?

I guess so. The thing is, most of our trsh methods are safe - they have limits. like rotor scans max 4 in the settings. TS guesses have a flag I implemented a year ago or more where it tries all relevant methods and then does 'all_attempted' to indicate exhaustion. Orca Mem has a guard if 'memory' is not in the ess_trsh_methods. and molpro has an elif chain of guards.

The issue now gaussian memory - no 'memory' not in ess trsh method guard. can keep doubling until hiting 95% node memory. I think that was for ATLAS? and then general ESS trsh where there is no global max attempt counter - so it relies entirely on ess_trsh_methods list eventually mathcing attempted_ess_trsh_methods.

So, I am not so sure having an ess trsh counter per method is really relevant here.

calvinp0:
Ok

> I think the trsh counter could be per trsh method (not to try the same one too many times), Maybe we can add a trsh_counter dict to Species?

Further investigating, I am hesitant about per trsh method cause then they means we need to set a limit for a fair few methods and the also allow the user to change it (could be overwhelming), and I know for ORCA memory - I misspoke. It doesn't have a guard cause I know we sometimes have to troubleshoot the mem multiple times as ORCA can keep saying 'okay allocate more' and then we do, then it complains for even more.

arc/scheduler.py
```diff
@@ -1444,10 +1444,16 @@ def run_sp_job(self,
...
                 return
-        mol = self.species_dict[label].mol
-        if mol is not None and len(mol.atoms) == 1 and mol.atoms[0].element.symbol == 'H' and 'DLPNO' in level.method:
-            # Run only CCSD for an H atom instead of DLPNO-CCSD(T) / etc.
-            level = Level(repr='ccsd/vtz', software=level.software, args=level.args)
+        if self.species_dict[label].is_monoatomic() and 'dlpno' in level.method:
+            species = self.species_dict[label]
+            if species.mol.atoms[0].element.symbol in ('H', 'D', 'T'):
+                logger.info(f'Using HF/{level.basis} for {label} (single electron, no correlation).')
+                level = Level(method='hf', basis=level.basis, software=level.software)
```
alongd: We should probably still pass `args=level.args` like in the deleted line (`TIGHTSCF`, etc.)

alongd (commented):
Thanks, one comment

arc/scheduler.py
calvinp0: Fixed

alongd (dismissed):
Thanks!

alongd (approved):
Thanks!

PR #866, opened 2026-04-11
Fix switch_ts to reset state & clean up IRC when switching TS guesses

calvinp0:
When a TS guess fails validation (e.g., NMD check), switch_ts picks the next guess but previously left stale state behind:

  1. IRC species from the invalidated guess were never cleaned up. delete_all_species_jobs('TS0') only deletes jobs under the TS0 label, but IRC species like IRC_TS0_1 are separate entries in running_jobs/species_dict/etc. These orphaned species
  continued running in parallel with the new guess, potentially interfering with job processing.
  2. job_types flags (freq, sp, opt) were never reset. After guess N's freq completed, job_types['freq'] = True carried over to guess N+1, causing the scheduler to skip re-running freq for the new geometry.
  3. convergence was never reset to None.
  4. The old line self.output[label]['geo'] = ... wrote to the wrong dict level (top-level keys instead of self.output[label]['paths']), making it dead code.
  5. Pending pipe batches from the old guess were never discarded.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -2763,7 +2763,6 @@ class Scheduler(object):
         logger.info(f'Switching a TS guess for {label}...')
         self.determine_most_likely_ts_conformer(label=label)  # Look for a different TS guess.
         self.delete_all_species_jobs(label=label)  # Delete other currently running jobs for this TS.
-        self.output[label]['geo'] = self.output[label]['freq'] = self.output[label]['sp'] = self.output[label]['composite'] = ''
         freq_path = os.path.join(self.project_directory, 'output', 'rxns', label, 'geometry', 'freq.out')
         if os.path.isfile(freq_path):
             os.remove(freq_path)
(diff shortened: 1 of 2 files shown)


alongd (commented):
Looks good, but I think the addition should be inside `delete_all_species_jobs()`

calvinp0:
> Looks good, but I think the addition should be inside `delete_all_species_jobs()`
@alongd I made the change, but also in an arc run found another TS bug, so I added it - so please be aware of this: 
https://github.com/ReactionMechanismGenerator/ARC/pull/866/changes#diff-556ad66547f14f3f9a3915808a05b59151ee8216b817845d8d716577c0aeb5ebR3557

arc/scheduler.py
```diff
@@ -3555,7 +3554,14 @@ def troubleshoot_ess(self,
                                f'log file:\n"{job.job_status[1]["line"]}".'
         logger.warning(warning_message)
         if self.species_dict[label].is_ts and conformer is not None:
-            xyz = self.species_dict[label].ts_guesses[conformer].get_xyz()
+            tsg = next((t for t in self.species_dict[label].ts_guesses
+                        if t.index == conformer or t.conformer_index == conformer), None)
```
alongd: From the docsting of TSGuess:
```
index (int): A running index of all TSGuess objects belonging to an ARCSpecies object.
conformer_index (int): An index corresponding to the conformer jobs spawned for each TSGuess object.
                               Assigned only if self.success is ``True``.
```
They are distinct (though confusing). I think here we should only check `conformer_index`

arc/scheduler.py
```diff
@@ -3705,6 +3711,35 @@ def delete_all_species_jobs(self, label: str):
                     job.delete()
         self.running_jobs[label] = list()
         self.output[label]['paths'] = {key: '' if key != 'irc' else list() for key in self.output[label]['paths'].keys()}
+        for job_type in self.output[label]['job_types']:
+            if job_type in ['rotors', 'bde']:
```
alongd: why skip those?

alongd (commented):
Thanks, added two comments

arc/scheduler.py
calvinp0: misunderstood how they work - fixed. thanks!

alongd (approved):

PR #867, opened 2026-04-11
Migrating ARC to Python 14

alongd:
Bump the target runtime from 3.12 to 3.14 and modernize the codebase accordingly.                                                                                            
   
  - Update Python, Cython (≥3.1), and RDKit (≥2026.03) pins across environment.yml, requirements.txt, Dockerfile, and pyproject.toml
  - Fix stale check_python() gate that still referenced Python 3.7 / RMG-Py
  - Remove dead from __future__ import print_function and deprecated assertDictEqual
  - Replace deprecated typing generics with built-in equivalents (list, dict, X | None, X | Y, collections.abc) across 70+ files
                                                                                                                                                                               
RMG's separate Python 3.9 environment for suprocess scripts (e.g. Arkane) is unchanged.


calvinp0:
I can't seem to properly comment on the line, and I know this has gone back to draft and I was not requested to review but I did see that the output.py still has Dict in it without either being imported or changed over

alongd:
Openbable seems to be the only package that inhibits the Py14 upgrade. The OB repo hasn't been updated since Dec 2024. There is a PR (number 2809) from July 2025 to bump into Py13, the discussion hasn't advanced since then.
Is OB dead/unmaintained?
I tried removing it as a dependency, it caused sever issues with perception and atom mapping. If we detach ourselves from OB, we'll need to carefully design alternatives for anything missing in RDKit, this could be a medium-large project.
The solution I have for now is for us to upgrade OB to Py14 [ourselves](https://github.com/DanaResearchGroup/openbabel/tree/py14). I'm placing it on our Anaconda channel (https://anaconda.org/danagroup/repo). Let's see if that works.

arc/scheduler_test.py
```diff
@@ -907,104 +907,6 @@ def test_switch_ts_cleanup(self, mock_run_opt):
         self.assertIsNone(sched.species_dict[ts_label].ts_checks['NMD'])
         self.assertIsNone(sched.species_dict[ts_label].ts_checks['E0'])
 
-        # Verify rotors convergence flag preserved as True (not blanket-reset to False).
```
calvinp0, line 910: Was this test completely removed or moved to another script?

arc/scheduler_test.py
```diff
@@ -907,104 +907,6 @@ def test_switch_ts_cleanup(self, mock_run_opt):
...
-        self.assertTrue(sched.output[ts_label]['job_types']['rotors'])
-
-    @patch('arc.scheduler.Scheduler.run_opt_job')
-    def test_switch_ts_rotors_reset(self, mock_run_opt):
```
calvinp0, line 914: Same question (removed or moved?)

arc/job/adapters/obabel.py
```diff
@@ -27,6 +28,24 @@
     settings['default_job_settings'], settings['global_ess_settings'], settings['input_filenames'], \
     settings['output_filenames'], settings['servers'], settings['submit_filenames'], settings['OB_PYTHON']
 
+# Fall back to current Python if ob_env is not available or has broken OB.
+# OB may be in the main env (e.g., danagroup build in arc_env).
+if OB_PYTHON is not None and OB_PYTHON != sys.executable:
+    import subprocess as _sp
+    try:
+        _result = _sp.run([OB_PYTHON, '-c', 'from openbabel import openbabel; '
+                           'assert openbabel.OBForceField.FindForceField("MMFF94s") is not None'],
```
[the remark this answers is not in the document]
calvinp0: Could put a + like github security says, or make a var called command first

```
cmd = (
    'from openbabel import openbabel; '
    'assert openbabel.OBForceField.FindForceField("MMFF94s") is not None'
)
```
And then place it in the `run` function

arc/scheduler_test.py
alongd: These two omissions weren't intentional. I also found other minor omissions. Restored now, force pushed. Thanks for catching this!

calvinp0:
I think we need to get the CodeQL back in order so it does not keep complaining
Also, I think we need to see what tests are missing as an 8% drop in codecov is quite substantial

alongd:
The CodeQL errs that are left are false positives.
I'll take a look at the tests and coverage

alongd:
@calvinp0, the 8% coverage drop is unavoidable until Cython's coverage plugin supports Python 3.14's `sys.monitoring`

calvinp0:
> @calvinp0, the 8% coverage drop is unavoidable until Cython's coverage plugin supports Python 3.14's `sys.monitoring`

Could you then turn it off, the sys monitoring and try the previous way:

https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives

calvinp0:
> > @calvinp0, the 8% coverage drop is unavoidable until Cython's coverage plugin supports Python 3.14's `sys.monitoring`
> 
> Could you then turn it off, the sys monitoring and try the previous way:
> 
> https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives


linetrace (True / False), default=False
Write line tracing hooks for Python profilers or coverage reporting into the compiled C code. This also enables profiling. Note that the generated module will not actually use line tracing, unless you additionally pass the C macro definition CYTHON_TRACE=1 to the C compiler (e.g. using the setuptools option define_macros). Define CYTHON_TRACE_NOGIL=1 to also include nogil functions and sections. Define CYTHON_USE_SYS_MONITORING to either 1 or 0 to control the mechanism used to implement these features on Python 3.13 and above. Note that neither profile nor linetrace work with any tool that uses sys.monitoring on Python 3.12.

Dockerfile
```diff
@@ -55,7 +55,7 @@ RUN micromamba run -n rmg_env bash -c "\
     "
 
 WORKDIR /home/mambauser/Code/ARC
-RUN micromamba create -y -v -n arc_env python=3.12 -f environment.yml && \
+RUN micromamba create -y -v -n arc_env -c conda-forge -c danagroup -f environment.yml && \
```
calvinp0: Sorry, I'm doing the comments slowly in parts. So bear with me as I get throught checking. 

So one, I think this change should be squashed with the first commit that originally touched the Docker file (that's what I see from the commits but I could be wrong)

Secondly, the python pin was not redundant. In micromamba an explicit pin of a package, like pythonf, in the create/install line can reduce the backtracking even if the same version pin is in the environment file. It's something to do with in the install/create like taking precedence before even considering other packages in the environment file. 

So I don't mind it being removed again but I think we need to be aware of the solve issue comes up again in Docker building that this is the solution

alongd: Good point, I'm restoring the python pin here

alongd:
@calvinp0, should be ready for a final look

arc/job/adapters/molpro_test.py
```diff
@@ -468,8 +468,8 @@ def tearDownClass(cls):
         A function that is run ONCE after all unit tests in this class.
         Delete all project directories created during these unit tests
         """
-        for folder in ['test_MolproAdapter_1', 'test_MolproAdapter_2']:
-            shutil.rmtree(os.path.join(ARC_TESTING_PATH, folder), ignore_errors=True)
+        for i in range(1, 8):
```
calvinp0: A bit pedantic but I am not sure I agree with using a set range number here on the off chance someone adds a test but doesn't realise they need to increase the range.

Would it be possible to do

```
    for attr in vars(cls).values():
        if isinstance(attr, MolproAdapter):
            shutil.rmtree(attr.project_directory, ignore_errors=True)
```

Or just even if a folder has the name `MolproAdapter` in it?

.github/workflows/gh-pages.yml
```diff
@@ -37,16 +37,24 @@ jobs:
         with:
           environment-name: arc_env
           environment-file: ARC/environment.yml
+          condarc: |
+            channels:
+              - conda-forge
+              - danagroup
           cache-environment: true
+          cache-environment-key: py314v3-arc-env
```
calvinp0, line 45: Is this intended to be set in `main` too?

alongd: Yes, the gh-pages workflow creates rmg_env first, which sets a condarc with only conda-forge. Without the explicit condarc on the `arc_env` step, the danagroup channel gets lost and openbabel can't be resolved for Py 3.14. CI doesn't need it because it only creates one env.

alongd:
@calvinp0, ready for another round

pyproject.toml
```diff
@@ -6,3 +6,6 @@ requires = [
     "numpy",
 ]
 build-backend = "setuptools.build_meta"
+
+[project]
+requires-python = ">=3.14"
```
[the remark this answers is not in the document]
calvinp0: So, this should be done as well since it does a big warning when doing `make compile`

Can drop the dead fields form `setup.py` - removing `description=` and `author=` and `name=` and `version=` since `pyproject.toml` is the source of metadata truth. I guess move these fields also from `setup` to `pyproject`

calvinp0:
As messaged on Slack - we need to compile Openbabel on the Danagroup channel to either GLIBC 2.17 or 2.28 as 2.32 is not always supported by HPC servers

calvinp0 (approved):
Look good to me. I do think we need to fix those last cyclic imports but that can be done in a new PR

PR #870, opened 2026-04-15
Fix switch_ts resetting rotors convergence flag and stale rotors_dict

calvinp0:
delete_all_species_jobs blanket-set all output job_types to False, including rotors and bde which are initialised to True by initialize_output_dict.  For species with no torsional modes (e.g. cyclic TS from THF), no scan jobs are ever spawned, so rotors stays False and check_all_done incorrectly marks the TS as unconverged — even when opt, freq, sp, and IRC all passed.

Additionally, switch_ts did not reset rotors_dict, so determine_rotors was never re-called for the new TS geometry and stale scan results from the previous guess carried over.

Changes:
- Preserve the True default for rotors/bde in delete_all_species_jobs, matching initialize_output_dict.
- Reset rotors_dict and number_of_rotors in switch_ts when job_types['rotors'] is enabled, so the new geometry gets fresh rotor detection.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -2773,6 +2773,11 @@ class Scheduler(object):
         if os.path.isfile(freq_path):
             os.remove(freq_path)
         self.species_dict[label].populate_ts_checks()  # Restart the TS checks dict.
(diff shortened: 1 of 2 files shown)


Lilachn91 (approved):
LGTM, thanks :)

PR #871, opened 2026-04-17
Improve trsh attempts reporting

Lilachn91:
- Added explicit reporting of ESS troubleshooting attempt count.
- Cleaned troubleshooting-method reporting 
- Replaced hardcoded “geometry optimization” wording in failure logs with dynamic job type.

### DIFF
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -1137,10 +1137,22 @@ def trsh_ess_job(label: str,
         couldnt_trsh = True
 
     if couldnt_trsh:
-        logger.error(f'Could not troubleshoot geometry optimization for {label}! '
-                     f'Tried troubleshooting with the following methods: {ess_trsh_methods}')
+        # Count and remove 'trsh_attempt' entries for cleaner reporting
+        trsh_attempt_count = ess_trsh_methods.count('trsh_attempt')
+        filtered_methods = [method for method in ess_trsh_methods if method != 'trsh_attempt']
+        
+        # Build the message with the count and filtered methods
+        if trsh_attempt_count > 0 and filtered_methods:
+            message = f'Tried troubleshooting {trsh_attempt_count} times, with the following methods: {filtered_methods}'
+        elif trsh_attempt_count > 0:
+            message = f'Tried troubleshooting {trsh_attempt_count} times'
+        else:
+            message = f'Tried troubleshooting with the following methods: {filtered_methods}'
+        
+        logger.error(f'Could not troubleshoot {job_type} for {label}! '
+                     f'{message}')
         output_errors.append(f'Error: Could not troubleshoot {job_type} for {label}! '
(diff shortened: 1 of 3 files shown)


arc/job/trsh.py
```diff
@@ -1137,10 +1137,22 @@ def trsh_ess_job(label: str,
...
-        logger.error(f'Could not troubleshoot geometry optimization for {label}! '
-                     f'Tried troubleshooting with the following methods: {ess_trsh_methods}')
+        # Count and remove 'trsh_attempt' entries for cleaner reporting
+        trsh_attempt_count = ess_trsh_methods.count('trsh_attempt')
+        filtered_methods = [method for method in ess_trsh_methods if method != 'trsh_attempt']
+        
+        # Build the message with the count and filtered methods
+        if trsh_attempt_count > 0 and filtered_methods:
+            message = f'Tried troubleshooting {trsh_attempt_count} times, with the following methods: {filtered_methods}'
+        elif trsh_attempt_count > 0:
```
calvinp0: Is there a time when there would be no methods? I mean if it did attempt to troubleshoot x number of times, then it would have used a method, I think?

calvinp0: Also, when there is no methods applied, so no troubleshoot could be done (an edge case I guess?) Should then the message be something like "No applicable troubleshooting methods found".

arc/scheduler.py
```diff
@@ -3593,6 +3594,8 @@ def troubleshoot_ess(self,
                         f'Reached max troubleshooting attempts ({max_ess_trsh}).')
             self.output[label]['errors'] += f'Error: ESS troubleshooting attempts exhausted for {label} {job.job_type}; '
             return
+        logger.warning(f'Troubleshooting {label} job {job.job_name} '
+                           f'(attempt number {trsh_attempts + 1}).')
```
calvinp0: This is more a readability suggestion but we could do something like this:
```
next_attempt = trsh_attempts + 1
logger.warning(
    f'Troubleshooting {label} job {job.job_name} '
    f'(attempt number {next_attempt}).'
)
job.ess_trsh_methods.append('trsh_attempt')
```

That way if someone were to look at this code in the future, it would be quick to see that its the next attempt

arc/job/trsh.py
Lilachn91: trsh_attempt can appear with no other methods because scheduler appends trsh_attempt before calling trsh_ess_job. In trsh_ess_job, when the software is 'terachem' or there are no implemented trsh methods, couldnt_trsh is set to true but no methods are added.
I modified the message, thanks!

arc/scheduler.py
Lilachn91: Changed, thanks.

calvinp0 (dismissed):

calvinp0 (approved):

PR #876, opened 2026-04-20
Skip 2D-graph isomorphism enforcement for TS species

calvinp0:
## Summary
- ARC was enforcing 2D-graph isomorphism on TSs, but TS connectivity/bond-orders are not uniquely defined (partial forming/breaking bonds), so valid TS geometries were being rejected.
- `ARCSpecies.mol_from_xyz` now short-circuits the isomorphism check for TSs and simply accepts the perceived molecule.
- Scheduler: `is_ts` guards added at the three remaining `check_xyz_isomorphism` call sites (`check_directed_scan`, `check_directed_scan_job`, `troubleshoot_scan_job`). Matches the existing guards in `parse_composite_geo` and `parse_opt_geo`.
- Secondary fix: the non-TS warning message in `mol_from_xyz` now logs `perceived_mol` in the "and:" branch (it was logging `self.mol` twice).
- Docs updated in `TS_search.rst` and `advanced.rst` to describe the policy.

## Design note for reviewer
I kept the policy at the **call sites** rather than pushing an early-return into `check_xyz_isomorphism` itself. Rationale: returning `True` from a method named "check..." when no check ran conflates "passed" with "skipped", and several callers propagate the value into logs/restart dicts (`self.output[label]['isomorphism'] += '... passed ...'`). The scheduler already used this call-site-guard pattern at `parse_composite_geo` (line 2434) and `parse_opt_geo` (line 2537), so this keeps the codebase consistent.

One thing worth a second look: at `check_directed_scan_job`, `is_isomorphic=True` is now stored into `rotors_dict[i]['directed_scan'][...]['is_isomorphic']` for TSs. If any downstream consumer trusts that field as evidence a real check ran, a sentinel (`None` for skipped) or a rename would be better. I didn't change it because the same semantic already exists in the docstring for `mol_from_xyz`, but @alongd please weigh in.

## Test plan
- [x] `arc.species.species_test.TestARCSpecies.test_ts_mol_from_xyz_skips_isomorphism_enforcement` — TS accepts perceived xyz-derived molecule even when stored 2D graph disagrees
- [x] `arc.scheduler_test.TestScheduler.test_check_directed_scan_job_skips_isomorphism_for_ts` — mock-based, asserts `check_xyz_isomorphism` is NOT called and `is_isomorphic=True` is recorded
- [x] `arc.scheduler_test.TestScheduler.test_troubleshoot_scan_job_skips_isomorphism_for_ts` — mock-based, asserts `check_xyz_isomorphism` is NOT called and `final_xyz` / `run_opt_job` path runs
- [x] Existing `test_mol_from_xyz` and `test_check_xyz_isomorphism` still pass (non-TS behavior unchanged)
- [x] Full suite run in CI

cc @alongd — would value your eyes on the design note above, especially whether the call-site-guard pattern is the right call or whether you'd prefer centralizing in `check_xyz_isomorphism` itself.

### DIFF
--- a/arc/scheduler_test.py
+++ b/arc/scheduler_test.py
@@ -6,7 +6,7 @@ This module contains unit tests for the arc.scheduler module
 """
 
 import unittest
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 import os
 import shutil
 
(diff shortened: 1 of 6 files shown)


arc/species/species.py
```diff
@@ -1639,28 +1641,32 @@ def mol_from_xyz(self,
...
+                                       f'{perceived_mol.copy(deep=True).to_adjacency_list()}')
+                        raise SpeciesError(f'XYZ and the 2D graph representation for {self.label} are not compliant.')
+                    if not self.keep_mol:
+                        if is_mol_valid(perceived_mol, charge=self.charge, multiplicity=self.multiplicity, n_radicals=self.number_of_radicals):
                             self.mol = perceived_mol
+                        else:
+                            try:
+                                order_atoms(ref_mol=perceived_mol, mol=self.mol)
+                            except SanitizationError:
+                                self.mol = perceived_mol
```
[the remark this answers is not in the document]
calvinp0: the `order_atoms()` fallback was already gated under `if not self.kee_mol` on main. So both `self.mol = perceived_mol` assignment and the `order_atoms` fallback sat inside the same `if not self.keep_mol` block. So, `keep_mol=True` has always been don't touch `self.mol`. The PR perserves that exactly for non-TSs.

The TS branch is deliberately different: a transition state is not a species in the RMG sense, and a 2D molecular graph cannot represent the partial forming/breaking bonds at a saddle point

arc/scheduler_test.py
```diff
@@ -1005,6 +1005,90 @@ def test_switch_ts_rotors_reset(self, mock_run_opt):
...
+        job_mock.torsions = [[3, 1, 2, 4]]
+        job_mock.job_name = 'scan_a200'
+
+        with patch('arc.species.species.ARCSpecies.check_xyz_isomorphism') as mock_iso, \
+                patch('arc.scheduler.Scheduler.delete_all_species_jobs'):
+            sched.troubleshoot_scan_job(job=job_mock, methods={'change conformer': new_xyz})
+
+        mock_iso.assert_not_called()
+        self.assertEqual(sched.species_dict['TS_trsh'].final_xyz, new_xyz)
+        mock_run_opt.assert_called_once()
```
alongd, line 1119: add an extra blank line

arc/scheduler_test.py
```diff
@@ -1005,6 +1005,90 @@ def test_switch_ts_rotors_reset(self, mock_run_opt):
...
+                patch('arc.scheduler.parser.parse_geometry', return_value=ts_xyz), \
+                patch('arc.scheduler.parser.parse_e_elect', return_value=-123.45):
+            sched.check_directed_scan_job(label='TS_dirscan', job=job_mock)
+
+        mock_iso.assert_not_called()
+        recorded = sched.species_dict['TS_dirscan'].rotors_dict[0]['directed_scan'][('45.00',)]
+        self.assertTrue(recorded['is_isomorphic'])
+
+    @patch('arc.scheduler.Scheduler.run_opt_job')
+    def test_troubleshoot_scan_job_skips_isomorphism_for_ts(self, mock_run_opt):
```
alongd, line 1076: Can you also add a test for troubleshoot_scan_job when not applying 'change conformer'?

arc/species/species_test.py
```diff
@@ -1684,6 +1684,43 @@ def test_mol_from_xyz(self):
...
+11 H u0 p0 c0 {5,S}
+12 H u0 p0 c0 {5,S} {17,vdW}
+13 H u0 p0 c0 {5,S}
+14 H u0 p0 c0 {6,S}
+15 H u0 p0 c0 {6,S}
+16 H u0 p0 c0 {6,S}
+17 H u0 p0 c0 {1,S} {12,vdW}"""
+
+        spc = ARCSpecies(label='TS0', adjlist=adj, xyz=xyz, is_ts=True, multiplicity=2, charge=0)
+        self.assertIn('3  H u0 p0 c0 {2,S}', spc.mol.to_adjacency_list())
```
alongd: I wonder if we can test while making it less brittle.
BTW, the atom order in an adjlist isn't always that of a Molecule anyway.

maybe change the test to:
just create a Molecuel object from the adjlist provided. See that ti has a vdW bond. then create the TS ARCSpecies, and see that ot doesn't have that bond in the .mol attribute?

alongd (commented):
Added small comments

alongd:
Thanks, @calvinp0, the design looks good

alongd (approved):
Thanks!

PR #879, opened 2026-04-25
species/zmat: import vectors as a module to break import cycle

calvinp0:
## Summary                                                                                                                                                                                                         
  - `arc.species.zmat` → `arc.species.vectors` → `arc.species.converter` → `arc.species.zmat` was a real module-level import cycle, latent since 2023. It only stayed harmless because `arc/species/__init__.py`
  happened to load `converter` before anything reaching `vectors`. CodeQL flagged both ends: `zmat.py:39` (cycle origin) and seven `converter.py:33-39` "may not be defined" errors on the from-name imports of zmat 
  symbols.
  - **Commit 1 (`f07d1f21`)**: switches `zmat.py` to module-style `from arc.species import vectors` and prefixes the 17 call sites. This was a *necessary but insufficient* step — module imports tolerate partial   
  initialization mid-cycle, but the structural cycle still existed.                                                                                                                                                  
  - **Commit 2 (`0f5d3ad9`)**: actually breaks the cycle by removing `vectors.py`'s module-level import of `converter`. The only call site (`get_vector`) was using `converter.xyz_to_x_y_z(xyz)` to build three
  full-length axis tuples just to read two atoms' coordinates — replaced with two direct `xyz['coords'][i]` lookups. Net: 6 lines → 3 lines, O(N) → O(1), and the heavy `check_xyz_dict` validation overhead is gone 
  (caller passes a well-formed dict).                                                                            
  - After this PR there is no module-level path from `vectors` back to `zmat`, so all 8 CodeQL alerts clear.

### DIFF
--- a/arc/species/vectors.py
+++ b/arc/species/vectors.py
@@ -9,7 +9,6 @@ from typing import List, Union
 from arc.common import logger
 from arc.exceptions import VectorsError
 from arc.molecule.molecule import Molecule
-from arc.species import converter
 
 
 def get_normal(v1: List[float],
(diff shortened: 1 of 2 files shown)


arc/species/vectors.py
```diff
@@ -361,11 +360,9 @@ def get_vector(pivot: int,
     Returns: list
          A vector pointing from the pivotal atom towards the anchor atom.
     """
-    x, y, z = converter.xyz_to_x_y_z(xyz)
-    dx = x[anchor] - x[pivot]
-    dy = y[anchor] - y[pivot]
-    dz = z[anchor] - z[pivot]
-    return [dx, dy, dz]
+    px, py, pz = xyz['coords'][pivot]
```
alongd, line 363: And it's even more efficient now :)

alongd (approved):
Excellent, thanks for getting to the bottom of this!

PR #881, opened 2026-04-27
Break down monolithic driver and heuristics tests

alongd:
Splits the catch-all `test_map_abstractions` and heuristics tests into one test per reaction case, so failures point at a specific scenario instead of one giant test. Shared XYZ fixtures live on `setUpClass`. No production-code changes.


ranbenayoun (approved):
Looks good to me

PR #883, opened 2026-04-30
env_run: isolate adapter subprocesses from arc_env activation vars

calvinp0:
## Summary

ARC runs inside `arc_env`, but the AutoTST, GCN, and TorchANI adapters shell out to scripts that live in their own envs (`tst_env`, `ts_gcn_env`, `tani_env`). The previous code path was:

```python
subprocess.run("source ~/.bashrc; <target_env_python> <script>", shell=True)
```

This invokes the target env's interpreter directly without deactivating `arc_env` first. ARC's exported activation vars (`BABEL_LIBDIR`, `LD_LIBRARY_PATH`, `CONDA_PREFIX`, ...) stay bound to `arc_env`'s paths in the child, so shared libraries in the target env then resolve plugins against the wrong tree — producing ABI-mismatch crashes (most visibly OpenBabel plugin loading in `tst_env`).

## What this PR does

- Adds `arc/job/env_run.py` exposing `run_in_conda_env(python, script, *args)`, which routes the subprocess through `<launcher> run -n <env>` so the target env's `activate.d` hooks fire and rebind the leaked vars to the correct paths.
- Detects an available launcher in preference order: the active one per `CONDA_EXE`/`MAMBA_EXE`, then `conda` → `mamba` → `micromamba` on PATH. Selects the right stdio flag per launcher (`--no-capture-output` for conda/mamba, omitted for micromamba which rejects it).
- Wires the three callers (`arc/job/adapters/torch_ani.py`, `arc/job/adapters/ts/autotst_ts.py`, `arc/job/adapters/ts/gcn_ts.py`) over to the helper and drops the now-unused `subprocess` imports.
- Adds `from __future__ import annotations` to `autotst_script.py` and `tani_script.py` so their PEP 604 union syntax (`X | Y`) imports cleanly under whatever Python version the target env ships (tst_env is 3.9; tani_env varies).

## Test plan

- [x] AutoTST TS-search run completes without OpenBabel plugin errors
- [x] GCN TS-search run completes and produces guess xyz
- [x] TorchANI single-point job completes and writes output yml
- [x] Verified on a host where `conda` is the active launcher; verify also under `mamba` / `micromamba` if available

### DIFF
--- a/arc/job/env_run.py
+++ b/arc/job/env_run.py
@@ -1,0 +1,144 @@
+"""Invoke a script under a sibling conda/mamba env, isolated from ARC's env.
+
+ARC runs inside ``arc_env``. Several adapters (AutoTST, GCN, TorchANI)
+shell out to scripts that live in their *own* envs (``tst_env``,
+``ts_gcn``, ``tani_env``). Running the target env's ``python``
+binary directly leaves ARC's exported activation vars (``BABEL_LIBDIR``,
+``LD_LIBRARY_PATH``, ``CONDA_PREFIX``, ...) bound to ``arc_env``'s
+paths in the child, which causes ABI-mismatch crashes when shared
+libraries in the child resolve plugins against the wrong env's tree.
+
+Routing through a launcher's ``run`` subcommand makes the launcher
+deactivate the caller env and re-activate the target, so the target
+env's own ``activate.d`` hooks fire and bind those vars to its paths.
+
+Three launchers are supported, in preference order:
+
+1. ``conda`` — needs ``--no-capture-output`` to avoid buffering child
+   stdio.
+2. ``mamba`` — same parser as conda for ``run``; also needs
+   ``--no-capture-output``.
+3. ``micromamba`` — independent C++ reimplementation; streams stdio by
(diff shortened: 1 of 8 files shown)


alongd:
The CI failure here is an env drift for our tani_env, [fix is here](https://github.com/ReactionMechanismGenerator/ARC/pull/695/changes/6f5704c61a3341206ea69583c978cdb9e881db59) (can cherry-pick it)

arc/job/env_run.py
```diff
@@ -0,0 +1,121 @@
...
+    env_prefix = env_prefix_from_python(python_executable)
+    launcher, extra_flags = _detect_launcher()
+    argv = [
+        launcher, "run", *extra_flags,
+        "-p", env_prefix,
+        "python", script_path,
+        *script_args,
+    ]
+    logger.debug("env-run: %s", " ".join(argv))
+    return subprocess.run(argv, check=check)
```
alongd: We should centralize here the capture of stdout and stderr. Let's put it all here and log for better maintainability

alongd (commented):
LGTM, thanks. one comment

alongd (approved):

Issue #694, opened 2023-08-21
ARC's report of currently running jobs


alongd:
ARC reports once an hour the "currently running jobs" in its log file.
For long jobs these reports are repetitive, and clutter the log file.
We should consider reporting the currently running jobs in a different file, and once an hour replace its content instead of appending.
It could also be nice to report the time before the rest of the content in this file.

PR #886, opened 2026-05-07
Write an hourly running jobs status into a separate YAML

alongd:
don't spam the ARC.log if the report repeats it self and ARC is idle. outsource the running jobs report into a separate YAML, only update it when needed, and drop a heartbeat line into ARC's log file.

addresses #694


LeenFahoum (approved):
Looks good to me

PR #887, opened 2026-05-07
Add GoFlow and RitS TS search adapters

alongd:
## Summary

Adds **GoFlow** and **RitS**, two flow-matching ML TS-guess adapters, sharing
the same lifecycle, the same Tier-1 / Tier-2 test split, and the same
"warn + skip" behavior when their respective env (`goflow_env` / `rits_env`)
or pretrained checkpoint is unavailable.

Both are **opt-in only** — their inference stacks are heavyweight, so neither
is in the default `ts_adapters` list (which is unchanged: `heuristics`,
`linear`, `AutoTST`, `GCN`, `xtb_gsm`, `orca_neb`). Enable either via
`ts_adapters: ['goflow', ...]` in the input file.

Both are also scoped to **unimolecular reactions**, exactly like the Linear
adapter: `all_families_ts_adapters` stays `[]`, both are registered per family
in `ts_adapters_by_rmg_family` for the same 58 families that list `linear`
(so `H_Abstraction` is excluded) and in `ts_adapters_for_unknown_unimolecular`,
and each adapter guards at runtime on `rxn.is_unimolecular()`, warning and
skipping otherwise. `A + B <=> C` qualifies; a 2↔2 abstraction such as
`CH4 + OH <=> CH3 + H2O` does not.

- **GoFlow** — flow-matching, E(3)-equivariant network that predicts TS
  geometries from atom-mapped reactant + product SMILES. RDB7-trained, so on
  top of the molecularity guard it enforces a runtime domain guard
  (`{H, C, N, O, F}`, ≤100 atoms).
- **RitS** — flow-matching network that predicts TS geometries directly from
  atom-mapped reactant + product 3D structures. Handles multi-bond
  rearrangements and charged species.

Sister-repo path discovery is consolidated into a single
`arc/settings/external_paths.py` (with matching test module). Both adapters
are inlined into the main CI lane (cached env + checkpoint per adapter), so
their Tier-2 end-to-end tests actually run on every PR.

## Sources

**GoFlow**
- Paper: Galustian et al., *Digital Discovery* 2025 — DOI [10.1039/D5DD00283D](https://doi.org/10.1039/D5DD00283D)
- Upstream code: <https://github.com/heid-lab/goflow_lean>
- Pretrained checkpoint (Dana Research Group reproduction): <https://zenodo.org/records/20073635> — `epoch_316.ckpt`, SHA-256-verified by `install_goflow.sh`.

**RitS**
- Paper: Isayev lab — DOI [10.26434/chemrxiv.15001681/v1](https://doi.org/10.26434/chemrxiv.15001681/v1)
- Upstream code: <https://github.com/isayevlab/RitS>
- Pretrained checkpoint (Dana Research Group reproduction): <https://zenodo.org/records/19474153> — `rits.ckpt`, SHA-256
`2c6fdacc13a304cb8bb4030881ff4198de647d8f8f7fdaa878414e2514be0668`. `install_rits.sh` downloads and verifies it automatically.


devtools/install_all.sh
```diff
@@ -26,16 +26,20 @@ run_devtool () { bash "$DEVTOOLS_DIR/$1" "${@:2}"; }
...
 while [[ $# -gt 0 ]]; do
     case "$1" in
-        --no-clean) SKIP_CLEAN=true ;;
-        --no-ext)   SKIP_EXT=true  ;;
-        --no-arc)   SKIP_ARC=true  ;;
+        --no-clean)  SKIP_CLEAN=true ;;
+        --no-ext)    SKIP_EXT=true  ;;
+        --no-arc)    SKIP_ARC=true  ;;
+        --no-goflow) SKIP_GOFLOW=true ;;
+        --no-rits)   SKIP_RITS=true ;;
```
kfir4444, line 42: Should we really add flag/software here? what about the other software?

arc/settings/settings_test.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
kfir4444, line 2: Great! We can probably expand this to other adapters. (out of scope for this PR, can you open an issue about that?)

arc/job/adapters/common.py
```diff
@@ -91,7 +91,7 @@
                              'nitrile_hydrolysis': ['heuristics']
                              }
 
-all_families_ts_adapters = []
+all_families_ts_adapters = ['goflow', 'rits']
```
kfir4444, line 94: Including bi-molecular?

arc/settings/external_paths_test.py
```diff
@@ -0,0 +1,211 @@
...
+                             external_paths.find_goflow_ckpt(repo_path=tmp))
+
+
+class TestFindGoFlowFeatDict(unittest.TestCase):
+    """find_goflow_feat_dict() — locates the atom-feature codebook pickle."""
+
+    def test_returns_none_when_no_repo_and_no_env_var(self):
+        os.environ.pop('ARC_GOFLOW_FEAT_DICT', None)
+        self.assertIsNone(external_paths.find_goflow_feat_dict(repo_path=None))
+
```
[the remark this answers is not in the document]
kfir4444: Can you look into that? that sounds like it may endanger testers' env.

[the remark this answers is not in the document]
kfir4444: I see many similar issues- I think they may be important.

arc/job/adapters/scripts/goflow_script.py
```diff
@@ -0,0 +1,454 @@
...
+    if requested != 'auto':
+        return requested
+    try:
+        import torch
+    except ImportError:
+        return 'cpu'
+    return 'cuda' if torch.cuda.is_available() else 'cpu'
+
+
+def _validate_ckpt(ckpt_path: str) -> None:
```
kfir4444: Does that occure per compute? isn't most of it duplicated later in `run_goflow_inference`? I think we can safley make this function handle the validation and return the loaded model, and not loading it twice with `torch.load`.

arc/job/adapters/scripts/goflow_script.py
```diff
@@ -0,0 +1,454 @@
...
+                           Optional keys: ``n_samples`` (default 10), ``num_steps`` (default 25), ``device`` (default ``'auto'``).
+
+    Returns:
+        list[dict]: One TSGuess-shaped dict per sample (or a single failure sentinel if the pipeline raised).
+                    Each entry has keys ``method``,  ``method_direction``, ``method_index``, ``initial_xyz``,
+                    ``success``, ``execution_time``.
+    """
+    t0 = datetime.datetime.now()
+    try:
+        # Late imports: this function only runs inside goflow_env.
```
kfir4444: I thought the entire script runs in the goflow env

kfir4444 (changes requested):
Thanks @alongd! please see the comments.

devtools/install_all.sh
alongd: These envs are large, so we don't install them regularly in CI using these flags, at least indirectly (through `install-ci`). Then CI installs them in separate cached steps

arc/settings/settings_test.py
alongd: [Done](https://github.com/ReactionMechanismGenerator/ARC/issues/929)

arc/job/adapters/common.py
alongd: Yes, I tried CH4 + OH <=> CH3 + H2O for example, it gives results with correct connectivity (GoFlow seems better). So bimolecular works, but we;ll assess after the ARC benchmark and decide.

arc/settings/external_paths_test.py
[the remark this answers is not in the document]
alongd: The copilot comments are outdated and were addressed. It should be already fixed (I just didn't mark the copilot comments are resolved)

arc/job/adapters/scripts/goflow_script.py
alongd: you're right, yes - it's per reaction. I'll fix

arc/settings/external_paths_test.py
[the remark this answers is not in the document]
alongd: Already handled — every `os.environ.pop` in `external_paths_test.py` is now inside a `mock.patch.dict(os.environ, ...)` block, which restores the full dict on exit (12 call sites; none of the other new test files touch `os.environ`).

Worth noting the blast radius was narrower than it reads: `os.environ` is the pytest process's own copy, so a tester's shell was never at risk — the real exposure was cross-test leakage within a run, which matters under `-n 6`.

Happy to go through the other Copilot comments if you can point at which ones you meant.

arc/job/adapters/scripts/goflow_script.py
alongd: Right. I'll reword

arc/job/adapters/scripts/goflow_script.py
```diff
@@ -0,0 +1,461 @@
...
+                    ``success``, ``execution_time``.
+    """
+    t0 = datetime.datetime.now()
+    try:
+        # Late imports: at run time this whole script executes inside goflow_env, so these are
+        # always available here. They are imported inside the function rather than at module
+        # scope so that the module stays importable from ARC's main env, where torch/hydra/
+        # goflow are absent — arc/job/adapters/scripts/goflow_script_test.py imports it there
+        # to unit-test the pure-Python helpers above.
+        import torch
```
kfir4444, line 338: As a convention, we import at the beginning of the script, unless there is a reason not to.

arc/settings/settings.py
```diff
@@ -89,7 +98,10 @@
 supported_ess = ['cfour', 'gaussian', 'mockter', 'molpro', 'orca', 'qchem', 'terachem', 'onedmin', 'xtb', 'torchani', 'openbabel']
 
 # TS methods to try when appropriate for a reaction (other than user guesses which are always allowed):
-ts_adapters = ['heuristics', 'AutoTST', 'GCN', 'xtb_gsm', 'orca_neb']
+# Note: 'RitS' is intentionally NOT in the default — its env (rits_env +
+# pretrained ckpt) is heavyweight, so users opt in explicitly via
+# ``ts_adapters: ['rits', ...]`` in their input.yml.
+ts_adapters = ['heuristics', 'AutoTST', 'goflow', 'orca_neb']
```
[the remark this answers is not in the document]
kfir4444: Was this adressed?

arc/job/adapters/scripts/goflow_script.py
```diff
@@ -0,0 +1,358 @@
...
+    n_atoms = int(lines[i].strip())
+    i += 2  # skip count + comment line
+    coords: List[List[float]] = []
+    for _ in range(n_atoms):
+        if i >= len(lines):
+            break
+        parts = lines[i].split()
+        if len(parts) >= 4:
+            coords.append([float(parts[1]), float(parts[2]), float(parts[3])])
+        i += 1
```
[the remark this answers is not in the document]
kfir4444: Was this adressed?

kfir4444 (changes requested):
Thanks @alongd, a few more comments.

alongd:
summary of offline discussion: just unimolecular

arc/settings/settings.py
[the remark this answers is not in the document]
alongd: Yes — addressed. At HEAD (`70523b8`) neither adapter is in the default:

- `arc/settings/settings.py:105` → `ts_adapters = ['heuristics', 'linear', 'AutoTST', 'GCN', 'xtb_gsm', 'orca_neb']` (GCN and xtb_gsm restored; no `goflow`/`rits`)
- `arc/settings/settings_test.py:31,37,43` assert the exact list and that neither `goflow` nor `rits` appears
- `arc/main_test.py:63` fixture matches the same list
- `docs/source/TS_search.rst:326,417` say opt-in only for both

The PR description is the one place still describing GoFlow as default — I'll fix that.

arc/job/adapters/common.py
alongd: removing bimolecular

arc/job/adapters/scripts/goflow_script.py
alongd: Deliberate, and consistent with the rest of `arc/job/adapters/scripts/`. `goflow_script.py` runs inside `goflow_env`, but `goflow_script_test.py` imports it from `arc_env` to unit-test the pure-Python helpers (`read_xyz_positions`, `format_xyz_block`, the YAML IO, `_resolve_device`). `torch`, `hydra`, `torch_geometric` and `goflow` do not exist in `arc_env`, so a module-scope import would make that test module un-importable and we would lose the helper coverage.

Same pattern already in the repo for the same reason:
- `gcn_script.py:60,66,155`
- `ase_script.py:68,99,107,260,361,380`
- `kinbot_script.py:233,244,278`

CLAUDE.md's rule is "don't import inside functions ... unless there's a strong exceptional reason" — env-gated deps are that reason, and the rationale is spelled out in the comment at the import site. Happy to split the helpers into their own module if you'd rather see it structurally enforced, but that would diverge from how the other adapter scripts are laid out.

arc/job/adapters/common.py
alongd: Update — reversed on this. GoFlow and RitS are now scoped to **unimolecular** reactions only, exactly like the Linear adapter:

- `all_families_ts_adapters` is back to `[]`
- both are registered per family in `ts_adapters_by_rmg_family`, for the same 58 families that list `linear` — so `H_Abstraction` is excluded
- both are in `ts_adapters_for_unknown_unimolecular`
- each adapter now guards at runtime with `rxn.is_unimolecular()` and warns+skips otherwise, mirroring `linear.py:877`

`is_unimolecular()` is true when either side has a single species, so `A + B <=> C` still runs; a 2↔2 abstraction like `CH4 + OH <=> CH3 + H2O` does not. The four RitS and one GoFlow H-abstraction end-to-end tests now assert the adapter declines before entering `goflow_env`/`rits_env`, and a new `[CH2]CO <=> CC[O]` fixture carries the GoFlow end-to-end geometry checks instead.

arc/job/adapters/scripts/goflow_script.py
alongd: Fixed — `_load_and_validate_ckpt()` now returns the loaded checkpoint, so `run_goflow_inference()` consumes it instead of calling `torch.load` a second time on the ~57 MB file. One deserialization per reaction.

arc/job/adapters/scripts/goflow_script.py
alongd: Reworded. The comment now says the whole script executes inside `goflow_env` at run time and that the imports are function-scoped only so the module stays importable from `arc_env` for the helper unit tests — see the reply on the import-convention thread for the full rationale and the in-repo precedent.

kfir4444 (dismissed):
Thanks @alongd!

kfir4444 (approved):
LGTM!

PR #888, opened 2026-05-13
Improve Orca Optimization and Frequency Convergence Settings

Ophir-Weisz:
Enhanced frequency calculations to use defgrid3 instead of defgrid2, for consistency with the DFT grid of the fine opt job.

### DIFF
--- a/arc/job/adapters/orca.py
+++ b/arc/job/adapters/orca.py
@@ -1,5 +1,5 @@
 """
-An adapter for executing Orca 5 jobs
+An adapter for executing Orca 5/6 jobs
 
 https://orcaforum.kofo.mpg.de/app.php/portal
 """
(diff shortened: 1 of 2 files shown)


arc/job/adapters/orca.py
```diff
@@ -325,6 +326,7 @@ def write_input_file(self) -> None:
                 self.add_to_args(val="""
 %geom
 Calc_Hess true # calculation of the exact Hessian before the first opt step
+Recalc_Hess 5
```
kfir4444: Please note that this is very expensive, and sometimes leads to stability issues. We should also reconsider the earlier line, and putting it as a trash and not default.

Lilachn91 (approved):

PR #889, opened 2026-05-14
scripts: harden rmg_kinetics & rmg_thermo

alongd:
Deep-copy DB kinetics before scaling so a second query can't read a doubly-scaled A, gate change_rate on Arrhenius/EP so non-Arrhenius forms (Chebyshev/PLOG/etc.) are skipped safely, add sys.path bootstrap so `from common import ...` works regardless of CWD, add an --output flag so the input YAML isn't overwritten, document the training-entry filter and output units, and harden the helper's debug print against reactions without reactants. Adds argparse + subprocess unit tests.

### DIFF
--- a/arc/scripts/rmg_kinetics.py
+++ b/arc/scripts/rmg_kinetics.py
@@ -2,12 +2,25 @@
 # encoding: utf-8
 
 """
-A standalone script to run RMG
-and get kinetic rate coefficients for reactions
+A standalone script to run RMG and get kinetic rate coefficients for reactions.
+
+Output units (per entry returned by ``get_kinetics_from_reactions``):
+    - ``A``:    cm^3/(mol*s) for bimolecular reactions, s^-1 for unimolecular
+                (3-body: cm^6/(mol^2*s)). Reported in the units stored on the
+                Arrhenius object after the SI->cm conversion below.
+    - ``n``:    dimensionless temperature exponent.
+    - ``Ea``:   kJ/mol (converted from SI J/mol).
+    - ``T_min``, ``T_max``: K.
 """
 
+import copy
 import os
+import sys
 from typing import Dict, List, Optional, Tuple
+
+# Make ``from common import ...`` work no matter how this script is invoked
+# (e.g. ``python /abs/path/to/rmg_kinetics.py``, ``cd elsewhere && python ...``).
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
 from common import parse_command_line_arguments, read_yaml_file, save_yaml_file
 
@@ -15,6 +28,7 @@ from rmgpy.data.kinetics.common import find_degenerate_reactions
 from rmgpy.data.kinetics.family import KineticsFamily
 from rmgpy.data.rmg import RMGDatabase
 from rmgpy import settings as rmg_settings
+from rmgpy.kinetics import Arrhenius, ArrheniusEP, KineticsModel
 from rmgpy.reaction import same_species_lists, Reaction
 from rmgpy.species import Species
 
@@ -35,11 +49,12 @@ def main():
     """
     args = parse_command_line_arguments()
     input_file = args.file
+    output_file = args.output or input_file
     reaction_list = read_yaml_file(path=input_file)
     if not isinstance(reaction_list, list):
         raise ValueError(f'The content of {input_file} must be a list, got {reaction_list} which is a {type(reaction_list)}')
     result = get_rate_coefficients(reaction_list)
-    save_yaml_file(path=input_file, content=result)
+    save_yaml_file(path=output_file, content=result)
 
 
 def get_rate_coefficients(reaction_list: List[Dict]) -> List[Dict]:
@@ -62,6 +77,42 @@ def get_rate_coefficients(reaction_list: List[Dict]) -> List[Dict]:
     return reaction_list
 
 
+def get_a_factor_si_to_cm(num_reactants: int) -> float:
+    """
+    Get the factor that converts an Arrhenius A-factor from SI (m-based) units to
+    cm-based units, based on the reaction molecularity.
+
+    Args:
+        num_reactants (int): The number of reactants.
+
+    Returns:
+        float: 1.0 for unimolecular (s^-1), 1e6 for bimolecular (m^3->cm^3),
+               1e12 for termolecular (m^6->cm^6). Defaults to 1.0 otherwise.
+    """
+    return {1: 1.0, 2: 1e6, 3: 1e12}.get(num_reactants, 1.0)
+
+
+def scale_kinetics_by_degeneracy(kinetics: KineticsModel,
+                                 degeneracy: float,
+                                 ) -> KineticsModel:
+    """
+    Scale Arrhenius-type kinetics in place by the reaction-path degeneracy.
+
+    Non-Arrhenius forms (e.g. Chebyshev, PDepArrhenius) are returned unchanged,
+    since ``change_rate`` would otherwise corrupt their parameters.
+
+    Args:
+        kinetics (KineticsModel): An RMG kinetics object.
+        degeneracy (float): The reaction-path degeneracy to scale by.
+
+    Returns:
+        KineticsModel: The (possibly scaled) kinetics object.
+    """
+    if isinstance(kinetics, (Arrhenius, ArrheniusEP)):
+        kinetics.change_rate(degeneracy)
+    return kinetics
+
+
 def determine_rmg_kinetics(rmgdb: RMGDatabase,
                            reaction: Reaction,
                            dh_rxn298: Optional[float] = None,
@@ -71,6 +122,13 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
     Determine kinetics for `reaction` (an RMG Reaction object) from RMG's database, if possible.
     Assigns a list of all matching entries from both libraries and families.
 
+    Note:
+        Family entries originating from the training set are intentionally filtered out
+        (an empty returned list therefore means "no matching libraries and only training-set
+        family hits", not necessarily "no match at all"). Database kinetics are deep-copied
+        before any in-place mutation (degeneracy scaling, unit conversion) so the loaded
+        ``rmgdb`` instance remains unchanged across calls.
+
     Args:
         rmgdb (RMGDatabase): The RMG database instance.
         reaction (Reaction): The RMG Reaction object.
@@ -79,6 +137,7 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
 
     Returns: list[dict]
         All matching RMG reactions kinetics (both libraries and families) as a dict of parameters.
+        Empty list if nothing matched (or only training-set entries matched).
     """
     rmg_reactions = list()
     # Libraries:
@@ -89,19 +148,22 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
                 library_reaction.comment = f'Library: {library.label}'
                 rmg_reactions.append(library_reaction)
                 break
-    # # Families:
-    A_units = "cm^3/(mol*s)" if len(reaction.reactants) == 2 else "s^-1"
+    # Families:
+    a_factor = get_a_factor_si_to_cm(len(reaction.reactants))
     fam_list = loop_families(rmgdb, reaction)
     dh_rxn298 = dh_rxn298 or get_dh_rxn298(rmgdb=rmgdb, reaction=reaction)  # J/mol
     for family, degenerate_reactions in fam_list:
         for deg_rxn in degenerate_reactions:
             kinetics_list = family.get_kinetics(reaction=deg_rxn, template_labels=deg_rxn.template, degeneracy=deg_rxn.degeneracy)
             for kinetics_detailes in kinetics_list:
-                kinetics = kinetics_detailes[0]
-                kinetics.change_rate(deg_rxn.degeneracy)
+                # Deep-copy before mutating so the database object isn't double-scaled
+                # if the same family rule is queried again for another reaction.
+                kinetics = copy.deepcopy(kinetics_detailes[0])
+                kinetics = scale_kinetics_by_degeneracy(kinetics, deg_rxn.degeneracy)
                 if hasattr(kinetics, 'to_arrhenius'):
                     kinetics = kinetics.to_arrhenius(dh_rxn298)  # Convert ArrheniusEP to Arrhenius
-                kinetics.A.value_si = kinetics.A.value_si * (1e6 if A_units == "cm^3/(mol*s)" else 1)
+                if a_factor != 1.0 and isinstance(kinetics, Arrhenius):
+                    kinetics.A.value_si = kinetics.A.value_si * a_factor
                 deg_rxn.kinetics = kinetics
                 deg_rxn.comment = f'Family: {deg_rxn.family}'
                 if 'training' in deg_rxn.kinetics.comment:
@@ -150,7 +212,11 @@ def get_kinetics_from_reactions(reactions: List[Reaction]) -> List[Dict]:
     """
     kinetics_list = list()
     for rxn in reactions:
-        print(f'rxn: {rxn}, kinetics: {rxn.kinetics}, comment: {rxn.comment}')
+        try:
+            rxn_repr = str(rxn)
+        except (TypeError, AttributeError):
+            rxn_repr = '<reaction without reactants/products labels>'
+        print(f'rxn: {rxn_repr}, kinetics: {rxn.kinetics}, comment: {rxn.comment}', file=sys.stderr)
         kinetics_list.append({
             'kinetics': rxn.kinetics.__repr__(),
             'comment': rxn.comment,
--- a/arc/scripts/rmg_kinetics.py
+++ b/arc/scripts/rmg_kinetics.py
@@ -2,12 +2,26 @@
 # encoding: utf-8
 
 """
-A standalone script to run RMG
-and get kinetic rate coefficients for reactions
+A standalone script to run RMG and get kinetic rate coefficients for reactions.
+
+Output units (per entry returned by ``get_kinetics_from_reactions``):
+    - ``A``:    cm/mol/s convention (s^-1, cm^3/(mol*s), cm^6/(mol^2*s), ...),
+                obtained from the SI value via RMG's
+                ``A.get_conversion_factor_from_si_to_cm_mol_s()`` (the same
+                dimensionality-based factor RMG uses to write Chemkin files),
+                so it is correct regardless of the matched entry's molecularity.
+    - ``n``:    dimensionless temperature exponent.
+    - ``Ea``:   kJ/mol (converted from SI J/mol).
+    - ``T_min``, ``T_max``: K.
 """
 
 import os
+import sys
 from typing import Dict, List, Optional, Tuple
+
+# Make ``from common import ...`` work no matter how this script is invoked
+# (e.g. ``python /abs/path/to/rmg_kinetics.py``, ``cd elsewhere && python ...``).
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
 from common import parse_command_line_arguments, read_yaml_file, save_yaml_file
 
@@ -35,11 +49,12 @@ def main():
     """
     args = parse_command_line_arguments()
     input_file = args.file
+    output_file = args.output or input_file
     reaction_list = read_yaml_file(path=input_file)
     if not isinstance(reaction_list, list):
         raise ValueError(f'The content of {input_file} must be a list, got {reaction_list} which is a {type(reaction_list)}')
     result = get_rate_coefficients(reaction_list)
-    save_yaml_file(path=input_file, content=result)
+    save_yaml_file(path=output_file, content=result)
 
 
 def get_rate_coefficients(reaction_list: List[Dict]) -> List[Dict]:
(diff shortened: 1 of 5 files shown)


arc/scripts/rmg_kinetics.py
```diff
@@ -2,19 +2,33 @@
...
 
 """
-A standalone script to run RMG
-and get kinetic rate coefficients for reactions
+A standalone script to run RMG and get kinetic rate coefficients for reactions.
+
+Output units (per entry returned by ``get_kinetics_from_reactions``):
+    - ``A``:    cm^3/(mol*s) for bimolecular reactions, s^-1 for unimolecular
+                (3-body: cm^6/(mol^2*s)). Reported in the units stored on the
+                Arrhenius object after the SI->cm conversion below.
```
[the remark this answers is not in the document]
alongd: Good catch — fixed in 342f51856. Replaced the binary unimolecular/bimolecular flag with `get_a_factor_si_to_cm(num_reactants)` (`{1: 1.0, 2: 1e6, 3: 1e12}`), so termolecular reactions now get the correct m^6→cm^6 (×1e12, `cm^6/(mol^2*s)`) scaling instead of being mislabeled `s^-1`. The code now matches the documented 3-body units.

arc/scripts_test.py
```diff
@@ -117,5 +119,152 @@ def test_cp_data_present(self):
...
+        self.assertEqual(args.output, '/tmp/out.yml')
+
+
+@unittest.skipUnless(RMG_ENV, 'rmg_env not available')
+class TestRmgKineticsHelpers(unittest.TestCase):
+    """
+    Unit tests for ``rmg_kinetics.py`` helpers that don't need a full RMG database load.
+
+    Each test runs a tiny ``python -c`` snippet inside ``rmg_env`` so we can import
+    rmgpy and the script module directly. Stdout is parsed as YAML.
```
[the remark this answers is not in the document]
alongd: Fixed in 342f51856 — the class docstring now reads "Stdout is parsed as JSON" to match the `json.dumps`/`json.loads` snippets.

arc/scripts_test.py
```diff
@@ -117,5 +119,152 @@ def test_cp_data_present(self):
...
+            # The guard logic itself is a one-line isinstance check; replicate it here
+            # so a regression that drops the guard would fail the test.
+            from rmgpy.kinetics import Chebyshev
+            cheb = Chebyshev(coeffs=[[1.0, 0.0], [0.0, 0.0]],
+                             kunits='cm^3/(mol*s)',
+                             Tmin=(300.0, 'K'), Tmax=(2000.0, 'K'),
+                             Pmin=(0.01, 'bar'), Pmax=(100.0, 'bar'))
+            assert not isinstance(cheb, (rk.Arrhenius, rk.ArrheniusEP))
+            arr = Arrhenius(A=(1.0, 's^-1'), n=0.0, Ea=(0.0, 'J/mol'))
+            assert isinstance(arr, (rk.Arrhenius, rk.ArrheniusEP))
```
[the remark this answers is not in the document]
alongd: Agreed, the old test was self-referential. Fixed in 342f51856: the guard is now factored into a real, called helper `scale_kinetics_by_degeneracy()` (invoked by `determine_rmg_kinetics`), and the test exercises that function directly — Arrhenius `A` must double while Chebyshev coeffs must stay put. Since `change_rate` provably shifts Chebyshev coeffs (-5.0 → -4.699), dropping the guard now corrupts them and fails the test.

arc/scripts_test.py
```diff
@@ -117,5 +119,152 @@ def test_cp_data_present(self):
...
+        save_yaml_file(path=input_path, content=original)
+
+        script = os.path.join(ARC_PATH, 'arc', 'scripts', 'rmg_thermo.py')
+        result = subprocess.run(
+            ['conda', 'run', '-n', 'rmg_env', 'python', script, input_path, '--output', output_path],
+            capture_output=True, text=True, timeout=300,
+        )
+        self.assertEqual(result.returncode, 0, f'thermo script failed: {result.stderr}')
+
+        # Input must be byte-identical (no overwrite).
```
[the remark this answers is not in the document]
alongd: Fixed in 342f51856 — the test now captures the raw input bytes before running the script and asserts the bytes are unchanged afterward, so the "byte-identical" claim is genuinely verified (rather than a parsed-YAML comparison).

arc/scripts/rmg_kinetics.py
```diff
@@ -62,6 +77,42 @@ def get_rate_coefficients(reaction_list: List[Dict]) -> List[Dict]:
...
+    cm-based units, based on the reaction molecularity.
+
+    Args:
+        num_reactants (int): The number of reactants.
+
+    Returns:
+        float: 1.0 for unimolecular (s^-1), 1e6 for bimolecular (m^3->cm^3),
+               1e12 for termolecular (m^6->cm^6). Defaults to 1.0 otherwise.
+    """
+    return {1: 1.0, 2: 1e6, 3: 1e12}.get(num_reactants, 1.0)
```
calvinp0: Minor: .get(num_reactants, 1.0) silently returns 1.0 for 0 or ≥4 reactants - prefer raising, since a silent 1.0 is indistinguishable from a legitimate unimolecular result. 
Also that the 3: 1e12 entry is currently unreachable: no termolecular family is in RMG's default set, which is what load_rmg_database() loads. Harmless to keep, but it's untested.

arc/scripts/rmg_kinetics.py
calvinp0, line 304: Loads DB but never builds the rule tree. RMG itself follows `load_kinetics` with, for ever non-auto-generated family: 
`family.add_rules_from_training(thermo_database=rmgdb.thermo)`
`family.fill_rules_by_averaging_up()`
Without, the estimator falls back to generic ancestor node

arc/scripts/rmg_kinetics.py
```diff
@@ -89,19 +148,22 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
...
+                kinetics = copy.deepcopy(kinetics_detailes[0])
+                kinetics = scale_kinetics_by_degeneracy(kinetics, deg_rxn.degeneracy)
                 if hasattr(kinetics, 'to_arrhenius'):
                     kinetics = kinetics.to_arrhenius(dh_rxn298)  # Convert ArrheniusEP to Arrhenius
-                kinetics.A.value_si = kinetics.A.value_si * (1e6 if A_units == "cm^3/(mol*s)" else 1)
+                if a_factor != 1.0 and isinstance(kinetics, Arrhenius):
+                    kinetics.A.value_si = kinetics.A.value_si * a_factor
                 deg_rxn.kinetics = kinetics
                 deg_rxn.comment = f'Family: {deg_rxn.family}'
                 if 'training' in deg_rxn.kinetics.comment:
```
calvinp0: Once `add_rules_from_training` is added, this filter deletes the answer. It's a substirng test on the kinetics comment, and that method writes "From training reaction {N} used for {template}" into the rule's own comment. Drop the filter and labelling instead, keyed off the source elemnt that `get_kinetics` already returns rather than a substring matchingg prose. Also gives `entry.rank`

arc/scripts/rmg_kinetics.py
```diff
@@ -89,19 +148,22 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
...
-                kinetics = kinetics_detailes[0]
-                kinetics.change_rate(deg_rxn.degeneracy)
+                # Deep-copy before mutating so the database object isn't double-scaled
+                # if the same family rule is queried again for another reaction.
+                kinetics = copy.deepcopy(kinetics_detailes[0])
+                kinetics = scale_kinetics_by_degeneracy(kinetics, deg_rxn.degeneracy)
                 if hasattr(kinetics, 'to_arrhenius'):
                     kinetics = kinetics.to_arrhenius(dh_rxn298)  # Convert ArrheniusEP to Arrhenius
-                kinetics.A.value_si = kinetics.A.value_si * (1e6 if A_units == "cm^3/(mol*s)" else 1)
+                if a_factor != 1.0 and isinstance(kinetics, Arrhenius):
```
calvinp0: RMG has the primitive:
```python
 A_reported = k.A.value_si * k.A.get_conversion_factor_from_si_to_cm_mol_s()
```
(`rmgpy/quantity.py:127-150`) Derives the factor from the unit's dimensionality (100 ** metres) and its what RMG uses to write Chemkin files. It defines the cm/mol/s convention ARC targets. MOlecularity of the query doesn't detemrine the units of the matched entry: it gives the wrong answer for reverse match libnrary entries for `m^2/(mol*2)` surface rates (1e4, not 1e6) and it has no entry at all for `cm^3/(molecule*s)` libraries of which arc loads three.

arc/scripts/rmg_kinetics.py
calvinp0: `reaction.is_isomoprhic(library_reaction)` defaults to `either_direction=True` (`rmgpy/reaction.py:564`) and the result is used without checking direction - so a library for the reverse reaction is reported as though it were forward.

Also note `library_reaction.kinetics` is the database object, `get_library_reactions` constructs `LibraryReaction(..., kinetics=entry.data)`. So whilst the `Reaction` wrapper is rebuilt per call so setting `.comment` is safe, if the unit conversion is ever applied to the library kinetics via mutation,  it will corrupt `rmgdb` for the rest of the run

arc/scripts/rmg_kinetics.py
```diff
@@ -62,6 +77,42 @@ def get_rate_coefficients(reaction_list: List[Dict]) -> List[Dict]:
...
+    return {1: 1.0, 2: 1e6, 3: 1e12}.get(num_reactants, 1.0)
+
+
+def scale_kinetics_by_degeneracy(kinetics: KineticsModel,
+                                 degeneracy: float,
+                                 ) -> KineticsModel:
+    """
+    Scale Arrhenius-type kinetics in place by the reaction-path degeneracy.
+
+    Non-Arrhenius forms (e.g. Chebyshev, PDepArrhenius) are returned unchanged,
```
calvinp0: Appears wrong. `Chebyshev.change_rate` does` coeffs.value_si[0,0] += log10(factor)` (RMG-Py `rmgpy/kinetics/chebyshev.pyx:257`)
`PDepArrhenius.change_rate` scales each pressure's Arrhenius (`arrhenius.pyx:943`). Neither corrupts anything

arc/scripts/rmg_kinetics.py
```diff
@@ -71,6 +122,13 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
     Determine kinetics for `reaction` (an RMG Reaction object) from RMG's database, if possible.
     Assigns a list of all matching entries from both libraries and families.
 
+    Note:
+        Family entries originating from the training set are intentionally filtered out
+        (an empty returned list therefore means "no matching libraries and only training-set
+        family hits", not necessarily "no match at all"). Database kinetics are deep-copied
```
calvinp0: RMG already returns copes. `rules.etimate_kinetics` does `kinetics = deepcopy(entry.data)`, `get_kinetics_from_depository` does `deepcopy(entry.data)` and `Reaction.copy()` deepcopes kinetics again. So at Line 161, there is the third copy of the same object. Although I suppose this is better to be safe than sorry?

arc/scripts/rmg_kinetics.py
calvinp0: This appears to be never read. It is overwritten in line 155 in the loop.

arc/scripts/rmg_kinetics.py
```diff
@@ -89,19 +148,22 @@ def determine_rmg_kinetics(rmgdb: RMGDatabase,
...
                 break
-    # # Families:
-    A_units = "cm^3/(mol*s)" if len(reaction.reactants) == 2 else "s^-1"
+    # Families:
+    a_factor = get_a_factor_si_to_cm(len(reaction.reactants))
     fam_list = loop_families(rmgdb, reaction)
     dh_rxn298 = dh_rxn298 or get_dh_rxn298(rmgdb=rmgdb, reaction=reaction)  # J/mol
     for family, degenerate_reactions in fam_list:
         for deg_rxn in degenerate_reactions:
             kinetics_list = family.get_kinetics(reaction=deg_rxn, template_labels=deg_rxn.template, degeneracy=deg_rxn.degeneracy)
```
calvinp0: Passing `degeneracy=deg_rxn.degeneracy` here is not a passive query: it routes to `rules.estimate_kinetics`, which mutates `kinetics.A.value_si *= degeneracy` and stamps the comment (`rules.py:371, 378, 487`).
Line 162 applies it a second time

The three training-depository entries - the only ones where get_kinetics genuinely ignores degeneracy, so scaling would be correct - are exactly the ones discarded at line 175.

arc/scripts_test.py
```diff
@@ -117,5 +119,153 @@ def test_cp_data_present(self):
...
+            rxn.kinetics = Arrhenius(A=(1.5e13, 'cm^3/(mol*s)'), n=0.0, Ea=(20.0, 'kJ/mol'),
+                                     Tmin=(300.0, 'K'), Tmax=(2500.0, 'K'))
+            rxn.comment = 'unit-test'
+            out = get_kinetics_from_reactions([rxn])
+            print(json.dumps(out[0]))
+        """)
+        import json
+        entry = json.loads(self._run_in_rmg_env(snippet))
+        self.assertEqual(entry['comment'], 'unit-test')
+        self.assertAlmostEqual(entry['A'], 1.5e13, delta=1e7)
```
calvinp0, line 187: This assertion passes because it bypasses `determine_rmg_kinetics`. It builds `A` in cm units and reads it straight back, so it validates nothing about the units contract in the new module docstring - the one interaction that's broken is the one it routes around.

arc/scripts_test.py
```diff
@@ -117,5 +119,153 @@ def test_cp_data_present(self):
...
+            arr = Arrhenius(A=(1.0, 's^-1'), n=0.0, Ea=(0.0, 'J/mol'))
+            scale_kinetics_by_degeneracy(arr, 2)
+            assert abs(arr.A.value_si - 2.0) < 1e-9, arr.A.value_si
+            cheb = Chebyshev(coeffs=[[1.0, 0.0], [0.0, 0.0]],
+                             kunits='cm^3/(mol*s)',
+                             Tmin=(300.0, 'K'), Tmax=(2000.0, 'K'),
+                             Pmin=(0.01, 'bar'), Pmax=(100.0, 'bar'))
+            before = cheb.coeffs.value_si.tolist()
+            scale_kinetics_by_degeneracy(cheb, 2)
+            assert cheb.coeffs.value_si.tolist() == before, 'Chebyshev coeffs were mutated'
```
calvinp0: Asserts that `Chebyshev coefficients` are left unscaled, which locks in dropped degeneracy as intended behavior. It's also unreachable in practice: the family loop only ever sees `Arrhenius`, `ArrheniusEP`, or `ArrheniusBM` rules, so no `Chebyshev` object can arrive at `scale_kinetics_by_degeneracy` via this path.

arc/scripts_test.py
```diff
@@ -117,5 +119,153 @@ def test_cp_data_present(self):
...
+        input_path = os.path.join(self.tmp_dir, 'in.yml')
+        output_path = os.path.join(self.tmp_dir, 'out.yml')
+        original = [{'label': 'H2', 'adjlist': self._h2_adjlist()}]
+        save_yaml_file(path=input_path, content=original)
+        with open(input_path, 'rb') as f:
+            input_bytes_before = f.read()
+
+        script = os.path.join(ARC_PATH, 'arc', 'scripts', 'rmg_thermo.py')
+        result = subprocess.run(
+            ['conda', 'run', '-n', 'rmg_env', 'python', script, input_path, '--output', output_path],
```
calvinp0: Are we able to use arc/job/env_run.py instead of the hardcoding?

calvinp0 (changes requested):

arc/scripts/rmg_kinetics.py
alongd: Removed. `get_a_factor_si_to_cm` is gone entirely — A is now converted with RMG's own `A.get_conversion_factor_from_si_to_cm_mol_s()`, which derives the factor from the matched entry's dimensionality. No molecularity map, no silent `1.0` default, and the unreachable `1e12`/termolecular branch goes with it. (`6316becda`)

arc/scripts/rmg_kinetics.py
alongd: Fixed. `load_rmg_database` now follows `load_kinetics` with `add_rules_from_training(thermo_database=rmgdb.thermo)` + `fill_rules_by_averaging_up()` for every non-`auto_generated` family, mirroring `rmgpy/rmg/main.py`. A CH4+OH smoke query now returns a `source: rate rules` entry, which it couldn't produce before the tree was built.

arc/scripts/rmg_kinetics.py
alongd: Fixed. The `'training'` substring filter and label are removed. Family hits are labeled from what `get_kinetics` returns — the source (`training` depository / `rate rules`) and `entry.rank` — via `format_kinetics_comment`. Training-set matches are reported, not dropped: the CH4+OH smoke run now yields two `H_Abstraction/training` entries (ranks 1 and 6).

arc/scripts/rmg_kinetics.py
alongd: Adopted. `get_kinetics_from_reactions` now reports `A = A.value_si * A.get_conversion_factor_from_si_to_cm_mol_s()`, uniformly for library and family entries, so the factor comes from each entry's own unit dimensionality rather than the query's molecularity.

arc/scripts/rmg_kinetics.py
alongd: Fixed. Library matching passes `either_direction=False`, and family results honor the `is_forward` flag from `get_kinetics` (reverse depository matches are skipped). Library kinetics is no longer mutated — the unit conversion now happens at report time from `value_si` — so `rmgdb` can't be corrupted.

arc/scripts/rmg_kinetics.py
alongd: Confirmed — `Chebyshev.change_rate` (`chebyshev.pyx:261`, `coeffs.value_si[0,0] += log10(factor)`) and `PDepArrhenius.change_rate` (`arrhenius.pyx:908`) are both correct scalings. The guard and `scale_kinetics_by_degeneracy` are removed; see the L157 thread — the manual scaling itself was the defect, not the kinetics form.

arc/scripts/rmg_kinetics.py
alongd: Removed. RMG already deep-copies (`rules.py:368/390/402`, `family.py:2640`, `Reaction.copy` at `reaction.py:1722`), and we no longer mutate the kinetics in place, so the extra `copy.deepcopy` was pure redundancy.

arc/scripts/rmg_kinetics.py
alongd: Removed. The `family` parameter was shadowed by `for family, ... in fam_list` and never read; dropped it from the signature and the caller no longer passes it.

arc/scripts/rmg_kinetics.py
alongd: Confirmed — this was the core bug. `get_kinetics(degeneracy=...)` already scales A on the rate-rule path (`rules.py:371/378/487`, comment "Multiplied by reaction path degeneracy"), so the second `change_rate(degeneracy)` double-counted it. The manual scaling is removed; degeneracy is applied once, by RMG. A CH4+OH smoke run confirms the rate-rule entry reports A=1e6 (degeneracy 4 applied once), not 4e6. Depository/training hits — which `get_kinetics` returns unscaled, exactly as you noted — are no longer discarded.

arc/scripts_test.py
alongd: Addressed. Added `test_get_kinetics_from_reactions_converts_si_to_cm`: it builds an Arrhenius in SI (`m^3/(mol*s)`) and asserts the reported A is ×1e6 (`cm^3/(mol*s)`), exercising the units contract in `get_kinetics_from_reactions` directly instead of reading a cm value straight back.

arc/scripts_test.py
alongd: Removed. With the guard gone the assertion is moot, and as you noted the family loop only yields Arrhenius/ArrheniusEP/ArrheniusBM, so a Chebyshev never reaches that path.

arc/scripts_test.py
alongd: Done. The tests now route through `rmg_env_command` from `arc/job/env_run.py` via a `_run_rmg_env` helper, replacing the hardcoded `conda run -n rmg_env` everywhere (availability probe, the arkane/thermo script runs, and the kinetics `-c` snippets).

alongd:
Rebased onto `main` and addressed the full review round (force-pushed `3d4d6bcd9` → `6316becda`). Every point verified against RMG-Py source before changing anything.

**Correctness (this changes the reported k(T), which was the point):**
- Removed the manual degeneracy scaling — `get_kinetics(degeneracy=...)` already applies it on the rate-rule path (`rules.py:371/378/487`), so the extra `change_rate` was double-counting. Degeneracy is now applied exactly once, by RMG.
- Build the rate-rule tree after load (`add_rules_from_training` + `fill_rules_by_averaging_up` per non-`auto_generated` family), so family estimates no longer fall back to a generic ancestor node.
- Dropped the `'training'` substring filter; family hits are labeled by source + `entry.rank` from what `get_kinetics` returns, so training/depository matches are reported instead of discarded.
- Report A via `A.get_conversion_factor_from_si_to_cm_mol_s()` (the factor RMG uses for Chemkin), uniformly for library and family entries, replacing the molecularity heuristic.
- Library matching is forward-only (`either_direction=False`), and family results honor `is_forward`.
- Removed the dead `family` param, the redundant deepcopy, and the guard whose premise (non-Arrhenius `change_rate` corrupts) was false.

**Tests:** added a real SI→cm units test, removed the unreachable Chebyshev test, and routed every script test through the production `rmg_env_command` helper instead of a hardcoded `conda run`.

Validated with a full-DB CH4 + OH smoke run: the rate-rule entry reports A=1e6 (degeneracy 4 applied once, not 4e6), two `H_Abstraction/training` entries now appear, and units are cm/mol/s.

@calvinp0 re-requesting your review.

calvinp0 (approved):

PR #897, opened 2026-06-10
Make adaptive_levels YAML-usable and reaction-consistent

alongd:
Two fixes to `adaptive_levels` (adaptive level-of-theory), from the roadmap.

  ### 1. YAML-usable schema
  `adaptive_levels` required Python tuple keys (`(1,6)`, `(7,'inf')`) that
  `yaml.safe_load` can't produce, so it was unusable from an input file. It now
  takes a list of entries, and round-trips through restart:

  ```yaml
  adaptive_levels:
    - atom_range: [1, 6]
      levels: {opt freq: wb97xd/def2tzvp, sp: ccsd(t)-f12/cc-pvtz-f12}
    - atom_range: [7, inf]
      levels: {opt freq: b3lyp/6-31g(d,p), sp: dlpno-ccsd(t)/def2-tzvp}

  The legacy tuple-dict form is removed. Also fixes the restart serializer
  (.items() bug).
```

 ### 2. Reaction-consistent levels

  Levels were picked per species from its own heavy-atom count, so a reaction's
  large TS and its small wells could land on different grains, mixing levels of
  theory across the barrier. Each reaction is now keyed by its (conserved)
  heavy-atom count, and all of its species are evaluated at that one level.

  New per-species thermo_at_own_level flag (default False): set True to
  compute a species' thermo at its own granular level — ARC then makes an
  autonomous relabeled copy for the reaction so the barrier stays consistent.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -97,20 +97,22 @@ class ARC(object):
         ess_settings (dict, optional): A dictionary of available ESS (keys) and a corresponding server list (values).
         bath_gas (str, optional): A bath gas. Currently used in OneDMin to calc L-J parameters.
                                   Allowed values are He, Ne, Ar, Kr, H2, N2, O2.
-        adaptive_levels (dict, optional): A dictionary of levels of theory for ranges of the number of heavy atoms in
-            the molecule. Keys are tuples of (min_num_atoms, max_num_atoms), values are dictionaries. Keys of the
-            sub-dictionaries are tuples of job types, values are levels of theory (str, dict or Level).
+        adaptive_levels (list, optional): A list of levels of theory for ranges of the number of heavy atoms in the
+            molecule. Each entry is a dictionary with an ``atom_range`` 2-list (min_num_atoms, max_num_atoms; the upper
+            bound may be ``'inf'``) and a ``levels`` mapping from job type to level of theory (str, dict or Level).
+            Job types sharing a level may be given as a whitespace- or comma-separated key (e.g. ``'opt freq'``).
             Job types not defined in adaptive levels will have non-adaptive (regular) levels.
             Example::
 
-                adaptive_levels = {(1, 5):      {('opt', 'freq'): 'wb97xd/6-311+g(2d,2p)',
-                                                 'sp': 'ccsd(t)-f12/aug-cc-pvtz-f12'},
-                                   (6, 15):     {('opt', 'freq'): 'b3lyp/cbsb7',
-                                                 'sp': 'dlpno-ccsd(t)/def2-tzvp'},
-                                   (16, 30):    {('opt', 'freq'): 'b3lyp/6-31g(d,p)',
-                                                 'sp': 'wb97xd/6-311+g(2d,2p)'},
-                                   (31, 'inf'): {('opt', 'freq'): 'b3lyp/6-31g(d,p)',
-                                                 'sp': 'b3lyp/6-311+g(d,p)'}}
+                adaptive_levels = [{'atom_range': [1, 5],
+                                    'levels': {'opt freq': 'wb97xd/6-311+g(2d,2p)',
+                                               'sp': 'ccsd(t)-f12/aug-cc-pvtz-f12'}},
+                                   {'atom_range': [6, 15],
+                                    'levels': {'opt freq': 'b3lyp/cbsb7',
+                                               'sp': 'dlpno-ccsd(t)/def2-tzvp'}},
+                                   {'atom_range': [16, 'inf'],
+                                    'levels': {'opt freq': 'b3lyp/6-31g(d,p)',
+                                               'sp': 'wb97xd/6-311+g(2d,2p)'}}]
 
         freq_scale_factor (float, optional): The harmonic frequencies scaling factor. Could be automatically determined
                                              if not available in Arkane and not provided by the user.
@@ -162,9 +164,8 @@ class ARC(object):
         ts_guess_level (Level): Level of theory for comparisons of TS guesses between different methods.
         irc_level (Level): The level of theory to use for IRC calculations.
         orbitals_level (Level): Level of theory for molecular orbitals calculations.
-        adaptive_levels (dict): A dictionary of levels of theory for ranges of the number of heavy atoms in
-            the molecule. Keys are tuples of (min_num_atoms, max_num_atoms), values are dictionaries. Keys of the
-            sub-dictionaries are tuples of job types, values are levels of theory (str, dict or Level).
+        adaptive_levels (dict): The processed adaptive levels, keyed by (min_num_atoms, max_num_atoms) tuples, each
+            mapping job-type tuples to ``Level`` objects (built from the user-facing ``adaptive_levels`` list).
             Job types not defined in adaptive levels will have non-adaptive (regular) levels.
         output (dict): Output dictionary with status and final QM file paths for all species. Only used for restarting,
                        the actual object used is in the Scheduler class.
@@ -430,8 +431,10 @@ class ARC(object):
         """
         restart_dict = dict()
         if self.adaptive_levels is not None:
-            restart_dict['adaptive_levels'] = {atom_range: {job_type: level.as_dict() for job_type, level in levels_dict}
-                                               for atom_range, levels_dict in self.adaptive_levels.items()}
+            restart_dict['adaptive_levels'] = [
+                {'atom_range': [atom_range[0], atom_range[1]],
+                 'levels': {' '.join(job_types): level.as_dict() for job_types, level in levels_dict.items()}}
+                for atom_range, levels_dict in self.adaptive_levels.items()]
         if self.allow_nonisomorphic_2d:
             restart_dict['allow_nonisomorphic_2d'] = self.allow_nonisomorphic_2d
         if self.arkane_level_of_theory is not None:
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: docs/source/advanced.rst [hunk 1]
```diff
@@ -251,25 +251,44 @@ The method returns a dictionary containing the ``'e_o'`` tuple (electrons, orbit
...
-                       (16, 30):    {('opt', 'freq'): 'b3lyp/6-31g(d,p)',
-                                     'sp': 'wb97xd/6-311+g(2d,2p)'},
-                       (31, 'inf'): {('opt', 'freq'): 'b3lyp/6-31g(d,p)',
-                                     'sp': 'b3lyp/6-311+g(d,p)'}}
-
-Note that job types which are not specified in ``adaptive_levels`` will use no-adaptive
-(defined separately e.g., via ``opt_level``, or using ARC's defaults.
+To do so, pass the ``adaptive_levels`` attribute, which is a list of entries.
+Each entry is a dictionary with an ``atom_range`` 2-list giving the heavy
+(non-hydrogen) atom count range (the upper bound may be ``inf``), and a ``levels``
```
Lilachn91: For clarity, I would rephrase to:

Each entry is a dictionary with two keys: ``atom_range``, a two-element list `[min, max]` giving a range for the species' size by counting the number of heavy (non‑hydrogen) atoms (the upper bound of the last range must be ``inf``); and ``levels``: mapping from job type(s) to the level of theory (a string or dictionary) to use for that size range.

docs/source/advanced.rst
```diff
@@ -251,25 +251,44 @@ The method returns a dictionary containing the ``'e_o'`` tuple (electrons, orbit
...
+
+When a species participates in a reaction, all of the reaction's species (the TS and the
+reactant/product wells) are energy-evaluated at a single, reaction-consistent level keyed
+by the largest participant (the TS), so the barrier is computed at one consistent level.
+By default (the per-species ``thermo_at_own_level`` flag, ``False``) a well whose own size
+falls on a finer grain than its reaction's is evaluated directly at the reaction-wide
+(coarser) level, and its thermochemistry uses that same level. Set ``thermo_at_own_level=True`` on a species
+to instead compute its thermochemistry at its own size-appropriate (granular) adaptive level:
+an autonomous, relabeled copy of the species is then created and used by the reaction (at the
+reaction-wide level), while the original keeps its own level for thermochemistry.
```
Lilachn91, line 292: Also for clarity, I'd rephrase, especially the last part:

By default (the per-species ``thermo_at_own_level`` flag, ``False``), a well whose own size falls on a finer grain than its reaction, is evaluated directly at the reaction-wide (coarser) level, and its thermochemistry uses that same level.
To compute its thermochemistry at its own size-appropriate (granular) adaptive level, set ``thermo_at_own_level=True`` on a species. ARC will then create an autonomous, relabeled copy of the species to be computed at the reaction-wide level and used by the reaction, while the original species will keep its own level for thermochemistry.

arc/main.py
```diff
@@ -1256,44 +1259,60 @@ def standardize_output_paths(self):
...
-    atom_ranges = sorted(list(adaptive_levels.keys()), key=lambda x: x[0])
+        for job_types, level in levels.items():
+            job_type_tuple = tuple(job_types.replace(',', ' ').split())
+            processed[atom_range][job_type_tuple] = Level(repr=level)
+    atom_ranges = sorted(processed.keys(), key=lambda x: x[0])
     for i, atom_range in enumerate(atom_ranges):
         if i and atom_ranges[i-1][1] + 1 != atom_ranges[i][0]:
-            raise InputError(f'Atom ranges of adaptive levels must be consecutive. '
-                             f'Got:\n{list(adaptive_levels.keys())}')
+            raise InputError(f'Atom ranges of adaptive levels must be consecutive. Got:\n{atom_ranges}')
```
Lilachn91, line 1314: What happens if the first range doesn't start at 0 or 1, or at the lowest heavy atoms count? Is it allowed (as opposed to not finishing with inf)?

arc/scheduler.py
```diff
@@ -3957,6 +3962,78 @@ def determine_adaptive_level(self,
...
+                        if spc.adaptive_lot_n_heavy is None \
+                                or self._adaptive_atom_range(spc.adaptive_lot_n_heavy) == reaction_range:
+                            spc.adaptive_lot_n_heavy = reaction_n_heavy
+                            continue
+                        logger.warning(f'Species {label} participates in reactions on different adaptive level '
+                                       f'grains, creating a dedicated copy of it for reaction {rxn.label} to keep '
+                                       f'the reaction internally consistent.')
+                    copy_label = check_label(f'{label}_TS{rxn_index}')[0]
+                    if copy_label not in self.species_dict:
+                        copy_spc = spc.copy()
```
Lilachn91, line 4032: The copied reaction participant should probably have `thermo_at_own_level` reset to False. Since `spc.copy()` preserves the original setting, a `thermo_at_own_level=True` source creates a copy that still opts into copy behavior. On restart, `_apply_adaptive_reaction_levels()` can see the copy as needing its own copy and relabel, duplicating species across restarts.

arc/scheduler.py
```diff
@@ -3957,6 +3962,78 @@ def determine_adaptive_level(self,
...
+                        continue
+                    if not spc.thermo_at_own_level:
+                        if spc.adaptive_lot_n_heavy is None \
+                                or self._adaptive_atom_range(spc.adaptive_lot_n_heavy) == reaction_range:
+                            spc.adaptive_lot_n_heavy = reaction_n_heavy
+                            continue
+                        logger.warning(f'Species {label} participates in reactions on different adaptive level '
+                                       f'grains, creating a dedicated copy of it for reaction {rxn.label} to keep '
+                                       f'the reaction internally consistent.')
+                    copy_label = check_label(f'{label}_TS{rxn_index}')[0]
```
Lilachn91, line 4024: This should guard against label collisions. If `{label}_TS{rxn_index}` already exists but is not the copy created for this reaction, the reaction will silently point to an unrelated species. Is there a way to validate that an existing label is the intended adaptive-level copy?

Lilachn91 (commented):
Thanks, I left a few comments (some are just minor wording comments).

arc/main.py
alongd: great point, I'll fix. thanks!

arc/scheduler.py
```diff
@@ -3936,6 +3941,70 @@
...
+            reactant_species = [self.species_dict[label] for label in rxn.reactants if label in self.species_dict]
+            if len(reactant_species) != len(rxn.reactants) \
+                    or any(spc.number_of_heavy_atoms is None for spc in reactant_species):
+                logger.warning(f'Could not determine reaction-wide adaptive levels for {rxn.label}, '
+                               f'using per-species levels.')
+                continue
+            reaction_n_heavy = sum(spc.number_of_heavy_atoms for spc in reactant_species)
+            reaction_range = self._adaptive_atom_range(reaction_n_heavy)
+            for participants in (rxn.reactants, rxn.products):
+                for i, label in enumerate(participants):
```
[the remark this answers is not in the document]
alongd: Resolved — the outer and inner loops now use distinct variables (`rxn_i` and `pos`); no shared loop variable remains.

arc/scheduler.py
```diff
@@ -3936,6 +3941,70 @@ def determine_adaptive_level(self,
...
+        Make every reaction's energetics internally consistent under adaptive levels of theory.
+
+        Heavy atoms are conserved across a reaction, so the reaction-wide heavy-atom count (the sum over the reactants,
+        equal to the TS supermolecule count) keys a single adaptive level for the whole reaction. A reaction
+        participant whose own heavy-atom count lands it on a *different* adaptive grain than the reaction would
+        otherwise be evaluated at an inconsistent level, mixing levels of theory across the barrier. To avoid this,
+        for each such participant whose ``thermo_at_own_level`` is ``True`` (the default), an autonomous relabeled
+        copy of the species is created from the outset and used by the reaction (evaluated at the reaction-wide
+        level), while the original species is left to compute its own thermochemistry at its own granular level. If
+        ``thermo_at_own_level`` is ``False``, the participant itself is evaluated at the reaction-wide level (no copy).
```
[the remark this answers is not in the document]
alongd: Fixed — the docstring now states `thermo_at_own_level` defaults to `False`, and the copy/no-copy branches are described in the correct direction.

arc/scheduler.py
```diff
@@ -3936,6 +3941,70 @@ def determine_adaptive_level(self,
...
+            reaction_range = self._adaptive_atom_range(reaction_n_heavy)
+            for participants in (rxn.reactants, rxn.products):
+                for i, label in enumerate(participants):
+                    spc = self.species_dict.get(label)
+                    if spc is None or spc.number_of_heavy_atoms is None \
+                            or self._adaptive_atom_range(spc.number_of_heavy_atoms) == reaction_range:
+                        continue
+                    if not spc.thermo_at_own_level:
+                        spc.adaptive_lot_n_heavy = reaction_n_heavy
+                        continue
```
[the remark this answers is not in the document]
alongd: The current logic detects this rather than silently overwriting: a no-copy species already pinned to a different grain by an earlier reaction falls through to a dedicated per-reaction copy (`{label}_TS{rxn_index}`), keeping each reaction internally consistent. Covered by `test_shared_species_across_grains_gets_copy`.

arc/main.py
```diff
@@ -1256,44 +1259,58 @@ def standardize_output_paths(self):
...
+    for entry in adaptive_levels:
+        if not isinstance(entry, dict) or 'atom_range' not in entry or 'levels' not in entry:
+            raise InputError(f'Each adaptive levels entry must be a dictionary with "atom_range" and "levels" '
+                             f'keys, got {entry} which is a {type(entry)} in:\n{adaptive_levels}')
+        atom_range = entry['atom_range']
+        if not isinstance(atom_range, (list, tuple)) or len(atom_range) != 2 \
+                or not all(isinstance(a, int) or a in ('inf', float('inf')) for a in atom_range):
+            raise InputError(f'The "atom_range" of each adaptive levels entry must be a 2-length list of integers '
+                             f'with an optional "inf" upper bound, got {atom_range} in:\n{adaptive_levels}')
+        atom_range = (atom_range[0], 'inf' if atom_range[1] in ('inf', float('inf')) else atom_range[1])
```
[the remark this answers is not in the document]
alongd: Addressed — `atom_range` now requires an `int` lower bound; only the upper bound may be `int`/`inf`. `[inf, inf]` and `[inf, 10]` are rejected (see `test_process_adaptive_levels`).

docs/source/advanced.rst
```diff
@@ -251,25 +251,44 @@ The method returns a dictionary containing the ``'e_o'`` tuple (electrons, orbit
...
+
+Note that job types which are not specified in ``adaptive_levels`` will use non-adaptive
+levels (defined separately, e.g., via ``opt_level``, or using ARC's defaults).
+
+When a species participates in a reaction, all of the reaction's species (the TS and the
+reactant/product wells) are energy-evaluated at a single, reaction-consistent level keyed
+by the largest participant (the TS), so the barrier is computed at one consistent level.
+By default (the per-species ``thermo_at_own_level`` flag, ``False``) a well that lands on a
+coarser grain than its reaction is evaluated directly at the reaction-wide level, and its
+thermochemistry uses that same (coarser) level. Set ``thermo_at_own_level=True`` on a species
```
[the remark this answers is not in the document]
alongd: Fixed — the docs now say a small well lands on a finer grain and is evaluated at the reaction-wide (coarser) level.

docs/source/advanced.rst
alongd: Rephrased along these lines; also tightened to "the upper bound of the last range must be `inf`".

docs/source/advanced.rst
alongd: Rephrased along these lines.

arc/main.py
alongd: Good question — it wasn't enforced. `process_adaptive_levels` now requires the first range to start at 1, so there's no bottom gap that would only surface later as a runtime `SchedulerError`.

arc/scheduler.py
alongd: Fixed — the copy is now created with `thermo_at_own_level=False`, and an idempotency test confirms a re-apply (as on restart) doesn't duplicate species.

arc/scheduler.py
alongd: Added a guard — an existing copy label is validated as the intended adaptive copy (`adaptive_lot_n_heavy == reaction_n_heavy`); an unrelated species already occupying that label now raises `SchedulerError` instead of being silently used.

Lilachn91 (commented):
LGTM, Thanks!

Lilachn91 (approved):

PR #900, opened 2026-06-16
Fix: detect SCRF polarization charge error and troubleshoot with nosymm

LeenFahoum:
Gaussian jobs using SMD/SCRF solvation can fail at l9999 with "Error on total polarization charges", which is a cavity construction issue unrelated to geometry convergence. ARC was classifying this as a generic Unconverged error and never attempting nosymm, which disables symmetry in the cavity and typically resolves the issue. Added detection of this error string in the l9999 block of determine_ess_status so that NoSymm is included in the parsed keywords, triggering the existing trsh_keyword_nosymm troubleshooting path.

### DIFF
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -94,6 +94,7 @@ def determine_ess_status(output_path: str,
                     if 'l9999.exe' in line or 'link 9999' in line:
                         cycle_issue = False
                         neg_eigenvalues = False
+                        polarization_error = False
                         for j in range(i, len_reversed_lines):
                             if 'Number of steps exceeded' in reverse_lines[j]:
                                 keywords = ['MaxOptCycles', 'GL9999']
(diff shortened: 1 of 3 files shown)


alongd:
@LeenFahoum, looks good! please rebase

alongd:
@LeenFahoum, please rebase and Slack me to merge

LeenFahoum:
Closing this PR, since after checking with Gaussian support we found that `"Error on total polarization charges"` is not a reliable indication of the convergence issue. It appears in SCRF calculations regardless of whether the job eventually converges, so using it as a detection signal in ARC is not the right approach.
The issue seems to be related to Gaussian reorienting the molecule during SCRF optimizations with tight thresholds, which can introduce numerical noise for C1 cases. Adding `NoSymm` avoids this by keeping the molecule in the input orientation.

I will implement the fix in the `qa_ts` branch instead, where ARC can add `NoSymm` more generally when a Gaussian SCRF calculation does not converge.

PR #906, opened 2026-06-30
Optimize Atom Mapping procedure

kfir4444:
# Atom Mapping Performance Improvements — `optimize_am` branch

## Overview

Reduce runtime of the atom mapping procedure, utilzing algorithmic and programmatic solutions.

---

### Problem

zmat_nerf implements the SN-NeRF atom placement algorithm: given three already-placed reference atoms A, B, C, it places a new atom D at bond length r from C, bond angle θ(B,C,D), and dihedral φ(A,B,C,D). This is the inner loop of _add_nth_atom_to_coords in zmat.py. However, this doesn't work if co-linearity exist.

`determine_d_atoms()` could select a reference atom triplet (B, C, D) that is nearly
collinear. `_add_nth_atom_to_coords()` later computes `cross(B−A, normalize(C−B))` for
NeRF placement; a near-zero cross product produces NaN coordinates and causes mapping
failures on linear-geometry intermediates.

### Fix

Added `_nerf_cross_sq()` — a double-precision function computing
`|cross(B−A, normalize(C−B))|²`. A threshold `_NERF_CROSS_SQ_MIN = 0.01`
(≈3.8° from linear at 1.5 Å bond) gates triplet acceptance:

```python
if _nerf_cross_sq(coords, ...) >= _NERF_CROSS_SQ_MIN:
    d_atoms.append(i)
    break
```

The same guard is applied to the `[n, 2, 1, 0]` fallback path, with a scan
for a valid 4th reference atom when the fallback itself is collinear.

---

## Change 2 — `arc/species/_zmat_kernels.py` + `arc/species/vectors.py`: C kernel wiring

One key bottleneck of atom mapping comes from vector computations:
The cost decomposition for C₈H₁₈ species to species mapping (cProfile, 0.770 s total, 1,284,756 calls)

| Component                              | Calls  | Cumul. time | Fraction |
|----------------------------------------|--------|-------------|----------|
| `fix_dihedrals_by_backbone_mapping`    | 2      | 0.764 s     | **99.2%** |
| `set_dihedral`                         | 20     | 0.651 s     | 84.5%    |
| `xyz_to_zmat`                          | 60     | 0.514 s     | 66.8%    |
| `_add_nth_atom_to_zmat`                | 1,560  | 0.481 s     | 62.5%    |
| `calculate_param`                      | 10,357 | 0.377 s     | 48.9%    |
| `np.cross`                             | 5,644  | 0.191 s     | 24.8%    |
| `get_angle`                            | 7,597  | 0.190 s     | 24.7%    |
| `get_dihedral`                         | 1,438  | 0.153 s     | 19.9%    |
| `zmat_to_coords`                       | 60     | 0.128 s     | 16.6%    |
| `numpy.moveaxis`                       | 16,932 | 0.110 s     | 14.3%    |
| `order_atoms` (inside copy)            | 4      | 0.091 s     | 11.8%    |
| `numpy.asarray`                        | 29,025 | 0.059 s     | 7.7%     |

Key optimizations of the vector coputations imperative for fast atom mapping. This is achieved by compiling kernels in C for these types of computations, and calling them via the `_zmat_kernels.py` (new file)

ctypes loader for the pre-compiled `_zmat_c_kernels.so`. Exposes:
- `zmat_r_f32` — interatomic distance
- `zmat_a_f32` — bond angle
- `zmat_d_f32` — dihedral angle
- `zmat_nerf` — NeRF atom placement (SN-NeRF formula)

Sets `available = True` only when the shared library loads without error; all callers
gate on this flag so the code degrades gracefully when the `.so` is absent, to avoid compiling errors.

### `vectors.py`

Fast paths in `calculate_distance` and `calculate_dihedral_angle`:

```python
if _ck_available:
    return _ck.zmat_r_f32(c0[0], c0[1], c0[2], c1[0], c1[1], c1[2])
```

`calculate_angle` was left as numpy — the C kernel uses float32 which caused a
precision failure in `test_get_zmat_param_value` at the ~0.001° level. This can be debated offline.

Also added `apply_rodrigues_rotation()` used by Change 3. (source: [rotation](https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula))

---

### Problem

`fix_dihedrals_by_backbone_mapping()` corrected each torsion angle by calling
`set_dihedral()`, which performs a full `xyz → zmat → modify → zmat → xyz` round-trip.
This is O(N²) per torsion and O(N³) overall for large molecules.

### Fix

Direct Cartesian rotation via Rodrigues' formula:

1. Identify downstream atoms using BFS from the far-side pivot (`_downstream_atoms_adj`).
2. Rotate only those atoms: `R(u,δ) = I cosδ + (1−cosδ)u⊗u + sinδ[u]×`
3. No zmat round-trip — O(N) per torsion, O(N²) overall.

```python
apply_rodrigues_rotation(coords, axis_origin=coords[pivot_i],
                         axis_unit=axis_unit, angle_rad=delta,
                         indices=downstream)
```

Additional changes in this file:
- `identify_superimposable_candidates`: capped at `_MAX_BACKBONE_CANDIDATES = 4`
  (was unbounded) and deduplicates inline, removing the `prune_identical_dicts` post-pass. No apparent performence loss.

---

### Problem

For species with pre-ordered mol_list, `order_atoms_in_mol_list` ran VF2
for every resonance structure even when atom IDs already matched positionally.

### Fix

```python
if (len(mol.atoms) == len(ref_mol.atoms)
        and all(m.id == r.id and m.id != -1
                for m, r in zip(mol.atoms, ref_mol.atoms))):
    continue
```

Also changed `order_mol_by_atom_map` to use `reordered.update(sort_atoms=False)`,
preventing implicit reordering after explicit atom positioning.

---

### 5a. Adjacency-list CPI cache in `determine_possible_reaction_products_from_family`

`check_product_isomorphism` (CPI) was called once per `product_list` from
`generate_products`. For large families (e.g. R_Addition on C10H8+H generates
2400 product_lists), most entries are graph-duplicates — same adjacency structure,
different atom-label mappings. The isomorphism result depends only on the graph,
so it is cached by adjacency-list key:

```python
adj_key = tuple(m.to_adjacency_list() for m in template_mols)
if adj_key in iso_cache:
    if not iso_cache[adj_key]: continue
else:
    result = check_product_isomorphism(products=template_mols, p_species=p_species)
    iso_cache[adj_key] = result
    if not result: continue
```

For reaction C10H9 ↔ C10H8+H: 2400 CPI calls → 150 unique per family (16× reduction).

### 5b. Shared `fwd_iso_cache`/`rev_iso_cache` across family calls

When `rmg_family_set='all'` is requested, 112 families are iterated, many duplicated
(same label appears as both RMG and ARC copy). With a shared cache, the second occurrence
of any family hits entirely from cache (0 CPI evaluations).

For C16H11 isomerization: CPI calls 794 → 406, time 19.55s → 9.47s.

`flip_reaction()` is now called once before the loop instead of once per family.
The pre-build is wrapped in `try/except` to preserve the original graceful-skip
behaviour when `to_smiles()` fails (e.g. openbabel mocked in tests).

### 5c. mol_list search limit for large molecules

For molecules with more than 2 resonance structures AND more than 20 atoms,
`generate_unimolecular_products` limits the search to the first 2 Kekulé forms:

```python
if len(mol_list) > 2 and len(mol_list[0].atoms) > 20:
    mol_list = mol_list[:2]
```

Additional resonance structures produce near-duplicate product graphs already
deduplicated by the adjacency-list cache at the CPI level. The VF2 cost of
generating them (3–4 s for a 27-atom polycyclic per family) is eliminated.

For C16H11, 27 atoms, 7 resonance structures: `generate_products`
for `Intra_R_Add_Exocyclic` drops from ~3s to ~0.9s per call.

---

## Change 6 — `arc/species/species.py`: `skip_conformers` on `scissors()`/`_scissors()`

Added `skip_conformers: bool = False` parameter propagated to the `generate_conformers`
calls inside `_scissors()`. Allows mapping-layer callers to get the cut species
structure without triggering conformer generation.

---


alongd:
@kfir4444, nice CI speedup!

arc/family/family.py
```diff
@@ -263,7 +263,14 @@ def generate_unimolecular_products(self,
         """
         isomorphic_subgraph_dicts = list()
         reactant_to_group_maps = reactant_to_group_maps[0]
-        for mol in reactants[0].mol_list or [reactants[0].mol]:
+        mol_list = reactants[0].mol_list or [reactants[0].mol]
+        # For large molecules (>20 atoms) with many resonance structures, limit the search
+        # to avoid combinatorial explosion. Most reaction sites are captured by the first
+        # few Kekulé forms; additional resonance structures produce equivalent sites already
```
alongd, line 269: resonance doesn't always mean Kekulé / Clar type. there are many others, including complex `N` and `S` chemistries, and conjugated allyls.
I'm not sure we can use this bypass. I'm OK to use it for most of the tests (keep some that follow the original branch).

Makefile
```diff
@@ -123,3 +123,7 @@ compile:
 	@ python utilities.py check-python
 	python setup.py build_ext main --inplace --build-temp .
 	@ python utilities.py check-dependencies
+	@echo "Compiling arc/species/_zmat_c_kernels.so…"
+	gcc -O3 -march=native -ffast-math -shared -fPIC \
+	    -o arc/species/_zmat_c_kernels.so \
+	    arc/species/_zmat_c_kernels.c -lm
```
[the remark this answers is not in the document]
alongd: I agree with the copilot comment. On a different x86-64 CPU with an older microarchitecture you could get SIGILL that kills Python. let's rethink it

arc/species/_zmat_c_kernels.c
```diff
@@ -0,0 +1,218 @@
+/*
+ * _zmat_c_kernels.c — fast geometry kernels for arc.species.zmat
+ *
+ * All functions take individual x,y,z doubles per atom.
+ * This lets Python callers use tuple-unpacking (*coords[i]) instead of
+ * building numpy arrays, eliminating the dominant overhead in the hot path.
+ *
+ * Compiled with:
+ *   gcc -O3 -march=native -ffast-math -shared -fPIC -o _zmat_c_kernels.so \
```
alongd, line 9: `-ffast-math` - if we compile with finite-math-only (NaN/Inf never occur), it will return `0` instead of true `NaN` results. If we try to guard against co-linearity in dihedrals, then this isn't assisting.
add `-fno-finite-math-only` or drop `-ffast-math`?

arc/mapping/engine.py
```diff
@@ -270,6 +271,9 @@ def fingerprint(spc: ARCSpecies,
     return fingerprint_dict
 
 
+_MAX_BACKBONE_CANDIDATES = 4
```
alongd, line 274: In symmetric molecules the best superimposable backbone map can be beyond the first 4.
Why is 4 safe? How do we know accuracy is preserved?

arc/mapping/engine.py
```diff
@@ -442,6 +447,130 @@ def remove_gaps_from_values(data: dict[int, int]) -> dict[int, int]:
     return new_data
 
 
+def _build_adj(mol: Molecule) -> tuple:
+    """Build adjacency as tuple-of-tuples[int] from mol, 0-indexed. O(N+E)."""
+    atom_to_idx = {atom: i for i, atom in enumerate(mol.atoms)}
+    return tuple(tuple(atom_to_idx[nb] for nb in atom.edges) for atom in mol.atoms)
+
+
+def _downstream_atoms_adj(adj: tuple, pivot_i: int, pivot_j: int) -> list[int]:
```
alongd, line 456: this cuts the pivot_i–pivot_j bond and BFSs the far side. But if that bond is in a ring, the graph stays connected. Needs fixing

arc/mapping/engine.py
```diff
@@ -442,6 +447,130 @@ def remove_gaps_from_values(data: dict[int, int]) -> dict[int, int]:
...
+    if bond_len < 1e-8:
+        return False
+    inv = 1.0 / bond_len
+    axis_unit = (dx * inv, dy * inv, dz * inv)
+    downstream = _downstream_atoms_adj(adj, pivot_i, pivot_j)
+    apply_rodrigues_rotation(coords, coords[pivot_i], axis_unit, delta, downstream)
+    return True
+
+
+def _set_dihedral_rodrigues(spc: ARCSpecies, torsion: list[int], deg_abs: float) -> bool:
```
alongd, line 506: is this func used anywhere?

arc/mapping/engine.py
```diff
@@ -442,6 +447,130 @@ def remove_gaps_from_values(data: dict[int, int]) -> dict[int, int]:
...
+
+def _downstream_atoms(mol: Molecule, pivot_i: int, pivot_j: int) -> list[int]:
+    """
+    BFS from pivot_j without crossing to pivot_i.
+    Returns all atom indices reachable from pivot_j when the pivot_i–pivot_j bond is cut.
+    """
+    return _downstream_atoms_adj(_build_adj(mol), pivot_i, pivot_j)
+
+
+def _rodrigues_on_coords(coords: list, adj: tuple, torsion: list[int], deg_abs: float) -> bool:
```
alongd, line 484: is this func used anywhere?

arc/mapping/engine.py
```diff
@@ -442,6 +447,130 @@ def remove_gaps_from_values(data: dict[int, int]) -> dict[int, int]:
...
+            continue
+        visited.add(atom)
+        downstream.append(atom)
+        for nb in adj[atom]:
+            if nb not in visited:
+                queue.append(nb)
+    return downstream
+
+
+def _downstream_atoms(mol: Molecule, pivot_i: int, pivot_j: int) -> list[int]:
```
alongd, line 476: is this func used anywhere?

arc/mapping/engine.py
```diff
@@ -284,16 +288,17 @@ def identify_superimposable_candidates(fingerprint_1: dict[int, dict[str, str |
...
         return []
     key_1 = list(fingerprint_1.keys())[0]
+    candidates: list[dict[int, int]] = []
     for key_2 in fingerprint_2.keys():
-        # Try all combinations of heavy atoms.
         result = iterative_dfs(fingerprint_1, fingerprint_2, key_1, key_2)
-        if result is not None:
+        if result is not None and result not in candidates:
             candidates.append(result)
-    return prune_identical_dicts(candidates)
```
alongd, line 296: if we don't use `prune_identical_dicts` here, is that function still used anywhere?

arc/species/_zmat_kernels.py
```diff
@@ -0,0 +1,50 @@
+"""
```
alongd, line 1: let's not ship it pre-compiled, add it to the `make compile` command?

alongd (commented):
Thanks! Added some comments

Makefile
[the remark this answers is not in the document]
kfir4444: You're both right, and the failure mode is worse than degraded performance. I disassembled the .so as it was committed on this branch: 60 FMA3 instructions and 346 VEX-encoded AVX instructions. On any pre-Haswell CPU that is an immediate SIGILL. Critically, the graceful-fallback design gave zero protection against it — ctypes.CDLL() succeeds, the illegal instruction only arrives at call time, and no Python-level except can catch a SIGILL. So the available flag was protecting against the one failure mode that couldn't happen (a missing file) and not against the one that could.
  
Both flags are dropped. The build is now $(CC) -O3 -fno-math-errno -fno-trapping-math -shared -fPIC (Makefile:133-142), which keeps the useful part — branch-free sqrt — without the ISA or NaN implications. Verified on the new build: 0 FMA3, 0 VEX-encoded instructions. Also switched gcc → $(CC) so the build honours the toolchain. Rationale is recorded as a comment in the Makefile so it doesn't get re-added.

arc/species/_zmat_c_kernels.c
kfir4444: Exactly right, and this was the more dangerous of the two flags — it would have silently broken correctness rather than crashing. -ffast-math implies -ffinite-math-only, which licenses the compiler to assume NaN never occurs, and the whole degenerate-geometry detection depends on a NaN dihedral propagating back to Python.

I dropped -ffast-math entirely rather than adding -fno-finite-math-only, since -ffast-math also pulls in -fassociative-math and friends, and none of them are worth the reasoning burden in a geometry kernel. Verified empirically after the change: zmat_d_f32 on collinear input (0,0,0) (1,0,0) (2,0,0) (3,0,0) returns nan. The header comment at  _zmat_c_kernels.c:12-14 now states this so the flags aren't "optimized" back in later.

arc/species/vectors.py
```diff
@@ -197,6 +201,10 @@ def calculate_angle(coords: list | tuple | dict,
     if not all([isinstance(a, int) for a in new_atoms]):
         raise VectorsError(f'all entries in atoms must be integers, got: {new_atoms} ({[type(a) for a in new_atoms]})')
     new_atoms = [a - index for a in new_atoms]  # convert 1-index to 0-index
+    if _ck_available:
+        c0, c1, c2 = coords[new_atoms[0]], coords[new_atoms[1]], coords[new_atoms[2]]
+        deg = _ck.zmat_a_f32(c0[0], c0[1], c0[2], c1[0], c1[1], c1[2], c2[0], c2[1], c2[2])
+        return deg if 'degs' in units else math.radians(deg)
```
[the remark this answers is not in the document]
kfir4444: Confirmed and fixed. Worth noting for the record that this bug predates the branch — main has angle = 0.5 * sum([torsion_dict['angle 1'], torsion_dict['angle 2']]) in the same function; the Rodrigues rewrite inherited it rather than introduced it.

It now takes the circular midpoint along the shorter arc: half = 0.5 * get_angle_in_180_range(a2 - a1, round_to=None) (engine.py:560-565). I deliberately did not use signed_angular_diff() here, even though it's the obvious helper — it rounds to 2 decimals, which would quantize every rotation to 0.01° and accumulate across up to 50 alignment iterations.

arc/mapping/engine.py
kfir4444: It isn't safe — you were right, and the cap is gone entirely.

Measured on jittered benzene, truncating to 4 discarded the best-RMSD candidate in half of the trials, by up to 0.04 Å. The cap was trading correctness for a bound that wasn't needed: the number of distinct candidates is bounded by the molecule's symmetry order, which was at most 6 across the aromatics, rings and cages tested. The dedup that made the cap look harmless is retained, but inline (engine.py:291-301), so duplicates are collapsed without dropping the tail.

arc/mapping/engine.py
kfir4444: The BFS wrapped around the ring and returned everything except pivot_i, and rotating that set about the pivot axis would tear the ring apart.

_downstream_atoms_adj now detects this and returns [] (engine.py:462-463): after the BFS, if any other neighbour of pivot_i was reached, the bond must lie in a cycle, so there is no well-defined downstream group. Callers treat [] as "skip this torsion".

In practice rotor torsions never contain ring bonds — find_internal_rotors in conformers.py:1690 filters on not mol.is_bond_in_cycle(bond) — so this is a guard for direct callers rather than a live bug on the rotor path. It's cheap and the function is now safe standalone.

arc/family/family.py
```diff
@@ -567,20 +618,43 @@ def determine_possible_reaction_products_from_family(rxn: ARCReaction,
...
+        if p_species is None:
+            p_species = rxn.get_reactants_and_products(return_copies=True)[1]
+        if iso_cache is None:
+            iso_cache = {}
         for group_labels, product_lists in products.items():
             for product_list in product_lists:
                 template_mols, r_label_dict = product_list[0], product_list[1]
-                if not isomorphic_products(rxn=rxn, products=template_mols):
-                    continue
+                # Cache check_product_isomorphism by adjacency-list key.
```
calvinp0: Can we double check this caching? I think since the key is coming form `to_adjacancey_list()`, it includes the reaction site labels (*1, *2) and I believe those would differ for every candidate which would in turn cases misses. So we should strip the labels before building the key

arc/species/_zmat_c_kernels.c
```diff
@@ -0,0 +1,222 @@
...
+    if (cosine <-1.f) cosine =-1.f;
+    /* arccos in float32, degree conversion in float32 (matches np.arccos(float32)*float) */
+    return (double)(acosf(cosine) * (180.0f / (float)M_PI));
+}
+
+double zmat_d_f32(double ax, double ay, double az,
+                  double bx, double by, double bz,
+                  double cx, double cy, double cz,
+                  double dx, double dy, double dz) {
+    /* Compute difference vectors in float32 (matches np.asarray(coords, float32)).
```
calvinp0, line 79: Should we do a length check too since in `get_dihedral` we do that?

arc/species/vectors.py
```diff
@@ -246,6 +253,12 @@ def calculate_dihedral_angle(coords: list | tuple | dict,
         raise VectorsError(f'all entries in torsion must be integers, got: {new_torsion} '
                            f'({[type(t) for t in new_torsion]})')
     new_torsion = [t - index for t in new_torsion]  # convert 1-index to 0-index if needed
+    if _ck_available:
+        c0, c1 = coords[new_torsion[0]], coords[new_torsion[1]]
+        c2, c3 = coords[new_torsion[2]], coords[new_torsion[3]]
+        deg = _ck.zmat_d_f32(c0[0], c0[1], c0[2], c1[0], c1[1], c1[2],
+                              c2[0], c2[1], c2[2], c3[0], c3[1], c3[2])
+        return deg if 'degs' in units else math.radians(deg)
```
calvinp0: so this function returns `get_dihedral` which can raise a `VectorsError` but the kernel does not have that

arc/job/adapters/ts/linear_test.py
```diff
@@ -4938,48 +4945,25 @@ def test_interpolate_intra_oh_migration(self):
...
-                if angle < 75.0:
-                    ok_angles = False
-                    break
-            if not ok_angles:
-                continue
-            found_good = True
-            break
-        self.assertTrue(found_good, msg='No TS guess has early-TS OH-migration characteristics: stretched O-O, moderate '
-                                        'C-O forming, near-bonded migrating H, and perpendicular spectator H orientation')
-        expected_ts = """C      -1.40886397    0.22567351   -0.37379668
```
calvinp0, line 4973: Was this intended to be deleted?

arc/species/_zmat_c_kernels.c
```diff
@@ -0,0 +1,222 @@
...
+
+double zmat_r_f32(double ax, double ay, double az,
+                  double bx, double by, double bz) {
+    float fax=(float)ax, fay=(float)ay, faz=(float)az;
+    float fbx=(float)bx, fby=(float)by, fbz=(float)bz;
+    float dx=fbx-fax, dy=fby-fay, dz=fbz-faz;
+    return (double)sqrtf(dx*dx + dy*dy + dz*dz);
+}
+
+double zmat_a_f32(double ax, double ay, double az,
```
calvinp0: Is this called anywhere? I saw a comment in line 204 of vectors.py about it sticking to numpy rather than using this

arc/species/_zmat_c_kernels.c
```diff
@@ -0,0 +1,222 @@
...
+/*
+ * Given atoms A, B, C already placed, add atom D with:
+ *   bond length C-D  = cd_len
+ *   angle B-C-D      = bcd_deg  (degrees)
+ *   dihedral A-B-C-D = abcd_deg (degrees)
+ *
+ * Implements the SN-NeRF algorithm (Parsons et al. 2005, J. Comput. Chem.).
+ * Mirrors _add_nth_atom_to_coords in zmat.py for i >= 3.
+ */
+void zmat_nerf(double ax, double ay, double az,
```
calvinp0: I also do not see where this is used

arc/statmech/arkane.py
```diff
@@ -566,9 +585,30 @@ def run_arkane(statmech_dir: str) -> bool:
...
+    elif RMG_PYTHON and os.path.isfile(RMG_PYTHON):
+        # Direct Python path — works in envs where MAMBA_EXE is absent (mambaforge, conda)
+        # but micromamba's condabin/conda shim would fail.
+        rmg_bin = os.path.dirname(RMG_PYTHON)
+        shell_script = rf'''bash -c 'set -euo pipefail
+cd "{statmech_dir}"
+export PATH="{rmg_bin}:$PATH"
+export RMG_DB_PATH="{rmg_db_path}"
+export RMG_DATABASE="{rmg_db_path}"
+"{RMG_PYTHON}" -m arkane input.py{arkane_suffix}' '''
```
calvinp0: I see in the comment about the LD_LIBRARY_PATH corruption inheritance. Should we not scrub the env here?

calvinp0: Or we can drop it and keep use of the `else`  at `610`?

arc/output.py
```diff
@@ -326,18 +327,30 @@ def _get_energy_corrections(arkane_level_of_theory, bac_type: str | None) -> tup
...
-                f'    micromamba run -n {rmg_env} python {script_path} {tmp_in} {tmp_out}; '
-                'elif command -v conda >/dev/null 2>&1; then '
-                f'    conda run -n {rmg_env} python {script_path} {tmp_in} {tmp_out}; '
-                'elif command -v mamba >/dev/null 2>&1; then '
-                f'    mamba run -n {rmg_env} python {script_path} {tmp_in} {tmp_out}; '
-                'else '
-                '    echo \'micromamba/conda/mamba required\' >&2; exit 1; '
-                'fi"',
-            ]
+            mamba_exe = os.environ.get('MAMBA_EXE', '')
```
calvinp0: With this and also in processor.py, maybe we should have a helper instead in common.py for the rmg_env_command?
I did implement something of a helper here for the TS scripts - maybe we could extract it to common for use here too?
https://github.com/ReactionMechanismGenerator/ARC/blob/4e75e7968a64856aee09b6d6ba15846b89373b68/arc/job/env_run.py

calvinp0 (approved):
amazing

PR #908, opened 2026-07-04
fix: replace removed distutils.spawn with shutil.which (Python 3.12+)

alongd:
## Problem

`distutils` was removed from the Python standard library in **3.12**, so `arc/main.py`'s
`from distutils.spawn import find_executable` raises `ModuleNotFoundError` on Python 3.14.
It only appears to work in environments where setuptools happens to install its `distutils`
compatibility shim — a fresh CI env without that shim fails.

Because `arc.main` is imported transitively by most of ARC (and by downstream consumers like
T3), this breaks at **import/collection time**. Concretely, the T3 test suite currently dies with:

```
ImportError while importing test module '.../tests/test_common.py'.
E   ModuleNotFoundError: No module named 'distutils'
```

(chain: `t3.common` → `arc.species.species` → `arc.main` → `from distutils.spawn import find_executable`)

## Fix

Replace the single stdlib usage with **`shutil.which`**, the documented drop-in replacement for
`distutils.spawn.find_executable` — both return the path to the executable found on `PATH`, or
`None`. `shutil` is already imported in `arc/main.py`, so no new import is added.

The 7 affected call sites all look up ESS executables on `PATH` (`g03`/`g09`/`g16`/`qchem`/`orca`/
`molpro`/`terachem`), and the downstream truthiness checks (`if g03 or g09 or g16:` …) are unchanged.

## Scope / safety

- This is the **only** `distutils` usage anywhere in `arc/*.py`, and there are no other
  removed-stdlib imports — this single change makes the ARC import graph Python-3.12+/3.14 clean.
- ARC's own `arc.settings.settings.find_executable(env_name, ...)` is a **different** function
  (it locates the python interpreter of a named conda env) and is intentionally left untouched;
  `arc/job/adapters/ase_adapter.py` imports that settings function, not the stdlib one.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -13,7 +13,6 @@ import logging
 import os
 import shutil
 import time
-from distutils.spawn import find_executable
 from IPython.display import display
 
 from arc.common import (VERSION,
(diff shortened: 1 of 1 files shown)


kfir4444 (approved):
LGTM. Was this tested on an env with python>3.12? 
If so lets update the env to use it

PR #909, opened 2026-07-07
Fix NMD crash on un-mappable reactions and OS memory over-allocation trsh

alongd:
Two independent, self-contained bug fixes surfaced and validated during an end-to-end run (`linear` TS-adapter validation, wB97XD/def2-TZVP, Gaussian). Each was **confirmed on real jobs** during the run, and each ships with unit tests.

## 1. NMD check crashes on un-atom-mappable reactions
`analyze_ts_normal_mode_displacement()` raised a `ReactionError` (`Cannot get bonds ...`) when `reaction.atom_map is None`, aborting the **entire** ARC run. This happens for reactions ARC cannot atom-map — e.g. `CH3 + CS <=> CCS` (`R_Addition_MultipleBond` with a charged `[C-]#[S+]`).

**Fix:** guard the check to log a warning and return `None` when there is no atom map, so the scheduler skips NMD validation for that one reaction and continues.

- `arc/checks/nmd.py`, test in `arc/checks/nmd_test.py`.
- Well-scoped: NMD is skipped **only** when mapping is impossible; when a map exists the check still runs and correctly rejects wrong-mode TSs.

## 2. Conflating OS memory over-allocation with Gaussian needing more memory
A Gaussian `galloc`/`malloc failed` error means the **node could not provide the requested memory** (we asked for too much) — the opposite of Gaussian internally requesting *more* memory. Both were classified as `Memory`, so troubleshooting **increased** the request, making a node-capacity failure strictly worse.

**Fix:** classify the OS-level failure as `MemoryOverallocation` and troubleshoot it by **halving** the request down to a 4 GB floor, giving up with a clear "use a higher-memory node" message once at the minimum.

- `arc/job/trsh.py`, tests in `arc/job/trsh_test.py`, fixture `arc/testing/trsh/gaussian/galloc.out`.
- Tests cover both the status classification and the reduce-then-give-up path.

## Validation
- New unit tests pass (verified against repo-default server settings).
- Both fixes were **live-confirmed** during a ~2-day zeus run: the NMD guard emitted the skip message and the run continued with **0 crashes**; the memory-overallocation path was observed reducing real jobs' memory request on the cluster.

### DIFF
--- a/arc/checks/nmd.py
+++ b/arc/checks/nmd.py
@@ -49,6 +49,12 @@ def analyze_ts_normal_mode_displacement(reaction: ARCReaction,
     """
     if job is None:
         return None
+    if reaction.atom_map is None:
+        # Without an atom map the formed/broken/changed bonds cannot be determined; skip the NMD
+        # check for this reaction rather than crashing the whole run (e.g. reactions ARC cannot map).
+        logger.warning(f'Cannot analyze the TS normal mode displacement for reaction {reaction.label} '
+                       f'without an atom map; skipping the NMD check for this reaction.')
+        return None
     ts_xyz = reaction.ts_species.get_xyz()
     n_ts = len(ts_xyz['symbols'])
     n_expected = sum(spc.number_of_atoms for spc in reaction.r_species)
--- a/arc/checks/nmd.py
+++ b/arc/checks/nmd.py
@@ -49,6 +49,15 @@ def analyze_ts_normal_mode_displacement(reaction: ARCReaction,
     """
     if job is None:
         return None
+    if reaction.atom_map is None:
+        # Without an atom map the formed/broken/changed bonds cannot be determined; skip the NMD
+        # check for this reaction rather than crashing the whole run (e.g. reactions ARC cannot map).
+        logger.warning(f'Cannot analyze the TS normal mode displacement for reaction {reaction.label} '
+                       f'without an atom map; skipping the NMD check for this reaction.')
+        if 'Atom map is None; skipped the TS normal mode displacement check; ' \
+                not in reaction.ts_species.ts_checks['warnings']:
+            reaction.ts_species.ts_checks['warnings'] += 'Atom map is None; skipped the TS normal mode displacement check; '
+        return None
     ts_xyz = reaction.ts_species.get_xyz()
     n_ts = len(ts_xyz['symbols'])
     n_expected = sum(spc.number_of_atoms for spc in reaction.r_species)
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -201,8 +201,10 @@ def determine_ess_status(output_path: str,
                     error = 'An operation on the check file was specified, but a .chk was not found or is incomplete.'
                     line = ''
                 elif 'malloc failed' in line or 'galloc' in line:
-                    keywords = ['Memory']
-                    error = 'Memory allocation failed (did you ask for too much?)'
+                    # A system-level allocation failure: the node could not provide the requested
+                    # memory (we asked for too much), distinct from Gaussian internally needing more.
+                    keywords = ['MemoryOverallocation']
+                    error = 'Memory allocation failed at the OS level - the node could not provide the requested memory.'
                     line = ''
                 elif 'PGFIO/stdio: No such file or directory' in line:
                     keywords = ['Scratch']
@@ -974,6 +976,25 @@ def trsh_ess_job(label: str,
         if 'no_qc' in ess_trsh_methods:
             logger_info.append('removed QC')
         
+
+        # A system-level allocation failure (galloc/malloc): the node could not provide the requested
+        # memory, so we asked for too much. Reduce the request rather than escalating - increasing
+        # memory on an over-allocation only makes a node-capacity failure worse.
+        if 'MemoryOverallocation' in job_status['keywords'] and server is not None:
+            reduced_memory = max(memory_gb / 2.0, 4)
+            if reduced_memory < memory_gb:
+                couldnt_trsh = False
+                memory = reduced_memory
+                logger.info(f'Troubleshooting {job_type} job in {software} for {label} using less memory: {memory} GB '
+                            f'instead of {memory_gb} GB (the node could not allocate the requested amount)')
+                ess_trsh_methods.append('memory')
+            else:
+                couldnt_trsh = True
+                output_errors.append(
+                    f'Error: Could not troubleshoot {job_type} for {label}! The node could not allocate even '
+                    f'{memory_gb} GB, and ARC is already at its minimum memory request; use a higher-memory node.')
+                logger.error(f'Could not troubleshoot {job_type} job in {software} for {label}. The node could not '
+                             f'allocate {memory_gb} GB (already at the minimum memory request).')
 
         # Check if memory is in the keyword
         if 'Memory' in job_status['keywords'] and 'too high' not in job_status['error'] and server is not None:
--- a/arc/job/trsh.py
+++ b/arc/job/trsh.py
@@ -13,6 +13,7 @@ from arc.common import (check_torsion_change,
                         convert_to_hours,
                         estimate_orca_mem_cpu_requirement,
                         get_logger,
+                        get_memory_headroom_fraction,
                         get_number_with_ordinal_indicator,
                         is_same_pivot,
                         is_same_sequence_sublist,
@@ -40,12 +41,12 @@ from arc.parser.parser import (parse_1d_scan_coords,
 
 logger = get_logger()
 
-delete_command, inconsistency_ab, inconsistency_az, maximum_barrier, preserve_params_in_scan, rotor_scan_resolution, \
-    servers, submit_filenames, default_job_settings = settings['delete_command'], settings['inconsistency_ab'], \
-                                                      settings['inconsistency_az'], settings['maximum_barrier'], \
-                                                      settings['preserve_params_in_scan'], \
-                                                      settings['rotor_scan_resolution'], settings['servers'], \
-                                                      settings['submit_filenames'], settings['default_job_settings']
+delete_command, gaussian_memory_headroom_fractions, inconsistency_ab, inconsistency_az, maximum_barrier, \
+    preserve_params_in_scan, rotor_scan_resolution, servers, submit_filenames, default_job_settings = \
+    settings['delete_command'], settings['gaussian_memory_headroom_fractions'], settings['inconsistency_ab'], \
+    settings['inconsistency_az'], settings['maximum_barrier'], settings['preserve_params_in_scan'], \
+    settings['rotor_scan_resolution'], settings['servers'], settings['submit_filenames'], \
+    settings['default_job_settings']
 
 
 def determine_ess_status(output_path: str,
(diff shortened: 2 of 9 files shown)


arc/job/trsh.py
```diff
@@ -1015,6 +1017,25 @@ def trsh_ess_job(label: str,
             logger_info.append('removed QC')
         
 
+        # A system-level allocation failure (galloc/malloc): the node could not provide the requested
```
calvinp0: I tried this patch recently and there is a bug where the halving has not stopping rule.
In one example of a job:

 starting at 42 GB:

 round  mem_in  mem_out  gave_up  ess_trsh_methods
 0        42      21      False   ['int=(Acc2E=14)', 'memory']
 1        21      10.5    False   [..., 'memory', 'memory']
 2        10.5    5.25    False   [..., 'memory' x3]
 3        5.25    4       False   [..., 'memory' x4]
 4        4       4       True    [..., 'all_attempted']

Every row is a full resubmission. The real job spent 9 h 17 m doing this and the TS was then abandoned.

Why it can't converge: my log had both initial_time and final_time, so the queue granted the reservation and the job ran. Gaussian then died in Link 1 with MaxMem=531496960 - that is 4055 MiB, exactly the %mem ARC had written. At that point the job had too little memory, and halving digs deeper.

There is also no ratio to be gained. %mem is always exactly 90% of the reservation at every size (adapter.py multiplies job_memory_gb by 1.10, or 1.05 when capped; gaussian.py then takes 0.90 of that). Halving keeps the same 90% and only shrinks the absolute slack - ~4.6 GiB at 42 GB, ~0.44 GiB at 4 GB. If the failure comes from Gaussian's memory outside %mem, each halving makes it more likely.

arc/job/trsh.py
```diff
@@ -1015,6 +1017,25 @@ def trsh_ess_job(label: str,
...
+            reduced_memory = max(memory_gb / 2.0, 4)
+            if reduced_memory < memory_gb:
+                couldnt_trsh = False
+                memory = reduced_memory
+                logger.info(f'Troubleshooting {job_type} job in {software} for {label} using less memory: {memory} GB '
+                            f'instead of {memory_gb} GB (the node could not allocate the requested amount)')
+                ess_trsh_methods.append('memory')
+            else:
+                couldnt_trsh = True
+                output_errors.append(
```
calvinp0, line 1035: By the time ARC is requesting 4GB, so 'use a higher-memory node' points at the one thing that cannot help - on a bigger node ARC would still ask for 4GB. Maybe we should write to the user which levers they need to adjust for next time as well?

arc/job/trsh.py
```diff
@@ -1015,6 +1017,25 @@ def trsh_ess_job(label: str,
             logger_info.append('removed QC')
         
 
+        # A system-level allocation failure (galloc/malloc): the node could not provide the requested
+        # memory, so we asked for too much. Reduce the request rather than escalating - increasing
+        # memory on an over-allocation only makes a node-capacity failure worse.
+        if 'MemoryOverallocation' in job_status['keywords'] and server is not None:
+            reduced_memory = max(memory_gb / 2.0, 4)
```
calvinp0: There is an edge case if `memory_gb` is already below 4 (say 3), `reduced memory` is 4, so you fall into the give up branch and report the node cannot allocate even 3Gb and arc is already at its min mem request. It is below, not at it. Also, 4 being a bare literal rather than a memory know from settings may be something to consider? although settings starts to really grow and I guess unreadable

arc/job/trsh.py
```diff
@@ -1015,6 +1017,25 @@ def trsh_ess_job(label: str,
...
+            if reduced_memory < memory_gb:
+                couldnt_trsh = False
+                memory = reduced_memory
+                logger.info(f'Troubleshooting {job_type} job in {software} for {label} using less memory: {memory} GB '
+                            f'instead of {memory_gb} GB (the node could not allocate the requested amount)')
+                ess_trsh_methods.append('memory')
```
calvinp0: worth knowing, increas path below (line 1049) appends the same `memory` marker. So if  ajob alternates - galloc, halve, Gaussian asks for more, doubel, galloc - neither brnach can tell it is oscillating. And because the list keeps growing, the `attempted_ess_trsh_methods == ess_trsh_methods` never trips. It is bounded by `max_ess_trsh` (25) but that is 25 submisisons.

arc/job/trsh_test.py
```diff
@@ -856,6 +856,40 @@ def test_trsh_ess_job_terachem_trsh_attempt_only(self):
...
+        label, level_of_theory, server = 'ethanol', {'method': 'wb97xd', 'basis': 'def2tzvp'}, 'server1'
+        job_type, software, fine, num_heavy_atoms, cpu_cores = 'opt', 'gaussian', False, 2, 8
+        job_status = {'keywords': ['MemoryOverallocation'],
+                      'error': 'Memory allocation failed at the OS level - the node could not provide the requested memory.'}
+        # Above the floor: the request is halved (32 -> 16), not increased.
+        *_, memory, shift, cpu_cores_out, couldnt_trsh = trsh.trsh_ess_job(
+            label, level_of_theory, server, job_status, job_type, software, fine, 32,
+            num_heavy_atoms, cpu_cores, [])
+        self.assertFalse(couldnt_trsh)
+        self.assertEqual(memory, 16)
```
calvinp0: the test covrs one step (32 to 16) and the floor case but nothign covers galloc happening again, which is where the behaviour can go wrong.

arc/checks/nmd.py
```diff
@@ -49,6 +49,12 @@ def analyze_ts_normal_mode_displacement(reaction: ARCReaction,
     """
     if job is None:
         return None
+    if reaction.atom_map is None:
+        # Without an atom map the formed/broken/changed bonds cannot be determined; skip the NMD
+        # check for this reaction rather than crashing the whole run (e.g. reactions ARC cannot map).
+        logger.warning(f'Cannot analyze the TS normal mode displacement for reaction {reaction.label} '
```
calvinp0, line 55: it may also be beneficial to appen a line to `reaction.ts_species.ts_checks['warnings']` so a skipped NMD check shows up in the run summary

calvinp0 (changes requested):

arc/job/trsh.py
alongd: @-

arc/job/trsh.py
alongd: @-

arc/job/trsh.py
alongd: @-

arc/job/trsh.py
alongd: @-

arc/job/trsh_test.py
alongd: @-

arc/checks/nmd.py
alongd: @-

arc/job/trsh.py
```diff
@@ -975,6 +977,25 @@ def trsh_ess_job(label: str,
...
+            else:
```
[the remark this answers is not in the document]
alongd: @-

arc/job/trsh.py
```diff
@@ -975,6 +977,25 @@ def trsh_ess_job(label: str,
...
+                couldnt_trsh = True
+                output_errors.append(
+                    f'Error: Could not troubleshoot {job_type} for {label}! The node could not allocate even '
+                    f'{memory_gb} GB, and ARC is already at its minimum memory request; use a higher-memory node. ')
+                logger.error(f'Could not troubleshoot {job_type} job in {software} for {label}. The node could not '
```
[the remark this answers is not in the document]
alongd: @-

alongd:
@calvinp0 thanks — your 42 GB run took the memory half of this PR apart, and it deserved it. Reworked and rebased onto `main`.

**The premise was wrong, not just the magnitude.** A `galloc` only ever appears *inside* a Gaussian output file, which exists only because the queue granted the reservation and g16 started. So it is never the node refusing an oversized request; it is a malloc failure inside an already-granted allocation. Combined with your ratio argument — `adapter.py` reserves `job_memory_gb * 1.10`, `gaussian.py` gives Gaussian `0.90` of that, so `%mem`/reservation is size-invariant — halving preserved the 90% and only shrank the absolute slack. `gaussian.py:503-508` already blamed insufficient headroom for `galloc`; this PR had been contradicting a comment three files away.

**What it does now.** Hold the queue reservation constant and give Gaussian more headroom by stepping `%mem` down `0.90 → 0.75 → 0.60`, then give up with the levers the user can actually pull. Neither more nor less total memory is requested, which is the only remedy that is safe under both readings of a `galloc`. On your 42 GB job that is headroom 4.6 → 11.6 → 18.5 GiB over 2 resubmissions, instead of 4.6 → 0.44 GiB over 5.

Also in this revision:
- keyword renamed `MemoryOverallocation` → `GaussianMemoryAllocation`; the old name asserted a cause ARC cannot establish
- the ladder lives in `settings.gaussian_memory_headroom_fractions`, which also absorbs the `0.90` that was a second hardcoded constant in `gaussian.py` — one knob, one fewer magic number
- the 4 GB floor is gone, so the `memory_gb < 4` edge case dissolves
- the marker is `memory_headroom_<fraction>`, distinct from the increase branch's `'memory'`; with a finite ladder the reduce/increase oscillation is structurally impossible. It is also kept strictly out of `trsh_keyword`, since `update_input_dict_with_args()` renders that verbatim into the Gaussian route section
- markers accumulate in the restart file, so the resolution rule is explicit: lowest fraction wins
- skipped NMD checks now surface in `ts_checks['warnings']`

**Tests:** repeated-galloc ladder stepping, ladder exhaustion and message format, a 12-round alternation-termination regression test (fails on the previous head), `get_memory_headroom_fraction` edge cases, `%mem` byte-identical with no marker, and marker-never-reaches-the-input-file. 1210 passed across `arc/job/`, `arc/checks/` and `arc/scheduler_test.py`.

One thing I did **not** do, and want your read on: Codex's first-choice remedy was to requeue once on a different node at identical memory before touching `%mem`, on the theory that an oversubscribed or sick node is fixed by neither more nor less memory. ARC already has the machinery (`trsh_job_on_server` + `server_nodes`) but only on the server-error path, and routing an ESS error into node reselection needs a new channel out of `trsh_ess_job` into `scheduler.py`. That felt like a feature rather than part of this bugfix, so I left it — worth a follow-up issue?

alongd:
CI note: the red run was flaky, not a regression. It failed on `arc/species/conformers_test.py::test_embed_rdkit` and `::test_rdkit_force_field_abandons_an_optimization_that_raises` — both in a module this PR doesn't touch, both with `cj_spc.mol` arriving with an over-full carbon valence, i.e. cross-test state contamination on the shared class fixture. Re-running the identical commit with no changes passed.

Evidence it isn't this branch: 6 full-suite runs on `main` and 6 on this branch, all under the repo's default `-n auto --dist=worksteal`. `main` fails `arc/job/adapter_test.py::test_determine_job_status` 6/6 locally (pre-existing, and separate from the CI failure). This branch shows the same, plus an occasional extra flake that lands in a *different* module each time — `xtbgsm_test` once, `ts_test::test_compute_rxn_e0` once, `conformers_test` in CI — which is the signature of latent order-dependent contamination being reshuffled by any change in test count, not of a defect in the diff.

The one place this PR could legitimately perturb that: a new test mutated the shared class-level `job_1` fixture and restored it afterwards. Rewritten to build fresh adapter instances instead, which also tests the production path (the marker arrives at construction). Squashed into the same commit; force-pushed as `5b5b11b`.

The pre-existing `test_determine_job_status` failure under xdist looks worth a separate issue — it reproduces on `main` without this branch.

calvinp0 (approved):

PR #912, opened 2026-07-14
Enrich running_jobs.yml snapshot with server identifiers

calvinp0:
## What

ARC's `<project>/running_jobs.yml` status snapshot previously listed only bare job_names (e.g. `tsg3`). To find the job on the cluster you had to cross-reference `arc.log` to learn its qstat name (e.g. `a3129`).

This adds `Scheduler._running_job_snapshot_entry`, which resolves each running `job_name` back to its `Job` object in `job_dict` (matching on `job.job_name`) and emits a self-describing entry: `{name, server_name (the qstat job name), job_id, adapter, server}`. Unresolvable entries fall back to `{'name': job_name}`. `report_running_jobs_snapshot` now builds `running_jobs` as `{label: [entry, ...]}`.

## Why

Surfaced while diagnosing benchmark reaction_22, where the snapshot alone couldn't map `tsg3` to its qstat job `a3129` without arc.log. This is a usability/observability improvement — no behavior change to job scheduling.

## Testing

`arc/scheduler_test.py::test_report_running_jobs_snapshot` updated to expect the new dict shape and to assert a resolvable job carries its server ids. Full `arc/scheduler_test.py` passes (29 passed).

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -871,6 +871,37 @@ class Scheduler(object):
         # Generate a TS report:
         self.generate_final_ts_guess_report()
 
+    def _running_job_snapshot_entry(self, label: str, job_name: str) -> dict:
+        """
+        Resolve a running ``job_name`` (as stored in ``self.running_jobs``, e.g. ``'tsg3'``)
+        back to its Job object and build a snapshot entry that carries the identifiers needed
+        to find the job on the server: ``server_name`` (the queue job name shown by ``qstat``,
+        e.g. ``a3129``), ``job_id`` (the queue's numeric id), the ``adapter`` (software), and the
+        ``server``. Falls back to just ``{'name': job_name}`` if the Job object can't be located
+        (e.g. the job was already popped, or a restart-loaded entry whose stored name doesn't
+        match any live ``job.job_name``). Resolved incore jobs still get an entry, carrying the
+        job number as their ``job_id``, while ``server`` may be ``None``.
+
+        Args:
+            label (str): The species/TS label the job belongs to.
+            job_name (str): The job name as stored in ``self.running_jobs``.
+
+        Returns:
+            dict: A self-describing snapshot entry for the running job.
+        """
+        entry = {'name': job_name}
+        for jobs in self.job_dict.get(label, dict()).values():
+            if not isinstance(jobs, dict):
+                continue
+            for job in jobs.values():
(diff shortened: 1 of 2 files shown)


calvinp0:
Closing — opened prematurely. Tracking via the branch only; not requesting review yet.

arc/scheduler.py
```diff
@@ -871,14 +871,47 @@ def schedule_jobs(self):
...
+    def _running_job_snapshot_entry(self, label: str, job_name: str) -> dict:
+        """
+        Resolve a running ``job_name`` (as stored in ``self.running_jobs``, e.g. ``'tsg3'``)
+        back to its Job object and build a snapshot entry that carries the identifiers needed
+        to find the job on the server: ``server_name`` (the queue job name shown by ``qstat``,
+        e.g. ``a3129``), ``job_id`` (the queue's numeric id), the ``adapter`` (software), and the
+        ``server``. Falls back to just ``{'name': job_name}`` if the Job object can't be located
+        (e.g. the job was already popped, or a restart-loaded entry whose stored name doesn't
+        match any live ``job.job_name``). Resolved incore/local jobs still get an entry, just with
+        ``server``/``job_id`` possibly ``None``.
```
Lilachn91: That's not quite right — incore jobs always get a job_id (it falls back to the job's internal number if nothing else is set). Only server can actually be None for incore jobs. Could you fix the wording so it doesn't say job_id can be None too?

arc/scheduler_test.py
```diff
@@ -1277,11 +1278,12 @@ def test_report_running_jobs_snapshot(self):
         self.sched1.running_jobs = {'spcA': ['opt_a1234']}
         self.sched1.active_pipes = {}
 
-        # First call: file doesn't exist, snapshot must be written.
+        # First call: file doesn't exist, snapshot must be written. An unresolvable job
```
Lilachn91, line 1281: This test covers two cases: the label has nothing in job_dict at all (falls back to just the name), and the label has a matching job (gets the full enriched entry). It's missing a third case: the label has other jobs in job_dict, just none of them matching the one we're looking for, e.g. we're looking for 'sp_b5678', but running_jobs is still just {'spcA': ['opt_a1234']}.

Lilachn91 (commented):
LGTM! See two minor comments.

Lilachn91 (approved):
Looks good, thanks!

PR #914, opened 2026-07-20
Fix multiplicity parity validation

calvinp0:
This pull request introduces a parity check for spin multiplicity and electron count in `ARCSpecies`, ensuring that the specified multiplicity is physically possible given the species' composition and charge. It also adds robust testing for this logic and fixes a test case for nitrogen to use the correct multiplicity and Molpro input. Additionally, it improves code clarity by importing `NUMBER_BY_SYMBOL` where needed.

Validation of multiplicity and electron count:

* Added a `check_multiplicity_parity` method to `ARCSpecies` that verifies the parity of the electron count and multiplicity, raising a `SpeciesError` if they are inconsistent. This method is called during initialization. [[1]](diffhunk://#diff-90a40ffecf9d6d48c85a21031d54d87a791aea24a88edffcb41d4fd1a250c223R1565-R1624) [[2]](diffhunk://#diff-90a40ffecf9d6d48c85a21031d54d87a791aea24a88edffcb41d4fd1a250c223R520)
* Implemented a `get_number_of_electrons` method in `ARCSpecies` to count electrons from the molecule or xyz.

Testing improvements:

* Added `test_check_multiplicity_parity` to `species_test.py` to ensure the parity check works and raises errors for impossible combinations.

Bug fixes and test corrections:

* Updated the nitrogen test case in `molpro_test.py` to use `multiplicity=4` (quartet) and adjusted the Molpro input file templates accordingly. [[1]](diffhunk://#diff-e2915f1ebb1a8146a93fa9c5e65906724a335658792b6872122292821bb06a1cL95-R95) [[2]](diffhunk://#diff-e2915f1ebb1a8146a93fa9c5e65906724a335658792b6872122292821bb06a1cL415-R433)

Code clarity:

* Added the missing import of `NUMBER_BY_SYMBOL` in `species.py` to support electron counting.

### DIFF
--- a/arc/checks/ts_test.py
+++ b/arc/checks/ts_test.py
@@ -847,7 +847,7 @@ class TestTSChecks(unittest.TestCase):
         xyz_2 = parse_geometry(os.path.join(ARC_TESTING_PATH, 'irc', 'rxn_1_irc_2.out'))
         # Use wrong product species
         rxn = ARCReaction(r_species=[ARCSpecies(label='R', smiles='O=[C]COO', xyz=xyz_1)],
-                          p_species=[ARCSpecies(label='P_wrong', smiles='CC', xyz=xyz_2)])
+                          p_species=[ARCSpecies(label='P_wrong', smiles='O=C(O)C[O]', multiplicity=2)])
         rxn.ts_species = ARCSpecies(label='TS', is_ts=True)
         ts.check_irc_species_and_rxn(xyz_1=xyz_1, xyz_2=xyz_2, rxn=rxn)
         self.assertFalse(rxn.ts_species.ts_checks['IRC'])
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -9,13 +9,14 @@ import os
 from math import isclose
 
 import arc.molecule.element as elements
-from arc.common import (SYMBOL_BY_NUMBER,
-                        almost_equal_coords,
+from arc.common import (almost_equal_coords,
                         convert_list_index_0_to_1,
+                        count_electrons,
                         dfs,
                         get_logger,
                         get_single_bond_length,
                         is_angle_linear,
+                        is_multiplicity_parity_valid,
                         is_xyz_linear,
                         read_yaml_file,
                         timedelta_from_str,
@@ -519,6 +520,7 @@ class ARCSpecies(object):
         if not isinstance(self.charge, int):
             raise SpeciesError(f'Charge for species {self.label} is not an integer. '
                                f'Got {self.charge} which is a {type(self.charge)}.')
+        self.check_multiplicity_parity()
         if not self.is_ts and self.initial_xyz is None and self.final_xyz is None and self.mol is None \
                 and not self.conformers:
             raise SpeciesError(f'No structure (xyz, SMILES, adjList, or Molecule) '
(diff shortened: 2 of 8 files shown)


arc/species/species.py
```diff
@@ -1560,6 +1562,61 @@ def determine_multiplicity_from_xyz(self):
                 self.multiplicity = 1
                 logger.debug(f'\nMultiplicity not specified for {self.label}, assuming a value of 1')
 
+    def get_number_of_electrons(self) -> int | None:
```
alongd, line 1604: please add a unit test

arc/species/species.py
```diff
@@ -1560,6 +1562,61 @@ def determine_multiplicity_from_xyz(self):
...
+            return
+        total_electrons = self.get_number_of_electrons()
+        if total_electrons is None:
+            return
+        n_electrons = total_electrons - self.charge
+        if n_electrons % 2 != (self.multiplicity - 1) % 2:
+            lower = self.multiplicity - 1
+            valid = [m for m in (lower, self.multiplicity + 1) if m >= 1]
+            parity_word = 'odd' if n_electrons % 2 == 0 else 'even'
+            logger.warning(
```
alongd: I think this should raise an error, not just warn

alongd (commented):
Thanks, @calvinp0 , Added two comments

arc/species/species.py
```diff
@@ -1606,6 +1608,69 @@ def determine_multiplicity_from_xyz(self):
                 self.multiplicity = 1
                 logger.debug(f'\nMultiplicity not specified for {self.label}, assuming a value of 1')
 
+    def get_number_of_electrons(self) -> int | None:
```
alongd, line 1604: I have a feeling that ARC does this or something similar in several places. would make sense to consolidate. can you please run an agent to find all similar functionalities and recommend how to consolidate?

arc/checks/ts_test.py
```diff
@@ -847,7 +847,7 @@ def test_check_irc_wrong_species(self):
         xyz_2 = parse_geometry(os.path.join(ARC_TESTING_PATH, 'irc', 'rxn_1_irc_2.out'))
         # Use wrong product species
         rxn = ARCReaction(r_species=[ARCSpecies(label='R', smiles='O=[C]COO', xyz=xyz_1)],
-                          p_species=[ARCSpecies(label='P_wrong', smiles='CC', xyz=xyz_2)])
+                          p_species=[ARCSpecies(label='P_wrong', smiles='O=C(O)C[O]', multiplicity=2)])
```
alongd, line 850: this changes the product itself, not just fixing the multiplicity. if we want it to be `P_wrong`, then maybe it should stay SMILES `CC`? and just the multiplicity should be `1`?

arc/species/species_test.py
```diff
@@ -608,6 +608,51 @@ def test_charge_and_multiplicity(self):
...
+        with self.assertRaises(SpeciesError):
+            ARCSpecies(label='HO2_bad', smiles='[O]O', multiplicity=1, compute_thermo=False)
+
+        with self.assertRaises(SpeciesError):
+            ARCSpecies(label='CH4_cation_bad', smiles='C', charge=1, multiplicity=1, compute_thermo=False)
+
+        ts = ARCSpecies(label='TS0', is_ts=True, multiplicity=2)
+        self.assertIsNone(ts.get_number_of_electrons())
+
+    def test_get_number_of_electrons(self):
```
alongd, line 694: please also test anions and cations, both radicals and non radicals
also add a test assertion on the error message

arc/species/species.py
```diff
@@ -1606,6 +1608,69 @@ def determine_multiplicity_from_xyz(self):
...
+        if total_electrons is None:
+            return
+        n_electrons = total_electrons - self.charge
+        if n_electrons % 2 != (self.multiplicity - 1) % 2:
+            lower = self.multiplicity - 1
+            valid = [m for m in (lower, self.multiplicity + 1) if m >= 1]
+            parity_word = 'odd' if n_electrons % 2 == 0 else 'even'
+            raise SpeciesError(
+                f'Impossible multiplicity for species {self.label}: a species with '
+                f'{total_electrons} electrons and a net charge of {self.charge} has '
```
alongd: Minor wording: "a species with 18 electrons ... has 17 electrons" reads oddly. Consider "a neutral species with {total_electrons} electrons at a net charge of {self.charge} has {n_electrons} electrons".

arc/species/species.py
alongd: CC reinforces me, saying:
```
This is now the **third** copy of symbol-to-electron counting in this module:

- `determine_multiplicity_from_xyz` (1595-1603) — nested `SYMBOL_BY_NUMBER` scan, subtracts charge
- `check_xyz` (3046-3054) — nested `SYMBOL_BY_NUMBER` scan, subtracts charge
- `get_number_of_electrons` (1620-1631) — `NUMBER_BY_SYMBOL` lookup, does not subtract charge

The PR imports `NUMBER_BY_SYMBOL`, which is exactly the dict that collapses the other two nested `O(len(SYMBOL_BY_NUMBER))` scans into a lookup — but only the new method uses it. Since `get_number_of_electrons()` now exists, both older sites could delegate to it and the duplication goes away:

```python
# determine_multiplicity_from_xyz
electrons = self.get_number_of_electrons()
if electrons is not None:
    electrons -= self.charge
    ...
```

That consolidation is what this PR set up; worth taking it the rest of the way so the three copies can't drift.
```

File: arc/species/species.py [hunk 7]
```diff
@@ -1606,6 +1608,69 @@ def determine_multiplicity_from_xyz(self):
...
+                           is not yet known (no Molecule and no xyz at this point).
+
+        Raises:
+            SpeciesError: If an xyz is present but contains an unrecognized atom symbol
+                          (a genuinely invalid composition, e.g. a typo'd element).
+        """
+        if self.mol is not None:
+            return sum(atom.element.number for atom in self.mol.atoms)
+        xyz = self.get_xyz()
+        if xyz is None and len(self.conformers):
```
alongd: CC says:
```
Dead branch, copy-pasted from lines 1592-1593. `get_xyz()` already folds `self.conformers[0]` into its `or` chain:

```python
# species.py:1256-1257
conf = self.conformers[0] if self.conformers else None
xyz = self.final_xyz or self.initial_xyz or self.most_stable_conformer or conf or self.cheap_conformer
```

So if `conformers[0]` is truthy, `xyz` is never `None` here; if it's falsy, this fallback assigns it and the very next `if xyz and xyz.get('symbols')` rejects it anyway. The branch can't change the result — safe to delete (in both places).
```

File: arc/species/species.py [hunk 8]
```diff
@@ -1606,6 +1608,69 @@ def determine_multiplicity_from_xyz(self):
...
+            Optional[int]: The total number of electrons, or ``None`` if the composition
+                           is not yet known (no Molecule and no xyz at this point).
+
+        Raises:
+            SpeciesError: If an xyz is present but contains an unrecognized atom symbol
+                          (a genuinely invalid composition, e.g. a typo'd element).
+        """
+        if self.mol is not None:
+            return sum(atom.element.number for atom in self.mol.atoms)
+        xyz = self.get_xyz()
```
alongd: CC says:
```
`get_xyz()` defaults to `generate=True`, so a method documented as a pure count can trigger force-field conformer generation as a side effect — and it's called from `__init__` via `check_multiplicity_parity`.

Not reachable today: the generate branch (species.py:1265) needs `self.mol is not None or self.mol_list is not None`, and we only get here when `self.mol is None`, with `set_mol_list()` not producing a `mol_list` in that state. So this is hardening, not a live bug.

Still worth `self.get_xyz(generate=False)` — it's free, and it keeps a constructor-time validator away from that path. When I forced the `mol is None` / `mol_list is not None` state manually I hit a `RecursionError` via `get_xyz` -> `get_cheap_conformer` -> `is_diatomic` -> `get_xyz` (a pre-existing cycle, not introduced here).
```

arc/species/species_test.py
```diff
@@ -608,6 +608,51 @@ def test_charge_and_multiplicity(self):
         self.assertEqual(spc7.multiplicity, 2)
         self.assertEqual(spc8.multiplicity, 1)
 
+    def test_check_multiplicity_parity(self):
```
alongd: CC says:
```

Coverage suggestions — the parity logic itself is right (I verified each of these behaves correctly), these just lock the behavior in:

```python
# anion — the `- self.charge` sign is where a bug would hide
ARCSpecies(label='OH_anion', smiles='[OH-]', multiplicity=1, compute_thermo=False)
with self.assertRaises(SpeciesError):
    ARCSpecies(label='OH_anion_bad', smiles='[OH-]', multiplicity=2, compute_thermo=False)

# high multiplicity — the case this PR had to fix in molpro_test.py
ARCSpecies(label='N_quartet', xyz={'symbols': ('N',), 'isotopes': (14,),
                                   'coords': ((0.0, 0.0, 0.0),)}, multiplicity=4)
with self.assertRaises(SpeciesError):
    ARCSpecies(label='N_bad', xyz={'symbols': ('N',), 'isotopes': (14,),
                                   'coords': ((0.0, 0.0, 0.0),)}, multiplicity=3)

# genuine xyz-only species at init (the existing tests mutate .mol after construction,
# so the xyz branch of get_number_of_electrons never runs during __init__)
h2o = {'symbols': ('O', 'H', 'H'), 'isotopes': (16, 1, 1),
       'coords': ((0.0, 0.0, 0.12), (0.0, 0.76, -0.48), (0.0, -0.76, -0.48))}
with self.assertRaises(SpeciesError):
    ARCSpecies(label='xyz_bad', xyz=h2o, multiplicity=2, compute_thermo=False)

# the restart / species_dict path IS covered, but only because the call site sits inside
# `if species_dict is None or self.yml_path is not None` -- pin it so a refactor can't
# silently drop the check
with self.assertRaises(SpeciesError):
    ARCSpecies(species_dict={'label': 'restart_bad', 'multiplicity': 2, 'charge': 0,
                             'mol': <ethane adjlist>, 'is_ts': False, ...})
```

Also worth asserting on the message (`assertIn('Impossible multiplicity', str(cm.exception))`) — the message is the whole user-facing value of this change.
```

arc/species/species.py
calvinp0: thanks. done now. `arc/common.py` is now the one that owns `count_electrons()` and `is_multiplicity_valid()`
but kept the `sum()` call on the `self.atom` cause delegating it meant convertin mol atoms to symbols and then -revalidating them. but happy to change that too

arc/checks/ts_test.py
calvinp0: So CC with multiplicity 1 wouldn't work, for two reasons. First, P_wrong was never ethane, it was SMILES CC with the C2H3O3 IRC coords, which I think was intentional originally. But because the wrongness came from the impossibility rather than from being a different product, it wasn't really testing "wrong product", and multiplicity 1 with that xyz violates parity anyway.

and if we kept as CC with the xyz_2 coords like in the original then ARC drops CC for the graph but keeps Ethans singlet on the C2H3O3 composition, producing a species that cannot exist and so we get a test for 'impossible species doesn't match' rather than 'wrong species doesn't match'. Its cause of the parity guard we see this issue now.

arc/species/species_test.py
calvinp0: done!

alongd (approved):
Thanks!!

PR #915, opened 2026-07-20
Guard TSGuess clustering against coordinate-less guesses

calvinp0:
A coordinate-less TS guess (e.g. a failed kinbot/queue guess, or a queue TS-search job whose .log produced no parseable geometry) reached almost_equal_coords(None, None) during cluster_tsgs, raising a TypeError that aborted the entire scheduler at init (benchmark reaction_08).

- almost_equal_tsgs: if either guess's get_xyz() is None, return False instead of comparing coordinates. A guess with no geometry can never be a geometric duplicate.
- process_completed_tsg_queue_jobs: when a queue .log yields no usable geometry, mark the guess failed and log a warning instead of silently dropping it, so it is never kept as a clusterable 'successful' guess.

Adds regression tests covering both paths.


LeenFahoum (approved):
looks good

PR #917, opened 2026-07-20
Export tool-neutral ARC results and parser evidence, and fix three Molpro defects

calvinp0:
## What changed

Re-homes the ARC-owned result-production pieces from the historical TCKDB integration work onto clean ARC main, without bringing the TCKDB adapter back into ARC, and fixes three Molpro defects in files this PR already owns.

- expands tool-neutral calculation, scan, thermo, correction, spin, and TS-path provenance in `output.yml`
- adds Cartesian Hessian, rich IRC, GSM energy, constraint, and spin parsers
- emits an atomic `tckdb_evidence.json` sidecar, schema 1.0, bound to output schema 1.1 by document ID
- preserves and transfers queued xTB-GSM node evidence safely
- delegates opt-in uploads to the standalone `tckdb-arc` package, with disabled/missing-package no-op behavior
- keeps all TCKDB enum, indexing, nesting, and payload translation out of ARC
- reads the Molpro version from modern logs, which reported none since 2021
- writes the Molpro command that exists for each open-shell method
- fixes `parse_geometry` for Molpro, which raised on every log and reported Bohr as Angstrom

10 commits, 45 files, no file touched by more than one commit.

## Why

The adapter is moving to its own repository, but ARC must continue to own parsing and publication of the scientific facts generated by ARC jobs. This creates that durable producer/consumer boundary without merging the old `tckdb-imp` branch wholesale.

# The result contract

## `freq_hessian_method`

`output.yml` gains a per-species/TS key stating how the frequency job's Hessian was built: `analytic`, `finite_difference_gradient`, `finite_difference_energy`, or null.

The archive consuming this record resolves its imaginary-mode noise floor from that fact, but records it only from an explicit keyword (Gaussian `Freq=Numer`/`Numerical`/`EnOnly`, ORCA `NumFreq`/`AnFreq`). Gaussian never states its analytic default, so the field was absent on every Gaussian record. ARC, as the producer, knows what it requested and which default applies, and can state it.

The value is derived from ARC's own request — the frequency input deck in preference to the log's echoed route — combined with the documented behaviour of the ESS for that method class. Where the request does not settle the answer, the field is null. It is never inferred from anything weaker:

- **Gaussian**, bare `freq`: analytic for HF (RHF/UHF), all DFT functionals including the documented double hybrids, MP2, CIS and CASSCF; gradient differencing for ROHF, MP3, MP4(SDQ), CCD, CCSD, CID, CISD, QCISD and BD; energy differencing for MP4 (which means MP4(SDTQ)), MP5, CCSD(T), QCISD(T), QCISD(TQ), BD(T) and BD(TQ). Composites, semi-empirical and force-field methods yield null, as does any method not on those lists.
- **ORCA**: `Freq` is documented as an alias of `AnFreq`, so it is analytic for HF and non-double-hybrid DFT without RI-JK, and null otherwise; `NumFreq` is gradient differencing, or energy differencing when combined with `NumGrad`.
- **PySCF**: keyed on multiplicity, matching what the script actually branches on.
- Every other ESS, and a job with neither deck nor log, yield null.

## Provenance for the applied corrections

`output.yml` records the atom-energy and bond-additivity corrections it applied, the level they were keyed to and the `bac_type`, but identified the software only as `arkane_git_commit` — a git HEAD read from `RMG_PATH`. A pip or conda installed RMG-Py has no git repository there, so that key came back empty and the record then named no source at all for the corrections.

A new top-level `arkane_version` sits beside it, mirroring the existing `arc_version` / `arc_git_commit` pair. Both fields resolve from **one** source so they cannot describe different installs: Arkane prints its version and the RMG-Py HEAD in its own log header, which ARC already produces under `calcs/statmech/`, so that is read first, with `rmgpy/version.py` under `RMG_PATH` as the fallback for both together. Logs older than the run's start are ignored, since that directory survives between runs. Either key is `null` rather than absent when nothing resolves.

`importlib.metadata` is not a usable source here: RMG-Py ships no dist-info, so it raises `PackageNotFoundError` even inside a working `rmg_env`.

The git-commit helper no longer catches every exception. A `RMG_PATH` that is set but is not a directory now logs a warning, so a misconfiguration is distinguishable from an ordinary non-git install.

These name the Arkane that ARC invoked. They are not a claim about what it produced: a run that started and then failed reports the same identity as one that completed, and the correction tables can be empty while both fields are populated.

The evidence sidecar's `producer` block carries the same two values beside ARC's own, under the same names the output document uses, so a reader holding only the sidecar can still name the install that produced the corrections it describes. Both are always present and `null` when unknown, following the block's existing convention for `git_commit`.

The level's `year` is carried through, and is part of this provenance rather than incidental to it: Arkane selects its atom-energy correction entry by year, so two levels identical in method and basis resolve to different correction tables when their years differ. A record that publishes the corrections without the key that chose them is incomplete.

## Which parameter block each correction came from

A consumer reading a corrected energy could see that a correction was applied but not which library block supplied it. Each per-species correction record reports `matched_arkane_key`, the Arkane key ARC matched in the section that correction's parameters live in.

The atom-energy and bond-additivity keys are matched independently and can differ, and the difference identifies the refit vintage. Against the current RMG-database — 68 atom-energy keys, 26 Petersson, 17 Melius — three pairs diverge exactly that way, for instance atom energies under `wb97mv` against bond corrections under `wb97mv2023`. The bond-additivity record previously published the atom-energy key, so in those cases it named the wrong block. It now reports its own, and `null` when nothing matched in that section rather than borrowing the other.

One qualification the field's definition carries: the atom-energy key is the model chemistry ARC hands Arkane, so both totals are computed under it. `matched_arkane_key` on the bond-additivity record names the section ARC matched and the key the run-level table was read under, not a second key the numbers were computed from. Both facts are in the document and neither impersonates the other.

The frequency scale factor now reports `freq_scale_factor_key`, the entry it resolved to in ARC's own `freq_scale_factors.yml`. The resolved citation stays alongside it in `freq_scale_factor_source`: every entry in that file carries an explicit integer index into a top-level `sources` mapping, written by the entry's author, so the attribution is a 1:1 pointer rather than an inference from neighbouring prose.

## Versions

`output.yml`'s `schema_version` goes 1.0 to 1.1. That is an earned increment: main emits 1.0 today.

Everything else introduced here starts at 1. The evidence sidecar is schema 1.0, and the parser-version strings are `arc-hessian-1`, `arc-irc-path-1` and `arc-gsm-stringfile-1`. These were bumped several times on unmerged branches while the shapes settled, but no consumer ever received an earlier one, so publishing them as v2 and v3 would distinguish them from nothing.

# Molpro

Three independent defects, all in files this PR already touches.

## Reading the version

`output.yml` records the ESS version per job type, so a version that fails to parse is a silently incomplete record. `arc/parser/adapters/molpro.py::parse_ess_version` matched only a `NAME      : 2015.1.37` header entry. Molpro 2015 emits that block; 2021 and 2022 do not, so ARC recorded no Molpro version at all for any modern run. Across the 11 Molpro fixtures in the repository, 6 returned nothing, including every 2021.x and 2022.x file.

The banner line `Version 2022.3 linked Wed Nov 30 06:40:34 2022` appears in all 11, so it is now the fallback. The `NAME` entry is still preferred where present, because it carries the patch level the banner truncates — `2015.1.37` against `2015.1`. In the 4 fixtures carrying both, the two agree and `NAME` always comes first.

| Molpro release | before | after |
| --- | --- | --- |
| 2015.1.11, 2015.1.37 | parsed | unchanged |
| 2021.2, 2021.3 | none | parsed |
| 2022.3 | none | parsed |

The scan is bounded to the first 200 lines, following the Gaussian adapter's precedent for a header field, but keeps the streaming read rather than Gaussian's `readlines()` so a large log is never loaded whole. The deepest header among the fixtures is line 69.

ORCA needed no change: it parses correctly across 4.1.2, 5.0.4 and 6.0.0.

## The open-shell method name

`arc/job/adapters/molpro.py` prefixed `u` to the method for any unrestricted species, unconditionally:

```python
if not is_restricted(self):
    input_dict['restricted'] = 'u'
```

`u<method>` is a Molpro command for some methods and not for others, so ARC could write an input Molpro has no command for.

A per-method table decides the open-shell command, with a suffix rule for the methods Molpro does not implement open-shell at all.

| requested method | ARC wrote | ARC now writes | basis |
| --- | --- | --- | --- |
| `ccsd-f12`, `ccsd(t)-f12`, `dcsd` | `uccsd(t)-f12` | unchanged | "Open-shell unrestricted UCCSD-F12 approximations ... Restricted open-shell Hartree-Fock (RHF) orbitals are used" |
| `mp2`, `df-mp2`, `mp2-f12`, `df-mp2-f12` | `ump2-f12` | `rmp2-f12` | no `UMP2-F12` command exists; "all methods except `ump2` use spin-restricted Hartree-Fock (RHF) reference functions" |
| anything ending `-f12c` | `uccsd(t)-f12c` | `NotImplementedError` | "Currently CCSD-F12c is not available for open-shell cases" |
| `casscf` | `ucasscf` | `casscf` | "The program is invoked by the command MULTI, MCSCF, CASSCF, or CASCI"; no `UCASSCF` exists |
| anything else | `u<method>` | `u<method>` | see below |

A rewrite is logged at warning level naming both forms, since it changes the command from the one the user wrote. The substitution is the same theory: `RMP2-F12` is open-shell MP2-F12, and the mapping is density-fitting consistent on both rows.

Unrecognised methods keep the `u` prefix deliberately rather than by omission. It is correct for the whole coupled-cluster and configuration-interaction family, and defaulting to no prefix would silently select Molpro's **closed-shell** program — a worse failure than an invalid command.

The refusal keys on the `-f12c` suffix rather than two exact names, so `df-ccsd(t)-f12c` and `mp2-f12c` are covered too. It raises `NotImplementedError` from job construction, matching the existing precedent for Molpro plus IRC in `check_argument_consistency`.

The multireference path is unaffected: `mrci`/`rs2` already clears the prefix and builds its own command. A test fails if that guard is removed.

### Not a change: the `u` prefix on coupled cluster is correct

Worth stating, because it looks like the same class of problem and is not. Molpro's `HF`/`RHF` is the spin-restricted program, and ARC's template hardcodes `{hf; ... wf,spin=N,charge=C;}`, so ARC never requests UHF orbitals. The `u` selects the open-shell correlation ansatz on an RHF reference, giving `RHF-UCCSD(T)-F12`, which the manual requires for open-shell F12. ARC's own checked-in output confirms it: `arc/testing/sp/TS_x118_sp_CCSD(T).out` pairs `wf,spin=2` with `uccsd(t)-f12;` and reports `!RHF-UCCSD(T)-F12a energy`. ARC's default single point, `ccsd(t)-f12/cc-pvtz-f12`, was correct before this change and is unchanged by it.

## Geometry parsing

`parse_geometry` iterated a generator that had already been consumed, so it raised on every Molpro log, and where it did return coordinates they were Molpro's Bohr values labelled as Angstrom. It now iterates a list and converts to Angstrom via `bohr_to_angstrom`.

## Validation

- output, output-schema, evidence and parser suites: 584 passed, 54 subtests
- xTB-GSM, scheduler, species and CLI suites: 539 passed, 205 subtests
- evidence, scripts and statmech suites: 98 passed
- Molpro adapter, parser and level suites: 175 passed, 2 subtests
- standalone consumer parity: 491 passed, 3 skipped, 34 subtests

The `freq_hessian_method` rule table was validated by a quantum-chemistry review against vendor documentation, then by an adversarial code review which mutation-tested it: forcing the rule to always-absolute, always-multiplicative, or with the donor and acceptor partners swapped each fails the tests, so they are not tautological.

A later adversarial review of the Arkane provenance was answered the same way. Removing the schema test's resolver patch, or reverting the append-log guard, each fails a specific test — so both are load-bearing rather than decorative.

https://claude.ai/code/session_01MndwJm4BEs4DN3vJQsEwD3


arc/parser/adapters/gaussian.py
```diff
@@ -491,6 +664,260 @@ def parse_irc_traj(self) -> list[dict[str, tuple]] | None:
             i += 1
         return traj if traj else None
 
+    def parse_irc_path(self) -> list[dict] | None:
```
alongd, line 694: the implementation of this function seems to be duplicated

arc/output.py
```diff
@@ -630,34 +1322,47 @@ def _statmech_to_dict(spc, project_directory: str, point_group: str | None = Non
...
-    for rotor in spc.rotors_dict.values():
+    for rotor_index, rotor in spc.rotors_dict.items():
         if rotor.get('success') is not True:
             continue
         scan = rotor.get('scan')       # 4-atom dihedral defining atoms, 1-indexed
         pivots = rotor.get('pivots')   # 2-atom rotation axis, 1-indexed
         symmetry = rotor.get('symmetry', 1)
         rotor_type = rotor.get('type', 'HinderedRotor')
         treatment = 'free_rotor' if 'Free' in str(rotor_type) else 'hindered_rotor'
+        scan_key = (
```
alongd: CC says:
_get_torsions emits source_scan_key='scan_rotor_N' whenever _scan_log_is_parseable() is true — but that helper only checks os.path.isfile() (1379). The producer _build_rotor_scans emits the matching record only for dimensions==1 (1444) and a successful _build_scan_result_for_rotor (1449). So any 2D/ND rotor, or a 1D rotor whose log exists but fails full parse, yields a torsion referencing a scan_rotor_N record absent from rotor_scans — the exact dangling reference the docstring at 1423-1425 promises won't happen.
Fix: gate the key on the same dimensions==1 + successful-scan-result predicate the producer uses (share one predicate).

ARC.py
```diff
@@ -6,13 +6,52 @@
...
+    if tckdb_settings.get('enabled') is False:
+        return
+    try:
+        import tckdb_arc
+    except ModuleNotFoundError as exc:
+        if exc.name == 'tckdb_arc':
+            _warn_missing_tckdb_package()
+            return
+        raise
+    from tckdb_arc.adapter import TCKDBAdapter
```
alongd, line 40: what's `tckdb_arc`? do we install it in our devscripts? or is it in the env?

arc/output.py
```diff
@@ -363,6 +658,188 @@ def _safe(fn, default=None):
...
+            'level_of_theory': lot_dict,
+            'total': {
+                'value': float(aec_block['value']),
+                'unit': aec_block.get('value_unit', 'hartree'),
+            },
+            'components': aec_block.get('components') or [],
+        }
+        if aec_table:
+            correction['parameter_table'] = {
+                'unit': aec_block.get('value_unit', 'hartree'),
```
alongd: CC says:
parameter_table.unit cross-sourced. Table values come from the AEC/BAC tables (AEC=Hartree, BAC=kJ/mol) but unit is taken from the Arkane total block's value_unit (default kcal/mol for BAC). If the two ever disagree, per-parameter values are mislabeled.
Fix: label with the table source's own unit.

arc/tckdb_evidence.py
```diff
@@ -0,0 +1,400 @@
...
+    output_directory.mkdir(parents=True, exist_ok=True)
+    descriptor, temporary = tempfile.mkstemp(dir=output_directory, suffix=".tckdb_evidence.json.tmp")
+    target = output_directory / EVIDENCE_FILENAME
+    try:
+        with os.fdopen(descriptor, "w", encoding="utf-8") as handle:
+            json.dump(evidence_doc, handle, indent=2, sort_keys=True, allow_nan=False)
+            handle.write("\n")
+            handle.flush()
+            os.fsync(handle.fileno())
+        os.replace(temporary, target)
```
alongd, line 418: parent dir isn't fsynced after os.replace

alongd (commented):
Thanks for this well-though of contribution! I left some minor comments

ARC.py
calvinp0: It's for https://github.com/calvinp0/tckdb-adapters/
I added in the devtools installation now but it's still a WIP. Splitting the adapter from ARC has somewhat been a challenge
https://github.com/ReactionMechanismGenerator/ARC/blob/56800ae93a52f93d20d8a1017b20413a5ca6629d/devtools/install_tckdb_arc.sh

arc/parser/adapters/qchem.py
```diff
@@ -124,6 +124,53 @@ def parse_t1(self) -> float | None:
         # Not implemented for Q-Chem.
         return None
 
+    def parse_s_squared(self) -> dict[str, float | None] | None:
```
alongd, line 127: is this tested?

arc/parser/adapters/orca.py
```diff
@@ -163,6 +168,121 @@ def parse_normal_mode_displacement(self) -> tuple[np.ndarray | None, np.ndarray
...
+        canonical = os.path.join(directory, 'input.hess')
+        if os.path.isfile(canonical):
+            return canonical
+        candidates = sorted(glob.glob(os.path.join(directory, '*.hess')))
+        if len(candidates) == 1:
+            return candidates[0]
+        # Zero (nothing to parse) or ambiguous (>1) — decline rather than guess.
+        return None
+
+    def parse_cartesian_hessian_lower_triangle(self) -> list[float] | None:
```
alongd, line 195: are the new funcs here tested?

Makefile
```diff
@@ -105,6 +107,9 @@ install-torchani:
 install-uma:
 	bash $(DEVTOOLS_DIR)/install_uma.sh
 
+install-tckdb-arc:
```
alongd, line 110: is this included in `make install-all`?

alongd (commented):
Thanks! added some comments, specifically about parser tests and makefile.
How does CI work with this new installation?

arc/parser/adapters/gaussian.py
calvinp0: fixed - extracted `_read_irc_geometry_table`

arc/parser/adapters/qchem.py
calvinp0: yeah, all three: parse_s_squared, ORCA Hessian and parse_orca_constraints

Makefile
calvinp0: it's a work in progress on the other repo side, so i would see this more as a placeholder - i still need to ensure that the adapter will do what it needs but this PR is not reliant on it.

alongd:
# Review — ARC PR #917, "Export tool-neutral ARC results and parser evidence"

| | |
|---|---|
| repo | ReactionMechanismGenerator/ARC |
| PR | [#917](https://github.com/ReactionMechanismGenerator/ARC/pull/917) (calvinp0, open) |
| branch | `feature_arc_result_export_contract` |
| head | `c6afceaed163fe87404d92d0a8556e0bf81aa5d9` |
| base | `e3cff36d748e004387a1dbc7809e7797879dc85c` (merge-base with `main`) |
| diff | 29 files, +5952/−185 — 15 source files across 10 packages |
| date | 2026-08-11 |
| edits | None. Review ran in a detached read-only worktree; the mutation pass used a separate disposable worktree and restored it byte-identically (`git status --short` empty on both). |

HEAD did not move during the review. Every finding below is against `c6afcea`.

## Provenance scale

| Score | Established by |
| --- | --- |
| 10 | Executed. Output pasted into the finding. |
| 8–9 | Read the specific lines and traced the path. |
| 6–7 | Someone read the lines and reasoned; nobody ran the failure. |
| ≤5 | Pattern match; check before acting. |

---

## Fix first

### 1. The evidence sidecar pairs a Hessian with a geometry in a different Cartesian frame — 10

`arc/tckdb_evidence.py:104-114`

`geometry_xyz_text` is `record["xyz"]` — ARC's exported geometry, which comes from
`parse_geometry`, i.e. Gaussian's **standard** orientation. `lower_triangle` is parsed from the
freq log, which Gaussian prints in the **input** orientation. Nothing reconciles them and there
is no `frame` field. The `len(triangle) != dimension*(dimension+1)//2` guard at line 102 catches
a size mismatch, which a rotation never trips.

Reconstructing the spectrum from the exported pair, on ARC's own fixture
`arc/testing/restart/3_restart_bde/calcs/Species/anilino_radical`:

```
Input orientation:       worst deviation vs ESS-reported freqs:     0.00 cm-1
Standard orientation:    worst deviation vs ESS-reported freqs:    95.36 cm-1
```

and against the geometry actually exported (from the opt log):

```
ESS-reported freqs   : ['200.5', '393.9', '428.1', '492.7', '539.5', '624.3', ...]
reconstructed freqs  : ['34.1',  '200.5', '393.9', '428.4', '492.7', '621.3', ...]
worst absolute deviation: 193.43 cm-1
```

The two logs hold the same molecule — max internal distance-matrix difference 0.000001 Å — related
by a pure rigid-body rotation of 4.08 Å in max coordinate displacement.

Consequence: any consumer re-deriving frequencies, ZPE, partition functions, or an Eckart barrier
from `lower_triangle` + `geometry_xyz_text` gets invented low-frequency modes (~30 cm⁻¹, and
low-frequency modes dominate S and Cp) and errors up to 200 cm⁻¹. For a TS the imaginary mode moves
by an order of magnitude (`-1131 → -72` cm⁻¹), destroying the tunneling correction. Nothing crashes.
The projection residual is 0.00 by construction, so a naive consumer self-check will not catch it.

Fix: export the geometry from the Hessian's own source — Gaussian's last `Input orientation:` table,
ORCA's `.hess` `$atoms` block (Bohr → Å) — or rotate the Hessian into the exported frame with the
block-diagonal Kabsch rotation (`H' = R H Rᵀ`). Add an explicit `frame:` field either way. Until
then the Hessian evidence should not ship.

The test that would have caught this is ~20 lines: mass-weight the exported pair, project out
trans/rot, compare to `parse_frequencies()` at <1 cm⁻¹.

### 2. Command injection through `project_directory` — 10

`arc/output.py:730-742`, dispatched via `arc/job/local.py:27`

`_compute_species_corrections` hand-builds a shell string and interpolates `tmp_in`/`tmp_out`
unquoted into a `bash -lc "..."` word, then hands it to `execute_command(..., executable='/bin/bash')`,
which runs `subprocess.run(shell=True)` — two shell layers, and `$(...)` still expands inside the
double quotes. Both temp paths derive from `tempfile.mkstemp(dir=os.path.join(project_directory, 'output'))`.

`project_directory` is unvalidated: `ARC.py:99-100` takes it straight from `input.yml`, and
`arc/main.py:281-283` assigns and `os.makedirs` it with no character check. `check_project_name`
(`arc/main.py:894`) guards only the project *name*.

Reproduced with the command construction copied byte-for-byte from `output.py` and the dispatch
copied from `local.py`, with `project_directory` containing a `$(touch …)` substring:

```
MARKER EXISTS AFTER RUN: True
```

It fires at the end of a normal run, as the invoking user. No ESS job, no server, no privilege
boundary. The same defect degrades silently for any path containing a space: argv splits wrong,
the helper exits non-zero, `except Exception` returns `{}`, and `energy_corrections` is quietly
absent from `output.yml`.

Fix: pass an argv list, or `shlex.quote()` every interpolated value. Better: this function
re-implements the micromamba/conda/mamba ladder that `rmg_env_command()` already provides — the
sibling `_get_energy_corrections` two functions earlier calls it. Fixing it centrally is the right
landing spot (note `rmg_env_command` has the same unquoted-`py_args` weakness).

### 3. `parameter_table` is a table of reference energies, not corrections — 10

`arc/output.py:801-805`, `arc/scripts/get_species_corrections.py:83-116`, `docs/output_yml_schema.md:148,285`

Documented as "the parameter table ARC actually used", `unit: hartree`, sitting beside `total` and
`components`. It is Arkane's `atom_energies` — bare atomic electronic energies that Arkane
**subtracts**, omitting the `atom_hf − atom_thermal` term entirely. Executed in `rmg_env`:

```
atom_energies[C] = -37.84772413025377 Hartree
CH4 total AEC = 40.414403477948255 Hartree = 106108.01060717815 kJ/mol
  comp C x 1 : parameter_value= -37.84772413025377 | contribution_value= 38.11820260724303
  comp H x 4 : parameter_value= -0.4933821648951652 | contribution_value= 2.296200870705227
```

`contribution_value` and `count * parameter_value` differ in **sign** and by ~0.3 Hartree/atom
(~700 kJ/mol/atom). `docs:148` ("Per-atom corrections in Hartree") is flatly wrong. A consumer
reconstructing the correction as `Σ count·parameter_value` — the only reading the field name and
the surrounding shape support — gets the wrong sign and ~10⁵ kJ/mol of error.

The contract also never states that `energy_corrections[atom_energy].total` is a standard-state
atomization reference shift of order +10⁵ kJ/mol rather than a small additive correction. That is
the convention leak the "tool-neutral" label conceals.

Fix: rename to `reference_atom_energies`, state `applied_as: subtracted`, document the basis shift.
Or drop the parameter table and keep only `components[].contribution_value`, the one self-consistent
quantity. (The sign of `total` itself is correct — `arkane/statmech.py:618` adds both corrections.)

### 4. The same BAC dict is emitted twice with contradicting units — 10

`arc/output.py:128-130` and `:766, :826-830`; `docs/output_yml_schema.md:149`

`_get_energy_corrections()` returns one `bac` dict. It is written to
`doc['bond_additivity_corrections']` (documented **kJ/mol**) and then passed as `bac_table=bac` into
`parameter_table` with `_BAC_TABLE_UNIT = 'kcal_mol'`. Identical numbers, two unit declarations, one
file. Executed against RMG: Petersson BAC parameters are kcal/mol
(`arkane/encorr/bac.py:422` → `ScalarQuantity(bac, 'kcal/mol')`; sampled `{'C#C': -0.373, 'C-C': 0.126,
'C-F': 0.539}`). So `kcal_mol` is right and the documented top-level unit is wrong by 4.184×.

A consumer trusting the doc introduces a 1–2 kJ/mol systematic error per bond — ~10 kJ/mol on a C₆
species, an order of magnitude in k at 500 K. The kJ/mol claim predates this PR, but this is where
the data acquired a second conflicting label and where the doc was rewritten.

Related, **8**: the BAC `parameter_table` is matched independently
(`find_best_across_files(..., PBAC_SECTION_START, ...)`, `output.py:609`) while `total`/`components`
are computed from the AEC-matched level string (`output.py:677-680`), and `level_of_theory`
(`output.py:787`) reports the ARC `Level`, not the matched Arkane key. Three numbers in one record
can originate from three different things, and the doc's "The Arkane level of theory the correction
was applied at" is not what the field holds.

---

## Then

### 5. `tunneling: "Eckart"` is asserted unconditionally — 8

`arc/output.py:1789` — `kinetics.get('tunneling') or ARKANE_TUNNELING_METHOD`.

`ARKANE_TUNNELING_METHOD` (`arc/statmech/arkane.py:44`) is the string ARC renders *into* the Arkane
template. It is not evidence that Eckart tunneling was applied to these numbers. Kinetics that did
not come from that template run — user-supplied in the input YAML, restart-restored, or from a
future non-Arkane statmech adapter (the `StatmechAdapter` ABC exists to allow one) — get stamped
anyway. A downstream user, told the correction is applied, does not apply their own, and
permanently under-predicts low-T rates by the tunneling factor (often >10× below 400 K). This is
the archetypal correction-applied-or-to-be-applied failure and it is unfalsifiable from the document.

Settle it by: constructing an `ARCReaction` with kinetics set from a non-Arkane source and checking
the emitted field. Fix: emit `tunneling` only when the value came from the Arkane run that produced
these kinetics; otherwise `null`. A provenance field must never default from a template constant.

### 6. The test that verifies this PR's central claim never runs, and is mocked where it matters — 10

`arc/tckdb_evidence_test.py:313`

`test_cross_repository_golden_contract` — "Producer assembly is exactly the sidecar consumed by
tckdb-adapters" — skips unless a sibling `tckdb-adapters` checkout is adjacent on disk, so it never
runs in CI:

```
SKIPPED [1] arc/tckdb_evidence_test.py:313: tckdb-adapters golden fixture checkout is not adjacent
```

And when it does run, it patches `_build_hessian`, `_build_irc`, and `_build_gsm` with the expected
values (lines 323-332) — the three heaviest and riskiest builders, including the one carrying
finding 1. The test cannot detect a defect in any of them. It also runs against the repo rule
"Avoid using MOCK classes when testing, always prefer real objects."

### 7. Rotor-scan absolute energy indexing is unpinned — 10

`arc/output.py:1743`. Mutation `float(abs_energies[source_index])` → `[source_index - 1]`:

```
170 passed, 14 subtests passed in 7.66s
```

`grep electronic_energy_hartree arc/output_test.py` gives one hit, `output_test.py:1848`, presence-only
(`assertIn`). The sibling `relative_energy_kj_mol` in the same test *is* pinned to an exact value.
An off-by-one would silently attach the wrong point's absolute energy to every exported sample.

### 8. `_parse_gradient` has zero coverage — 10

`arc/tckdb_evidence.py:206-217`. Mutation dropping the `/ len(components)` RMS normalization,
inflating RMS by √N:

```
18 passed, 1 skipped in 0.86s
```

All IRC/gradient coverage in the test file goes through a mocked `parse_irc_path`, bypassing this
function. The sibling `_parse_energy` *is* exercised (mutating its `split()[1]` → `[0]` was caught),
so this is a specific gap, not a test-infra limitation. Any bug here ships silently into
`max_gradient_hartree_per_bohr` / `rms_gradient_hartree_per_bohr`.

Related, **7**: `_parse_gradient` derives `atom_count = int((len(lines) - 3) / 2)` and slices
positionally, so any Turbomole `gradient` file with more than one `cycle` block yields a
wrong-but-parseable slice — only `ValueError/IndexError/OSError/ZeroDivisionError` are caught.

### 9. `cp_data` → `thermo_points` renamed under a bump documented as additive — 10

`arc/output.py:1259`; `docs/output_yml_schema.md:112` reads "`1.1` adds the optional evidence
descriptor". Confirmed against both commits: `cp_data` exists at `e3cff36` (`output.py:561-570`,
`species.py:2692`, `docs:163`), is absent at head, with no alias anywhere in the tree. `schema_version`
does bump 1.0 → 1.1, which is more migration discipline than most of this codebase has — but a
rename is not a minor additive change, and the changelog line doesn't mention it.

**Correction, from the data-migration pass:** this was initially flagged as a restart hazard. It is
not. `output.yml` is write-only — zero readers in-repo — and `ThermoData` never enters
`ARCSpecies.as_dict()`/`from_dict()`. The `arc/statmech/arkane.py:522` read is from `thermo.yaml`,
regenerated every run and consumed within the same run. External consumers of `output.yml` are the
only ones affected.

### 10. Schema drift on TS frequency fields — 10

`output.py:1033,1038` emits `imaginary_frequencies_cm1`; the field does not appear anywhere in
`docs/output_yml_schema.md` (grep). Separately, `docs:379` declares a converged TS has
`freq_n_imag: 1` while `output.py:1036` emits `len(imaginary_freqs)` — a TS with frequencies
`[-1500.0, -120.0, 80.0]` emits `2`, contradicting the documented contract.

Found independently by Codex and re-derived here by grep against both files.

### 11. The installer pulls unpinned code from a personal account into ARC's own environment — 9

`devtools/install_tckdb_arc.sh:24,27,99` — `pip install "git+https://github.com/calvinp0/tckdb-adapters.git@main#subdirectory=tckdb_arc"`.

Three compounding properties: unpinned (`@main`, no tag, no SHA, no hash), a personal account rather
than `ReactionMechanismGenerator`, and deliberately installed into `arc_env` — which `ARC.py:38` then
imports in-process, alongside `~/.arc/settings.py` server definitions, SSH key paths, and live SFTP
sessions. pip executes the repo's build backend at install time.

Mitigating and worth saying: not wired into `install_all.sh`, opt-in via a separate Make target,
`set -eo pipefail` present, no `sudo`, no writes outside the conda env, no `curl|bash`. This is a
supply-chain design objection, not an active hole. Gate before landing: pin a SHA or tag, and move
the repo under the org. Minor in the same file: no `set -u`, so `--env` with a missing value yields
`conda run -n  python`; `$ENV_RUNNER` is unquoted at lines 68/77/82/96/99.

### 12. Upload is default-on with an unvalidated destination — 8

`ARC.py:29-52, 101-105` — `if tckdb_settings.get('enabled') is False: return`. The check defaults to
allow: any `tckdb:` block in `input.yml` that doesn't explicitly say `enabled: false` uploads the
run's full scientific record, to whatever endpoint the block names, handed verbatim to
`TCKDBConfig.from_dict`. No allowlist, no confirmation, no log line naming the target before data
leaves the machine. The only warning that fires (`_warn_missing_tckdb_package`) fires when the
package is *absent*; the success path is silent.

Settle it by: installing the adapter and running with a `tckdb:` block pointing at a local listener.
Fix: require `enabled: true` explicitly, and log the resolved destination host at INFO.

**Good news on the payload**, traced and clean: `arc/tckdb_evidence.py` embeds only geometries,
Hessian triangles, IRC/GSM points, and `source_log` strings; all paths flow through `_make_rel_path`
and are project-relative; no module in the diff makes a network call; `test_paths_are_relative_and_
operational_fields_do_not_leak` asserts it.

---

## Smaller

- **`arc/parser/adapters/orca.py:236-243` — 10.** `n_rows` comes from the `.hess` file with a `< 6`
  floor, no ceiling, and no cross-check against atom count, before two O(n²) allocations. A 17-byte
  file gives `MemoryError: Unable to allocate 29.1 TiB for an array with shape (2000000, 2000000)`.
  Contained today by the sole caller's `except Exception`, but this is a newly-public parser method.
  Validate against `3 * atom_count`, which the evidence builder already has. The Gaussian sibling
  (`gaussian.py:288`) derives `n_rows` from observed indices and is not vulnerable.
- **GSM node evidence — 7.** `arc/tckdb_evidence.py:228-257,305-316` assumes the `ograd` invocation id
  equals the final `stringfile.xyz0000` node index. GSM re-evaluates gradients many times per node
  while growing the string and `cp -p` overwrites, so each file holds the last evaluation under that
  id; an offset attaches node *k*'s energy to node *k±1*. Separately `path_coordinate_angstrom`
  (line 299) is a running sum of consecutive Kabsch RMSDs — not a Cartesian arc length (`√N · RMSD`),
  not mass-weighted, and not commensurable with the IRC record's `reaction_coordinate_sqrt_amu_bohr`
  sitting in the same schema. Two different "reaction coordinates" under one contract, undefined.
- **`arc/output.py:1716,1731-1733` — 8.** `requested_end` is computed from `len(rel_energies)`, i.e.
  *completed* points, while the comment directly above says "describe the requested grid, not the
  completed-point spacing" and `parse_scan_args` returns `step_count`, which is read for `step_size`
  and discarded. A Gaussian-truncated scan reports its truncated span as the request.
- **`arc/testing/freq/orca_hessian_h2o/input.hess` — 10.** The fixture diagonal is
  `0.55, 0.50, 0.48, 0.30, 0.28, 0.26, 0.31, 0.29, 0.27` — hand-written round numbers satisfying
  neither translational invariance nor the six-zero-eigenvalue condition, paired with a 10-line
  `output.out` stub reporting one `0.00 cm**-1` frequency. Assertions are `len(triangle) == 45` and
  `all(0.1 < v < 1.0 for v in diag)`. A transpose, a mis-paged column block, a frame mismatch, or a
  wrong ORCA unit all pass. This is where the tests agree with the misunderstanding.
- **`arc/job/adapters/ts/xtb_gsm.py:353-372` — 8.** Credit: `_extract_gsm_evidence_archive` is the
  best-defended new code in the PR — normpaths members, rejects absolute paths, `..` escapes,
  symlinks, hardlinks, and non-file/dir members, with a two-entry name allowlist. No traversal bypass
  found. Missing: any cap on member count or uncompressed size (decompression bomb), and
  `extractall` without `filter='data'`, which is a `DeprecationWarning` on 3.12 and a silent
  semantic change on 3.14.
- **`arc/output.py:1300` — 7.** `harmonic_frequencies_cm1` is `spc.freqs`, unscaled, while
  `statmech.e0_kj_mol` has `freq_scale_factor` applied. The doc says "as parsed" but never
  "unscaled"; a consumer recomputing ZPE will not reproduce `e0_kj_mol`.
- **`arc/output.py:1750` — 8.** `'relaxed': True` is hardcoded for every rotor scan, but ARC supports
  rigid directed scans (`'brute_force_sp'`, `scheduler.py:807`). For those it is a false statement.
- **`arc/tckdb_evidence.py:308` — 9.** `"node_label": index if index else None` gives node 0 a `null`
  label. Falsy-vs-`None`.
- **Version literals — 8.** `output.py:103` (`'1.1'`), `tckdb_evidence.py:369`
  (`"output_schema_version": "1.1"`), `tckdb_evidence.py:28` (`EVIDENCE_SCHEMA_VERSION = "1.0"`). The
  sidecar restates output.yml's version as an independent literal instead of importing it; they will
  drift on the first bump and a consumer gating on `output_schema_version` will then mis-gate.
- **`arc/scripts/get_species_corrections.py:79` — 10.** `_lot_from_string` is a lossy regex
  round-trip: `args=('scf=xqc',)` is silently dropped. Harmless today because AEC keys carry no
  `args`, but it makes the reconstructed level not provably equal to the matched key.
- **Repo style rules — 10.** Six inner-function imports in `arc/output.py` (492, 499, 669, 1523, 1526,
  1529) against CLAUDE.md's explicit "Don't import inside functions". `write_output_yml` and ~20 other
  new functions missing type hints on parameters and/or return. 14 functions in `tckdb_evidence.py`
  without docstrings (`_run_relative`, `_resolve`, `_canonical_xyz`, `_unavailable`, `_finite_json`,
  `_build_hessian`, `_direction`, `_build_irc`, `_parse_energy`, `_parse_gradient`, `_parse_xtbout`,
  `_node_outputs`, `_path_search_method`, `_build_gsm`). `tckdb_evidence.py:342` is 130 chars.
  `_run_relative` (line 39) is defined and never called — `grep` returns only its definition — which
  suggests an intended path-sanitizing guard got dropped.

## Scope

The `xtb_gsm` / `ograd` rider is defensible on its merits: the GSM evidence cannot exist without
preserving node outputs. But it changes the **production submit path** for every GSM job — `gsm.orca`
becomes a bash trap wrapper and the real binary moves to `gsm.orca.bin` — to serve an optional export.
That is a live-jobs regression risk riding in an export PR; it deserves its own commit boundary and
functional verification on a real queue. Ordering is at least correct: `write_input_file()`
(`xtb_gsm.py:275`) runs before `_write_queue_gsm_wrapper()` (`:281`), so the wrapper is not clobbered.

---

## Checked and found sound

Don't re-tread these — each was a live theory that died.

- **`output.yml` breaks campaign restart.** It doesn't. `output.yml` is write-only (zero readers
  in-repo), built entirely in memory and written once after `scheduler.execute()` returns, wrapped in
  try/except, atomic via `tempfile.mkstemp` + `os.replace` (`output.py:200-210`). `restart.yml` is a
  separate path. `TSGuess.from_dict` fed a base-shaped dict without `method_source_paths`
  reconstructs it from `log_path`/`method` without crashing; `_parse_scan_constraints` on a rotor dict
  without `scan_software` returns `[]`; the new `paths` keys (`irc_directions`, `gsm`) are read with
  `.get()` and written with `.setdefault()`, so an old `restart.yml` cannot `KeyError`.
- **AEC/BAC signs are inverted relative to what Arkane applies.** No — `arkane/statmech.py:618` is
  `e_electronic + atom_corrections + bond_corrections`, and ARC records the added value.
- **Hessian units or packing are wrong / Arkane's SI factor leaked in.** No — unprojected
  mass-weighted eigenvalues of the parsed triangle reproduce the log frequencies to 0.01 cm⁻¹ across
  six fixtures. Only the frame is wrong (finding 1).
- **ARC never sets `IOp(7/33=1)`, so the Gaussian Hessian export is a permanent no-op.** No —
  `arc/job/adapters/gaussian.py:317,320` set it for both `freq` and `optfreq`.
- **`_lot_from_string` flattens a `CompositeLevelOfTheory`, so corrections are computed at the wrong
  level.** No — 0 of 66 `atom_energies` keys are composite, so the flattening path is unreachable for
  AEC.
- **`_build_gsm`'s `int((len(frames)-1)/2)+1` invents a midpoint TS node.** It mirrors ARC's own
  pre-existing choice at `xtb_gsm.py:474`. Scientifically debatable, not introduced here.
- **`_write_queue_gsm_wrapper` is clobbered by `write_input_file`'s `safe_copy_file`.** No — ordering
  is correct.
- **Composite (CBS-QB3) runs export a Hessian from the opt+freq log.** No — `Scheduler.run_freq_job`
  (`scheduler.py:1435-1454`) spawns a separate freq job for composite runs. The frame mismatch turned
  out to be general to plain opt → freq pairs, which is worse.
- **`arc/parser/adapter.py`'s interface change breaks untouched adapters.** No — `parse_1d_scan_energies_hartree`
  and `parse_s_squared` are concrete base-class methods with `None`/`(None, None)` defaults, so all
  nine adapters (cfour, molpro, psi_4, qchem, terachem, xtb, yaml, gaussian, orca) still satisfy it.
- **Multi-species freq jobs export one species' geometry against the supermolecule's Hessian.** No —
  guarded at `tckdb_evidence.py:102`; degrades to `parse_failed`.
- **The new scheduler code runs per polling tick.** No — `scheduler.py`'s diff is per-job-callback
  bookkeeping; `write_output_yml` and the evidence builder are one-shot at end of campaign. Nothing
  here compounds over a multi-day run.
- **`freq_scale_factor_source` mishandles `data/freq_scale_factors.yml`'s non-uniform keys.** Not
  attributable — `_resolve_freq_scale_factor_source` (`output.py:265`) is unchanged by this PR. The
  hazard is real but pre-existing.
- **`thermo_points` and `h298_kj_mol` disagree at 298 K.** Unfalsifiable, not a bug — `_CP_TEMPS`
  excludes 298.15 so they never overlap. Which also means the thermo block has no internal cross-check.
- **Transposing the ORCA Hessian assignment (`matrix[row,col]` → `[col,row]`) survives the tests.**
  True but not a defect: the Hessian is symmetric and the parser's symmetric-fallback lookup masks a
  pure transpose.
- **Schema declared vs emitted, field by field.** Every documented key — `schema_version`,
  `tckdb_evidence`, `composite_method`, `neb_level` omission semantics, conformers omission,
  `energy_corrections`, `rotor_scans`, `sp_spin_diagnostic`, opt settings, `thermo_points`,
  `tunneling`, `long_kinetic_description`, `ts_guesses`, `gsm_log`, `irc_log_directions` — traces to
  its emission site with matching nullable-vs-omitted semantics. The only drift found is finding 10.
- **`get_species_corrections.py` ↔ `output.py` subprocess contract.** Input/output YAML shapes match
  on both sides.
- **`ARC.py::run_tckdb_upload` hiding sub-dependency import errors.** Deliberate, and pinned by
  `test_internal_dependency_import_error_is_not_hidden`.
- **Performance.** Nothing above informational. Worst measured: `parse_1d_scan_full_result`
  (`parser.py:321`) re-reads its log three times despite a docstring claiming it saves callers from
  exactly that — 0.070 s/rotor on a 5.6 MB scan log, one-shot at end of run. Gaussian's Hessian
  assembly uses a dict-of-tuples where ORCA uses dense numpy; 0.31 s at 400 atoms, irrelevant at
  kinetics-scale molecules.
- **Payload hygiene.** Grepped every added line across the diff for hostnames, usernames,
  `/home/`|`/Users/` paths, IPs, cluster names, credentials. One external literal in the whole diff:
  the `REPO_URL` in finding 11. No `~/.arc/` writes.

---

## What roots the provisional findings

Three causes generate most of the list. Fix them and the rest stop being independent problems:

1. **The contract has no mechanical definition.** It exists only as a 426-line prose table in
   `docs/output_yml_schema.md`. There is no JSON Schema, no round-trip validation test, and no stated
   policy for what a consumer does on `1.2` versus `2.0`. Drift between doc and code is therefore
   undetectable — and has already happened twice (findings 4 and 10) before the contract has a single
   external consumer. Ship a JSON Schema in-repo, validate `output.yml` against it in a test, and
   derive both version strings from one constant. That closes 4, 9, 10, and the version-literal item.

2. **Nothing checks a physical invariant.** Every new test asserts shape — key present, length 45,
   value between 0.1 and 1.0 — and none asserts that the exported numbers reproduce something known.
   One test that mass-weights the exported Hessian against the exported geometry and matches
   `parse_frequencies()` would have caught finding 1, would catch a transpose, a unit slip, and a
   mis-paged column block, and would make the fabricated ORCA fixture impossible to keep. This is the
   single highest-value change in the list.

3. **Provenance fields are defaulted from constants rather than observed.** Finding 5 stamps
   `Eckart` from a template constant; finding 3 labels reference energies as corrections; finding 8's
   related item labels a rotor `relaxed: True` unconditionally; the BAC `level_of_theory` reports the
   ARC `Level` rather than the Arkane key actually matched. In an evidence document, a field that
   cannot be false carries no information — and worse, it reads as a measurement. Every provenance
   field should be `null` when unobserved.

## Thin coverage

- **Never mutated, only read:** `arc/scripts/get_species_corrections.py`, `arc/output.py::_get_energy_corrections`
  (AEC/BAC lookup, ~578-591), the `E_h_kJmol` division sites at `output.py:347` and `:969`,
  `arc/parser/adapters/qchem.py`, `arc/parser/constraints_test.py`. "No finding" there means
  unexamined, not clean.
- **Never executed:** the TCKDB upload path in `ARC.py` — the `tckdb_arc` package is not installed in
  this environment, so findings 11 and 12 are read-only.
- **Not run:** `functional/` (out of scope), and no ESS job was submitted to any server, per the
  repo's guardrails.
- **The golden-contract test** was skipped in every run of this review, for the same reason it skips
  in CI (finding 6).

---

## Method

Eight passes, dispatched in parallel so none read another's conclusions: testing/mutation,
security, api-contract, maintainability, performance, data-migration, red team, and a cross-model
adversarial pass (Codex). Tier T3 by breadth — 15 source files across 10 packages — which unlocked
the red team; the security pass was force-added anyway by the new shell installer and the Makefile
change.

Findings were deduplicated by file:line and claim. Where two passes converged independently, it is
noted in place: finding 10 (Codex, then re-derived by grep here), finding 9 (Codex flagged the
rename; data-migration independently found and then *corrected* its severity by proving `output.yml`
is not the restart file), finding 6 (api-contract found the skip; verified here and sharpened — the
test also mocks the three builders it claims to verify), and the `rmg_env_command` duplication
(security and red team, from opposite directions).

Findings 1 and 2 were re-derived by the reviewing session before being reported, not adopted on the
specialist's confidence. For finding 1 the decisive control — reconstructing the spectrum from the
input orientation versus the standard orientation against the same Hessian — was run here and is the
0.00 / 95.36 cm⁻¹ pair quoted above.

Environment for every gate: `conda activate arc_env` (interactive; `conda run -n arc_env` mis-sets
`BABEL_LIBDIR` and breaks OpenBabel), `HOME` overridden to an empty directory so `~/.arc/settings.py`
could not influence results, `PYTHONPATH` set to the review worktree, `pytest -n0`. Compiled Cython
`.so` files were copied in from the primary checkout, without which `import arc` fails in a fresh
worktree. Baseline over every touched module: **782 passed, 1 skipped, 0 failed**.

calvinp0:
> GSM node evidence — 7. arc/tckdb_evidence.py:228-257,305-316 assumes the ograd invocation id
equals the final stringfile.xyz0000 node index. GSM re-evaluates gradients many times per node
while growing the string and cp -p overwrites, so each file holds the last evaluation under that
id; an offset attaches node k's energy to node k±1. Separately path_coordinate_angstrom
(line 299) is a running sum of consecutive Kabsch RMSDs — not a Cartesian arc length (√N · RMSD),
not mass-weighted, and not commensurable with the IRC record's reaction_coordinate_sqrt_amu_bohr
sitting in the same schema. Two different "reaction coordinates" under one contract, undefined.

Kabsch returns RSSD, not RMSD (https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.align_vectors.html). So the $\sqrt{N}$ claim isn't correct. THe GSM coords are already a cartesion path lenght. But yes, there was a defect. `kabsch` superposes at the mass centroid, so the value is isotope sentive - renamed to `cumulative_com_superposed_displacement_angstromg)

Also, with the shifting, good catch but it turns out it was worse. Appears that GSM showed the IDs arent frame numbers at all, they're slots in an internal working array. 

So the force calculation files contain the usefule stuff: the molecule coords at the moment the calc ran. So will look at the geom and which fram it is instead.

> 1. The evidence sidecar pairs a Hessian with a geometry in a different Cartesian frame — 10 arc/tckdb_evidence.py:104-114 geometry_xyz_text is record["xyz"] — ARC's exported geometry, which comes from parse_geometry, i.e. Gaussian's standard orientation. lower_triangle is parsed from the freq log, which Gaussian prints in the input orientation. Nothing reconciles them and there is no frame field. The len(triangle) != dimension*(dimension+1)//2 guard at line 102 catches a size mismatch, which a rotation never trips.
Reconstructing the spectrum from the exported pair, on ARC's own fixture
arc/testing/restart/3_restart_bde/calcs/Species/anilino_radical:
Input orientation:       worst deviation vs ESS-reported freqs:     0.00 cm-1
Standard orientation:    worst deviation vs ESS-reported freqs:    95.36 cm-1 
and against the geometry actually exported (from the opt log):
ESS-reported freqs   : ['200.5', '393.9', '428.1', '492.7', '539.5', '624.3', ...]
reconstructed freqs  : ['34.1',  '200.5', '393.9', '428.4', '492.7', '621.3', ...] worst absolute deviation: 193.43 cm-1
The two logs hold the same molecule — max internal distance-matrix difference 0.000001 Å — related
by a pure rigid-body rotation of 4.08 Å in max coordinate displacement.
Consequence: any consumer re-deriving frequencies, ZPE, partition functions, or an Eckart barrier
from lower_triangle + geometry_xyz_text gets invented low-frequency modes (~30 cm⁻¹, and
low-frequency modes dominate S and Cp) and errors up to 200 cm⁻¹. For a TS the imaginary mode moves
by an order of magnitude (-1131 → -72 cm⁻¹), destroying the tunneling correction. Nothing crashes.
The projection residual is 0.00 by construction, so a naive consumer self-check will not catch it.
Fix: export the geometry from the Hessian's own source — Gaussian's last Input orientation: table,
ORCA's .hess $atoms block (Bohr → Å) — or rotate the Hessian into the exported frame with the
block-diagonal Kabsch rotation (H' = R H Rᵀ). Add an explicit frame: field either way. Until
then the Hessian evidence should not ship.
The test that would have caught this is ~20 lines: mass-weight the exported pair, project out trans/rot, compare to parse_frequencies() at <1 cm⁻¹.

Fixed this. In TCKDB, it has its own Gaussian parser which skips the standard orientation matreicies exactly for this reason,  and it would have rejected the old records at FRAME_CONSISTENCY_TOLERANCE_CM1 = 100.0.s

alongd:
Re-reviewed at `1f84d56` against base `af4740f8`. **The review is satisfied** — all 20 findings resolved, and the two P0s are fixed in a way that's provably load-bearing rather than just changed.

Verified end-to-end by building the actual emitted evidence record and reconstructing the spectrum from `geometry_xyz_text` + `lower_triangle`:

```
Gaussian (anilino_radical freq_a4345):  frame=gaussian_input_orientation  worst deviation 0.0020 cm-1
ORCA     (orca_hessian_h2o):            frame=orca_hess_atoms            worst deviation 0.0047 cm-1
```

against 95.36 / 193.43 cm⁻¹ before. It also fails closed — renaming the `Input orientation:` header yields `hessian_frame_unavailable` rather than silently falling back to the standard orientation.

Mutation testing confirms the new tests would catch a regression. Reintroducing the exact original defect (`'Input orientation:'` → `'Standard orientation:'`) fails three independent tests, one of them by **95.36 cm⁻¹** against a `< 1.0` tolerance — the precise number from the original review. Corrupting a Hessian element, the Bohr→Å factor, the gradient RMS normalization, and the rotor energy index were all caught too. **554 passed, 0 skipped** across the touched modules; the golden-contract test that never ran in CI now runs unconditionally against real staged fixtures.

Special credit for `arc/schemas/output_yml_schema.json` — 25 `additionalProperties: false`, 35 `required` blocks, correct `unevaluatedProperties` handling for the `allOf` composition, and a meta-test proving the closure constraints can actually fail. That answers the root cause, not just the symptom.

---

## One thing to fix before merge

**`arc/parser/adapters/gaussian.py:325` — the Hessian geometry is selected independently of the Hessian.**

`parse_cartesian_hessian_lower_triangle` takes the **last** `Force constants in Cartesian coordinates:` block (`:255`). `parse_cartesian_hessian_geometry` scans backward from **EOF** for `Input orientation:`. Nothing ties the two. The docstring promises "the frame that actually matches the Hessian" and the returned label `gaussian_input_orientation` asserts a guarantee the implementation doesn't provide.

Ten logs in ARC's own test corpus already pair the wrong way:

```
arc/testing/rotor_scans/scan_2D_N3H5.out
  FC blocks=1 IO blocks=85  | picked line 43204 | correct line 187 | max|picked-correct| = 2.5875 Angstrom
arc/testing/rotor_scans/CH2OOH.out
  FC blocks=1 IO blocks=238 | picked line 98772 | correct line 177 | max|picked-correct| = 1.5080 Angstrom
arc/testing/restart/1_restart_thermo/calcs/scan_a19049.out
  FC blocks=1 IO blocks=198 | picked line 99379 | correct line 184 | max|picked-correct| = 1.8864 Angstrom
```

2.59 Å is not a rotation — it's a different point on the PES.

This is **latent, not live**: the sole production caller (`arc/tckdb_evidence.py:185`) is fed `record["freq_log"]`, always a dedicated freq job with a single FC block. But `parse_cartesian_hessian_geometry` is now public API on the adapter base class (`arc/parser/adapter.py:233`), and any caller passing an `opt=(ts,calcfc)` log — FC computed at step 1, optimization continuing after — gets the converged geometry paired with the step-1 Hessian. `_build_hessian`'s two guards (atom count, triangle length) both still pass, because neither is sensitive to *which* geometry it got.

Suggested fix, ~5 lines: find the FC block index first, then search backward from *that index* for `Input orientation:`. Better still, return the pair from one method so they cannot drift. A regression test is cheap — `arc/testing/rotor_scans/CH2OOH.out` already exhibits it.

## Follow-ups — not blockers

- **`arc/testing/freq/orca_hessian_h2o/` is synthetic**, which makes the ORCA leg of the Hessian test self-consistent rather than independent. Three tells: the geometry is the textbook experimental water structure to 10 significant figures (r = 0.9578 Å, θ = 104.51° exactly); `$atoms` carries exact isotope masses (`15.99491`/`1.00783`) where real ORCA writes averaged ones — compare `arc/testing/freq/orca_example_freq.log`, which has `15.999`/`1.008`; and `output.out` has an `ORCA SCF GRADIENT CALCULATION` banner over what should be a freq log. The premise being tested is correct (I verified `$atoms` really is Bohr against the real log), and this fixture is a large improvement on the hand-written one it replaced — but a real 3-atom `.hess`/`.out` pair would make the test independent. At minimum, a comment saying it's synthetic.
- **`arc/scripts/arkane_levels.py` has no tests.** It's new, load-bearing, and exactness-critical — a wrong reconstruction silently looks up a different level's parameters. It's a genuine improvement on the regex it replaced (113/113 Arkane DB keys round-trip, 0 declines, including nested `CompositeLevelOfTheory` keys the regex would have flattened), which is exactly why it deserves a test pinning that. Also missing type hints on both signatures.
- **The schema pins correction units hard but leaves the rate coefficient's free.** `parameter_unit` is `const`-pinned to `hartree`/`kcal_mol` with a comment explaining that a kJ-for-kcal substitution must fail validation — but `A_units`, `Ea_units`, `dEa_units` and `tunneling` are `{"type": ["string", "null"]}`. Those are regex-scraped from Arkane output with a silent default; if Arkane's format shifts, a kcal number labelled `kJ/mol` validates clean. Enums would close it. (The scraping predates this PR.)
- **Both Hessian geometry parsers can return silently truncated geometries** — `gaussian.py:329-338` breaks out of the row loop on a malformed line leaving already-collected atoms intact; `orca.py:326-346` truncates when the `$atoms` count is smaller than the real list. Currently masked because `_build_hessian` derives `atom_count` from the geometry and the triangle-length check fires, but the guard lives in the caller rather than the parser.
- **`arc/parser/adapters/hessian_test.py:81`** projects 6 rigid DOF unconditionally while its own docstring says "six (five for a linear species)". On a linear molecule the axial-rotation column is identically zero, `qr` fills it arbitrarily, and a real vibration gets projected out — ~3309 cm⁻¹ on the NH log. Not triggered by the current fixtures, but it will read as a parser bug when it fires.
- Minor: `arc/scripts/get_qm_corrections.py:42` `import re` is now unused. `arc/tckdb_evidence.py:544` duplicates the TS-node selection expression from `xtb_gsm.py:494` verbatim — they agree today, but a shared constant would keep `selected_source_point_index` honest. `orca.py` reads the `$atoms` mass column and discards it, so a deuterated run's Hessian can't be reproduced from the stored evidence.

## Corrections to my original review

**You were right about `kabsch`, and I was wrong.** `arc/species/converter.py:2657` returns scipy's second value, which is RSSD:

```
n=  4  scipy=0.302349  rssd=0.302349  rmsd=0.151174  score/rmsd=2.0000  sqrt(n)=2.0000
n=256  scipy=2.854113  rssd=2.854113  rmsd=0.178382  score/rmsd=16.0000  sqrt(n)=16.0000
```

RSSD already carries the √N, so summing consecutive values is dimensionally a Cartesian displacement in Å. My "the arc length would be √N · RMSD" presumed the function returned an RMSD needing that factor — it doesn't, and the number never needed rescaling. The COM-superposition caveat you raised is the real defect, and `cumulative_com_superposed_displacement_angstrom` plus the documented isotope-sensitivity and upper-bound behaviour is a better statement of the field than either of us started with.

Worth noting what seeded my error: `arc/species/species.py:2352` and `:2360` both call this "the Kabsch RMSD" in the docstring, and `converter.py:2641` calls it a "similarity score". It returns RSSD. Pre-existing and outside this PR, but it's a live trap for the next reader — a one-line docstring fix somewhere would be worth it.

One other correction: I flagged that an unparseable level-of-theory key would leave `matched_arkane_key` asserted against null corrections. That's wrong — `arc/output.py:874` guards on `if aec_block and aec_block.get('value') is not None:`, so no correction record is emitted at all. The residual is only that the null outcome is silent (exit 0, empty stderr).

## Out of scope — belongs to `main`

Three defects surfaced that are **not** this PR's. `arc/settings/external_paths.py` and `arc/settings/settings.py` arrived via the rebase from #887, and `git diff af4740f8 1f84d56 -- arc/job/env_run.py` is empty:

1. `rmg_env_command` interpolates `cwd` and `env_vars` with bare double quotes (`f'cd "{cwd}"'`), which don't stop command substitution. Reachable from `arc/statmech/arkane.py` with `cwd=statmech_dir`, which descends from the input YAML's project directory — the same vector as F2. Nothing in the suite pins the quoting.
2. An explicit `ARC_*_REPO`/`ARC_*_CKPT` override that's invalid falls through to the next candidate instead of being authoritative, so a mistyped path silently runs a different checkpoint.
3. Overriding `GOFLOW_REPO_PATH` in `~/.arc/settings.py` doesn't update the eagerly-derived ckpt/feat-dict paths, so repo A's configs run against repo B's weights.

I'm opening a separate PR for these three against `main` — they shouldn't ride on this one.

arc/scripts/get_species_corrections.py
```diff
@@ -0,0 +1,233 @@
...
+import sys
+import traceback
+
+import rmgpy.constants as constants
+
+from arkane.encorr.corr import get_atom_correction, get_bac
+from arkane.encorr.data import atom_energies, pbac
+
+from arkane_levels import lot_from_string
+from common import read_yaml_file, save_yaml_file
```
alongd, line 70: this imports from ARC? does this imply that we have both Arkane (RMG) and ARC in the same env?

calvinp0: https://github.com/ReactionMechanismGenerator/ARC/pull/917/changes#diff-49cd9f82811313427e88378beec4688ae3e4e4b2e00015ea38f4776683137f1dR1445

it's a script so should be subprocess called with rmg env

alongd (commented):
Another review pass at `ddc3081` (testing/mutation, schema contract, security, plus a cross-model adversarial pass). The test suite held up well: 9 targeted mutations (Molpro fixes, `_get_torsions` scan-key guard, Hessian source flag, parent-dir fsync, PBAC gate, ORCA Hessian values) were all caught, and the changed modules pass (655 tests). Previous-round points (duplicate doc rows, AEC/BAC unit sourcing, parent-dir fsync) are resolved.

Points to fix before merge. Each of these was reproduced by running the code:

1. **Broken test for the `cp_data` → `thermo_points` rename.** `arc/scripts/save_arkane_thermo_test.py:156` still reads `content['P1']['cp_data']`. The file isn't touched by this PR. Running it under `rmg_env` gives `KeyError: 'cp_data'`. CI doesn't catch it because this test skips in `arc_env` and no workflow runs it in `rmg_env`.

2. **Gaussian constraint parser loses the target of the constraints ARC itself writes.** `arc/job/adapters/gaussian.py:393` emits two lines per constraint: `B 1 2 =1.45 B`, then `B 1 2 F`. `_parse_gaussian_constraint_line` returns `None` for the first line (`B` action). It returns `target_value: None` for the second:
   ```
   'B 1 2 =1.45 B' -> None
   'B 1 2 F'       -> {'coordinate_type': 'distance', 'atom_indices': [1, 2], 'target_value': None, ...}
   ```
   Carrying the value from a preceding `=x B` line on the same atoms into the matching `F` would fix it. Please add a test that feeds the parser exactly what the adapter writes.

3. **IRC reaction coordinate sign.** `docs/output_yml_schema.md` (~L250) says `reaction_coordinate_sqrt_amu_bohr` is "signed about the TS". `parse_irc_path` copies Gaussian's cumulative value unchanged, though, so the reverse branch is positive too: `rxn_1_irc_2.out` gives `('reverse', 1, 0.07236), ('reverse', 2, 0.14469), ...`. Either negate for `direction == 'reverse'` or change the doc to say it's unsigned and to use `direction`.

Smaller points:

4. **TCKDB URL logged verbatim** (`ARC.py` ~L64). If the URL carries userinfo (`https://user:token@host`), it lands in `arc.log`. Logging only the hostname would avoid that.
5. **Where does `tckdb_arc` come from?** Commit d5353ca21's message mentions `devtools/install_tckdb_arc.sh` and a `make install-tckdb-arc` target, but neither is in the commit. The warning text points users to a "tckdb-arc package", which invites a guessed `pip install tckdb-arc`. Please name a pinned, canonical install source, or say explicitly that it's not on PyPI. Also, the adapter call shape is tested only against mocks.
6. **`_make_rel_path` (`arc/output.py` ~L514)** returns things like `../../../home/<user>/...` for logs outside the project directory. That leaks the local directory layout into the exports meant for sharing. Consider emitting `null`, or the basename plus an "outside project" flag.
7. **`parse_1d_scan_full_result` docstring says it never raises**, but it raises `TypeError` on an empty log and `FileNotFoundError` on a missing path. The only caller wraps it in `try`, so either fix the docstring or add the guard.
8. **Silent skip on restart.** The `tckdb` block is popped before `ARC()` is built, so it never reaches `restart.yml`, and a restarted project silently uploads nothing. A one-line INFO log or a doc note would help.

Worth checking. These came from the adversarial pass and were not reproduced:

- `arc/output.py` ~L205 computes per-species corrections from the global `bac_type`. On a kinetics-only run (the kinetics adapter uses `bac_type=None`), `output.yml` might report a BAC that Arkane never applied.
- GSM: if xTB-GSM writes all-zero stringfile energies to mean "no energy", `parser_evidence.py` ~L544 exports them as a flat 0 kcal/mol profile.
- `_read_irc_geometry_table` doesn't require the closing separator. A log truncated mid-table would export a point with only some of its atoms.
- Gaussian IRC point energies come only from `SCF Done`. For an MP2 or double-hybrid IRC, `electronic_energy_hartree` would then be the SCF component.

calvinp0:
@alongd 

Thanks, all addressed at cefbbadc.

1. cp_data test — now reads thermo_points. Verified under rmg_env (10 tests, OK). You're right that CI never runs it there, since it skips in arc_env.

2. Gaussian constraint targets — the parser now carries the value from a =x B line onto the matching F line for the same atoms. The round-trip test builds the deck with GaussianAdapter.write_input_file() rather than hand-typed lines, so it can't drift from the writer. Doing that exposed a writer bug too: the F line had no trailing newline, so two constraints came out as B 1 2 FA 1 2 3 =104.50 B. That's fixed in its own commit, with a two-constraint test.

3. IRC sign — reverse-branch values are now negated, so the coordinate is signed about the TS and the two branches form one axis. direction is still emitted, and the magnitudes are unchanged.

4. URL — the log shows host and port only (urllib.parse.urlsplit), so userinfo, path and query never reach arc.log.

5. tckdb_arc source — the commit message no longer mentions the dropped install script. The warning now names https://github.com/calvinp0/tckdb-adapters and says it isn't on PyPI. The adapter call itself is still only tested against mocks in ARC, because the package isn't installed in ARC's CI.

6. _make_rel_path — a log outside the project directory now gives null, and the schema and docs say so.

7. parse_1d_scan_full_result — I changed the code to match the docstring rather than the reverse. It guards a missing path and degrades each sub-parser to None, so the caller's try is gone.

8. Restart — ARC.__init__ has no tckdb parameter, so the block can't reach restart.yml, and keeping a possibly credential-bearing block out of the project directory is desirable anyway. It now logs at INFO when the block is dropped, and the docstring notes it.

Worth checking - all four reproduced, all four fixed:

- BAC on kinetics-only runs — confirmed: the kinetics run passes bac_type=None. A species' bond_additivity record is now emitted only when thermo was computed for it. bac_type is documented as the request, and TSs never get one.

- GSM all-zero energies — confirmed on the repo's own fixture. An all-zero column across several frames is now treated as absent.

- Truncated IRC table — confirmed: it produced a 4-atom point among 8-atom ones. The table reader is now all-or-nothing, and the point is dropped with a warning. The Hessian geometry reader was changed to fail closed the same way.

- MP2 / double-hybrid IRC energies — confirmed, off by about 231 kcal/mol, and the error varied along the path. parse_irc_path now takes the post-SCF energy using the same helper as parse_e_elect, extracted so both use one implementation.

Two other changes since ddc3081 that you'll see in the diff: the sidecar is renamed to parser_evidence.json, since it holds ARC's parsed facts and nothing TCKDB-specific. And the thermo block now records standard_state_pressure_pa (101325, i.e. 1 atm). It's derived from RMG's partition function rather than written in as a literal.

PR #924, opened 2026-07-26
Add PySCF as a local in-core ESS adapter (sp + opt + freq)

alongd:
## Summary

Adds **PySCF as a local (in-core) electronic-structure adapter** so ARC can run DFT on the
workstation with **no PBS/zeus round-trip**. Uses the **YAML-handoff** in-core pattern (mirroring
`ase_adapter.py`): the engine script runs PySCF in a dedicated `pyscf_env` conda env, writes an
`output.yml` in ARC's internal schema, and ARC's `YAMLParser` re-parses it (the on-disk
`output.yml` is the entire contract).

**Scope: exactly `sp` + `opt` + `freq` (+`optfreq`), closed- and open-shell (RKS/UKS).**
IRC (no native PySCF path — needs Sella) and scans/rotors are **out of scope**, deferred.

## Level of theory & frequency method

- **ωB97M-V / def2-TZVP** — native VV10 nonlocal dispersion, no external dispersion package.
  (ωB97X-D3, the originally-requested level, is **not available in PySCF** — `wb97x_d3` is
  black-listed in `pyscf/scf/dispersion.py` and raises `NotImplementedError`. ωB97M-V was chosen
  as the replacement.)
- **Closed-shell (RKS): analytic Hessian** (VV10 second derivative is implemented).
- **Open-shell (UKS): numerical Hessian** (central finite difference of analytic gradients),
  because PySCF raises `NotImplementedError: UKS Hessian for NLC functional` — there is no analytic
  Hessian for NLC functionals in the unrestricted case.

## Validation (raw PySCF 2.14.0, ωB97M-V/def2-TZVP, before wiring the adapter)

| species | job | result |
|---|---|---|
| H₂O (RKS) | sp+opt+**analytic** Hess | E=−76.42960 Ha; r(O–H)=0.9611 Å; freqs [1617.5, 3823.7, 3927.5] cm⁻¹ (all real); ZPE 56.04 kJ/mol; |μ|=2.07 D |
| OH· (UKS) | sp+opt+**numerical** Hess | E=−75.74005 Ha; r(O–H)=0.9738 Å; freq 3728.6 cm⁻¹ (real) |

## End-to-end round-trip (through the full ARC adapter path, local, no zeus)

A real `optfreq` on H₂O via `PySCFAdapter.execute_incore()` → `output.yml` → `YAMLParser`:
`e_elect=−200665.877 kJ/mol`, geometry, freqs `[1617.5, 3823.7, 3927.5]`, ZPE `56.038 kJ/mol`,
dipole `2.0657 D`, normal modes `(3,3,3)` — all ingest cleanly.

## Schema decisions (output.yml, schema v1)

- **Energy units:** `sp` and `zpe` in **kJ/mol** (matching what `YAMLParser` consumes); scan
  `energies` would be Hartree — documented, but scans are out of scope here.
- **Geometry:** always emitted as an ARC xyz **dict** under `opt_xyz` (opt jobs) / `xyz` (sp/freq),
  never a string — addressing the geometry key/format drift.
- **Frequencies:** emitted as `freqs` (cm⁻¹, imaginary as negative reals) + `modes` + `zpe`, i.e.
  exactly the keys `YAMLParser` already ingests. **`hessian`/`reduced_masses`/`force_constants` are
  deliberately NOT emitted** since `YAMLParser` drops them (no written-but-dropped fields). Ingesting
  the analytic Hessian into Arkane is a follow-up (see below).
- **Error surfacing:** on failure the script writes `success: false` + `error`, and
  `YAMLParser.logfile_contains_errors()` now returns that error (previously always `None`), so a
  crashed local job isn't silently treated as done. Producers that don't emit `success` are unaffected.

## Files

- `arc/job/adapters/scripts/pyscf_script.py` — engine script (RKS/UKS sp/opt/freq via geomeTRIC +
  analytic/numerical Hessian; emits schema-v1 `output.yml`). Runs in `pyscf_env`.
- `arc/job/adapters/pyscf_adapter.py` — the `JobAdapter` (YAML in/out; launches the script via
  `run_in_conda_env` to isolate `pyscf_env`).
- `arc/job/adapters/pyscf_test.py` — unit tests (IO/logic mocked for CI; engine test skipped unless
  PySCF is importable). 7 passed, 1 skipped locally.
- `arc/parser/adapters/yaml.py` — error-surfacing fix (above).
- Registration: `supported_ess`, `global_ess_settings`, `levels_ess` (`wb97m-v`),
  `input_filenames`, `output_filenames`, `PYSCF_PYTHON` (settings.py); `JobEnum.pyscf` (adapter.py);
  adapter import (`adapters/__init__.py`); `register_job_adapter('pyscf', ...)`.

## Reviewer notes / follow-ups

- **Auto-routing:** `levels_ess['pyscf'] = ['wb97m-v']` routes wb97m-v jobs to PySCF. A local
  `~/.arc/settings.py` (if present) overrides `levels_ess`/`supported_ess`, so add `pyscf` there too
  to exercise routing on a machine that uses a local settings override.
- **Environment:** needs a `pyscf_env` conda env with `pyscf`, `geometric`, `pyyaml`.
- **Deferred:** IRC, scans/rotors; ingesting the (closed-shell) analytic Hessian into Arkane
  (`YAMLParser` currently reads freqs+modes+zpe, not the full Hessian).
- **TS opt** (`is_ts`) attempts a geomeTRIC saddle search (`transition=True`); validated here only
  for minimization.

### DIFF
--- a/arc/job/adapters/pyscf_adapter.py
+++ b/arc/job/adapters/pyscf_adapter.py
@@ -1,0 +1,287 @@
+"""
+An adapter for executing PySCF jobs locally (a local in-core ESS, no PBS/queue round-trip).
+
+PySCF runs DFT (sp, opt, freq, optfreq) on the workstation and hands off an ``output.yml`` in
+ARC's internal format, which ARC's ``YAMLParser`` re-parses (the YAML-handoff pattern, mirroring
+``ase_adapter.py``). The engine script runs inside the dedicated ``pyscf_env`` conda environment
+via ``run_in_conda_env`` so ARC's activation variables do not leak into the child process.
+
+Scope: sp + opt + freq (closed- and open-shell, RKS/UKS). IRC and scans/rotors are out of scope.
+"""
+
+import datetime
+import os
+from typing import TYPE_CHECKING, List, Optional, Tuple, Union
+
+from arc.common import ARC_PATH, get_logger, read_yaml_file, save_yaml_file
+from arc.job.adapter import JobAdapter
+from arc.job.adapters.common import _initialize_adapter
+from arc.job.env_run import run_in_conda_env
+from arc.job.factory import register_job_adapter
+from arc.settings.settings import PYSCF_PYTHON
+
+if TYPE_CHECKING:
+    from arc.level import Level
+    from arc.reaction import ARCReaction
+    from arc.species.species import ARCSpecies
+
+logger = get_logger()
+
+PYSCF_SCRIPT_PATH = os.path.join(ARC_PATH, 'arc', 'job', 'adapters', 'scripts', 'pyscf_script.py')
+
+
+class PySCFAdapter(JobAdapter):
+    """
+    A class for executing PySCF jobs.
+
+    Args:
+        project (str): The project's name. Used for setting the remote path.
+        project_directory (str): The path to the local project directory.
+        job_type (list, str): The job's type, validated against ``JobTypeEnum``.
+        args (dict, optional): Methods (e.g., troubleshooting) to be used in input files.
+        bath_gas (str, optional): A bath gas.
+        checkfile (str, optional): The path to a previous Gaussian checkfile to be used in the job.
+        conformer (int, optional): Conformer number if optimizing conformers.
+        constraints (list, optional): A list of constraints to use during an optimization or scan.
+        cpu_cores (int, optional): The total number of cpu cores requested for a job.
+        dihedral_increment (float, optional): The degrees increment to use when scanning dihedrals.
+        dihedrals (list, optional): The dihedral angles of a directed scan job.
+        directed_scan_type (str, optional): The type of the directed scan.
+        ess_settings (dict, optional): A dictionary of available ESS and a corresponding server list.
+        ess_trsh_methods (list, optional): A list of troubleshooting methods already tried out.
+        execution_type (str, optional): The execution type, 'incore', 'queue', or 'pipe'.
+        fine (bool, optional): Whether to use fine geometry optimization parameters. Default: ``False``.
+        initial_time (datetime.datetime or str, optional): The time at which this job was initiated.
+        irc_direction (str, optional): The direction of the IRC job (`forward` or `reverse`).
+        job_id (int, optional): The job's ID determined by the server.
+        job_memory_gb (int, optional): The total job allocated memory in GB (14 by default).
+        job_name (str, optional): The job's name (e.g., 'opt_a103').
+        job_num (int, optional): Used as the entry number in the database, as well as in ``job_name``.
+        job_server_name (str, optional): Job's name on the server (e.g., 'a103').
+        job_status (list, optional): The job's server and ESS statuses.
+        level (Level, optional): The level of theory to use.
+        max_job_time (float, optional): The maximal allowed job time on the server in hours.
+        run_multi_species (bool, optional): Whether to run a job for multiple species in the same input file.
+        reactions (list, optional): Entries are ARCReaction instances, used for TS search methods.
+        rotor_index (int, optional): The 0-indexed rotor number (key) in the species.rotors_dict dictionary.
+        server (str, optional): The server to run on.
+        server_nodes (list, optional): The nodes this job was previously submitted to.
+        species (list, optional): Entries are ARCSpecies instances.
+                                  Either ``reactions`` or ``species`` must be given.
+        testing (bool, optional): Whether the object is generated for testing purposes, ``True`` if it is.
+        times_rerun (int, optional): Number of times this job was re-run with the same arguments.
+        torsions (list, optional): The 0-indexed atom indices of the torsion(s).
+        tsg (int, optional): TSGuess number if optimizing TS guesses.
+        xyz (dict, optional): The 3D coordinates to use in the job.
+    """
+
+    def __init__(self,
+                 project: str,
+                 project_directory: str,
+                 job_type: Union[List[str], str],
+                 args: Optional[dict] = None,
+                 bath_gas: Optional[str] = None,
+                 checkfile: Optional[str] = None,
+                 conformer: Optional[int] = None,
+                 constraints: Optional[List[Tuple[List[int], float]]] = None,
+                 cpu_cores: Optional[str] = None,
+                 dihedral_increment: Optional[float] = None,
+                 dihedrals: Optional[List[float]] = None,
+                 directed_scan_type: Optional[str] = None,
+                 ess_settings: Optional[dict] = None,
+                 ess_trsh_methods: Optional[List[str]] = None,
+                 execution_type: Optional[str] = None,
+                 fine: bool = False,
+                 initial_time: Optional[Union[datetime.datetime, str]] = None,
+                 irc_direction: Optional[str] = None,
+                 job_id: Optional[int] = None,
+                 job_memory_gb: float = 14.0,
+                 job_name: Optional[str] = None,
+                 job_num: Optional[int] = None,
+                 job_server_name: Optional[str] = None,
+                 job_status: Optional[List[Union[dict, str]]] = None,
+                 level: Optional['Level'] = None,
+                 max_job_time: Optional[float] = None,
+                 run_multi_species: bool = False,
+                 reactions: Optional[List['ARCReaction']] = None,
+                 rotor_index: Optional[int] = None,
+                 server: Optional[str] = None,
+                 server_nodes: Optional[list] = None,
+                 queue: Optional[str] = None,
+                 attempted_queues: Optional[List[str]] = None,
+                 species: Optional[List['ARCSpecies']] = None,
+                 testing: bool = False,
+                 times_rerun: int = 0,
+                 torsions: Optional[List[List[int]]] = None,
+                 tsg: Optional[int] = None,
+                 xyz: Optional[dict] = None,
+                 ):
+
+        self.job_adapter = 'pyscf'
+        self.execution_type = execution_type or 'incore'
+        self.incore_capacity = 100
+        self.command = None
+        self.url = 'https://pyscf.org/'
+
+        self.sp = None
+        self.opt_xyz = None
+        self.freqs = None
+
+        self.script_path = PYSCF_SCRIPT_PATH
+
+        _initialize_adapter(obj=self,
+                            is_ts=False,
+                            project=project,
+                            project_directory=project_directory,
+                            job_type=job_type,
+                            args=args,
+                            bath_gas=bath_gas,
+                            checkfile=checkfile,
+                            conformer=conformer,
+                            constraints=constraints,
+                            cpu_cores=cpu_cores,
+                            dihedral_increment=dihedral_increment,
+                            dihedrals=dihedrals,
+                            directed_scan_type=directed_scan_type,
+                            ess_settings=ess_settings,
+                            ess_trsh_methods=ess_trsh_methods,
+                            fine=fine,
+                            initial_time=initial_time,
+                            irc_direction=irc_direction,
+                            job_id=job_id,
+                            job_memory_gb=job_memory_gb,
+                            job_name=job_name,
+                            job_num=job_num,
+                            job_server_name=job_server_name,
+                            job_status=job_status,
+                            level=level,
+                            max_job_time=max_job_time,
+                            run_multi_species=run_multi_species,
+                            reactions=reactions,
+                            rotor_index=rotor_index,
+                            server=server,
+                            server_nodes=server_nodes,
+                            queue=queue,
+                            attempted_queues=attempted_queues,
+                            species=species,
+                            testing=testing,
+                            times_rerun=times_rerun,
+                            torsions=torsions,
+                            tsg=tsg,
+                            xyz=xyz,
+                            )
+
+    def determine_settings(self) -> dict:
+        """
+        Build the ``settings`` block passed to pyscf_script.py: the level's method (the xc
+        functional) and basis set, the job memory, the ``fine`` flag and cpu core count (which
+        pins PySCF's thread count), plus any user ``args['keyword']`` knobs.
+
+        Returns:
+            dict: The resolved PySCF run settings.
+        """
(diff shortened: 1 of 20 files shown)


calvinp0:
The issue I had with pyscf or rather with any locally run ESS is the serial run of jobs and since we spawn so many conformers we would be waiting weeks to months. I had tried to implement PySCF with utilising parallel runs by "submitting" to the system and then holding onto their PID and checking it every so often, so emulating HPC systems but the issue then came down to the hardware of the local PC

https://github.com/ReactionMechanismGenerator/ARC/tree/pyscf

This was written before claude code etc so there may be nuances or other things I could have done

alongd:
## End-to-end validation: real ARC runs through the PySCF adapter, CPU vs GPU

Ran this branch on real species to check the adapter works end-to-end and to measure whether the GPU
path is worth using. Reporting timings and the bugs this shook out.

### Setup

- **Level:** ωB97M-V/def2-TZVP, geomeTRIC optimization, executed **in-core** (`software: pyscf`, server `local`).
- **CPU:** 28 cores. **GPU:** NVIDIA RTX 4080 SUPER 16 GB, `gpu4pyscf-cuda12x` 1.8.0, PySCF 2.14.
- **Timing** = wall time of one ARC `opt` job, from `input.yml` written to `output.yml` written.
- `compute_thermo: false`, `calc_freq_factor: false`, explicit input geometries (no conformer search).

### Geometry optimization

| Species | SMILES | Atoms | Basis functions | Spin | CPU (28 cores) | GPU | GPU speedup |
|---|---|---|---|---|---|---|---|
| hydrogen peroxide | `OO` | 4 | 74 | singlet | 78–92 s | 112 s | **0.8× (slower)** |
| 2-aminoethyl radical | `[CH2]CN` | 9 | 129 | doublet | 582–713 s | 284 s | **~2.3×** |
| n-propanol | `CCCO` | 12 | 172 | singlet | 426–481 s | 161–163 s | **~2.7×** |
| n-dodecane | `CCCCCCCCCCCC` | 38 | 528 | singlet | not measured | 1449 s | — |

**The GPU only pays off above roughly 100 basis functions**, and the advantage grows with system size.
For H₂O₂ the GPU is *slower* — kernel-launch and transfer overhead dominates a 74-function problem.

`[CH2]CN` costs more than propanol on CPU despite having fewer basis functions: it is open-shell (UKS)
and needed more geometry steps. Note it is the **2-aminoethyl radical** (·CH₂CH₂NH₂) — bare `N` in
SMILES is an amine, not a nitrile.

**Reproducibility.** GPU timings are near-identical across three independent runs (propanol
163/161/161 s; H₂O₂ 112/112 s; `[CH2]CN` 284/284 s). CPU timings vary with machine load, hence the
ranges — these were measured on a shared workstation whose load average swung between 4 and 52. One
badly contended run (propanol 836 s) is excluded as unrepresentative. **Treat the CPU column as
indicative, not benchmark-grade.**

ARC's second `fine_grid` optimization stage, CPU: H₂O₂ 40 s, propanol 238 s, `[CH2]CN` 414 s.

### `fine` → PySCF grid mapping

ARC's Gaussian adapter emits `integral=(grid=ultrafine)` for a fine job, i.e. the pruned (99,590) grid.
Measured PySCF equivalents for a second-row atom:

| PySCF `grids.level` | radial × angular (C/N/O) | Gaussian equivalent |
|---|---|---|
| 3 (PySCF default) | (75, 302) | FineGrid — exact match |
| **4** ← used for `fine` | **(90, 590)** | **UltraFine (99,590)** |
| 5 | (105, 770) | overshoots UltraFine by ~1.5× in grid points |

Level 4 is therefore the honest analogue. The NLC (VV10) grid is left at PySCF's default level 3, since
the nonlocal kernel is smooth and conventionally integrated more coarsely than the XC term.

### Not yet measured

- **Frequencies on either device.** The opt → fine-opt → freq pipeline now runs end-to-end (verified —
  freq jobs spawn and execute), but no freq job has been carried to completion yet.
- **n-dodecane on CPU.** The 1-core and 10-core runs were interrupted. At the point of interruption
  **10 cores had already passed the GPU's 1449 s without converging**, so the GPU is >1.2× faster at
  this size — a lower bound, not a measurement.

### GPU limitations worth knowing before routing production jobs

- **Closed-shell analytic Hessians are forced back to the CPU.** `gpu4pyscf` 1.8.0 fails with
  `assert a.flags.c_contiguous` (`lib/cupy_helper.py:527`, via `hessian/rhf.py:753`). This reproduces
  with B3LYP too, so it is **not** NLC-specific — it looks like a general gpu4pyscf bug worth reporting
  upstream. Consequence: singlet `freq` gets no GPU acceleration.
- **Open-shell frequencies use a numerical Hessian**, because PySCF raises
  `NotImplementedError: UKS Hessian for NLC functional` (`hessian/uks.py:44`). RKS NLC Hessians are
  implemented. This is expensive — `[CH2]CN`'s numerical Hessian was still running after 25 min.
- **Scaling is sub-linear.** A 10-thread job sustained only ~650% CPU, not ~1000%. For throughput this
  argues for several concurrent modest-core jobs (what the local pipe backend does) over one wide job.
- Use the **cuda12x** wheel stack, not cu13 — the `nvidia-*-cu13` pip wheels are unpopulated stubs
  (`nvidia-nvjitlink-cu13` is 0.0.1) and cupy fails to find `libnvJitLink.so`.

### Bugs this shook out

Running a local in-core ESS end-to-end had evidently never been done before; five distinct failures had
to be fixed to get a single species through `opt` → `freq`. Four are pre-existing ARC issues, not
PySCF-specific.

| # | Site | Symptom |
|---|---|---|
| 1 | `arc/common.py:156` | `check_ess_settings` rejects server `local` unless a `local` entry exists in `servers`. Any in-core ESS needs one. |
| 2 | `arc/scheduler.py:1235` | `conf_opt: false` is ignored for a species defined only by SMILES — the gate is `job_types['conf_opt'] ... or species.get_xyz(generate=False) is None`, so conformers are generated anyway and dispatched to the conformer-level ESS. |
| 3 | `arc/job/local.py:186` | `ValueError: Server cluster software local is not supported` — a machine with no queueing system has no queue job IDs to report. |
| 4 | `arc/job/adapter.py:799` | `ValueError: Unrecognized cluster software: local` — tries to collect scheduler stdout/stderr files that cannot exist without a scheduler. |
| 5 | `arc/job/trsh.py` | **The significant one.** `determine_ess_status()` only understands the text logs of gaussian/qchem/orca/molpro/terachem; anything else falls through to `return '', ...`. Status `''` is not `'done'`, so ARC marked every successfully completed in-core job as failed, tried to troubleshoot it, found no methods, and declared the species unconverged — while `output.yml` said `success: True` the whole time. **This is not PySCF-specific: `ase`, `torchani`, `gcn` and `mockter` all write `output.yml` and hit the identical fallthrough.** Fixed generally via a new `determine_ess_status_from_yaml()` that treats the adapter's own `success`/`error` keys as authoritative. |

Fixes for 3, 4 and 5 are in this branch, with tests. `check_running_jobs_ids()` had no test coverage at
all previously. Item 1 is a settings/documentation gap — the shipped `servers['local']` template now
documents `'local'` as a valid `cluster_soft` for a machine with no queue. **Item 2 is untouched and
still open**; it is orthogonal to this PR and probably deserves its own issue.

calvinp0:
@alongd Missing the installation of PySCF scripts - both with and without gpu support

calvinp0:
I think also pin it to 1.7.6 for GPU4PySCF since there is a regression:
https://github.com/pyscf/gpu4pyscf/issues/836
I even got the error when running the GPU example scripts on PySCF

calvinp0:
Also, please write up an update to the documentation on the utilisation of PySCF so a user can look on the website to understand what is required, what they need to change and what each change does

arc/job/pipe/pipe_run.py
```diff
@@ -770,15 +813,46 @@ def _ingest_rotor_scan_1d(run_id, pipe_root, spec, state, species_dict, label):
 # ===========================================================================
 
 
+def local_worker_limit(cpus_per_worker: int, memory_mb_per_worker: int) -> int:
```
calvinp0, line 884: The problem that I see here is that this function is stateless, so to speak. It only takes how big one worker, and not how many workers are already running. So every batch asks it the same questions and gets the same answers because it is recomputing form the full machine as if it were one batch only.

so we could have a sp batch and a freq batch and a ts batch, and each time we ask 'how many fit?', it would do 
`os.cpu_count() // 5` and say 4 for each cause it still believes 20 cores are free.

arc/job/adapters/pyscf_adapter.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def execute_queue(self):
+        """
+        PySCF is a local in-core ESS; fall back to incore execution when queued.
+
+        ``execution_type`` is updated to reflect what actually happened, so that downstream
+        status handling (e.g., ``determine_job_status()``) does not query a queueing system
+        for a job that never reached one.
+        """
+        logger.warning('Queue execution is not supported for PySCF in ARC; running incore instead.')
```
calvinp0: This gets spammed quite a bit, I don't know if we should have this warning?

arc/job/adapters/pyscf_adapter.py
```diff
@@ -0,0 +1,283 @@
...
+                            attempted_queues=attempted_queues,
+                            species=species,
+                            testing=testing,
+                            times_rerun=times_rerun,
+                            torsions=torsions,
+                            tsg=tsg,
+                            xyz=xyz,
+                            )
+
+    def determine_settings(self) -> dict:
```
calvinp0, line 174: `settings_dict` is seeded from `args['keyword']` before ARC setdefaults its own keys, so a user arg named method, basis, memory_mb, cpu_cores or device silently wins. 
`args: {keyword: {method: 'pbe'}}` would run PBE while the level, logs and restart file all still say b3lyp.

calvinp0:
```
┌────────────────────────────────────────────────────────────┬───────────────────────────────────┐
│                            step                            │      settings_dict['method']      │
├────────────────────────────────────────────────────────────┼───────────────────────────────────┤
│ user args seed the dict                                    │ 'pbe'                             │
├────────────────────────────────────────────────────────────┼───────────────────────────────────┤
│ setdefault('method', 'b3lyp')                              │ 'pbe' ← no-op, key already exists │
├────────────────────────────────────────────────────────────┼───────────────────────────────────┤
│ what PySCF actually runs                                   │ PBE                               │
├────────────────────────────────────────────────────────────┼───────────────────────────────────┤
│ what ARC's logs, restart.yml, and the Arkane level all say │ B3LYP                             │
└────────────────────────────────────────────────────────────┴───────────────────────────────────┘
```

calvinp0 (changes requested):

alongd:
This PR leans on ARC's existing local pipe worker pool to fan the conformer jobs out across cores in parallel (in-core PySCF calls, no external queue) , with a single machine-wide CPU budget `(servers['local']['cpus'])`.

alongd:
Added devtools/pyscf_environment.yml + devtools/install_pyscf.sh following the same pattern as the other ESS installers, and wired it into install_all.sh. GPU support is opt-in via --cuda (installs the gpu4pyscf-cuda12x wheel on top of the CPU base); without the flag you get a CPU-only pyscf_env.

alongd:
Thanks - I agree re pining pyscf to 1.74.6

arc/job/adapters/pyscf_adapter.py
alongd: nice catch. I agree, fixed

arc/job/adapters/pyscf_adapter.py
alongd: downgrading to `logger.debug`

arc/job/pipe/pipe_run.py
alongd: Agreed — it's stateless by design: a single pool is bounded (workers × cores/worker ≤ `cpus`), but two concurrent pools/processes each derive independently and can jointly overshoot. A proper cross-process fix needs a shared semaphore, which is out of scope for this PR; it's noted as a known limitation and is worth a follow-up issue to track.

alongd:
Thanks for the context. This PR deliberately avoids the PID-holding / HPC-emulation approach from the `pyscf` branch: PySCF runs in-core, and parallelism comes from ARC's local pipe worker pool (parallel in-core jobs) bounded by a single machine-wide CPU budget (`servers['local']['cpus']`) — so no PID bookkeeping, and the local core count is respected by construction.

arc/job/adapters/pyscf_adapter.py
alongd: Good catch — dropped this to `logger.debug` in `execute_queue` (pyscf_adapter.py). PySCF is in-core, so the queue→in-core fallback fires on every job; at `warning` it flooded the log with a non-actionable message. `debug` keeps it available for troubleshooting without the spam. Resolving.

alongd:
Done — added a **Run PySCF In-Core (Queueless)** section to `docs/source/running.rst`. It covers: installing into `pyscf_env` via `bash devtools/install_pyscf.sh` (`--cuda` for the gpu4pyscf stack); configuring the reserved `local` server with `cluster_soft: 'local'` and the machine-wide `cpus` budget; routing `pyscf` via `global_ess_settings`; the supported job types (`sp`/`opt`/`freq`); GPU selection (`PYSCF_DEVICE` globally or the per-job `device` keyword, with the gpu4pyscf pin that avoids the Hessian regression); method/basis always coming from the level of theory; and the transparent queue→in-core fallback.

arc/job/pipe/pipe_run.py
alongd: Agreed — this is a real limitation, but the fix needs cross-process coordination (the arrays are independent processes with no shared counter), which is out of scope for this PR. Tracked in #961 for a follow-up; `pipe_settings['local_max_workers']` is the manual cap in the meantime. Resolving here.

calvinp0 (approved):

PR #925, opened 2026-07-26
Fix Mockter freq-parsing regression on diatomic/linear species

alongd:
## What's broken

On `main`, any ARC job with a `freq` job type on a **diatomic/linear** species routed to the Mockter adapter crashed with:

```
File "arc/scheduler.py", line 2703, in check_negative_freq
    for freq in vibfreqs:
TypeError: 'NoneType' object is not iterable
```

`examples/mockter/input.yml` reproduces it (its `OH` species is diatomic). The minimal repro (a single `[OH]` species with `freq`) crashes the same way.

## Root cause (probed empirically — the original premise was inverted)

The dispatch note suspected a runtime **path** bug (parser handed a good file returns `None`). Probing the actual runtime artifact inverted that: `calcs/Species/OH/freq_*/output.yml` contained **`freqs: []`** (empty), not a populated list. C2H6 (8 atoms) converged fine in the same run — only the diatomic crashed. Two combining defects:

1. **Origin — `MockAdapter.execute_incore` (`arc/job/adapters/mockter.py`)** generated `3N-6` frequencies unconditionally. For a diatomic (N=2), `3*2 - 6 = 0` → an **empty** freqs list. A diatomic/linear molecule physically has `3N-5` modes, so OH should have 1 frequency.
2. **Regression — YAML parser adapter (`arc/parser/adapters/yaml.py`)**, introduced in the parser refactor (`9224a8c6`):
   ```python
   freqs = self.data.get('freqs') or self.data.get('frequencies')
   return np.array(freqs, dtype=np.float64) if freqs else None
   ```
   An empty list is falsy, so `[]` was treated as *missing* and returned `None`. `check_negative_freq` then iterated `None` unguarded → crash.

So the runtime value really was `None`, but because the file legitimately held an empty list that the parser collapsed to `None` — not because of a wrong path.

## Fix

- **`mockter.py`** (origin): emit the physically-correct mode count — `0` monatomic, `3N-5` linear (via `is_xyz_linear`), `3N-6` nonlinear. A diatomic now keeps its 1 real frequency instead of an empty list. Existing nonlinear cases (e.g. the 3-atom TS test → `3N-6 = 3`) are unchanged.
- **`yaml.py`** (regression): distinguish an empty freqs list (→ empty array) from a genuinely missing key (→ `None`), so empty is no longer collapsed to `None`.
- **`scheduler.py`** (defense-in-depth): `check_negative_freq` guards `vibfreqs is None` by logging and returning `False` (routes to troubleshooting) rather than crashing — and, per the "don't mask it" directive, it does **not** silently pass a freq-less job.

## Tests

- `mockter_test.py::test_mockter_freq_count_matches_geometry` — mock freq count is `0`/`3N-5`/`3N-6` for monatomic/linear/nonlinear (incl. the diatomic regression case).
- `parser_test.py::test_parse_frequencies` — YAML `freqs: []` → empty array (not `None`); missing key → `None`.

## Verification (arc_env)

- `python ARC.py examples/mockter/input.yml` → **exits 0**, no `NoneType` crash. All four species converge; `OH` now carries `harmonic_frequencies_cm1: [500.0]` (point group `Cinfv`) through to statmech — freqs are not silently dropped.
- Minimal `[OH]` + `freq` repro completes cleanly.
- `arc/job/adapters/mockter_test.py` + `arc/parser/parser_test.py`: **38 passed** (`-n0`).

> `TS0` in the example still fails, but that is pre-existing and orthogonal: the example sets `ts_adapters: []`, so no TS guess geometry is generated (`opt_log: null`, `xyz: null`) — the TS never reaches a freq job. Not caused by this change.

Please review — **do not merge yet**.


alongd:
Addressed the Copilot review (folded into `12a131e`):

- **`mockter.py`** — guarded `freqs[0] = -500` behind `and freqs` so a TS with an empty freq list (e.g. a monatomic edge case) no longer raises `IndexError`.
- **`scheduler.py`** — updated the `check_negative_freq` docstring so `vibfreqs` documents the `list | np.ndarray | None` type and the new None-as-parse-failure behavior.
- **`scheduler_test.py`** — extended `test_check_negative_freq` to assert `vibfreqs=None` returns `False` (locks in the guard against the original NoneType crash).

arc/scheduler.py
```diff
@@ -2692,8 +2692,12 @@ def check_negative_freq(self,
         Args:
             label (str): The species label.
             job (JobAdapter): The optimization job object.
-            vibfreqs (list): The vibrational frequencies.
+            vibfreqs (list | np.ndarray | None): The vibrational frequencies, or ``None`` if they could not be parsed.
         """
+        if vibfreqs is None:
+            logger.error(f'Could not parse frequencies for {label} from the freq job output file '
+                         f'{job.local_path_to_output_file}. Treating the freq job as unsuccessful.')
+            return False
```
calvinp0, line 2700: After this maybe add a second guard where we do len(vibfreqs) == 0 and is not ts and is not monoatomic then log an error? That way a species that for some reason yields no freqs whatsoever and is not monoatomic cannot sneak through

calvinp0 (changes requested):

arc/scheduler.py
alongd: Good call — added exactly that second guard (folded into `b2c19da`):

```python
if len(vibfreqs) == 0 and not self.species_dict[label].is_ts and self.species_dict[label].number_of_atoms > 1:
    logger.error(f'The freq job for {label} yielded no frequencies, but {label} is a polyatomic non-TS '
                 f'species and should have vibrational modes. Treating the freq job as unsuccessful.')
    return False
```

So a polyatomic non-TS species that yields no frequencies is now treated as an unsuccessful freq job rather than silently passing, while a monatomic species (legitimately 0 modes) is exempted. Extended `test_check_negative_freq` to cover it.

calvinp0 (approved):

PR #935, opened 2026-07-29
Map reactions in their family discovery direction; anchor XH3 frames on a non-linear pair

calvinp0:
Base: `main`. This branch absorbs **#948** ("Assign every hydrogen and reject an incomplete atom map"), which was flattened into this PR and is now closed. Its former dependency **#947** (`fix_map_two_species_candidate`) is **merged into `main`**, so there is nothing to read first — this branch is 0 commits behind `main` and stands alone.

Six commits, no file touched by more than one of them. Three parts, reviewable separately:

| part | what | files |
|---|---|---|
| 1 | map in the direction the family was discovered in | `arc/mapping/driver.py` |
| 2 | XH3 local frames: non-linear anchors, parity-preserving assignment | `arc/mapping/engine.py`, `arc/species/vectors.py` |
| 3 | assign every hydrogen; reject an incomplete map instead of shifting it | `arc/mapping/engine.py` |

---

## Part 1 — map in the discovery direction

- When every product dict of a reaction carries `discovered_in_reverse`, the first-orientation `map_rxn()` attempt **cannot** succeed: it fails on a missing template map or on `get_template_product_order()`, returns `None`, and `map_reaction()` then recovers through its existing flip fallback. `map_reaction()` now goes straight to the flipped orientation in that case.
- The two intermediate `logger.error()` calls therefore fired as false alarms on reactions that mapped fine. Both are downgraded to `logger.debug` and the "cannot atom map" phrasing dropped — it read as a hard failure for something orientation-specific and routinely recovered. Those two branches still return `None`, so no atom map changes because of the log edit.
- A genuine total-mapping failure (both orientations exhausted) is still surfaced at ERROR level by `ARCReaction.atom_map` ("could not be atom mapped").
- Tested by driving the first-orientation failure through `map_rxn()` and asserting that no ERROR is logged while `map_reaction()` still returns a valid map, and that a retro-Diels-Alder reaction, whose product dicts are all `discovered_in_reverse`, is mapped with the expected bond changes.

---

## Part 2 — XH3 local frames

Unlike part 1, this **does** change hydrogen assignments — for XH3 groups whose two species list their hydrogens in different orders, and for molecules with a linear heavy-atom skeleton that previously could not be mapped at all.

### The defect

`_select_ch3_anchors()` promised a second anchor B forming a non-linear B–X–A angle, but never tested for it and never fell back to hydrogens. For a linear heavy-atom skeleton (propyne, `CH3-C#CH`) every heavy candidate lies on the X→A axis, so `_construct_local_axes()` raised `Anchors are colinear; cannot define unique plane.`, `map_rxn()` returned `None`, and the whole reaction could not be atom mapped. Surfaced by a Retroene benchmark reaction, `C=C=CCC <=> C#CC + C=C`, once the species carried optimized geometries.

Anchor selection now walks candidates in priority order and returns the first that passes the new `_anchors_define_a_plane()`, so it can never hand `_construct_local_axes()` anchors that function rejects. `_construct_local_axes()` itself remains strict.

### Why there are two gates, not one

This is the part most likely to look redundant on a first read, so, explicitly:

`_construct_local_axes()` raises when `|e_x × v_XB| = |v_XB| · sin θ < 1e-8`. That quantity is **length-weighted**, and the two failure modes it conflates are independent:

- **A purely angular criterion cannot bound it.** A short second anchor at a perfectly comfortable angle still fails: `|v_XB| = 1e-7 Å` at `θ = 2°` gives `3.5e-9` and raises.
- **A purely length-weighted criterion is not chemical.** On an RDKit-embedded propyne the near-linear tier-1 heavy pair scores `1.028918e-07` at an A–X–B angle of `179.999998°` — ten times the `1e-8` threshold — so it is *accepted*, and a 0.001 Å jitter on the far carbon then swings the frame's azimuthal reference through the full 0–360°. That figure is reproducible rather than embedding-dependent: propyne has no rotatable heavy-atom dihedral, so the embedding converges to the same near-linear minimum every time (bit-identical across ten embeddings, in-process and across separate interpreters).

So `_anchors_define_a_plane()` applies both:

1. **Angular** — `is_angle_linear()` with `zmat.TOL_180` (0.9°), the tolerance ARC already uses to decide an angle is too close to 0 or 180° to define a dihedral, computing the angle with `vectors.get_angle()`. This is the chemical requirement.
2. **Cross-product `>= 1e-8`** — the *same* test `_construct_local_axes()` applies internally, reproduced exactly, so a pair accepted here can never make it raise. This is the numerical precondition.

Measured over a 4,732-point sweep of anchor lengths (1e-12 to 1e3 Å) and angles (0 to 180°): the angular criterion alone accepts 1,400 configurations, of which **350 still make `_construct_local_axes()` raise**. The shipped pair accepts 1,050 and raises on none. `test_anchors_define_a_plane_agrees_with_construct_local_axes` pins that agreement — widened here to sweep near-threshold *angles* as well as near-threshold lengths, and to assert rejection as well as acceptance — and `test_anchors_define_a_plane_angular_threshold` pins both sides of the angular gate (0.89° rejected, 0.91° accepted) at anchor distances from 0.05 to 100 Å, so a length-weighted criterion can no longer pass it.

### The anchor tiers, and why the hydrogen tier is kept

1. The preferably-heavy neighbour of A other than X.
2. Any other neighbour of A other than X.
3. Any atom bonded to X other than A — i.e. one of the XH3 hydrogens.

- **Tier 3 is what makes propyne mappable.** A linear heavy-atom skeleton has no off-axis heavy anchor; only the group's own hydrogens do. Removing it would reinstate the original failure. A debug log records that the phase is undetermined when that tier is used.
- **The old tier offering "any remaining atom in the molecule" is removed.** An XH3 centre has exactly one heavy neighbour and is therefore always terminal, so the hydrogen tier always succeeds for a genuine XH3 centre and that tier only ever fired for a non-bonded, chemically meaningless anchor. Anchors are now never selected from atoms not bonded to the centre.
- **Replacing tier 3 with a positional hydrogen assignment was considered and rejected** — see below. It is a no-op on the propyne case that motivates tier 3, and it is actively unsafe.

### Parity, and why a positional fallback is unsafe

- The geometric path always yields an **even** permutation: `e_z = e_x × v_XB` is right-handed for any anchor, and `e_x` corresponds across species, so the azimuthal *cyclic order* is intrinsic. Only the *phase* is arbitrary — and phase is physically meaningless for a terminal XH3, whose three hydrogens are interconverted by its own C3 rotation. **Parity is not meaningless.**
- A positional assignment is an odd permutation in **three of the six** relabelings. `linear.py` interpolates Z-matrix internals and copies spectator variables verbatim (`arc/job/adapters/ts/linear_utils/math_zmat.py`), so for a methyl at or adjacent to the reactive centre an odd permutation blends two dihedrals to the same value: measured H–C–H **0.5°** and H···H **0.010 Å** at weight 0.5 for propanol — two nuclei on one point, an unconvergeable guess. Any cyclic phase yields a perfect methyl at every weight (H–C–H 108°, H···H 1.77 Å).
- The three hydrogens of an XH3 group are therefore now always assigned, and by **cyclic azimuthal order** rather than by list position. `_map_xh3_group_by_azimuthal_order()` needs only the X→A bond, so it also covers the case where no secondary anchor exists.
- The pre-existing `len(spc_1.mol.atoms) <= 5` shortcut carried the same hazard — it made the assignment positional for species such as CH3F, an odd permutation whenever the two species list the methyl hydrogens in a different order — and is removed. Both paths now have a parity test that maps a species onto a copy of its own geometry with the coordinates of two methyl hydrogens exchanged, and asserts that the signed triple product of the three centre→H vectors keeps its sign — exactly the invariant the linear TS adapter's Z-matrix blend depends on. Against the previous code both report an odd permutation.
- `_map_xh3_group()` now also rejects a `(None, None)` anchor tuple — a non-empty tuple is truthy, so the previous guard missed it — and returns `None` instead of letting a `ValueError` propagate.

### `vectors.get_perpendicular_axes()`

The two axes completing a right-handed orthonormal frame with a given direction were constructed inline in `arc/mapping/engine.py`. Moved to the module that owns geometry maths, where it is reusable and Cython-bound. It delegates the azimuthal-phase choice to `vectors.get_perpendicular_unit_vector()` — the module's existing least-aligned-axis construction, which `main` gained (merged in #944) *after* this branch was written and which owns exactly the cross product this had open-coded — and keeps only its own right-handed pair construction and `tol` length guard:

```python
e_x = unit_vector(vector)
e_z = get_perpendicular_unit_vector(e_x)
return get_normal(e_z, e_x), e_z
```

It also validates its input, which the inline construction did not: non-finite input previously returned NaN axes that propagated silently into an atom map.

The refactor was checked against degenerate input rather than the unit tests alone (the shared primitive normalizes, so zero length is the case that could newly raise). Old and new were compared bit-exactly — packed IEEE-754 doubles — over 125,435 `(vector, tol)` pairs: 25,087 vectors × 5 tolerances (default, `1e-8`, `0.0`, `-1.0`, `1e-3`), covering ordinary vectors, all 27 sign/axis-aligned combinations, magnitude ties where the argmin is ambiguous, magnitudes from `5e-324` to `1e308`, exact zero, non-finite input, wrong-length input, and 25,000 random vectors spanning 14 decades.

**All 125,435 agree except 30**, from 6 vectors, all of the same class: `|vector| > ~1.34e154`, where the squared length overflows to `inf`, `unit_vector` returns `[0, 0, 0]`, and the perpendicular step is asked for a perpendicular of the zero vector. Old raised `ZeroDivisionError`; new raises `VectorsError`. No input returns a value in one and raises in the other, and none returns different values. That change is an improvement at the only call site: `_map_xh3_group_by_azimuthal_order()` catches `VectorsError` and degrades to the cyclic-order fallback, so what was an uncaught `ZeroDivisionError` is now the documented graceful path. The input is unreachable anyway — it needs a Cartesian displacement above 1e154 Å. Tests assert orthonormality and `e_x × e_y = e_z` over axis-aligned, general and badly scaled inputs, pin the one case that follows from the convention alone (a vector along +z yields `e_y = +x`, `e_z = +y`), assert scale invariance, and assert rejection of a non-finite, too-short or wrong-length input.

---

### The gap

- `map_hydrogens()` handled an XH2 group with `if mapped: … continue` and no `else`, so whenever `_map_xh2_group()` returned `{}` — no mapped heavy neighbour, or a heavy atom not carrying exactly two hydrogens — **both hydrogens were silently left out of the map**.
- A heavy atom carrying four or more hydrogens alongside a heavy neighbour matched no branch at all and was dropped the same way.

### Why omission was not a clean failure

`map_two_species()` builds its list map with `[v for k, v in sorted(atom_map.items(), …)]` and no length check, so a missing key **shifts every subsequent index by one** — every atom past the gap silently refers to the wrong one, and the reaction is mapped with a plausible-looking but wrong map. It surfaces much later as an `IndexError` in `glue_maps()`, which `try_mapping()` does not catch, since it catches only `ValueError`.

### The fix

- Both cases now fall back to assigning the hydrogens in the order they appear in the two species, each with a DEBUG log naming the centre.
- `map_two_species()` checks that the map is a permutation of both species' atoms before returning it, raising `ValueError` otherwise — so any future gap degrades to a mapping failure instead of corrupting indices.
- **XH2 needs no parity guard.** Unlike an XH3 group, the two hydrogens of an XH2 group are related by a reflection rather than by a rotation, so there is no handedness for the order-based fallback to preserve. Contrast part 2, where the parity of an XH3 assignment is load-bearing.
- The completeness check can only be unconditional because part 2 makes the XH3 branch always return an assignment for all three hydrogens; on `main` it would fire on ordinary XH3 groups.

Against the previous code, the new tests show the XH2-refinement case and the four-hydrogen case each returning a map missing the affected hydrogens, and `map_two_species()` returning a silently shifted list map.

### Behaviour change to be aware of

`map_two_species()` can now raise `ValueError` where it previously returned a shifted map. Not every caller catches that, so the failure mode differs by path:

**Already guarded** — the raise degrades to a clean mapping failure:
- everything reached from `map_rxn()` (`map_pairs`, `reorder_p_label_map`), which sits inside `try_mapping()`'s `except ValueError`
- `get_new_zmat_2_map`, reached via `combine_coordinates_with_redundant_atoms`, inside `h_abstraction`'s `except (ValueError, SpeciesError)`

**Not guarded** — an incomplete map now surfaces as an uncaught exception rather than as silently wrong indices:
- `map_isomerization_reaction` → `map_general_rxn` → `map_reaction`; `ARCReaction.atom_map` does not catch, so it escapes the property
- `h_abstraction` (`arc/job/adapters/ts/heuristics.py`), which the Heuristics TS adapter calls without a guard

This is the intended trade — an exception beats a corrupted map — but it is a real change in failure mode, and reviewers should decide whether those two paths want their own guard. With the fallbacks added here a gap should not be reachable through the normal paths; the check is a backstop.

---

### Test housekeeping

The XH3 centre is looked up rather than hardcoded at index 13, and the `[CH2]CON` geometries of `test_map_xh2_group()` moved to the fixtures so the tests that need them share them.

### DIFF
--- a/arc/mapping/engine.py
+++ b/arc/mapping/engine.py
@@ -13,15 +13,17 @@ from collections import deque
 from itertools import product
 from typing import TYPE_CHECKING
 
-from arc.common import extremum_list, get_angle_in_180_range, logger, signed_angular_diff
-from arc.exceptions import AtomTypeError, ConformerError, InputError, SpeciesError
+from arc.common import extremum_list, get_angle_in_180_range, is_angle_linear, logger, signed_angular_diff
+from arc.exceptions import AtomTypeError, ConformerError, InputError, SpeciesError, VectorsError
 from arc.family import ReactionFamily
 from arc.molecule import Molecule
 from arc.molecule.resonance import generate_resonance_structures_safely
 from arc.species import ARCSpecies
 from arc.species.conformers import determine_chirality
 from arc.species.converter import compare_confs, sort_xyz_using_indices, xyz_from_data
-from arc.species.vectors import apply_rodrigues_rotation, calculate_dihedral_angle, get_delta_angle
+from arc.species.vectors import (apply_rodrigues_rotation, calculate_dihedral_angle, get_angle, get_delta_angle,
+                                 get_perpendicular_axes, get_vector, get_vector_length, unit_vector)
+from arc.species.zmat import TOL_180
 
 if TYPE_CHECKING:
     from arc.molecule.molecule import Atom
@@ -44,6 +46,9 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
     All indices are 0-indexed.
     If a dict type atom map is returned, it could conveniently be used to map ``spc_2`` -> ``spc_1`` by doing::
         ordered_spc1.atoms = [spc_2.atoms[atom_map[i]] for i in range(len(spc_2.atoms))]
+
+    A ``ValueError`` is raised if the resulting map is not a permutation of the atoms of the two
+    species.
 
     When several superimposable backbone candidates are identified, each is scored by the RMSD
     between the two backbone distance matrices, so candidates tie within ``RMSD_TIE_TOLERANCE``
@@ -139,6 +144,11 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
             atom_maps = dict()
             for i in tied_indices:
                 atom_maps[i] = map_hydrogens(fixed_spcs[i][0], fixed_spcs[i][1], candidates[i])
+                if sorted(atom_maps[i].keys()) != list(range(spc_1.number_of_atoms)) \
+                        or sorted(atom_maps[i].values()) != list(range(spc_2.number_of_atoms)):
+                    raise ValueError(f'The atom map of {spc_1.label} and {spc_2.label} is not a permutation of their '
+                                     f'{spc_1.number_of_atoms} and {spc_2.number_of_atoms} atoms, '
+                                     f'got:\n{atom_maps[i]}')
             if len(tied_indices) > 1:
                 displacements = {i: fixed_spcs[i][0].kabsch(fixed_spcs[i][1],
                                                             [v for k, v in sorted(atom_maps[i].items(),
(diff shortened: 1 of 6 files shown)


arc/mapping/engine.py
```diff
@@ -644,43 +644,101 @@ def _find_preferably_heavy_neighbor(heavy_index: int,
     return None
 
 
+def _anchors_define_a_plane(center_index: int,
```
kfir4444, line 656: Ideally, this would be imported from `arc.species.vectors` for standardization. The implementation issue is that we don't have an entryway function in the `calculate_dihedral` pipeline that takes in 3 vectors and outputs the dihedral angle. I suggest adding a function to `arc/species/vectors.py` (call it `get_vector_dihedral` or something), then change the `get_dihedral` to use it after creating the 3 vectors, and import it here too. This is the best method to ensure SOTA computations.

kfir4444 (commented):
Thanks @calvinp0! please see one comment.

arc/mapping/engine.py
```diff
@@ -12,15 +12,18 @@
 from itertools import product
 from typing import TYPE_CHECKING
 
-from arc.common import convert_list_index_0_to_1, extremum_list, get_angle_in_180_range, logger, signed_angular_diff
+from arc.common import (convert_list_index_0_to_1, extremum_list, get_angle_in_180_range, is_angle_linear, logger,
```
kfir4444: style: parentheses are not required here.

arc/mapping/engine.py
```diff
@@ -865,6 +936,117 @@ def _map_xh3_group(heavy_idx_1: int,
...
+        The e_y and e_z unit vectors, or ``None`` if X and A coincide.
+    """
+    v_xa = get_vector(pivot=center_index, anchor=anchor_index, xyz={'coords': coords})
+    if get_vector_length(v_xa) < 1e-8:
+        return None
+    e_x = np.array(unit_vector(v_xa), dtype=float)
+    reference = np.zeros(3)
+    reference[int(np.argmin(np.abs(e_x)))] = 1.0
+    e_y = np.array(get_normal(get_normal(e_x, reference), e_x), dtype=float)
+    return e_y, np.cross(e_x, e_y)
```
kfir4444: This function needs to be in vectors too.

kfir4444 (commented):
Thanks @calvinp0, see some comments.

arc/mapping/engine.py
calvinp0: done. its in `vectors.get_perpendicular_axes()` after the `get_normal`

kfir4444 (approved):
LGTM! great job @calvinp0!

PR #937, opened 2026-08-02
Make the IRC verdict meaningful, enforce it, and never publish a rate silently from an IRC-invalid TS

calvinp0:
Make the IRC verdict meaningful, enforce it, and never publish a rate silently from an IRC-invalid TS.

This PR combines what were previously two stacked PRs (#937 and #938). They were never independently reviewed, and splitting them made the story harder to follow than it was worth: the verdict semantics, the enforcement that consumes them, and the labelling of anything that still slips through are one change. #938 is closed in favour of this.

## What went wrong

Benchmark reaction `r3_16` (nitroethane ⇌ ethyl nitrite, `intra_NO2_ONO_conversion`) produced a fitted Arrhenius expression of `Ea = 65.8 kJ/mol`. The correct barrier is ~250 kJ/mol — the published rate coefficient was wrong by **~9 orders of magnitude in k at 1000 K**. For scale, the published rate implies a room-temperature half-life of about 23 ms for nitroethane.

ARC had already determined the TS was invalid: `ts_checks['IRC'] = False`, because both optimized IRC endpoints re-perceive as the *product* (the "TS" is a torsional saddle of ethyl nitrite, connecting product↔product). That verdict was written to `restart.yml` — and then the Arrhenius fit was written to `RMG_kinetics.yml`, `rate_plots.pdf` and the kinetics library **with no indication that validation had failed**.

Arkane is not at fault. Its `E0` values (`R1 −81.43`, `P1 −73.74`, `TS −18.07 kJ/mol`) are self-consistent with the fitted expression: `TS − R1 = 63.36` against the fitted `Ea = 65.84`, a 2.5 kJ/mol gap that is unremarkable for an `n = 0.70` fit with Eckart tunnelling. Arkane correctly computed the rate for the structure it was given; the TS search supplied the wrong structure.

## The saddle is falsifiable without the IRC

Three independent lines of evidence say the rejected structure is a torsional saddle of ethyl nitrite, not the migration TS. The first needs no IRC, no atom map and no energy reference at all.

**1. The imaginary frequency is 3–4× too soft.** ARC's rejected saddle has a single imaginary mode at **−270.3 cm⁻¹**. A nitro→nitrite migration TS is a tight, strained C···O···N three-membered arrangement whose reaction coordinate is a heavy-atom bond reorganisation: literature values are **846i cm⁻¹ (B3LYP)** to **1063i cm⁻¹ (CASSCF)**. An alkyl-nitrite O–N torsional saddle is a soft internal rotation: a literature analogue gives **236i**, and an independent GFN2-xTB Hessian computed on the ethyl nitrite O–N torsional saddle gives a single imaginary mode at **−348 cm⁻¹**. −270.3 is squarely in the torsion class and nowhere near the migration class. A curvature that soft cannot be a C–N/C–O bond exchange.

**2. The geometry is ordinary ethyl nitrite.** C–O is **1.428 Å** — fully formed. C and N are **1-3 neighbours through the ether oxygen**: with d(C–O) = 1.428, d(O–N) = 1.40 and ∠C–O–N = 114.3°, the geminal C···N distance *is* **2.372 Å**. There is no stretched or partially broken C–N bond here; there is no C–N bond at all, and 2.376 Å is simply what 1-3 geminal separation looks like. A genuine migration TS has C–N ~1.9–3.4 Å **while** C–O is also partial at ~1.9–3.7 Å — both bonds in flight. Here C–O is complete. The structure sits in the product well, not between the wells.

**3. The barrier height matches a torsion, not an isomerisation.** Measured from the product well the saddle is a **55.7 kJ/mol** barrier (`TS −18.07` against `P1 −73.74`). The measured syn/anti barrier of ethyl nitrite is **ΔG‡ = 46.0 ± 0.4 kJ/mol** (gas-phase ¹H NMR, Conboy/Chauvel/True 1986); CH₃ONO is 45.3 kJ/mol (Van der Veken/Durig 1990). 55.7 is ~10 kJ/mol above the measured ethyl nitrite value — the right magnitude class for this torsion, and nowhere near the ~250 kJ/mol of a C–N-breaking isomerisation. Measured from the reactant well the same structure gives the 63–66 kJ/mol that was published as the reaction barrier.

A re-run that rejected that guess and found the true saddle gives `Ea = 253.639 kJ/mol`, against the RMG NOx2018 library's `253.8`. Both are ~250 kJ/mol. They are *not* the same quantity and should not be compared to a percentage: they are fitted modified-Arrhenius `Ea` parameters from fits with different `n` (the library's `n = 1`), so the closeness of the two numbers is coincidental. The point is the order of magnitude, which is the point at issue.

### 1. `ts_checks['IRC'] = False` now means only "checked and failed"

`check_irc_species_and_rxn` pessimistically set `False` up front and only wrote `True` on success. An early return — `rxn.get_bonds()` raising in the fallback — left `False` in place **having compared nothing**, so `False` conflated *"checked and failed"* with *"could not check"*, which makes it unusable as a gate. It now starts at `None` and writes `False` only where a comparison actually happened. (A second no-comparison path existed purely *because* of the up-front `False`: if `check_xyz_dict()` or `get_reactants_and_products()` raised, the exception propagated out with a "failed" verdict already recorded.)

A negative result from the **isomorphism tier alone** does not write `False` either. That tier perceives each endpoint's molecular graph from coordinates without the expected species' spin or charge, so it can disagree for a chemically correct endpoint. The code already treated that as inconclusive by falling through to the bond-list comparison; the verdict now follows, and stays `None` when that comparison cannot run.

### 2. The bond-list tier no longer invents connectivity

The bond-list tier is the only tier that can write `False`, so it must not perceive bonds that are not there. It was calling `get_bonds_from_dmat()` with the default `n_fragments=1`, which bonds **every hydrogen to its absolute nearest neighbour with no distance test at all** — `CH4 + H` at 50 Å comes back "bonded". Meanwhile `_perceive_irc_fragments()` — the *inconclusive* tier — already passed `n_fragments=0` for exactly this reason. The tiers were inverted: the tier that cannot reject anything used the safe setting, and the tier with the authority to reject used the lossy one.

Both call sites now pass `n_fragments=0`. Reproduced before and after on this branch:

| reaction | before | after |
|---|---|---|
| `O(3P) + H2 ⇌ OH + H` | `False` | `True` |
| `S(3P) + H2 ⇌ SH + H` | `False` | `True` |

These reach the bond-list tier because the triplet atom is not perceived as a triplet, so the isomorphism tier cannot match it. The `r3_16` true negative, the mismatched-species case and the identical-endpoints case are all unchanged.

**Measured exposure.** Triggering this needed *both* the isomorphism tier to fail *and* a bare atom in an endpoint. Across the 509 benchmark reactions, 4 have a bare H and 9 have a bare O/S/C/N, and **the two sets do not intersect** — so benchmark exposure was zero and no benchmark result changes. The risk was to small-molecule users: `O(3P) + H2` is a top-tier combustion reaction.

### 3. The verdict is enforced, and anything that still slips through is labelled

`ts_checks['IRC']` was previously **computed correctly and read nowhere.** It was explicitly exempted — `ts_passed_checks(species=..., exemptions=['E0', 'warnings', 'IRC'])` — and `check_ts`'s docstring still carried `Todo: check IRC`. By contrast an NMD failure **does** act: it logs and calls `switch_ts` to try another guess. There was no IRC equivalent.

- IRC is un-exempted (`exemptions=['E0', 'warnings']`) and the stale `Todo` is dropped.
- New `Scheduler.process_irc_verdict`: on `IRC is False`, log at ERROR and call `switch_ts`, mirroring the NMD path, with the same E0-style exhaustion fallback (restores the verdict and marks the TS unconverged when no guesses remain).
- `check_all_done`'s "stay unconverged" guard extends from `E0` to `['E0', 'IRC']`.

And when a rate is nonetheless produced from an IRC-invalid TS — on restart, or in a project that does not enforce — every artifact carrying it is marked via the new `common.get_ts_validation_comment()`: the Arrhenius `comment`, the `RMG_kinetics.yml` entry, the kinetics-library `longDesc`, and the `rate_plots.pdf` title, each with an `ERROR` logged naming the reaction.

**The rate is deliberately NOT suppressed** — it is diagnostically valuable, and it is how this bug was found. It is computed, recorded, and clearly labelled.

## Three-valued logic — the trap this had to avoid

`ts_checks` values are `True` / `False` / `None`, and `ts_passed_checks` used `not value`, which treats `None` and `False` identically. **Naively un-exempting IRC would have failed every run that skips IRC** (`job_types: irc: false`). So:

```python
if check in exemptions or (check == 'IRC' and value is None):
    continue
```

Scoped to IRC only; `None` semantics for the other keys are unchanged. `IRC is None` is likewise not marked and not suppressed. Many users skip IRC because it is expensive; they are unaffected.

## Is switching the TS after IRC actually supported? Yes — verified

IRC runs later in the pipeline than NMD, so this needed checking. `switch_ts` → `delete_all_species_jobs` resets `output[ts]['paths']`, all `job_types`, `convergence`, discards pending pipe work, **deletes the IRC species spawned from the rejected guess** and sets `irc_label = None` — so `spawn_post_opt_jobs` re-enqueues the full chain for the new guess. No stale state survives. Termination is bounded by `len(ts_guesses)` via `chosen_ts_list`.

This surfaced one latent bug that only appears once switching is enabled: `check_irc_species` is reached while the main loop iterates the *IRC species'* label; rejecting the TS deletes that label, and the loop then did `del self.running_jobs[label]` unconditionally → `KeyError`. Now guarded — and now covered: the guard was necessary *and* sufficient, but removing it left the entire suite green, so `test_schedule_jobs_tolerates_a_label_deleted_while_it_is_iterated` drives `schedule_jobs()` with a `check_all_done()` that deletes the iterated label the same way `delete_all_species_jobs()` does, and reproduces the `KeyError`.

## Reuse before writing

Searched before adding `is_ts_check_exempt()`: `git grep -n "e_elect.*E0"` and `git grep -n "ts_checks\["` across `arc/`, plus the callers of `ts_passed_checks` via the impact graph. That search is what found the problem — the `e_elect`/`E0` exemption was already implemented **twice**, at `arc/checks/ts.py` and in `common.get_ts_validation_comment()`, and the two had already drifted: one indexes `ts_checks['E0']` (raises `KeyError` when the key is absent), the other uses `.get('E0')`. Rather than leave a second copy, the predicate is extracted once into `arc/common.py` — the module that owns it, and the only direction that does not create a cycle, since `arc/plotter.py` → `arc/checks/ts.py` is a real import cycle via `arc.statmech.factory` → `arc.statmech.arkane` → `arc.plotter`. The shared helper takes the `.get()` semantics, and says so in its docstring.

## Measured impact across the completed benchmark runs

- **Exactly one run changes**: `r3_16` — the invalid TS is rejected, the search continues, and the wrong `Ea = 65.8` is no longer published.
- The four runs with `IRC = None` (`r2_11`, `r2_19`, `r3_01`, `r3_10`) are **unaffected by construction**.
- `IRC = True` runs: behaviour identical, one extra INFO line.
- The `n_fragments` correction changes no benchmark run (see the exposure measurement above).

## Test evidence

All measured on this branch, serially, against a baseline established on `origin/main` (`adf58ad9`) in the same worktree:

| | passed | failed | skipped | subtests |
|---|---|---|---|---|
| baseline, `origin/main` | 2745 | 5 | 36 | 208 |
| this branch | **2760** | 5 | 36 | 210 |

The 5 failures are identical in both: pre-existing `arc/job/adapters/torch_ani_test.py` path failures, unrelated to every module touched here. The five touched test files (`ts_test`, `common_test`, `processor_test`, `arkane_test`, `scheduler_test`) run **192 passed, 2 subtests passed**. Each of the three commits is independently green at the same 5-failure baseline (2749 / 2754 / 2760 passed), so the branch is bisectable.

Each fix is covered by a test that fails without it, verified by reverting the fix in place:

- `test_check_irc_bond_list_tier_does_not_bond_a_dissociated_atom` → `AssertionError: False is not True` on both subtests with `n_fragments` restored to the default.
- `test_schedule_jobs_tolerates_a_label_deleted_while_it_is_iterated` → `KeyError: 'IRC_TS_irc_deleted_label_1'` with the guard removed.

`test_check_irc_wrong_species` asserted with `assertFalse()`, which also passes on `None` — the very conflation this PR removes — and is changed to `assertIs(..., False)`.

## Known limitations (not fixed here)

- **Degenerate reactions get no IRC protection.** Tier 1 short-circuits on a graph match, so for a reaction whose reactants and products are isomorphic — `CH3 + CH4 → CH4 + CH3`, `H + H2 → H2 + H` — the check returns `True` even when *both* endpoints are the reactant. This is pre-existing and permissive rather than newly introduced, but it means symmetric H-abstractions, degenerate H-shifts and self-disproportionations are not protected by this gate.
- `infer_multiplicity`'s triplet-atom table has no O or S, which is why the two reactions in the table above fall through to the bond-list tier at all. Fixing that would make the isomorphism tier succeed for them; it is a separate change.
- The 1.2× bond-length threshold in `get_bonds_from_dmat` misses genuinely long bonds such as N₂O₄'s N–N.
- `check_ts(checks=['IRC'])` is an accepted no-op — the IRC check is performed by `check_irc_species_and_rxn`, not `check_ts`.
- `process_irc_verdict` reads the verdict off `ts_checks` rather than taking a returned value.
- The IRC verdict is **not** exported to `output.yml`: `_rxn_to_dict` uses an explicit kinetics allow-list that excludes `ts_validation` and `comment`, and `ts_checks` appears nowhere in `output.py`. A consumer of `output.yml` still cannot see whether a rate came from a validated TS.
- The TCKDB upload gate is out of scope: `arc/tckdb/` does not exist on `main`.

## Note for review

`except Exception` in the fallback was **kept** (not narrowed) and made loud rather than silent. `get_bonds()` reaches `atom_map` → `map_reaction`, `spc.copy()` and `.index()` lookups, so the raisable surface is genuinely unbounded; narrowing it wrongly would turn a benign "cannot atom-map" into a scheduler crash mid-run.

### DIFF
--- a/arc/statmech/arkane_test.py
+++ b/arc/statmech/arkane_test.py
@@ -6,11 +6,14 @@ This module contains unit tests for ARC's statmech.arkane module
 """
 
 import os
+import re
 import shutil
 import tempfile
 import unittest
+from unittest.mock import patch
 
-from arc.common import ARC_PATH, ARC_TESTING_PATH
+from arc.checks.common import TS_IRC_FAILED_MARKER
+from arc.common import ARC_TESTING_PATH
 from arc.exceptions import InputError
 from arc.level import Level
 from arc.reaction import ARCReaction
@@ -32,8 +35,14 @@ from arc.statmech.arkane import (
     check_arkane_aec,
     check_arkane_bacs,
     get_arkane_model_chemistry,
+    parse_e0,
+    parse_reaction_kinetics,
+    parse_thermo_data_block,
+    run_arkane,
+    _classify_arkane_stderr,
+    _parse_conformer_statmech,
+    _summarize_arkane_stderr,
 )
-from unittest.mock import patch
 
 
 class TestEnumerationClasses(unittest.TestCase):
@@ -429,9 +438,33 @@ class TestArkaneAdapter(unittest.TestCase):
 class TestArkaneOutputParsing(unittest.TestCase):
     """Tests for parsing functions that read Arkane output.py content."""
 
+    kinetics_output_content = """
+conformer(label='TS0', E0=(50.0, 'kJ/mol'), modes=[], spin_multiplicity=2, optical_isomers=1)
+
+kinetics(
+    label = 'nitroethane <=> ethyl_nitrite',
+    kinetics = Arrhenius(
+        A = (5.0, 's^-1'),
+        n = 1.0,
+        Ea = (20.0, 'kJ/mol'),
+        T0 = (1, 'K'),
+        Tmin = (300, 'K'),
+        Tmax = (2000, 'K'),
+        comment = 'Fitted to 10 data points; dA = *|/ 1.1, dn = +|- 0.01, dEa = +|- 0.1 kJ/mol',
+    ),
+)
+"""
+
+    @staticmethod
+    def isomerization_reaction() -> ARCReaction:
+        """Build the real nitroethane <=> ethyl nitrite reaction with a TS species named 'TS0'."""
+        rxn = ARCReaction(r_species=[ARCSpecies(label='nitroethane', smiles='CC[N+](=O)[O-]')],
+                          p_species=[ARCSpecies(label='ethyl_nitrite', smiles='CCON=O')])
+        rxn.ts_species = ARCSpecies(label='TS0', is_ts=True)
+        return rxn
+
     def test_parse_e0(self):
         """Test parse_e0 extracts E0 from conformer blocks."""
-        from arc.statmech.arkane import parse_e0
         content = """
 conformer(
     label = 'CH4',
@@ -445,14 +478,11 @@ conformer(
         self.assertIsNone(parse_e0('missing_species', content))
 
     def test_parse_e0_positive(self):
-        from arc.statmech.arkane import parse_e0
         content = "conformer(label='CHO', E0=(44.0971, 'kJ/mol'), modes=[], spin_multiplicity=2, optical_isomers=1)"
         self.assertAlmostEqual(parse_e0('CHO', content), 44.0971)
 
     def test_parse_conformer_statmech(self):
         """Test extraction of external_symmetry and optical_isomers."""
-        from arc.statmech.arkane import _parse_conformer_statmech
-        from unittest.mock import MagicMock
         content = """
 conformer(
     label = 'H2O',
@@ -467,18 +497,13 @@ conformer(
     optical_isomers = 1,
 )
 """
-        spc = MagicMock()
-        spc.label = 'H2O'
-        spc.optical_isomers = None
-        spc.external_symmetry = None
+        spc = ARCSpecies(label='H2O', smiles='O')
         _parse_conformer_statmech(spc, content)
         self.assertEqual(spc.optical_isomers, 1)
         self.assertEqual(spc.external_symmetry, 2)
 
     def test_parse_conformer_statmech_linear(self):
         """Test with LinearRotor."""
-        from arc.statmech.arkane import _parse_conformer_statmech
-        from unittest.mock import MagicMock
         content = """
 conformer(
     label = 'CO2',
@@ -488,23 +513,18 @@ conformer(
     optical_isomers = 1,
 )
 """
-        spc = MagicMock()
-        spc.label = 'CO2'
-        spc.optical_isomers = None
-        spc.external_symmetry = None
+        spc = ARCSpecies(label='CO2', smiles='O=C=O')
         _parse_conformer_statmech(spc, content)
         self.assertEqual(spc.external_symmetry, 2)
         self.assertEqual(spc.optical_isomers, 1)
 
     def test_parse_reaction_kinetics_with_uncertainties(self):
         """Test that dA, dn, dEa, n_data_points are parsed from the comment."""
-        from arc.statmech.arkane import parse_reaction_kinetics
-        from unittest.mock import MagicMock
         content = """
 conformer(label='TS0', E0=(50.0, 'kJ/mol'), modes=[], spin_multiplicity=2, optical_isomers=1)
 
 kinetics(
-    label = 'A + B <=> C + D',
+    label = 'CH4 + OH <=> CH3 + H2O',
     kinetics = Arrhenius(
         A = (1.2e10, 'cm^3/(mol*s)'),
         n = 2.5,
@@ -516,11 +536,11 @@ kinetics(
     ),
 )
 """
-        rxn = MagicMock()
-        rxn.label = 'A + B <=> C + D'
-        rxn.ts_species = MagicMock()
-        rxn.ts_species.label = 'TS0'
-        rxn.ts_species.e0 = None
+        rxn = ARCReaction(r_species=[ARCSpecies(label='CH4', smiles='C'),
+                                     ARCSpecies(label='OH', smiles='[OH]')],
+                          p_species=[ARCSpecies(label='CH3', smiles='[CH3]'),
+                                     ARCSpecies(label='H2O', smiles='O')])
+        rxn.ts_species = ARCSpecies(label='TS0', is_ts=True)
         parse_reaction_kinetics(rxn, content)
         self.assertIsNotNone(rxn.kinetics)
         self.assertAlmostEqual(rxn.kinetics['A'][0], 1.2e10)
@@ -534,13 +554,11 @@ kinetics(
 
     def test_parse_reaction_kinetics_no_comment(self):
         """Kinetics without a comment should still parse A, n, Ea."""
-        from arc.statmech.arkane import parse_reaction_kinetics
-        from unittest.mock import MagicMock
         content = """
 conformer(label='TS0', E0=(50.0, 'kJ/mol'), modes=[], spin_multiplicity=2, optical_isomers=1)
 
 kinetics(
-    label = 'X <=> Y',
+    label = 'nitroethane <=> ethyl_nitrite',
     kinetics = Arrhenius(
         A = (5.0, 's^-1'),
         n = 1.0,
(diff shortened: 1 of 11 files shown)


arc/statmech/arkane_test.py
```diff
@@ -561,6 +578,38 @@ def test_parse_reaction_kinetics_no_comment(self):
         self.assertAlmostEqual(rxn.kinetics['n'], 1.0)
         self.assertNotIn('dA', rxn.kinetics)
 
+    def test_parse_reaction_kinetics_marks_an_irc_invalid_ts(self):
+        """Kinetics of a TS that failed the IRC check are reported, and are labeled as invalid."""
+        from arc.statmech.arkane import parse_reaction_kinetics
```
alongd: please import at the module level (also same comment for other tests here)

arc/statmech/arkane_test.py
```diff
@@ -561,6 +578,38 @@ def test_parse_reaction_kinetics_no_comment(self):
...
+        self.assertIn(TS_IRC_FAILED_MARKER, rxn.kinetics['comment'])
+        self.assertIn('Fitted to 10 data points', rxn.kinetics['comment'])
+
+    def test_parse_reaction_kinetics_does_not_mark_an_unchecked_or_valid_ts(self):
+        """Kinetics of a TS for which the IRC check was not performed or was passed are not labeled."""
+        from arc.statmech.arkane import parse_reaction_kinetics
+        from unittest.mock import MagicMock
+        content = self.kinetics_output_content
+        for irc_value in [None, True]:
+            rxn = MagicMock()
```
alongd: if possible, try to use real ARCReaction objects instead of a mock

alongd: same for the test above

arc/common.py
```diff
@@ -1934,6 +1936,58 @@ def convert_to_hours(time_str:str) -> float:
     return h + m / 60 + s / 3600
 
 
+def is_ts_check_exempt(check: str,
```
alongd: why are these functions under arc.common? I think they belong in arc.checks.ts instead. is there an import issue?

alongd (commented):
please see some comments

arc/common.py
calvinp0: yes. arc/checks/ts.py imports arc.statmech.factory, whose package __init__ import arc.statmech.arkane, and arkane is one of the two consumers of this helper. the other is arc/plotter.py whic gets pulled in earl via arc.job.trsh --> arc.species.conformers. with both importing from arc.checks.ts, plain import ar dies with ImportError: cannot import name get_ts_validation_comment from partially initialised modeul arc.checks.ts. So i have no moned them into arc/checks/common.py instead - smae package and it only imports datetime. can move elsewhere in arc.checks if preferred

alongd (approved):
Thanks!

PR #938, opened 2026-08-02
Enforce the IRC check — reject a TS whose IRC endpoints do not match the wells

calvinp0:
> **Top of stack #939 — depends on #937 (`fix_irc_gated_kinetics`), which must merge first.**
> Enforcing IRC before #937 lands would reject valid TSs whenever the check merely *could not run*, because `False` currently also means "could not check". This PR's diff shows only its own layer.
>
> **This is where the user-visible behaviour change ships.** Once this lands, a fresh run never reaches #937's failed-validation markers: `processor.py` drops any reaction whose TS is unconverged, which is what this PR sets for an IRC-invalid TS. #937 remains meaningful on restart, for projects that do not enforce, and as defence in depth — but on a clean run the TS is rejected here, before a rate is ever fitted.

## What went wrong

`ts_checks['IRC']` was **computed correctly and read nowhere.** It is written in `arc/checks/ts.py` and consumed by no production code. It was explicitly exempted:

```python
ts_passed_checks(species=reaction.ts_species, exemptions=['E0', 'warnings', 'IRC'])
```

and `check_ts`'s docstring still carried `Todo: check IRC`. By contrast, an NMD failure **does** act — it logs and calls `switch_ts` to try another guess. There was no IRC equivalent.

Consequence: `r3_16` accepted a TS that connects product↔product, and published a rate wrong by ~9 orders of magnitude (see #937). A re-run that happened to reject that guess found the true saddle and produced `Ea = 253.639 kJ/mol`, matching the literature.

## The fix

- Un-exempt IRC (`exemptions=['E0', 'warnings']`), and drop the stale `Todo`.
- New `Scheduler.process_irc_verdict`: on `IRC is False`, log at ERROR and call `switch_ts`, mirroring the existing NMD path, with the same E0-style exhaustion fallback (restores the verdict and marks the TS unconverged when no guesses remain).
- `check_all_done`'s "stay unconverged" guard extended from `E0` to `['E0', 'IRC']`.

## Three-valued logic — the trap this had to avoid

`ts_checks` values are `True` / `False` / `None`, and `ts_passed_checks` used `not value`, which treats `None` and `False` identically. **Naively un-exempting IRC would have failed every run that skips IRC** (`job_types: irc: false`). So:

```python
if check in exemptions or (check == 'IRC' and value is None):
    continue
```

Scoped to IRC only; `None` semantics for the other keys are unchanged.

## Is switching the TS after IRC actually supported? Yes — verified

IRC runs later in the pipeline than NMD, so this needed checking. `switch_ts` → `delete_all_species_jobs` resets `output[ts]['paths']`, all `job_types`, `convergence`, discards pending pipe work, **deletes the IRC species spawned from the rejected guess** and sets `irc_label = None` — so `spawn_post_opt_jobs` re-enqueues the full chain for the new guess. No stale state survives. Termination is bounded by `len(ts_guesses)` via `chosen_ts_list`.

This surfaced one latent bug that only appears once switching is enabled: `check_irc_species` is reached while the main loop iterates the *IRC species'* label; rejecting the TS deletes that label, and the loop then did `del self.running_jobs[label]` unconditionally → `KeyError`. Now guarded.

## Measured impact across all 61 completed benchmark runs

- **Exactly one run changes**: `r3_16` — the invalid TS is rejected, the search continues, and the wrong `Ea = 65.8` is no longer published.
- The four runs with `IRC = None` (`r2_11`, `r2_19`, `r3_01`, `r3_10`) are **unaffected by construction**.
- `IRC = True` runs: behaviour identical, one extra INFO line.

## Evidence

60 tests pass (`ts_test`, `scheduler_test`), including exhaustion-termination and the `None` → no-switch case. Full suite: `2501 passed, 5 failed` — the 5 are pre-existing TorchANI path failures, unrelated to the touched modules.

**174 tests pass with both stack layers applied together** — the first time these two changes were exercised as a unit.

### DIFF
--- a/arc/checks/ts.py
+++ b/arc/checks/ts.py
@@ -507,6 +507,16 @@ def check_irc_species_and_rxn(xyz_1: dict,
     bond-list comparison if perception fails for either endpoint or if the expected
     reactant/product ``Molecule`` objects are unavailable.
 
+    Sets ``rxn.ts_species.ts_checks['IRC']`` to ``True`` if the endpoints match the reactants
+    and products, to ``False`` only if a comparison was actually performed and the endpoints
+    did not match, and leaves it as ``None`` (unknown) if no comparison could be performed
+    (e.g., the expected connectivity of the reaction is unavailable). ``False`` means
+    "checked and failed", never "could not check".
+
+    A negative isomorphism result alone does not set ``False``: it is treated as inconclusive
+    and the bond-list comparison decides. The verdict remains ``None`` if that comparison
+    cannot be carried out.
+
     Args:
         xyz_1 (dict): The coordinates of IRC species 1.
         xyz_2 (dict): The coordinates of IRC species 2.
@@ -514,7 +524,7 @@ def check_irc_species_and_rxn(xyz_1: dict,
     """
     if rxn is None:
         return None
-    rxn.ts_species.ts_checks['IRC'] = False
+    rxn.ts_species.ts_checks['IRC'] = None
     xyz_1, xyz_2 = check_xyz_dict(xyz_1), check_xyz_dict(xyz_2)
 
     # Primary check: molecular graph isomorphism
@@ -541,8 +551,10 @@ def check_irc_species_and_rxn(xyz_1: dict,
     # Fallback: bond-list connectivity comparison
     try:
         r_bonds, p_bonds = rxn.get_bonds()
-    except Exception:
-        logger.debug('Could not get reaction bonds for IRC fallback check.')
+    except Exception as e:
+        logger.warning(f'Could not get the reaction bonds of {rxn} for the IRC fallback check, '
+                       f'got:\n{e.__class__.__name__}: {e}\n'
+                       f'The IRC check of {rxn.ts_species.label} is therefore left undetermined.')
         return
     dmat_1, dmat_2 = xyz_to_dmat(xyz_1), xyz_to_dmat(xyz_2)
     dmat_bonds_1 = get_bonds_from_dmat(dmat=dmat_1, elements=xyz_1['symbols'])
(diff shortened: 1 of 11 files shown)


calvinp0:
Closing in favour of #937, which now carries both layers.

These two PRs were a stack: #937 made `ts_checks['IRC']` mean "checked and failed" rather than also "could not check", and this PR consumed that verdict to actually reject a TS. Neither had received any human review, so there was nothing to lose by combining them, and two things to gain:

- **They are one change.** Enforcing the verdict is only safe *because* the verdict was made three-valued first; splitting them meant neither PR could be evaluated on its own terms, and this PR's stack note had to spend a paragraph explaining that its own diff was not the whole story.
- **A must-fix spanned both.** Two independent reviews found that the bond-list tier — the only tier permitted to write `False`, and therefore the only one that can reject a TS — was calling `get_bonds_from_dmat()` with the default `n_fragments=1`, which bonds every H to its absolute nearest neighbour with no distance test. That made `O(3P) + H2 <=> OH + H` and `S(3P) + H2 <=> SH + H` fail the check with correct IRC endpoints. Both call sites now pass `n_fragments=0`, with a regression test. That fix lives in `arc/checks/ts.py`, which both PRs touched — it could not have been landed cleanly in either one alone without violating the one-file-one-commit rule.

#937 has been rewritten as a single coherent narrative and force-pushed with three commits over `main` with disjoint file sets (`common`/`plotter`/`processor`/`arkane`; `checks/ts`; `scheduler`). Everything that was in this PR's body — the enforcement layer, the three-valued trap, the `switch_ts` verification, the `running_jobs` `KeyError` (now with a regression test that reproduces it) and the benchmark impact — is in #937.

The branch `fix_enforce_irc_check` is left in place; nothing is lost.

PR #940, opened 2026-08-02
Stop conflating a TS guess's identity with its position in a list

calvinp0:
> **Layer 2 of a four-PR stack** — sits on #817. This PR's diff shows only its own layer.
>
> ```
> #943  Troubleshoot the opt job that failed, not the highest-numbered one
>  └── #941  Only let a geometry-determining job reject a TS guess
>       └── #940  Stop conflating a TS guess's identity with its position   ← you are here
>            └── #817  Match TS guesses by identity, not list position      (base: main)
> ```

## What went wrong

#817 established that a TS guess has a stable identity: it fixed the places where a guess's *position in a list* was written onto its *identity*, and made the identity itself unique. This PR is the audit that followed: the same conflation exists in several more places, because **four different integers are used interchangeably to mean "which guess"**:

| integer | what it actually is |
|---|---|
| `TSGuess.index` | the guess's stable identity / provenance |
| position in `species.ts_guesses` | ordering, changes when guesses are clustered or dropped |
| `conformer_index` | position in `successful_tsgs`, a *different* list |
| the `i` in a `tsg<i>` job name | which **adapter** was dispatched — not a guess at all |

They are all `int`, so nothing catches a mix-up. One adapter can contribute several guesses or none, and guesses from different adapters interleave, so the adapter number is not even an index into anything.

The worst instance is `arc/scheduler.py`, which used an adapter dispatch number directly as a list subscript:

```python
ts_guesses[get_i_from_job_name(job.job_name)]
```

When the adapter number exceeds the number of guesses generated so far this raises `IndexError` and kills the run; when it doesn't, it silently annotates an unrelated guess.

## The fix

Match guesses by **identity** (`TSGuess.index`) rather than by list position at the audited sites, across `arc/scheduler.py` and `arc/output.py`.

## Why it matters

Provenance is only meaningful if it survives. Guess indices are how we attribute a solved TS to the adapter that produced it — the statistic the benchmark's per-adapter comparison is built from. Every position-as-identity site is a place that attribution can silently move to the wrong guess.

## Evidence

Regression tests accompany the changed sites. **380 tests pass across all four stack layers applied together** (`species_test`, `scheduler_test`, `output_test`, `job/pipe`), run serially — this suite is flaky under `xdist` because its workers collide on shared `Projects/` directories.

### DIFF
--- a/arc/job/adapters/ts/autotst_ts.py
+++ b/arc/job/adapters/ts/autotst_ts.py
@@ -277,7 +277,7 @@ class AutoTSTAdapter(JobAdapter):
                                                        execution_time=tok,
                                                        xyz=xyz,
                                                        success=True,
-                                                       index=len(rxn.ts_species.ts_guesses),
+                                                       index=rxn.ts_species.next_ts_guess_index(),
                                                        )
                                     rxn.ts_species.ts_guesses.append(ts_guess)
                                     save_geo(xyz=xyz,
@@ -294,7 +294,7 @@ class AutoTSTAdapter(JobAdapter):
                                                t0=tic,
                                                execution_time=tok,
                                                success=False,
-                                               index=len(rxn.ts_species.ts_guesses),
+                                               index=rxn.ts_species.next_ts_guess_index(),
                                                )
                             rxn.ts_species.ts_guesses.append(ts_guess)
                             i += 1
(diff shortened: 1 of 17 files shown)


calvinp0:
Folded into #817 — same defect class, same files; the split only produced duplicated hunks.

The three sites this PR audited (`Scheduler.troubleshoot_ess` matching on the adapter that produced each guess, `Scheduler.run_ts_conformer_jobs` reporting the optimized guess as its own provenance, and `output._get_ts_imag_freq` looking up by identity) are carried in #817 unchanged, along with this PR's tests. Nothing was dropped.

Stack #942 has been dissolved: #817 keeps base `main`, and #941 and #943 are now independent PRs off `main`.

PR #944, opened 2026-08-02
Don't emit NaN coordinates for a collinear zmat reference

calvinp0:
## What went wrong

`zmat_to_coords()` could return a geometry containing `NaN` **without failing**, and `xyz_to_zmat()` could store a `NaN` in `zmat['vars']`. Three defects sit behind that, all in how an atom is placed against its zmat reference atoms.

They are **one commit**, because they are in one file — and they could not usefully be split anyway. Each hides the next, and two of the three are harmful on their own. Measured over 1,200 relabelled and reoriented ketene and propyne geometries, counting how many put **every** dummy atom at its own declared `RX = 1.0 Å` / `AX = 90°`:

| | dummies correctly placed | hard failures | worst round-trip error |
|---|---|---|---|
| `origin/main` | 106 / 1200 | 0 | 4.23 Å |
| guard only | 106 / 1200 | 120 | 3.73 Å |
| guard + index space | 334 / 1200 | 120 | 3.20 Å |
| guard + frame | 225 / 1200 | 0 | 1.92 Å |
| **all three** | **1200 / 1200** | **0** | **0.0000 Å** |

### 1. Index space

`add_dummy_atom()` is handed `coords`, the `'coords'` part of the xyz dict, indexed by the **mol/xyz** atom order. It passed that straight to `_add_nth_atom_to_coords()`, which resolves its A/B/C reference atoms out of the zmat's own R/A/D parameter strings — i.e. in **zmat** indices — and then used them to index the xyz-ordered list. The function's own docstring notes the two spaces differ.

They coincide only for the identity atom order, and **a dummy atom guarantees they diverge from its own index on**, since every later atom sits one zmat index further along than its xyz index. Geometry read back from an ESS carries an arbitrary atom order, so this is the ordinary case, not a contrived one.

The dummy was therefore placed against three atoms chosen **by number rather than by identity**. On the six-dummy polyyne `OC#CC#CC#CS` already in `zmat_test.py`, not one of the six dummies honoured its own declared internals: against 1.0 Å / 90°, the placements measured up to **4.54 Å and 167°**. On a linear backbone the misread triad is frequently three atoms of the chain itself, whose cross product vanishes — so the dummy landed at `(nan, nan, nan)` and every later dihedral measured against it became `nan`.

### 2. Frame

The `i == 2` branch placed the third atom at

```python
c_y = bc_length * math.sin(alpha)
c_z = b_z - bc_length * math.cos(alpha) if b_z else bc_length * math.cos(alpha)
coords.append((0.0, c_y, c_z))
```

That is the **canonical** frame — atom 0 at the origin, atom 1 on +z — hard-coded, right down to the `if b_z` test that picks the sign of the cos term by asking whether atom B happens to sit at `z = 0`. Every other branch from index 2 onwards places its atom relative to whatever is already in `coords`, so the function was frame-covariant everywhere except here.

`zmat_to_coords()` never notices: it builds `coords` from scratch and its atoms 0 and 1 really are at the origin and on +z. **`add_dummy_atom()` does notice** — it supplies the coordinates of a real molecule, in the molecule's own frame. A dummy marking a linear segment that begins at the *first* zmat atom is the third zmat atom, so it was written into those lab-frame coordinates at `(0, 1.0, b_z)` no matter where the molecule was. Measured for allene by reading back what `add_dummy_atom()` actually wrote:

| backbone on | `RX`, should be 1.0 Å | `AX`, should be 90.0° |
|---|---|---|
| z | 1.000000 | 90.0000 |
| y | **0.308000** | **0.0000** |
| x | **1.646470** | **37.3989** |

On z the canonical and lab frames coincide, which is why this was never seen. On x the marker is silently misplaced — no NaN, no warning, just a wrong reference for every dihedral that uses it. On y it lands **on the molecular axis** (`AX = 0°`), making the reference triad of the *next* dummy genuinely collinear; `vectors.calculate_param()` then stored a `NaN` in `zmat['vars']` — `DX_4_0_1_2` for ketene, `DX_4_2_1_0` for allene. A `NaN` that is already in `vars` is upstream of any placement guard.

### 3. A collinear reference produced NaN coordinates

For atom D the reference frame comes from the normal of the A-B-C plane:

```python
n = np.cross(ab, ubc)
un = n / vectors.get_vector_length(n)
```

Exactly collinear A-B-C makes `n` the zero vector and `un` `NaN`, reported only as `RuntimeWarning: invalid value encountered in divide`.

**The NaN is then laundered into zeros.** `translate_to_center_of_mass()` keeps each coordinate as `x[i] if abs(x[i]) > 1e-10 else 0.0`, and `abs(nan) > 1e-10` is `False`, so **every NaN coordinate is silently rewritten to `0.0`**. A guard written to clean up rounding dust reclassifies "not a number" as "small". The result is not a crash and not a `NaN` — it is a plausible-looking geometry with every atom on the origin, which would be handed to an ESS as coincident nuclei.

## The fix

**Index space.** The coordinates are translated into zmat order before the call, and only the resulting dummy position is appended back onto the xyz-ordered list. The `'X<i>'` map value already carries the dummy's index in that list; `map_index_to_int()` decodes it — the same decoding `vectors.calculate_dihedral_angle()` performs — so the two lists stay in step for any number of dummies.

**Frame.** Atom 2 is placed from the frame it is given:

```
C = B + R * (cos(alpha) * u_BA + sin(alpha) * u_perp)
```

Two atoms fix an axis but not a plane, so only the **azimuth** about the 0-1 axis is a convention; it comes from `vectors.get_perpendicular_unit_vector()` of that axis. That the azimuth is free is measured, not assumed: substituting a different valid perpendicular changes the round-tripped geometry of 5,152 cases by at most **2.2e-11 Å**.

`get_perpendicular_unit_vector()` returns `+y` for the canonical axis, so the previous choice is reproduced, and the new expression is **bit-identical to the old formula for every canonical frame** — 500,016 randomized cases plus the `alpha = 0 / 180 / 360` boundaries, zero differences. The third atom of every zmat built by `zmat_to_coords()` is unaffected to the last bit; only `add_dummy_atom()`, the one caller supplying a non-canonical frame, sees a change.

**Guard.** A collinear A-B-C has no reference plane, so the dihedral about it is genuinely undefined: any vector perpendicular to BC is an equally valid reference, and D's position is fully determined by the remaining distance and angle. `get_perpendicular_unit_vector()` supplies one deterministically, by crossing with the Cartesian axis the input is least aligned with, which cannot itself be degenerate. The guard triggers on the result not being finite rather than on a length threshold, so it **fires exactly where the previous code produced NaN and nowhere else**.

The guard is last, and least, of the three. **On its own it is a regression**, which is the main reason the three are landing together:

- On the ketene corpus it fires **120 times during reconstruction**, and all 120 then raise `VectorsError: vector must be finite` — a crash where `origin/main` returned a (wrong) geometry.
- On the propyne corpus it fires 11 times during *generation*, and there is 1 case where `origin/main` stored a `NaN` and the guard alone instead returns a finite geometry that is wrong by more than 1e-4 Å.
- Every one of those firings is a symptom of the two defects underneath, not of a genuinely degenerate input.

In fairness to it, I could not reproduce a case where the guard alone is *numerically worse* than `origin/main` — over 1,200 cases that count is 0, because `main`'s laundered collapse is itself a large error. The objection is that it substitutes an arbitrary-plane answer for a loud one, and floods the log.

With the other two fixed, **the guard fires zero times in either direction** across 6,230 round trips plus the 1,200-case permutation corpus. It is kept because `zmat_to_xyz()` is public and accepts Z-matrices ARC did not build, where a degenerate reference is a legitimate input, and because a linear segment can genuinely go unrecognized when the connectivity is unavailable (`mol=None`). Its message now says exactly that, rather than asserting the reference atoms must have been misresolved.

## Measurements

Two round-trip sweeps of xyz → `xyz_to_zmat` → `zmat_to_xyz`, comparing the interatomic-distance matrix (invariant under every isometry) and superposing with `converter.kabsch` (proper rotations only, so a reflection cannot hide):

- **Sweep A** — 77 RDKit-embedded molecules × 7 orientations (as-is, principal axis snapped to x/y/z, 3 random rotations) × {with `mol`, without} = **1,078 cases**.
- **Sweep B** — 28 linear-segment molecules × 46 orientations (6 axis-snapped, 40 random) × 4 atom orders (identity + 3 shuffles) = **5,152 cases**.
- **Permutation corpus** — ketene over all 120 relabelings and propyne over 120, each × 5 orientations = **1,200 cases**, additionally instrumenting `add_dummy_atom()` to check every dummy against its own declared internals, and counting guard firings separately for generation (`xyz_to_zmat`) and reconstruction (`zmat_to_xyz`).

| | `origin/main` | this PR |
|---|---|---|
| NaN in `zmat['vars']` (sweeps A+B) | 4 | **0** |
| geometries collapsed onto the origin | 2 | **0** |
| guard firings, generation / reconstruction | n/a | **0 / 0** |
| worst distance-matrix error | **3.42 Å** | **0.036 Å** |
| dummies at their declared 1.0 Å / 90° | 106 / 1200 | **1200 / 1200** |

Per case, sweep B: **5,118 of 5,152 bit-for-bit unchanged, 34 improve, 0 regress.** Sweep A: 17 improve, 0 regress. I looked specifically for a previously-correct geometry made worse and there is none. The residual 0.036 Å is the near-collinear regime named under "still out of scope", unchanged from `origin/main`.

## Fixture changes

Two hard-coded expectations in `test_zmat_from_xyz` move. Both are dihedrals measured **against a dummy atom**, and a dummy's azimuth about the linear axis it decorates is a **free gauge** — correcting where the dummy sits rotates the zero they are measured from.

| var | before | after | old vs new, Kabsch | old vs new, distance matrix | each vs the source geometry |
|---|---|---|---|---|---|
| `DX_14_0_1_12` (`OC#CC#CC#CS`) | 23.389 | **180.728** | 2.4e-07 Å | 8.9e-16 Å | 8.806e-04 Å, both |
| `DX_8_7_5_6` (`c4linear`) | 360.0 | **217.028** | 0.0 Å | 3.6e-15 Å | 1.014e-02 Å, both |

Neither is an internal coordinate of the molecule at all. Sweeping `DX_14_0_1_12` over 0°, 23.389°, 90°, 180°, 270° and both stored values reconstructs **the same geometry** every time, to within 3.4e-07 Å under proper rotations. It is exactly determined and frame-dependent — not noise-sensitive, and not ill-conditioned: its reference angle O-C-X is 159.9°, twenty degrees off collinear.

## Two commits

1. **`vectors.py`** — adds `get_perpendicular_unit_vector()`. This construction is open-coded at more than a dozen sites across `converter.py` and the transition-state `linear_utils` package under several inconsistent criteria; two of them, in `converter.py`, test `np.allclose(unit, [1, 0, 0])`, which matches only `+x` and returns `[nan, nan, nan]` for a bond on that axis. Choosing the axis by magnitude floors the cross product at `sqrt(2/3) = 0.816` instead of the `0.436` the common `abs(dot) > 0.9` idiom allows. It lives in `vectors.py` because that module owns geometry maths and is the intended Cython target. **The existing sites are deliberately not converted here** — most are fallback branches in transition-state guess code, where changing which arbitrary plane is chosen changes the guess geometry. That needs its own branch and review.
2. **`zmat.py`** — all three fixes above, plus tests.

`fix_dummy_atom_index_space` (`a90fe782`), which carried the index-space fix as a standalone branch, is **folded into this PR** and will not be opened separately. Landing it apart would have moved `DX_14_0_1_12` a second time and left the sequencing between the two branches to be resolved by hand.

## Scope: what is still not fixed

- **Near-collinear references still degrade silently.** At `|n| ~ 1e-16` the frame is built from cancelling digits and the realized distance and angle drift from the requested ones with no warning — up to 0.20 Å / 18.7° in randomized testing, and it is what the residual 0.036 Å above consists of. The guard does not fire there. Orthogonalizing `un` against `ubc` fixes the whole regime unconditionally and is a no-op on good input, but it changes the stored coordinates of an existing test fixture, so it belongs in its own branch with a chemistry review.
- **A non-finite value that reaches `zmat['vars']` from any other source still bypasses the guard.** The guard inspects the reference plane at placement time; it cannot see a `NaN` that was already stored. This PR removes the two sources we know of, but it does not add a check that `xyz_to_zmat()` never stores one, and `zmat_to_coords()` does not reject a `vars` entry that is not finite.
- **`get_perpendicular_unit_vector()` raises `ZeroDivisionError`, not `VectorsError`, below ‖v‖ ≈ 1e-162**, where the squared length underflows. This was raised in review as a broken contract; it is not — the helper's docstring documents both, deliberately and by name. It is left as documented rather than converted: the threshold is unreachable for any quantity in Ångström, and converting it would mean reopening `vectors.py`, whose sole commit is otherwise untouched by this round.

## Tests

Every new test was checked against the unfixed code. Across the three test classes that cover these defects, subtest failures are: `origin/main` **153**, guard only **146**, guard + index space **146**, guard + frame **6**, all three **0**.

| test | fails without the fix |
|---|---|
| `test_every_relabeling_of_ketene_round_trips` | yes — `DX_4_2_1_0 is nan` on `main`; on guard+frame, `1.9207 not less than 0.0001` for 6 of the 120 relabelings |
| `test_an_explicit_atom_order_round_trips` | yes — `zmat var DX_6_5_3_4 is nan` |
| `test_a_dummy_is_mapped_to_its_index_in_the_augmented_coordinates` | pins the invariant the placement reads through |
| `test_third_atom_is_placed_relative_to_the_existing_frame` | yes — `0.3999999 != 1.4` (distance), `0.0 != 90.0` (angle); 23 of its 30 subtests |
| `test_a_leading_dummy_atom_does_not_land_on_the_molecular_axis` | yes — `zmat var DX_4_2_1_0 is nan` |
| `test_ketene_on_the_y_axis_round_trips` | yes — `zmat var DX_4_0_1_2 is nan` |
| `test_canonical_frame_placement_is_unchanged` | mutation-tested (below) |
| `test_propyne_round_trip_is_not_nan_or_collapsed` | yes — `zmat var DX_6_5_3_4 is nan` |
| `test_collinear_reference_does_not_produce_nan` | yes — `Atom 3 was placed at (nan, nan, nan)` |
| `test_collinear_reference_honours_distance_and_angle` | yes — `nan != 1.5 / 1.09 / 2.0` |
| `test_non_collinear_reference_is_unchanged` | mutation-tested (below) |
| `test_get_perpendicular_unit_vector*` (3, in `vectors_test.py`) | yes — helper does not exist |

`test_every_relabeling_of_ketene_round_trips` is the one that closes the coverage gap: **every** test in the previous revision of this PR used the identity atom order, which is exactly the special case in which the index-space defect is invisible. It sweeps all 120 relabelings of the reported trigger.

Two tests assert an unchanged path and so cannot fail without the fix; both are verified by mutation.

- `test_canonical_frame_placement_is_unchanged` pins the exact float tuple the *original* formula produces. Replacing `get_perpendicular_unit_vector()` with a different but equally valid perpendicular makes all four of its subtests fail while every other test still passes — which is also the demonstration that the azimuth is a free convention.
- `test_non_collinear_reference_is_unchanged` asserts the guard does not steal the well-defined case. Replacing the guard condition with `if True:` so the real A-B-C plane is always discarded makes it fail with `82.717 != 60.0`. An earlier version of this test was **vacuous** — its reference geometry made the arbitrary plane coincide with the real one, so it passed under that mutant; it now uses a general-position triple and asserts the dihedral.

`arc/species/`: 314 passed. Full suite `pytest arc/ -n 6 --dist worksteal`: 2529 passed, 6 failed — the same 6 as `origin/main` (5 × `torch_ani_test`, `TypeError: … os.PathLike`, and `test_arc_families_path`, which asserts `'ARC'` is in the checkout path).

### DIFF
--- a/arc/species/zmat.py
+++ b/arc/species/zmat.py
@@ -1061,7 +1061,9 @@ def add_dummy_atom(zmat: dict,
     if d_str is not None:
         zmat['vars'][d_str] = 180
     # Update xyz with the dummy atom (useful when this atom is used to define dihedrals of other atoms).
-    coords = _add_nth_atom_to_coords(zmat=zmat, coords=list(coords), i=n)
+    zmat_ordered_coords = [tuple(coords[map_index_to_int(zmat['map'][zmat_index])]) for zmat_index in range(n)]
+    zmat_ordered_coords = _add_nth_atom_to_coords(zmat=zmat, coords=zmat_ordered_coords, i=n)
+    coords = list(coords) + [zmat_ordered_coords[n]]
     if connectivity is not None:
         # Update the connectivity dict to reflect that X is connected to the respective atom (r_atoms[1]),
         # this will help later in avoiding linear angles in the last three indices of a dihedral.
@@ -1157,11 +1159,26 @@ def _add_nth_atom_to_coords(zmat: dict,
     """
     Add the n-th atom to the coords (n >= 0).
 
+    The reference atoms are read from the zmat's R/A/D parameter strings, so ``coords`` is indexed
+    in the zmat's index space: entry ``k`` holds the coordinates of the atom at zmat index ``k``,
+    which is not the same as its index in the mol/xyz whenever the two orders differ.
+
+    Atoms 0 and 1 define the frame: they are placed at the origin and on the +z axis, respectively.
+    Given zmat-ordered ``coords``, every atom from index 2 onwards is placed relative to the
+    coordinates already present, so the frame they are in is honoured rather than replaced. Atom 2
+    has no reference plane to fix its azimuth about the 0-1 axis, which is therefore taken from
+    ``vectors.get_perpendicular_unit_vector()`` of that axis; for the canonical frame this
+    reproduces the +y half plane.
+
     Args:
         zmat (dict): The zmat.
-        coords (list): The coordinates to be updated (not the entire xyz dict).
+        coords (list): The coordinates to be updated (not the entire xyz dict), ordered by zmat index.
         i (int): The atom number in the zmat to be added to the coords (0-indexed)
         coords_to_skip (list, optional): Entries are indices to skip.
+
+    Raises:
+        VectorsError: If atom 2 is being added and atoms 0 and 1 are coincident, or if the B and C
+                      reference atoms of the atom being added are coincident.
 
     Returns:
         list: The updated coords.
@@ -1180,25 +1197,37 @@ def _add_nth_atom_to_coords(zmat: dict,
         bc_length = zmat['vars'][r_key]
         alpha = zmat['vars'][a_key]
         alpha = math.radians(alpha if alpha < 180 else 360 - alpha)
-        b_index = [indices for indices in get_atom_indices_from_zmat_parameter(r_key) if indices[0] == i][0][1]
-        b_z = coords[b_index][2]
-        c_y = bc_length * math.sin(alpha)
+        a_indices = [indices for indices in get_atom_indices_from_zmat_parameter(a_key) if indices[0] == i][0]
+        b_index, a_index = a_indices[1], a_indices[2]
         r"""
-        We differentiate between two cases for c_z:
-        Either atom A is at the origin (case 1), or atom B is at the origin (case 2).
-        One of them has to be at the origin (0, 0, 0), since we're adding the 3rd atom (so either A or B were 1st).
-        
-         y
-         ^                    C                         C
-         |           (1)       \        or     (2)     /
-         L__ > z           A -- B                    B -- A
-        
-        In case 1, we need to deduct len(B-C) from the z coordinate of atom B,
-        but in case 2 we need to take the positive value of len(B-C).
-        The above is also true if alpha(A-B-C) is > 90 degrees.
+        Atom C is placed relative to atom B, at a distance of bc_length from it, and rotated by alpha
+        away from the direction B -> A, where alpha is the A-B-C angle folded into [0, 180] just above:
+
+            C = B + bc_length * (cos(alpha) * u_BA + sin(alpha) * u_perp)
+
+        u_BA is the unit vector pointing from B towards A, and u_perp is a unit vector perpendicular to
+        the A-B axis. The two are orthonormal, so they span the plane C is placed in, and the component
+        along each is what makes the A-B-C angle come out at exactly alpha. Atoms A and B fix an axis
+        but not a plane, so the azimuth of u_perp about that axis is a convention.
+
+                              u_perp                                      u_perp
+                                ^                                           ^
+                             C  |                                           |  C
+                              \ |                                           | /
+                         alpha \|                                     alpha |/
+                    A <---------B                               A <---------B
+                         u_BA                                        u_BA
+
+                       alpha < 90                                  alpha > 90
+
+        Both terms are taken relative to atom B and to the direction B -> A, so the placement holds for
+        atoms A and B wherever they lie. For alpha > 90 degrees cos(alpha) is negative, the u_BA term
+        points away from A, and C falls on the far side of B, as drawn on the right.
         """
-        c_z = b_z - bc_length * math.cos(alpha) if b_z else bc_length * math.cos(alpha)
-        coords.append((0.0, c_y, c_z))
+        u_perp = vectors.get_perpendicular_unit_vector([coords[1][k] - coords[0][k] for k in range(3)])
+        u_ba = vectors.unit_vector([coords[a_index][k] - coords[b_index][k] for k in range(3)])
+        coords.append(tuple(coords[b_index][k] + bc_length * (math.cos(alpha) * u_ba[k]
+                                                              + math.sin(alpha) * u_perp[k]) for k in range(3)))
     elif i not in coords_to_skip:
         d_indices = [indices for indices in get_atom_indices_from_zmat_parameter(zmat['coords'][i][2])
                      if indices[0] == i][0]
@@ -1212,6 +1241,10 @@ def _add_nth_atom_to_coords(zmat: dict,
         bc_length = vectors.get_vector_length([coords[c_index][0] - coords[b_index][0],
                                        coords[c_index][1] - coords[b_index][1],
                                        coords[c_index][2] - coords[b_index][2]])
+        if not bc_length:
+            raise VectorsError(f'Cannot place atom {i} of the zmat: its reference atoms {b_index} and {c_index} '
+                               f'are coincident, so the direction from atom {b_index} to atom {c_index} is '
+                               f'undefined, and with it the angle and the dihedral that place atom {i}.')
         # A vector pointing from atom A to atom B:
         ab = [(coords[b_index][0] - coords[a_index][0]),
               (coords[b_index][1] - coords[a_index][1]),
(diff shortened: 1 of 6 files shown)


arc/species/zmat.py
```diff
@@ -1180,25 +1196,12 @@ def _add_nth_atom_to_coords(zmat: dict,
...
-        c_y = bc_length * math.sin(alpha)
-        r"""
-        We differentiate between two cases for c_z:
-        Either atom A is at the origin (case 1), or atom B is at the origin (case 2).
-        One of them has to be at the origin (0, 0, 0), since we're adding the 3rd atom (so either A or B were 1st).
-        
-         y
-         ^                    C                         C
-         |           (1)       \        or     (2)     /
-         L__ > z           A -- B                    B -- A
```
alongd, line 1194: I really liked this text figure... is it not relevant anymore? Tell CC to be more considerate if possible :)

alongd (approved):

PR #947, opened 2026-08-04
Map hydrogens using the backbone map of the chosen candidate

calvinp0:
**Stack position:** bottom of a three-PR atom-mapping stack, based on `main`. #935 sits on this, and #948 sits on #935 — that is the reading order. This PR is self-contained: it can be reviewed and reverted without either of the others.

### The bug

- `map_two_species` scores every superimposable backbone candidate by the RMSD of its dihedral-corrected superposition and picks the best one — then hands `map_hydrogens` the leftover loop variable `candidate`, the **last** iterate, instead of `candidates[chosen_candidate_index]`.
- So the best-RMSD *geometries* are paired with a different candidate's *backbone map*. Every anchor lookup downstream then runs in an inconsistent frame.
- The mismatch is invisible whenever there is exactly one candidate, because the last iterate is then also the chosen one — which is why it lasted.
- The new test uses 2-methylbutane, which fingerprints into two superimposable candidates with the best-scoring one not last. Mapped onto a copy of itself with identical coordinates it must return the identity; without the fix it returns backbone `{0:2, 1:1, 2:0, …}`.

### Why it read as intentional

- `candidate = None` is a sentinel set before the loop, and `if candidate is not None:` reads as a deliberate "did the loop body execute at least once?" check. The accident is that the *same variable* is then passed as the value, since after the loop it holds a perfectly valid-looking candidate. One name is doing double duty as a flag and as the chosen item.
- The guard is also unreachable-false. Emptiness is excluded three lines above, and `rmsds.index(min(rmsds))` would raise `ValueError` on an empty list a line earlier in any case.
- The `map_type == 'list'` conversion sat *inside* that guard, so had it ever been False the function would have returned a dict where a list was requested. Unreachable, but it confirms the guard had drifted past its purpose.
- The sentinel is therefore removed, not merely the argument swapped.

### Second defect: the RMSD was not a reproducible number

Consulting the score for the first time exposed that it had never been trustworthy.

- `map_isomerization_reaction` sets every bond order of its reactant and product copies to single so the two skeletons can be superimposed — but left `mol_list` holding the localized structures generated from the **original** bond orders.
- `determine_rotors()` reads `mol_list`, so the rotatable bonds, and with them the backbone torsions `fix_dihedrals_by_backbone_mapping` aligns, came from a molecule that no longer matched the one being mapped. It was aligning on the wrong hinges.
- Which localized structure ends up at `mol_list[0]` is not stable across interpreter runs, so the aligned geometries — and the RMSD scored for each candidate — moved from run to run. On `main` this was invisible: the score was computed, a winner chosen, and then the last candidate used regardless. "Always the last one" is a stable answer, not a right one.
- New `flatten_bond_orders()` discards `mol_list` and the rotors together with the bond orders, so the rotors follow the flattened molecule and every candidate is scored against the same reproducible alignment.

**Reproducing the instability:** it is only visible on the two middle commits of this PR — after `Map hydrogens using the backbone map of the chosen candidate` and before `Determine the mapping rotors from the flattened molecule`. `main` is stable because it discards the score entirely, and the finished branch is stable because the fix landed, so sweeping `PYTHONHASHSEED` on either endpoint shows nothing. Checked here: `test_interpolate_birad_recombination` passes on all of 11 seeds on `main` and on this branch's head. That is the expected result, not a refutation — but note that the intermediate window itself was **not** re-measured, so treat the instability as diagnosed from the code path rather than demonstrated by a seed sweep.

Measured on the birad recombination in `linear_test.py`:

| | before | after |
|---|---|---|
| candidate RMSDs (lower is better) | 3.09 / 3.14 | **0.151 / 0.127** |
| gap between them | 1.5% — noise | **19%** — a real winner |

Alignment improves about twenty-fold, and it selects **the same answer `main` produced**, so the geometry that test has always expected is unchanged. Nothing was re-blessed to make a test pass.

### Third: the tie between symmetry-equivalent candidates

- The score is the RMSD between the two **backbone distance matrices**, so it is identical up to floating-point noise for any permutation that leaves every backbone interatomic distance unchanged — which is every symmetry of the backbone. RMSD cannot discriminate them.
- A strict argmin would therefore hand the choice — and every reaction map built on it — to floating-point noise. `main` was insulated only because `candidates[-1]` is deterministic, which is coincidence rather than correctness.
- `RMSD_TIE_TOLERANCE = 1e-6`: candidates within that of the lowest score are treated as tied.
- **The backbone score is blind to hydrogens**, so tied candidates are not interchangeable — they are precisely the ones that place the hydrogens differently. Mapping neopentane onto a copy of itself with identical coordinates, the four tied candidates yield atom maps of 0.000, 9.432, 5.235 and 9.157 Å all-atom Kabsch RMSD. Selecting either end of the tied list by position picks one of those blind.
- Tied candidates are therefore scored a second time by the Kabsch RMSD of the full atom map each yields, and the last candidate that also ties on that displacement is used. Every symmetric species measured now maps onto a copy of itself at 0.000 Å, and `test_get_atom_map_5`'s reaction map is unchanged, so no existing expectation moved.

### Scope

- Behaviour change: for species with more than one superimposable candidate the hydrogen map changes — that is the fix. Where the backbone RMSDs differ genuinely, the best-scoring candidate is now used; where they tie, the candidate whose full atom map has the lowest displacement is used. Single-candidate species are unaffected.
- **Not fixed here:** `reorder_p_label_map` has no way to request the candidate consistent with the reactant labelling. That is the real defect behind the tie and deserves its own change.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/arc/mapping/driver.py
+++ b/arc/mapping/driver.py
@@ -32,6 +32,7 @@ from arc.species.converter import check_molecule_list_order
 if TYPE_CHECKING:
     from arc.molecule.molecule import Molecule
     from arc.reaction import ARCReaction
+    from arc.species.species import ARCSpecies
 
 MAX_PDI = 25
 
@@ -128,6 +129,28 @@ def map_general_rxn(rxn: ARCReaction,
     return atom_map
 
 
+def flatten_bond_orders(*species: ARCSpecies) -> None:
+    """
+    Set all bond orders of the given species' molecules to single.
+
+    Also discards the localized structures, the rotors and the directed rotor scans of each species,
+    which were derived from the previous bond orders, so that they are re-determined from the
+    flattened molecule. Species marked to skip rotor scans (a ``None`` ``rotors_dict``) keep that mark.
+
+    Args:
+        species (ARCSpecies): The species to flatten.
+    """
+    for spc in species:
+        for atom in spc.mol.atoms:
+            for bond in atom.bonds.values():
+                bond.order = 1
+        spc.mol_list = None
+        if spc.rotors_dict is not None:
+            spc.rotors_dict = dict()
+            spc.number_of_rotors = 0
+            spc.directed_rotors = dict()
+
+
 def map_isomerization_reaction(rxn: ARCReaction) -> list[int] | None:
     """
     Map isomerization reaction that has no corresponding RMG family.
@@ -210,9 +233,7 @@ def map_isomerization_reaction(rxn: ARCReaction) -> list[int] | None:
                 success = False
 
             if success:
-                for atom in r_copy.mol.atoms + p_copy.mol.atoms:
-                    for bd in atom.bonds.values():
-                        bd.order = 1
+                flatten_bond_orders(r_copy, p_copy)
                 return map_two_species(r_copy, p_copy, map_type='list')
 
         # 3) If fingerprint branch *didn’t* yield pairs, still do a deterministic edge‐removal mapping for *any* ring opening
@@ -232,9 +253,7 @@ def map_isomerization_reaction(rxn: ARCReaction) -> list[int] | None:
             except ValueError:
                 pass
             else:
-                for atom in r_copy.mol.atoms + p_copy.mol.atoms:
-                    for bd in atom.bonds.values():
-                        bd.order = 1
+                flatten_bond_orders(r_copy, p_copy)
                 return map_two_species(r_copy, p_copy, map_type='list')
 
     # 4) Fallback to the general mapping
--- a/arc/mapping/driver_test.py
+++ b/arc/mapping/driver_test.py
@@ -846,6 +846,47 @@ class TestMappingDriver(unittest.TestCase):
         self.assertIn(atom_map[7], [3, 4, 5, 8])
         self.assertIn(atom_map[8], [3, 4, 5, 8])
 
+    def test_flatten_bond_orders(self):
+        """Test the flatten_bond_orders() function."""
+        spc_1 = ARCSpecies(label='2-butene', smiles='CC=CC')
+        spc_2 = ARCSpecies(label='1,3-butadiene', smiles='C=CC=C')
+        spc_1.determine_rotors()
+        self.assertIn(2, [bond.order for atom in spc_1.mol.atoms for bond in atom.bonds.values()])
+        self.assertIsNotNone(spc_1.mol_list)
+        self.assertEqual(spc_1.number_of_rotors, 2)
+
+        flatten_bond_orders(spc_1, spc_2)
+
+        for spc in [spc_1, spc_2]:
+            self.assertTrue(all(bond.order == 1 for atom in spc.mol.atoms for bond in atom.bonds.values()))
+            self.assertIsNone(spc.mol_list)
+            self.assertEqual(spc.rotors_dict, dict())
+            self.assertEqual(spc.number_of_rotors, 0)
+
+        spc_1.determine_rotors()
+        self.assertIn([2, 3], [rotor['pivots'] for rotor in spc_1.rotors_dict.values()])
+
+    def test_flatten_bond_orders_discards_the_directed_rotors(self):
+        """Test that flatten_bond_orders() discards the directed rotor scans along with the rotors."""
+        spc = ARCSpecies(label='nbutane', smiles='CCCC', directed_rotors={'brute_force_sp': [[[1, 2], [2, 3]]]})
+        spc.determine_rotors()
+        self.assertEqual(spc.directed_rotors, {'brute_force_sp': [[[5, 1, 2, 3], [1, 2, 3, 4]]]})
+
+        flatten_bond_orders(spc)
+
+        self.assertEqual(spc.directed_rotors, dict())
+        spc.determine_rotors()
+        self.assertEqual([rotor['pivots'] for rotor in spc.rotors_dict.values()], [[1, 2], [2, 3], [3, 4]])
+
+    def test_flatten_bond_orders_keeps_the_skip_rotor_scans_mark(self):
+        """Test that flatten_bond_orders() keeps a None rotors_dict, which marks a species to skip rotor scans."""
+        spc = ARCSpecies(label='2-butene', smiles='CC=CC')
+        spc.rotors_dict = None
+
+        flatten_bond_orders(spc)
+
+        self.assertIsNone(spc.rotors_dict)
+
     def test_map_isomerization_reaction(self):
         """Test the map_isomerization_reaction() function."""
         reactant_xyz = """C  -1.3087    0.0068    0.0318
--- a/arc/mapping/engine.py
+++ b/arc/mapping/engine.py
@@ -28,6 +28,7 @@ if TYPE_CHECKING:
     from arc.reaction import ARCReaction
 
 RESERVED_FINGERPRINT_KEYS = ['self', 'chirality', 'label']
+RMSD_TIE_TOLERANCE = 1e-6
 
 
 def map_two_species(spc_1: ARCSpecies | Molecule,
@@ -43,6 +44,14 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
     All indices are 0-indexed.
     If a dict type atom map is returned, it could conveniently be used to map ``spc_2`` -> ``spc_1`` by doing::
         ordered_spc1.atoms = [spc_2.atoms[atom_map[i]] for i in range(len(spc_2.atoms))]
+
+    When several superimposable backbone candidates are identified, each is scored by the RMSD
+    between the two backbone distance matrices, so candidates tie within ``RMSD_TIE_TOLERANCE``
+    when the permutation leaves every backbone interatomic distance unchanged. That score covers
+    the backbone only, so the tied candidates are further scored by the Kabsch RMSD of the full
+    atom map each of them yields, and the last candidate that also ties on that displacement is
+    used. Both the backbone map and the dihedral-corrected geometries of that same candidate are
+    used to map the hydrogen atoms.
 
     Args:
         spc_1 (ARCSpecies | Molecule): Species 1.
@@ -110,7 +119,6 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
                 return None
         else:
             rmsds, fixed_spcs = list(), list()
-            candidate = None
             for candidate in candidates:
                 fixed_spc_1, fixed_spc_2 = fix_dihedrals_by_backbone_mapping(spc_1, spc_2, backbone_map=candidate)
                 fixed_spcs.append((fixed_spc_1, fixed_spc_2))
@@ -126,12 +134,23 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
                 xyz2 = sort_xyz_using_indices(xyz2, indices=[v for k, v in sorted(no_gap_candidate.items(),
                                                                                   key=lambda item: item[0])])
                 rmsds.append(compare_confs(xyz1=xyz1, xyz2=xyz2, rmsd_score=True))
-            chosen_candidate_index = rmsds.index(min(rmsds))
-            fixed_spc_1, fixed_spc_2 = fixed_spcs[chosen_candidate_index]
-            if candidate is not None:
-                atom_map = map_hydrogens(fixed_spc_1, fixed_spc_2, candidate)
-                if map_type == 'list':
-                    atom_map = [v for k, v in sorted(atom_map.items(), key=lambda item: item[0])]
+            lowest_rmsd = min(rmsds)
+            tied_indices = [i for i, rmsd in enumerate(rmsds) if rmsd - lowest_rmsd <= RMSD_TIE_TOLERANCE]
+            atom_maps = dict()
+            for i in tied_indices:
+                atom_maps[i] = map_hydrogens(fixed_spcs[i][0], fixed_spcs[i][1], candidates[i])
+            if len(tied_indices) > 1:
+                displacements = {i: fixed_spcs[i][0].kabsch(fixed_spcs[i][1],
+                                                            [v for k, v in sorted(atom_maps[i].items(),
+                                                                                  key=lambda item: item[0])])
+                                 for i in tied_indices}
+                lowest_displacement = min(displacements.values())
+                tied_indices = [i for i in tied_indices
+                                if displacements[i] - lowest_displacement <= RMSD_TIE_TOLERANCE]
+            chosen_candidate_index = max(tied_indices)
+            atom_map = atom_maps[chosen_candidate_index]
+            if map_type == 'list':
+                atom_map = [v for k, v in sorted(atom_map.items(), key=lambda item: item[0])]
 
     if inc_vals is not None:
         atom_map = [value + inc_vals for value in atom_map]
--- a/arc/mapping/engine_test.py
+++ b/arc/mapping/engine_test.py
@@ -572,6 +572,55 @@ class TestMappingEngine(unittest.TestCase):
                           H                  0.34179135    2.75680752   -1.36901816
                           H                  1.78229521    1.76788614   -1.37607384
                           H                  0.34142026    1.21854517   -2.19776146"""
+        cls.methylbutane_xyz = """C      -0.97433227    1.17065473   -0.35442290
+                                  C      -0.45192007   -0.26483788   -0.28088601
+                                  C      -0.33943943   -0.84708671   -1.69187722
+                                  C      -1.34349197   -1.16914868    0.58768626
+                                  C      -1.40105018   -0.74536522    2.04847109
+                                  H      -2.00419427    1.19988237   -0.72602982
+                                  H      -0.35559338    1.77478281   -1.02687519
+                                  H      -0.95159170    1.65360691    0.62712610
+                                  H       0.55534772   -0.24133144    0.15407433
+                                  H       0.32353843   -0.23606931   -2.31364700
+                                  H       0.07332070   -1.86107875   -1.66298158
+                                  H      -1.31775660   -0.89266154   -2.18248872
+                                  H      -0.96063413   -2.19652174    0.54586182
+                                  H      -2.36224242   -1.19617367    0.18190102
+                                  H      -1.90298172    0.21908306    2.16797716
+                                  H      -0.39634721   -0.67043363    2.47612199
+                                  H      -1.96252182   -1.48324241    2.63070314"""
+        cls.dimethylbutane_xyz = """C      -0.82461851    1.14673080   -0.77842678
+                                    C      -0.42610728   -0.06063487    0.08629698
+                                    C      -0.47034470   -1.33504078   -0.77290194
+                                    C      -1.45116287   -0.20442011    1.22812003
+                                    C       0.97587301    0.14090229    0.71707616
+                                    C       2.12880735    0.30329591   -0.26614701
+                                    H      -0.76102821    2.07968065   -0.20742874
+                                    H      -0.18120395    1.24559899   -1.65876690
+                                    H      -1.85453719    1.04872389   -1.14068352
+                                    H       0.17558199   -1.25377401   -1.65320345
+                                    H      -0.14861173   -2.21043461   -0.19787833
+                                    H      -1.48632406   -1.53069938   -1.13494088
+                                    H      -2.46459886   -0.34995913    0.83714693
+                                    H      -1.21500982   -1.06356074    1.86613835
+                                    H      -1.46522531    0.68925624    1.86223603
+                                    H       0.95650220    1.02607021    1.36632791
+                                    H       1.20453979   -0.71149012    1.37019493
+                                    H       2.24605190   -0.57899874   -0.90132288
+                                    H       1.99489999    1.18037559   -0.90523893
+                                    H       3.06651626    0.43837793    0.28340204"""
+        cls.benzene_xyz = """C      -1.39298622   -0.07016743    0.01432698
+                             C      -0.63555191   -1.24137615    0.02446132
+                             C       0.75743431   -1.17120872    0.01013428
+                             C       1.39298622    0.07016743   -0.01432699
+                             C       0.63555191    1.24137616   -0.02446130
+                             C      -0.75743431    1.17120872   -0.01013444
+                             H      -2.47828288   -0.12483594    0.02548953
+                             H      -1.13072003   -2.20855112    0.04351957
+                             H       1.34756286   -2.08371519    0.01803002
+                             H       2.47828289    0.12483592   -0.02548935
+                             H       1.13072003    2.20855114   -0.04351930
+                             H      -1.34756286    2.08371519   -0.01803033"""
 
     def test_map_two_species(self):
         """Test the map_two_species() function."""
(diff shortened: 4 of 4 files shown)


kfir4444 (dismissed):
LGTM, excited to see the next stage in this stack.

alongd:
Fleet review of `44a6b112d...3ec50c6b7` — four passes (mutation/testing, security, api-contract, and a cross-model adversarial pass), each run independently in a detached worktree at the head SHA so none could read another's conclusions. Baseline at head: `arc/mapping/` = 81 passed.

**The fix is right.** Head vs base `map_two_species` over 12 species: 10 identical, and the 2 that differ (naphthalene, cyclopentane) change from a base permutation to the identity. No case where head is worse. The leaked-loop-variable diagnosis reproduces, and reverting `candidates[chosen_candidate_index]` to `candidate` fails the new engine test. Everything below is coverage and documentation, not correctness.

---

### 1. `RMSD_TIE_TOLERANCE = 1e-6` is unpinned, and the tie it names is a property of the inputs rather than of symmetry

Two passes reached this independently — one by mutation, one by measurement — and I re-derived both halves.

**Unpinned.** Moving the constant and re-running the suite:

```
--- RMSD_TIE_TOLERANCE = 1e-3 ---   81 passed in 30.09s
--- RMSD_TIE_TOLERANCE = 1e-4 ---   81 passed in 30.68s
--- RMSD_TIE_TOLERANCE = 1e-2 ---   81 passed in 30.01s
```

`0.0` fails (via the pre-existing `test_get_atom_map_5`) and `1e9` fails (via the new engine test). So the whole constraint CI places on this constant is `0 < TOL < 1e9` — a value 10,000x larger is invisible.

**Knife-edge.** The docstring says *"Symmetry-equivalent candidates superimpose equally well and therefore always tie."* On idealized geometries that holds: measured tie residuals are 1.6e-8 (benzene), 1.3e-7 (neopentane), 2.4e-7 (cyclohexane), 2.8e-7 (naphthalene). But the residual scales linearly with geometric asymmetry:

```
c1ccc2ccccc2c1   (tol=1e-06)
  distort=0.0    max-delta=2.806e-07  n_tied=4/4  argmin=0 chosen=3  band < tol: tie holds
  distort=0.0001 max-delta=2.903e-04  n_tied=1/4  argmin=0 chosen=0  BAND > TOL: tie set collapses
C1CCCCC1         (tol=1e-06)
  distort=0.0    max-delta=2.369e-07  n_tied=6/6  argmin=0 chosen=5  band < tol: tie holds
  distort=0.0001 max-delta=2.065e-04  n_tied=1/6  argmin=0 chosen=0  BAND > TOL: tie set collapses
```

A 1e-4 A perturbation — orders below the convergence threshold of any ESS optimization — collapses the tie set from 4 candidates to 1 and reverts selection to plain argmin. The tie branch fires only when the two geometries agree to roughly **3e-7 A**, i.e. when both derive from the same coordinates.

That bears directly on the PR description's argument. "Degenerate ties resolve as they did before" is true in the same-source regime and stops being true outside it, with nothing to signal the branch you designed is no longer being taken. Worth deciding which regime it is for: if same-source only, say so in the docstring and `1e-6` absolute is fine; if it must survive genuinely different geometries, an absolute tolerance cannot do it and a relative form (`rmsd <= lowest * (1 + rtol) + atol`) is needed.

### 2. `test_map_two_species_uses_the_lowest_rmsd_candidate` cannot pin the tie-break

Instrumented, the fixture's scores are `rmsds=[0.0, 0.498]` — no tie, so `max(...)` and `min(...)` pick the same index and the test passes against either. Mutating `max` -> `min` *is* caught, but by `test_get_atom_map_5` (`AssertionError: Lists differ: [3, 4, 5, 10, 6] != [6, 5, 4, 9, 3]`), a legacy test that knows nothing about this change. The central new behaviour is pinned by luck.

Findings 1 and 2 share one root: no fixture anywhere constructs a genuine near-tie. One test with >=2 candidates whose RMSDs differ by less than the tolerance, asserting the **last** is chosen, closes both — and settles the open question below.

### 3. `flatten_bond_orders` discards `rotors_dict` but leaves `directed_rotors` referring to it — **please fix in this PR**

Clearing `rotors_dict` makes `determine_rotors()` run again, which re-runs `initialize_directed_rotors()`. That method is not idempotent: the first run rewrites `self.directed_rotors` from pivot pairs into four-atom scans (`species.py:1409-1417`), and the second rejects its own output:

```
directed_rotors after determine_rotors: {'brute_force_sp': [[[5, 1, 2, 3], [1, 2, 3, 4]]]}
HEAD RAISED SpeciesError The pivots [1, 5] do not represent a rotor in species nbutane.
                         Valid rotor pivots are: [[1, 2], [2, 3], [3, 4]]
```

This is **not a regression** — base raises too on the same input, as `TypeError: list indices must be integers or slices, not list`. But this is the commit where rotor re-derivation became a deliberate act, and `flatten_bond_orders` is already discarding the rotor set those entries point at, so clearing `spc.directed_rotors` alongside them is one line and coherent with what the function claims to do. It also matters for the ND-scan work.

### 4. The `if spc.rotors_dict is not None` guard is untested

Deleting it — always assigning `dict()` and `0` — leaves the suite green. `None` means "skip rotor scans" and is semantically distinct from `{}`; the guard is deliberate and nothing defends it.

---

### Negative results — checked and sound, so nobody re-treads them

- **Index alignment.** `rmsds`, `fixed_spcs` and `candidates` stay in lockstep on every path — exactly one append each per candidate, no `continue`/filter between them. Cleared independently by two passes; this was the review's prime suspicion going in.
- **Removing `if candidate is not None:`** is not fail-open. Emptiness already returns `None` three lines above, `map_hydrogens` has no `None`-returning path, and `compare_confs(rmsd_score=True)` always returns a float. Your reading of it as dead code is correct.
- **Aliasing.** `ARCSpecies.copy()` round-trips through `as_dict`/`__init__`, so `flatten_bond_orders`' mutation cannot reach caller-owned species. A full `map_isomerization_reaction` run left the caller's bond orders, `mol_list` and `number_of_rotors` unchanged.
- **Determinism.** Three passes independently swept `PYTHONHASHSEED` (4, 5 and 5 seeds): byte-identical maps throughout. `candidates` order comes from an int-keyed dict, not string-set iteration.
- **Chemical validity of the tie-break.** For 8 symmetric species every selected map is a permutation, element-preserving and bond-preserving — "last tied wins" picks a genuine automorphism, so it chooses *which* member of the symmetry class, never a wrong map.
- **`mol_list = None`** is safe: documented lazy-regenerate field, and consumers guard.
- **Wider sweep** (`arc/mapping/ arc/species/ arc/reaction/ arc/molecule/`) = 1012 passed, 3 failed; all 3 (`conformers_test` x2, `zmat_test`) reproduce at base in untouched files.

### Open judgment, not a defect

`max(...)` makes `map_two_species(spc, spc.copy())` return a non-identity permutation for symmetric species, where `min(...)` would return the identity — equally correct, and it makes the canonical case canonical. If the reason for `max` is minimal churn against base's effective `candidates[-1]`, that belongs in the docstring; the sentence there now explains the tolerance but not the preference for the last one.

### Coverage disclosure

Sized T1 (2 source files, 1 package, +92/-12), cap 3 passes. **Not run:** maintainability, performance, red team. The PR description's reproducibility claim — `test_get_atom_map_5` failing 8 runs in 20 under strict argmin, and the seed-dependent window living only between the two middle commits — was **not** independently reproduced; the passes confirmed the direction (`TOL=0.0` breaks that test) but nobody ran the 20-seed sweep or checked out the intermediate commits.

calvinp0:
> Fleet review of `44a6b112d...3ec50c6b7` — four passes (mutation/testing, security, api-contract, and a cross-model adversarial pass), each run independently in a detached worktree at the head SHA so none could read another's conclusions. Baseline at head: `arc/mapping/` = 81 passed.
> 
> **The fix is right.** Head vs base `map_two_species` over 12 species: 10 identical, and the 2 that differ (naphthalene, cyclopentane) change from a base permutation to the identity. No case where head is worse. The leaked-loop-variable diagnosis reproduces, and reverting `candidates[chosen_candidate_index]` to `candidate` fails the new engine test. Everything below is coverage and documentation, not correctness.
> ### 1. `RMSD_TIE_TOLERANCE = 1e-6` is unpinned, and the tie it names is a property of the inputs rather than of symmetry
> 
> Two passes reached this independently — one by mutation, one by measurement — and I re-derived both halves.
> 
> **Unpinned.** Moving the constant and re-running the suite:
> 
> ```
> --- RMSD_TIE_TOLERANCE = 1e-3 ---   81 passed in 30.09s
> --- RMSD_TIE_TOLERANCE = 1e-4 ---   81 passed in 30.68s
> --- RMSD_TIE_TOLERANCE = 1e-2 ---   81 passed in 30.01s
> ```
> 
> `0.0` fails (via the pre-existing `test_get_atom_map_5`) and `1e9` fails (via the new engine test). So the whole constraint CI places on this constant is `0 < TOL < 1e9` — a value 10,000x larger is invisible.
> 
> **Knife-edge.** The docstring says _"Symmetry-equivalent candidates superimpose equally well and therefore always tie."_ On idealized geometries that holds: measured tie residuals are 1.6e-8 (benzene), 1.3e-7 (neopentane), 2.4e-7 (cyclohexane), 2.8e-7 (naphthalene). But the residual scales linearly with geometric asymmetry:
> 
> ```
> c1ccc2ccccc2c1   (tol=1e-06)
>   distort=0.0    max-delta=2.806e-07  n_tied=4/4  argmin=0 chosen=3  band < tol: tie holds
>   distort=0.0001 max-delta=2.903e-04  n_tied=1/4  argmin=0 chosen=0  BAND > TOL: tie set collapses
> C1CCCCC1         (tol=1e-06)
>   distort=0.0    max-delta=2.369e-07  n_tied=6/6  argmin=0 chosen=5  band < tol: tie holds
>   distort=0.0001 max-delta=2.065e-04  n_tied=1/6  argmin=0 chosen=0  BAND > TOL: tie set collapses
> ```
> 
> A 1e-4 A perturbation — orders below the convergence threshold of any ESS optimization — collapses the tie set from 4 candidates to 1 and reverts selection to plain argmin. The tie branch fires only when the two geometries agree to roughly **3e-7 A**, i.e. when both derive from the same coordinates.
> 
> That bears directly on the PR description's argument. "Degenerate ties resolve as they did before" is true in the same-source regime and stops being true outside it, with nothing to signal the branch you designed is no longer being taken. Worth deciding which regime it is for: if same-source only, say so in the docstring and `1e-6` absolute is fine; if it must survive genuinely different geometries, an absolute tolerance cannot do it and a relative form (`rmsd <= lowest * (1 + rtol) + atol`) is needed.
> ### 2. `test_map_two_species_uses_the_lowest_rmsd_candidate` cannot pin the tie-break
> 
> Instrumented, the fixture's scores are `rmsds=[0.0, 0.498]` — no tie, so `max(...)` and `min(...)` pick the same index and the test passes against either. Mutating `max` -> `min` _is_ caught, but by `test_get_atom_map_5` (`AssertionError: Lists differ: [3, 4, 5, 10, 6] != [6, 5, 4, 9, 3]`), a legacy test that knows nothing about this change. The central new behaviour is pinned by luck.
> 
> Findings 1 and 2 share one root: no fixture anywhere constructs a genuine near-tie. One test with >=2 candidates whose RMSDs differ by less than the tolerance, asserting the **last** is chosen, closes both — and settles the open question below.
> ### 3. `flatten_bond_orders` discards `rotors_dict` but leaves `directed_rotors` referring to it — **please fix in this PR**
> 
> Clearing `rotors_dict` makes `determine_rotors()` run again, which re-runs `initialize_directed_rotors()`. That method is not idempotent: the first run rewrites `self.directed_rotors` from pivot pairs into four-atom scans (`species.py:1409-1417`), and the second rejects its own output:
> 
> ```
> directed_rotors after determine_rotors: {'brute_force_sp': [[[5, 1, 2, 3], [1, 2, 3, 4]]]}
> HEAD RAISED SpeciesError The pivots [1, 5] do not represent a rotor in species nbutane.
>                          Valid rotor pivots are: [[1, 2], [2, 3], [3, 4]]
> ```
> 
> This is **not a regression** — base raises too on the same input, as `TypeError: list indices must be integers or slices, not list`. But this is the commit where rotor re-derivation became a deliberate act, and `flatten_bond_orders` is already discarding the rotor set those entries point at, so clearing `spc.directed_rotors` alongside them is one line and coherent with what the function claims to do. It also matters for the ND-scan work.
> ### 4. The `if spc.rotors_dict is not None` guard is untested
> 
> Deleting it — always assigning `dict()` and `0` — leaves the suite green. `None` means "skip rotor scans" and is semantically distinct from `{}`; the guard is deliberate and nothing defends it.
> ### Negative results — checked and sound, so nobody re-treads them
> 
>     * **Index alignment.** `rmsds`, `fixed_spcs` and `candidates` stay in lockstep on every path — exactly one append each per candidate, no `continue`/filter between them. Cleared independently by two passes; this was the review's prime suspicion going in.
> 
>     * **Removing `if candidate is not None:`** is not fail-open. Emptiness already returns `None` three lines above, `map_hydrogens` has no `None`-returning path, and `compare_confs(rmsd_score=True)` always returns a float. Your reading of it as dead code is correct.
> 
>     * **Aliasing.** `ARCSpecies.copy()` round-trips through `as_dict`/`__init__`, so `flatten_bond_orders`' mutation cannot reach caller-owned species. A full `map_isomerization_reaction` run left the caller's bond orders, `mol_list` and `number_of_rotors` unchanged.
> 
>     * **Determinism.** Three passes independently swept `PYTHONHASHSEED` (4, 5 and 5 seeds): byte-identical maps throughout. `candidates` order comes from an int-keyed dict, not string-set iteration.
> 
>     * **Chemical validity of the tie-break.** For 8 symmetric species every selected map is a permutation, element-preserving and bond-preserving — "last tied wins" picks a genuine automorphism, so it chooses _which_ member of the symmetry class, never a wrong map.
> 
>     * **`mol_list = None`** is safe: documented lazy-regenerate field, and consumers guard.
> 
>     * **Wider sweep** (`arc/mapping/ arc/species/ arc/reaction/ arc/molecule/`) = 1012 passed, 3 failed; all 3 (`conformers_test` x2, `zmat_test`) reproduce at base in untouched files.
> 
> 
> ### Open judgment, not a defect
> 
> `max(...)` makes `map_two_species(spc, spc.copy())` return a non-identity permutation for symmetric species, where `min(...)` would return the identity — equally correct, and it makes the canonical case canonical. If the reason for `max` is minimal churn against base's effective `candidates[-1]`, that belongs in the docstring; the sentence there now explains the tolerance but not the preference for the last one.
> ### Coverage disclosure
> 
> Sized T1 (2 source files, 1 package, +92/-12), cap 3 passes. **Not run:** maintainability, performance, red team. The PR description's reproducibility claim — `test_get_atom_map_5` failing 8 runs in 20 under strict argmin, and the seed-dependent window living only between the two middle commits — was **not** independently reproduced; the passes confirmed the direction (`TOL=0.0` breaks that test) but nobody ran the 20-seed sweep or checked out the intermediate commits.

1. In regards to the `RMSD_TIE_TOLERANCE` being unpinned. Both halves were repoduced with one correction and now have fixed the underlying problem (a structural fix)

```
TOL=1e-2  81 passed     TOL=1e-4  81 passed
TOL=1e-3  81 passed     TOL=1e-5  81 passed
TOL=0.0    failed        TOL=1e9    failed
```

I could not reproduce the idea that 'residual scales with geometric asymmetry, so distortion collapses the tie'. I did perturbing of all atoms by a seeded uniform displacement

```
                distort=0     1e-6      1e-5      1e-4      1e-3
c1ccccc1        2.41e-08  1.06e-08  1.07e-08  1.07e-08  1.07e-08   tie holds 6/6 throughout
C1CCCCC1        2.18e-07  1.44e-07  1.35e-07  1.33e-07  1.33e-07   tie holds 6/6 throughout
CC(C)(C)C       1.63e-07  7.07e-08  6.21e-08  5.99e-08  5.99e-08   tie holds 4/4 throughout
```

The band shrunk under distortion and the tie survives 1e-3 Angstrom, three orders above where the initial review comment collapse was measured. The reason is that in `compare_confs(rmsd_score=True)`: it is the RMSD of the two distance matrices  (`converter.py:1979, calc_rmsd(np.triu(dmat1), np.triu(dmat2))`), not a cartesian RMSD after superposition. But the review conclusion did reproduce: naphthalene, both species built from identical coords and no distortion at all, gives band `2.66e-06` over tolerance, so the tie set collapses 4 to ` and the branch silently degrades to plain argmin. So the instinct is right that `1e-6` sits on a knifes edge; it's just that distortion is not what pushes it over.

Regarding the ESS-threshold aside: Gaussian's default `opt` Maximum Displacement trheshold is 0.001800 Bohr (`9.5e-4 A`) (G16 FAQ), so `1e-4 A` is below it but not by orders. More to the point, in the FAQ, it is noted that a structure is also accepted when the forces are two orders below threshold "regardless of the values of the displacements", and its worked example shows the same structure reporting 0.000635 A displacement from the optimisation and 0.005385 A from the freq job. So convergence does not pin geom to a displacement scale in the first place. The symmetry survives anyway, for the distance matrix reason above.

So, decided to keep `1e-6` absolute; not go relative. A relative form fails concretely - lowest is exactly `0.0` in most tie firing calls, so `lowest * (1 + rtol)` contributes nothing precisley where it is needed; and where th etie legitmately breaks, the band tracks noise rather than lowest, so you would need `rtol approx. 0.5`, wide neough to merge genuinely distinct candidates.

In regards to changes - finding 1 and the `max/min` question have the same root and fixing that root also de-fangs the constant (will discuss `max` vs `min` below). After the change,t he constant is pinnned from below to ` (3e-7, 1e-6]`:

```
TOL=1e-7 / 2e-7 / 3e-7   1 failed        TOL=1e-6 .. 0.5   85 passed
```
And, the important part, `TOL=0.5` five orders too large, now leaves the suite green, where at head it broke two tests. The tolerance no longer decides the answer by itself, it only decides how many candidates reach a second, phyiscally meaningful criterion. 

The review bracket (`0 … 1.7e-7`, then nothing until `4.98e-1`) is from the test corpus, and raising it to fix naphthalene mis-split should be confirmed on the benchmark corpus first. 

2. The new test cannot pin the tie-break. Confirmed and fixed it. The 2-methylbutane fixture scores `[0.0, 0.498]` - no tie, so the test passed under either selector, exactly as you found

However, there is now a genuine near tie fixture. 2,2 dimethylbutane fingerprints into three candidates scoring `[0.0, 2.32e-07, 0.4617]`: candidates 0 and 1 tie within tolerance, candidate 2 does not. That separates all three plausible rules - argmin picks 0, last of tied picks 1, base's `candidates[-1]` picks 2. A second fixture, benzene, ties on both criteria (all six canidates are exact ring symmetries) and pins the residual rule
The new behaviour is now pinned by the new tests

3. `flatten_bond_orders` leaves `directed_rotos` dangling is now fixed. `spc.directed_rotors = dict()` alongside `rotors_dict`

This is a regression introducted by this PR, not pre-existing. I could not reproduce a base-side `TypeError` on any wellformed directed rotor spec. `determine_rotos(0` returns early whilst `rotors_dict` is populated, so on the base the second call is a silent no-op

```
spec                                          base(no clear)  head(clear)                     with fix
{'brute_force_sp': [[[1,2],[2,3]]]}           ok              SpeciesError: pivots [1,5] …    ok
{'brute_force_opt': [['all']]}                ok              SpeciesError: pivots must be …  ok
{'cont_opt': [[[1,2],[2,3]]]}                 ok              SpeciesError: pivots [1,5] …    ok
{'brute_force_sp_diagonal': [[[1,2],[2,3]]]}  ok              SpeciesError: pivots [1,5] …    ok
{'cont_opt_diagonal': [['all']]}              ok              SpeciesError: pivots must be …  ok
```

Base raises only if `initialized_directed_rotors()` is invoked directly. That makes the fix more necessary. Test added, proved failing pre-fix

```python
AssertionError: {'brute_force_sp': [[[5, 1, 2, 3], [1, 2, 3, 4]]]} != {}
```

4. The `if spc.rotors_dict is not None` guard is untested - fixed. `test_flatten_bond_orders_keeps_the_skip_rotor_scans_mark` covers that a species marked to skip rotor scans keeps its `None` rather than being handed `{}`. Proved failing with the guard removed.


-----

`max` vs `min`

So, they are not equally correct from my finding. The reason is seen in the code. The score is computed over `backbone_1`/`backbone_2` only (`engine.py`) so hydrogens are excluded from the tie criterion. Tie candidates are therefore precisely the ones tha tplace hydrogens different. They are not interchangable.And choosing between them by list position would be choosing blindly.

Measure, mapping each species into a copy of itself with identical coords, so the identity map costs 0. Cos is `ARCSpecies.kabsch()`, all atoms:


| species | argmin | `max` (head) | worst tied |
|---|---|---|---|
| `CC(C)(C)C` | 0.000 Å | **9.157 Å** | 9.432 Å |
| `CC(C)(C)CC` | 0.000 Å | **9.427 Å** | 9.427 Å |
| `C1CCCCC1` | 0.000 Å | **5.240 Å** | 5.240 Å |
| `C1CCC1` | 0.000 Å | **2.832 Å** | 2.832 Å |
| `Cc1ccc(C)cc1` | 0.000 Å | **2.293 Å** | 2.293 Å |

So head returns a 9.2 A map for neopentane where 0 A map is available and equally valid

BUT `min` is not the answer either. On `test_get_atom_map_5`' reaction (C4H7 + O2 <> HO2 + C4H6), the two competing maps differ by butadienes C2 symmetry and measuring both:

| | all-atom | heavy-atom |
|---|---|---|
| `max` → `[6, 5, 4, 9, 3]` | **2.674 Å** | **0.506 Å** |
| `min` → `[3, 4, 5, 10, 6]` | 3.384 Å | 0.864 Å |

`max` is the better map here. So the legacy test is not enshrining a worse anser. `max` is wrong for self maps, `min` is wrong for the reaction; both are positional heuristics dressed up as a rule

So, to resolve this, break the tie on the quantity the backbone score cannot see. Tied candidates are scored a second time by the Kabsch RMSD of the full atom map each yiels, and the last candidate that also ties on that is used. Residual position based selection stays `max` which reproduces base where nothing discriminates.


This helps with both regimes and changed no existing test expectations:
- every self-map in the table above now costs 0.000 Å. Four of the five return the literal identity permutation; cyclobutane returns `[1, 0, 3, 2, 7, 6, 5, 4, 10, 11, 8, 9]`, a C2 automorphism of the ring, which is an equally valid zero-cost answer rather than a residual defect;
- `test_get_atom_map_5` still returns `[6, 5, 4, 9, 3]`, unchanged;
- `arc/mapping/`: 85 passed


Two things that were not fixed. Benzene still returns a ring roation that is not automorphism of the stored Kekule Molecule. All six candidates costs 0.0 A, so the displacement ciretion cannot discrimante them and bond order preservation would need its own criterion. Was not added due to it being speculative. And the change adds one new `UserWarning` in `mapping` where near-collinear fragment has a genuine multi-candidate tie. The residual returns is still correct, only the rotation is non-unique and we discard the rotation. But it is a new warning.

alongd (approved):

PR #948, opened 2026-08-04
Assign every hydrogen and reject an incomplete atom map

calvinp0:
**Stack position:** top of the atom-mapping stack — sits on #935, which sits on #947. Read them in that order.

The order matters: #935 makes the XH3 branch of `map_hydrogens` always return an assignment for all three hydrogens. Only once that hole is closed can the completeness check added here be made unconditional — on `main` it would fire on ordinary XH3 groups. This PR closes the two remaining holes and then makes any future hole fail loudly.

### The gap

- `map_hydrogens` handled an XH2 group with `if mapped: … continue` and no `else`, so whenever `_map_xh2_group` returned `{}` — no mapped heavy neighbour, or a heavy atom not carrying exactly two hydrogens — **both hydrogens were silently left out of the map**.
- A heavy atom carrying four or more hydrogens alongside a heavy neighbour matched no branch at all and was dropped the same way.

### Why omission was not a clean failure

- `map_two_species` builds its list map with `[v for k, v in sorted(atom_map.items(), …)]` and no length check, so a missing key **shifts every subsequent index by one** — every atom past the gap silently refers to the wrong one, and the reaction is mapped with a plausible-looking but wrong map.
- It surfaces much later as an `IndexError` in `glue_maps`, which `try_mapping` does not catch, since it catches only `ValueError`.

### The fix

- Both cases now fall back to assigning the hydrogens in the order they appear in the two species, each with a DEBUG log naming the centre.
- `map_two_species` checks that the map is a permutation of both species' atoms before returning it, raising `ValueError` otherwise — so any future gap degrades to a mapping failure instead of corrupting indices.
- **XH2 needs no parity guard.** Unlike an XH3 group, the two hydrogens of an XH2 group are related by a reflection rather than by a rotation, so there is no handedness for the order-based fallback to preserve. (Contrast #935, where the parity of an XH3 assignment is load-bearing.)

### Behaviour change to be aware of

`map_two_species` can now raise `ValueError` where it previously returned a shifted map. Not every caller catches that, so the failure mode differs by path:

**Already guarded** — the raise degrades to a clean mapping failure:
- everything reached from `map_rxn` (`map_pairs`, `reorder_p_label_map`), which sits inside `try_mapping`'s `except ValueError`
- `get_new_zmat_2_map`, reached via `combine_coordinates_with_redundant_atoms`, inside `h_abstraction`'s `except (ValueError, SpeciesError)`

**Not guarded** — an incomplete map now surfaces as an uncaught exception rather than as silently wrong indices:
- `map_isomerization_reaction` → `map_general_rxn` → `map_reaction`; `ARCReaction.atom_map` does not catch, so it escapes the property
- `h_abstraction` (`arc/job/adapters/ts/heuristics.py:897`), which the Heuristics TS adapter calls without a guard

This is the intended trade — an exception beats a corrupted map — but it is a real change in failure mode, and reviewers should decide whether those two paths want their own guard. With the fallbacks added here a gap should not be reachable through the normal paths; the check is a backstop.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/arc/mapping/driver.py
+++ b/arc/mapping/driver.py
@@ -71,6 +71,9 @@ def map_reaction(rxn: ARCReaction,
         if general_map is not None:
             return check_atom_map_and_return(general_map)
         return map_reaction(rxn, backend=backend, flip=True)
+    if rxn.product_dicts and all(product_dict.get('discovered_in_reverse', False)
+                                 for product_dict in rxn.product_dicts):
+        return map_reaction(rxn, backend=backend, flip=True)
     raw_map = try_mapping(rxn)
     if raw_map is None:
         return map_reaction(rxn, backend=backend, flip=True)
@@ -303,7 +306,7 @@ def map_rxn(rxn: ARCReaction,
         p_label_map = rxn.product_dicts[pdi]['p_label_map']
         template_products = rxn.product_dicts[pdi]['products']
     except (IndexError, KeyError) as e:
-        logger.error(f"No valid template maps for reaction {rxn} ({rxn.family}), cannot atom map. Got:\n{e}")
+        logger.debug(f"No valid template maps for reaction {rxn} ({rxn.family}) in this orientation. Got:\n{e}")
         return None
     try:
         template_order = get_template_product_order(rxn, template_products)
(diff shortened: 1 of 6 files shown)


calvinp0:
Flattened into #935. `fix_mapping_reverse_discovery` now carries both, rebased onto main (0 behind) and restructured to six commits with no file in more than one; the XH3-anchoring, XH3-parity and hydrogen-completeness rationale is preserved in the merged `engine.py`/`engine_test.py` messages.

Content identity across the rebase was clean in both directions. Note the local ref for this branch had drifted from origin and held a different commit set, so #935's content is what was flattened, not the local variant.

PR #962, opened 2026-08-12
Fix shell quoting in rmg_env_command and silent path-override fallthrough

alongd:
Three independent pre-existing defects on `main`, surfaced while reviewing #917 but not caused by it — `arc/job/env_run.py` is untouched by that PR, and `arc/settings/external_paths.py` / `arc/settings/settings.py` arrived via #887. Keeping them off #917 so they can be judged on their own.

One commit per fix, each test-first.

---

### 1. `env_run`: shell-quote `cwd` and `env_vars` in `rmg_env_command`

`rmg_env_command` built its preamble with bare double quotes:

```python
preamble.append(f'cd "{cwd}"')
preamble.append(f'export {key}="{val}"')
```

Double quotes don't stop command substitution — `$(...)` and backticks still expand inside them. `arc/statmech/arkane.py` calls this with `cwd=statmech_dir`, and `statmech_dir` descends from the project directory, which `ARC.py` takes from the user's `input.yml` and `arc/main.py` never validates (`check_project_name` guards only the project *name*). So a `project_directory` of `/tmp/run$(...)work` executes at Arkane time, as the invoking user.

Now `shlex.quote()` on `cwd` and on each `env_vars` value, plus `mamba_exe` and `rmg_python`. `suffix` is deliberately left verbatim — its docstring documents it as taking redirections and process substitution by contract, and quoting it would break every caller; instead the docstring now states the trust boundary. `py_args` is left alone too, since callers already quote it (`' '.join(shlex.quote(a) for a in ...)`) — that contract was implicit and is now written down.

Three of the new tests actually execute the generated script through bash with a `touch <marker>` payload and assert the marker was never created, rather than just string-matching. Reverting the implementation fails 6 tests:

```
FAILED arc/job/env_run_test.py::TestRmgEnvCommand::test_cwd_command_substitution_is_neutralized
FAILED arc/job/env_run_test.py::TestRmgEnvCommand::test_cwd_with_backticks_is_neutralized
FAILED arc/job/env_run_test.py::TestRmgEnvCommand::test_env_vars_command_substitution_is_neutralized
FAILED arc/job/env_run_test.py::TestRmgEnvCommand::test_cwd_with_space_is_quoted_in_script
FAILED arc/job/env_run_test.py::TestRmgEnvCommand::test_cwd_with_single_quote_is_quoted_in_script
FAILED arc/job/env_run_test.py::TestRmgEnvCommand::test_env_vars_with_space_is_quoted_in_script
6 failed, 23 deselected in 0.94s
```

A happy-path test pins that a plain `cwd` still `cd`s correctly, so the quoting didn't trade a bug for a regression.

### 2. `external_paths`: treat an `ARC_*` override as authoritative

All five finders (`ARC_GOFLOW_REPO`, `ARC_GOFLOW_CKPT`, `ARC_GOFLOW_FEAT_DICT`, `ARC_RITS_REPO`, `ARC_RITS_CKPT`) treated an explicit override as merely the *first candidate*. When it failed validation they fell through to `~/Code/<repo>` and the ARC-sibling location, silently.

The failure mode this produces: someone A/B-tests a new checkpoint, mistypes the path or has the mount missing, and ARC runs the **old** checkpoint and reports success. Nothing logged, and the results look fine.

An explicit override is now authoritative — validated against that finder's existing criterion, returned if it passes, and if it fails a warning is logged naming the variable, the path and the check, and `None` is returned without consulting the other candidates. The validation criteria themselves are unchanged.

This also adds the resolution-order tests that were missing. The previous suite passed even with all four env-var tiers demoted to last and the epoch sort inverted — the documented precedence wasn't pinned by anything.

### 3. `imports`: re-derive dependent paths when a repo path is overridden

`arc/settings/settings.py` derives dependents eagerly at import:

```python
GOFLOW_REPO_PATH = find_goflow_repo()
GOFLOW_CKPT_PATH = find_goflow_ckpt(GOFLOW_REPO_PATH)
```

`arc/imports.py` then overlays `~/.arc/settings.py` onto the settings dict. Overriding only `GOFLOW_REPO_PATH` gave you the new repo with the checkpoint still derived from the auto-discovered one — GoFlow running repo A's Hydra configs against repo B's weights, which yields silently wrong TS guesses rather than an error.

Fixed centrally in `arc/imports.py` rather than in the adapter, since `goflow_ts.py` already reads through `arc.imports.settings`. A small `(parent, dependent, finder)` table drives the re-derivation, and only when the parent was overridden and the dependent was not — an explicit override of the dependent always wins. The re-derivation is an importable named function so it's testable without touching a real `~/.arc/`.

---

**Testing.** Full `arc/` suite on this branch:

```
2667 passed, 6 skipped, 4 warnings, 29 subtests passed in 469.90s
```

The 6 skips are the pre-existing `kinbot_env` / `uma_env` environment gates, unchanged from `main`.

Each fix was written test-first, and each test was confirmed to fail against the unfixed implementation before the fix landed.

### DIFF
--- a/arc/imports.py
+++ b/arc/imports.py
@@ -2,12 +2,58 @@
 This module contains functionality to import user settings and fill in default values from ARC's settings.
 """
 
+import logging
 import os
 import sys
 
 import arc.settings.settings as arc_settings
+from arc.settings.external_paths import (find_goflow_ckpt,
+                                          find_goflow_feat_dict,
+                                          find_rits_ckpt,
+                                          queue_deferred_warning)
 from arc.settings.inputs import input_files
 from arc.settings.submit import incore_commands, pipe_submit, submit_scripts
+
+logger = logging.getLogger('arc')
+
+
+# (parent_key, dependent_key, finder_fn) triples: when a local ~/.arc/settings.py
+# overrides parent_key without also overriding dependent_key, dependent_key was
+# derived from the OLD parent_key at auto-discovery time and must be re-derived
+# from the new one, or it silently keeps pointing at a mismatched checkout/artifact.
+_OVERRIDDEN_DEPENDENTS = (
+    ('GOFLOW_REPO_PATH', 'GOFLOW_CKPT_PATH', find_goflow_ckpt),
+    ('GOFLOW_REPO_PATH', 'GOFLOW_FEAT_DICT_PATH', find_goflow_feat_dict),
+    ('RITS_REPO_PATH', 'RITS_CKPT_PATH', find_rits_ckpt),
+)
+
+
+def resolve_overridden_dependents(settings: dict, local_settings_dict: dict) -> None:
+    """
+    Re-derive dependent settings whose parent was overridden without the dependent itself.
+
+    A local ~/.arc/settings.py may override e.g. GOFLOW_REPO_PATH while leaving
+    GOFLOW_CKPT_PATH unset; since arc/settings/settings.py derives GOFLOW_CKPT_PATH
+    from GOFLOW_REPO_PATH at auto-discovery time (before the overlay), leaving it
+    untouched here would mix the new repo with the old, auto-discovered checkpoint.
+    An explicit override of the dependent itself always wins and is left alone.
+
+    Args:
+        settings (dict): The merged settings dict, updated in place.
+        local_settings_dict (dict): The keys explicitly set by the local
+            ~/.arc/settings.py overlay.
+
+    Returns:
+        None: `settings` is mutated in place.
+    """
+    for parent_key, dependent_key, finder in _OVERRIDDEN_DEPENDENTS:
+        if parent_key in local_settings_dict and dependent_key not in local_settings_dict:
+            settings[dependent_key] = finder(settings[parent_key])
+            if settings[dependent_key] is None:
+                msg = f'{parent_key} was overridden to "{settings[parent_key]}", but no valid ' \
+                      f'{dependent_key} could be re-derived from it; {dependent_key} is unset.'
+                logger.warning(msg)
+                queue_deferred_warning(msg)
 
 
 # Common imports where the user can optionally put a modified copy of settings.py or submit.py file under ~/.arc
@@ -27,6 +73,7 @@ if os.path.isfile(local_arc_settings_path):
     if local_settings:
         local_settings_dict = {key: val for key, val in vars(local_settings).items() if '__' not in key}
         settings.update(local_settings_dict)
+        resolve_overridden_dependents(settings, local_settings_dict)
         # Set global_ess_settings to None if using a local settings file (ARC's defaults are dummies)
         settings['global_ess_settings'] = local_settings_dict['global_ess_settings'] \
             if 'global_ess_settings' in local_settings_dict and local_settings_dict['global_ess_settings'] else None
--- a/arc/job/env_run.py
+++ b/arc/job/env_run.py
@@ -30,6 +30,7 @@ The launcher is detected at call time, with the active one (per
 """
 
 import os
+import shlex
 import shutil
 import subprocess
 from pathlib import Path
@@ -181,7 +182,7 @@ _ARC_ENV_ACTIVATION_VARS = (
 _NO_LAUNCHER_MSG = 'micromamba, mamba, or conda is required to run RMG helper scripts'
 
 
-def rmg_env_command(py_args: str,
+def rmg_env_command(py_args: str | list[str],
                     cwd: str | None = None,
                     env_vars: dict[str, str] | None = None,
                     suffix: str = '',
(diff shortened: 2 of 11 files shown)


arc/job/env_run.py
```diff
@@ -201,12 +202,16 @@ def rmg_env_command(py_args: str,
 
     Args:
         py_args (str): Everything after ``python``, e.g. ``'-m arkane input.py'``
-                       or ``f'{script_path} {in_path} {out_path}'``.
+                       or ``f'{script_path} {in_path} {out_path}'``. Callers are
```
calvinp0: says callers are responsible for quoting py_args, but no caller does. `processor.py`, `output.py`, `arkane.py` do not import shlex.

calvinp0: so in `processor.py` at line 261, `species_thermo_path` comes from `output_directory` -> `project_directory` which is invalidated by user input. its split into py_args unquoted, so `$(...)` in the project directoy still exists

same for `reactions_kinetics_path`

arc/settings/external_paths.py
```diff
@@ -29,8 +29,14 @@
...
 """
 
+import logging
 import os
 
+# Using logging.getLogger('arc') directly (matching arc.common's own
+# instantiation) rather than importing arc.common.get_logger(): this module
+# is imported by arc.settings.settings, which arc.common imports at module
+# load time, so importing arc.common here would create a circular import.
+logger = logging.getLogger('arc')
```
calvinp0, line 39: runs at import time before `initialize_log()` attaches handlers, so this warning only goes to stderr and not `arc.log`

arc/imports.py
```diff
@@ -6,10 +6,45 @@
...
+    Args:
+        settings (dict): The merged settings dict, updated in place.
+        local_settings_dict (dict): The keys explicitly set by the local
+            ~/.arc/settings.py overlay.
+
+    Returns:
+        None: `settings` is mutated in place.
+    """
+    for parent_key, dependent_key, finder in _OVERRIDDEN_DEPENDENTS:
+        if parent_key in local_settings_dict and dependent_key not in local_settings_dict:
```
calvinp0, line 50: if the new parent path has no valid checkpoint, this silently sets the dependent to None.

calvinp0 (changes requested):

alongd:
Thanks @calvinp0 — all four points addressed. Rebased onto `main` and each fix squashed into the commit it amends, so the branch is still the same three commits.

### `py_args` is not quoted by any caller (`env_run.py:205`, and the `processor.py:261` / kinetics follow-up)

Right, and the docstring contract was the actual defect: a guarantee that no call site honours isn't one. Rather than sprinkling `shlex.quote` across the callers, `rmg_env_command` now takes `py_args: str | list[str]` and quotes each element of a list itself, so the property is structural instead of documentary. A plain `str` is still spliced verbatim, so any out-of-repo caller is unaffected. All six in-repo call sites pass lists — including `['-m', 'arkane', 'input.py']` as three separate tokens, which is the one trap in the migration: collapsed into a single element it would be quoted into one argv token and python would fail with "No module named".

Worth being precise about what this buys in production. `shlex.quote` leaves `[A-Za-z0-9_@%+=:,./-]+` untouched, so a normal run generates byte-identical bash. The wins are the non-adversarial cases: a project directory containing a space silently split into two argv tokens, so the RMG-side script received a truncated path and failed with a confusing `FileNotFoundError` from inside `rmg_env`; and one containing parentheses was a bash *syntax error* that killed the script before python started. New tests cover both, alongside `$(...)` and backticks, and follow the existing style in this file — executing the generated script through bash with a `touch <marker>` payload and asserting the marker was never created, rather than string-matching.

### Import-time warning only reaches stderr, never `arc.log` (`external_paths.py:39`)

Confirmed — `arc/settings/external_paths.py` is imported through `settings.py` well before `initialize_log()` attaches its handlers, so those warnings never landed in the artifact anyone debugging a run actually reads. All five override warnings are now also queued in a module-level buffer that `initialize_log()` drains once the handlers exist. Draining clears the buffer, so a second `initialize_log()` call (tests do this) doesn't duplicate lines. The import direction is `arc.common` → `external_paths`, which is the safe one: `external_paths` still uses `logging.getLogger('arc')` directly rather than importing `arc.common`, to avoid the cycle through `arc.common` → `arc.settings.settings`.

### Silent `None` when the overridden parent has no valid checkpoint (`imports.py:44`)

Fixed — that was the same silent-failure class the commit set out to remove. The re-derivation now warns, naming the parent key, the new parent path and the dependent key left unset, through the same deferred buffer so it reaches `arc.log` too.

Also renamed `_resolve_overridden_dependents` → `resolve_overridden_dependents`, since the test imports it.

Tests on the touched modules (`env_run`, `imports`, `external_paths`, `common`, `processor`, `output`, `arkane`), run with a clean `HOME` so a local `~/.arc/settings.py` can't mask override behaviour: **274 passed**.

calvinp0 (approved):

PR #963, opened 2026-08-12
Preserve Level args through the as_dict round-trip

calvinp0:
### Symptom

- A user sets extra keyword args on a level, e.g. `args={'keyword': {'opt': 'opt=(verytight)'}}`.
- The **initial** job carries the keyword correctly.
- Every **troubleshooting re-run** of that job drops it — the ESS input file comes back without `verytight`. The loss is silent: nothing is logged.
- The same loss applies to `Level.copy()` and to levels written to / restored from `restart.yml`.

### Root cause

- `Level.as_dict()` gated the `args` entry on `all(self.args.values())`.
- `Level.__init__` always normalises args into **both** slots: `{'keyword': {...}, 'block': {}}`.
- So supplying only keyword args (the common case) leaves `block` empty → `all(...)` is `False` → the whole `args` entry is dropped from the dict.
- That dict is the reconstruction path:
  - `arc/level.py:248` — `Level(repr=<Level>)` rebuilds via `self.repr.as_dict()`
  - `arc/level.py:166` — `Level.copy()` rebuilds via `as_dict()`
  - `arc/scheduler.py:3659` — `troubleshoot_ess()` re-wraps `job.level` with `Level(repr=level_of_theory)` before re-running

### The fix

**1. `Level.as_dict()` — `all(...)` → `any(...)`**

- `args` is kept when **at least one** slot has content, and still omitted when both are empty.

Preserving `args` then exposed two further defects. Both are pre-existing, but the predicate had been accidentally masking them, and without these two changes this PR would turn each into a regression.

**2. Stop aliasing the Level's args into the job args**

- `arc/scheduler.py:988` did `args.update(level_of_theory.args)` — a top-level update that **binds the Level's nested `keyword`/`block` dicts by reference** into `job.args`.
- `set_job_args()` did `args = level.args`, aliasing the same way.
- Adapters then mutate those dicts in place (e.g. Orca's `setdefault` of `scf_convergence` / `dft_grid`), so adapter-injected keywords ended up **inside the Level itself**.
- Previously invisible, because `as_dict()` discarded them. With `args` preserved they would reach `restart.yml` and `output.yml`, and on restart be promoted to genuine **user** args — after which `setdefault` could no longer upgrade `dft_grid` to `defgrid3` for a fine/freq job.
- Both sites now copy.

**3. Stop a false "ARC ignores your options" warning**

- `set_job_args()` warned whenever the job args and the level args were both non-empty.
- The scheduler already merges the level args into the job args, so with `args` preserved this fired on ordinary **first** runs — while the options were in fact applied. That branch discards nothing.
- The warning is now limited to options the job args do not carry, i.e. only options genuinely being dropped.

**Where the copy belongs — settled by experiment.** A `deepcopy` inside `as_dict()` does **not** fix the pollution: adapters mutate `level.args` in place, so by the time `as_dict()` runs the Level is already polluted and `as_dict()` faithfully serializes it. Copying at the boundary (scheduler / `set_job_args`) is what actually prevents it, which is what this PR does. A `deepcopy` in `as_dict()` would solve a *different* problem — `d['args'] is level.args`, so a caller can mutate the Level through the returned dict — but that aliasing is **not specific to args**: `d['compatible_ess'] is level.compatible_ess` too. Hardening only `args` would be inconsistent, so it is left for a separate change.

### Test expectation that encoded the old behaviour

`arc/main_test.py::TestARC::test_as_dict` (Dec 2022) sets a **keyword-only** arg at line 64 and then expected the serialized dict to omit `args` entirely — i.e. it asserted the bug. Calling that out rather than changing it quietly: the expectation now includes the `args` entry.

### Verification

Full `arc/` suite, run the way CI runs it (`-n 6 --dist worksteal`), base vs. this branch:

```
BASE (origin/main): 5 failed, 2640 passed, 36 skipped
THIS BRANCH:        5 failed, 2646 passed, 36 skipped
```

Identical failure set (5 pre-existing `torch_ani_test` environment failures, `TypeError: argument should be a str or an os.PathLike object`), +6 = the six new tests. No test goes PASS → FAIL.

**Every new test was proven to fail without the fix.** Reverting only `arc/level.py`:

```
FAILED arc/level_test.py::TestLevel::test_as_dict_preserves_args_when_one_slot_empty
  - AssertionError: 'args' not found in {'method': 'wb97xd', 'basis': 'def2tzvp', ...}
FAILED arc/level_test.py::TestLevel::test_copy_preserves_args
  - AssertionError: {'keyword': {}, 'block': {}} != {'keyword': {'opt': 'opt=(v...
FAILED arc/job/adapters/gaussian_test.py::TestGaussianAdapter::test_user_keyword_args_survive_a_level_round_trip
  - KeyError: 'opt'
```

Reverting only `arc/job/adapters/common.py`:

```
FAILED arc/job/adapters/common_test.py::TestJobCommon::test_set_job_args_warns_only_for_dropped_options
  - AssertionError: Unexpected logs found: ["WARNING:arc:When troubleshooting j...
FAILED arc/job/adapters/common_test.py::TestJobCommon::test_set_job_args_does_not_alias_level_args
  - AssertionError: 'dft_grid' unexpectedly found in {'opt': 'opt=(verytight)',...
FAILED arc/job/adapters/orca_test.py::TestOrcaAdapter::test_writing_input_does_not_pollute_level_args
  - AssertionError: {'keyword': {'scf_convergence': 'tightscf', 'dft_grid'...
```

The Gaussian test is the one that pins the motivating symptom end to end: it rebuilds the level via `Level(repr=level)` exactly as troubleshooting does, writes the input file, and asserts `verytight` is in the route section.

### Compatibility

- **Reading** existing `restart.yml` files is unaffected.
- **Newly written** `restart.yml` / `output.yml` do change: a level with user args now serializes them (which is the point of the fix). Adapter-injected keywords are *not* serialized — item 2 above.

### Reuse check

Searched before writing, by behaviour rather than by name: whether anything already compensates for a Level's args being lost across the dict round-trip, and where levels are serialized and reconstructed. The round-trip sites are `Level.as_dict`, `Level.copy`, and `Scheduler.troubleshoot_ess`; nothing compensated anywhere. No new helper was added — this is a predicate correction inside the method that already owns the behaviour, plus copies at two existing call sites.

### Deliberately out of scope

- `Level.__str__` (`arc/level.py:151`) carries the same `all([val for val in self.args.values()])` pattern, so args are also omitted from the human-readable string when only one slot is filled. **Deferred deliberately, with evidence** — `str(level)` is the lookup key into `data/freq_scale_factors.yml` (`arc/level.py:471`, `arc/output.py:191`), and all 67 keys there are plain `method/basis[, software: x]` strings carrying no args. Flipping the predicate was measured to change `assign_frequency_scale_factor()` for a level with keyword-only args from **0.967 to `None`**, i.e. it silently disables frequency scaling for exactly the users this PR is meant to help. The full suite stays green through that regression, so it needs its own change with the lookup normalised (args stripped for the key) and its own tests. This PR leaves `str(level)` byte-identical.
- `arc/level.py:291-292` uses the local `args` dict instead of `self.args`, so the `str` / iterable forms of the `args` argument that the docstring and type hint advertise raise `AttributeError`. Pre-existing on both base and this branch.
- `arc/scheduler.py:986-988`: because the update replaces the whole `keyword` slot, `args['keyword']['scan_trsh']` set one line earlier is always clobbered. Pre-existing and unchanged here.


alongd:
Two-axis review of `official/main...fix_level_args` — **Standards** (does it conform to the repo's documented conventions?) and **Spec** (does it faithfully implement what was asked?). The axes were reviewed independently and are reported separately, unmerged.

There is no linked issue or PRD, so the Spec axis was run against the PR description. Since that description is self-authored alongside the code, its factual claims were checked against the code rather than taken at face value.

---

## Standards

### Hard violations (documented standard)

- **Multi-line docstrings must start on the line after `"""`** — `CLAUDE.md`, Code Style & Conventions. `arc/job/adapters/gaussian_test.py:1168` and `arc/level_test.py:130` both open with text on the `"""` line followed by a body paragraph. The single-line test docstrings elsewhere in the diff are correct.
- **Missing extreme case** — `CLAUDE.md`, Testing: "trivial → normal → wild-type → extreme". The new comprehension at `arc/job/adapters/common.py:512` assumes both `level.args[key]` and `args[key]` are dicts. The pre-existing test three lines above the new ones (`common_test.py:166`) uses the flat shape `args={'keyword': 'k1'}`, which raises `AttributeError: 'str' object has no attribute 'get'` the moment the level also carries args. Adapters type `args` as a bare `dict`, so nothing normalises this.
- **Untested line** — `arc/scheduler.py:989` (`args.update(copy.deepcopy(...))`) gains no test; all six new tests exercise the `common.py` path.

### Judgement calls

- **Duplicated Code + Feature Envy** — `common.py:520` and `scheduler.py:989` both reach into `Level.args` and defensively deep-copy it, and both modules had to add `import copy` for it. The copy arguably belongs on `Level` (a `get_args() -> dict` returning a deep copy), leaving one owner of the aliasing invariant instead of every future call site having to remember.
- **Divergent predicate** — `level.py:190` becomes `any(...)`, but `level.py:151` still reads `all([val for val in self.args.values()])`. The same question — "does this level carry meaningful args?" — is now answered two ways: a keyword-only level serialises its args but stays silent about them in `__str__`. (The PR explains why `__str__` is deliberately out of scope; the note here is only that the two predicates would benefit from a single named helper once that follow-up lands.)
- **Mysterious Name** — `common.py:512-515` uses `key_2`/`val_2` in a nested comprehension, then rebinds `ignored_args` to a filtered version of itself. It is the densest hunk in the diff and the one the new docstring paragraph exists to explain; a named helper (`dropped_level_args(args, level_args) -> dict`) would carry the explanation and admit a direct test.
- **Fragile assertion** — `gaussian_test.py:1187` pins the route section to `content.splitlines()[4]`. Locating the route line by its `#` prefix would survive an unrelated header change.

---

## Spec

### Verified

Fix 1 (the `any` predicate, `level.py:190`), fix 2 (`scheduler.py:989`, `common.py:520`) and fix 3 (the reworked warning) are each present at every site the PR names.

**The aliasing claim holds.** Both boundary copies are `copy.deepcopy`, not shallow — a shallow copy would not have stopped mutation of the nested `keyword`/`block` slots. The third round-trip site is safe for a reason the description does not state: `as_dict()` returns `clean_dict['args'] = self.args` *by reference*, but `Level.__init__` builds a fresh `{'keyword': {}, 'block': {}}` and refills it without mutating the source, so `Level(repr=<Level>)` and `copy()` come out deep. Every adapter mutation site (`orca.py:284`/`297`, `xtb_adapter`, `common.py:276`) touches job args only. No remaining path binds a Level's nested dicts into job args.

**The compatibility claim holds.** `build()` maps a missing `args` key to `None` → `self.args = args or {'keyword': {}, 'block': {}}`, and `'args'` is already in `allowed_keys`, so neither an old nor a new restart dict trips the illegal-key check.

### (a) Partial

> "Options from `level` that are already present in `args` are applied, and are not reported."

`set_job_args` applies nothing in that branch — it relies on `Scheduler.run_job` having merged already. A caller that passes non-empty args without that merge (e.g. direct adapter construction) gets the level args dropped, not applied.

> "only for options which are actually being dropped"

Converse of the same clause: when `args` is non-empty but every value is falsy (`{'keyword': {}, 'block': {}}`), neither branch runs — the level args are dropped silently, with no warning and no copy. The structure is pre-existing, but this PR newly states it as the contract.

### (b) Scope creep

None. The diff is exactly the three fixes, six tests, and the one `main_test.py` expectation the description calls out.

### (c) Implemented but looks wrong

- **`args.get(key, dict()).get(key_2, None)` assumes `args[key]` is a dict.** `common_test.py:166` asserts that `set_job_args(args={'keyword': 'k1'}, ...)` is valid input — it passes today only because that level (`CBS-QB3`) carries no args, so the new branch is never entered. With an args-carrying level the same shape raises `AttributeError`. The base code never indexed into `args`, so this is new. An `isinstance(args.get(key), dict)` guard would close it.
- **The message no longer matches what it detects.** `f'When troubleshooting {job_name}, ARC ignores the following user-specified options:'` — `set_job_args` runs at every job init, and the reworked condition now fires precisely on the *non*-troubleshooting paths (the scheduler pre-merges, so trsh re-runs no longer warn). Item 3 reworks the condition but leaves the sentence naming the wrong context.

---

**Summary** — Standards: 7 findings; worst is the flat-`args` shape breaking the new comprehension (the untested extreme case). Spec: 4 findings; worst is the same `AttributeError` on `args={'keyword': 'k1'}` with an args-carrying level. The two axes converged on that one independently, which is the strongest signal in the review.

alongd (approved):

PR #964, opened 2026-08-12
Recover ARC thermo when Arkane's thermo-library save fails (A+A reactions)

calvinp0:
## What a user sees go wrong

Run any reaction whose two reactants are the **same species** — `OH + OH <=> H2O + O` — and:

- `output.yml` reports `thermo: null` for **every** species in the run, reactants and products alike.
- The project thermo library comes out **empty**.
- `arc.log` shows Arkane ran fine, then a `DatabaseError` traceback at the very end.

Arkane had **already computed** the full NASA polynomials. They are sitting in `calcs/statmech/thermo/output.py`. ARC just never picked them up, so a completed, expensive job reported nothing.

## Why

- Arkane's `save_thermo_lib` keys thermo-library entries by adjacency list + multiplicity and raises `DatabaseError` on a duplicate. Two identical reactants = a duplicate.
- It is the **second-to-last statement of `Arkane.execute()`** — it runs **after** `output.py` is written and **before** `RMG_libraries/thermo.py` is.
- `save_arkane_thermo.py` guarded on exactly that missing library file and returned early:
  ```python
  if not os.path.isfile(thermo_lib_path):
      return
  ```
- No `thermo.yaml` → `parse_arkane_thermo_output` leaves `spc.thermo` unset → null thermo everywhere.

One duplicate species loses thermo for the **whole run**, because the library save is all-or-nothing.

## The fix

**1. Reload from `output.py`.** When `RMG_libraries/thermo.py` is absent, reconstruct each species' thermo straight from `output.py` and write `thermo.yaml` exactly as the library path would.

- Every job has already written `output.py` by the time `save_thermo_lib` runs, so each species' **own** thermo is recovered — nothing is inferred from another species.
- Blocks are located with `ast.parse`, walking `Call` nodes named `thermo`, then evaluated with the real rmgpy thermo classes in scope — the same context Arkane itself uses to read these files back.
- Recovers from **any** `save_thermo_lib` failure, not only A+A.
- Happy path untouched: when the library file exists it is used and `output.py` is not read. Covered by a test that passes before and after.

**2. Write each library entry once.** This is required *by* the reload, not optional alongside it: once both A+A duplicates carry thermo, `save_thermo_lib` emits two entries whose adjacency list and multiplicity match, and `rmgpy`'s `ThermoLibrary.load_entry` refuses the file:

```
DatabaseError: Adjacency list and multiplicity of R2 matches that of existing
molecule R1 in thermo library. Please correct your library.
```

Before the fix the library was *empty*; after the reload alone it would be *unloadable*. `plotter.save_thermo_lib` now applies the **same isomorphic + equal-multiplicity test RMG applies** and omits the later entry with a warning.

- This is a **library-file uniqueness** constraint, not a claim about thermochemical identity.
- It changes the library file **only**. Every species keeps its own computed thermo in `output.yml` and everywhere else it is reported.

## What is deliberately *not* here

An earlier revision of this PR also deduplicated species **before** the Arkane input was written, collapsing isomorphic same-multiplicity species into one block and copying the representative's thermo onto the other. **That has been removed.** It skipped a real calculation and assigned one species' thermodynamics to another, and ARC's `Molecule` is a 2D graph with no stereochemistry and no geometry — so it collapsed *cis*/*trans*-2-butene, *cis*/*trans*-1,2-dimethylcyclohexane, R/S enantiomers, and two distinct conformers of the same SMILES submitted as separate species. Errors of 3.1× to 17.3× in K_eq, with the sign set by list order, and it fired project-wide rather than only on A+A.

Its only real benefit was suppressing a stderr traceback. Since `save_thermo_lib` runs after `output.py` is written, the reload above already recovers every one of those cases correctly.

## Verification

- `arc/plotter_test.py`, `arc/scripts_test.py`, `arc/statmech/`, `arc/processor_test.py`, `arc/output_test.py` — **158 passed, 7 skipped**.
- `arc/scripts/save_arkane_thermo_test.py` — 7 passed under `rmg_env`; skips cleanly under `arc_env` pytest.

**Every new test was proven to fail on unfixed `main`.** With only `plotter.py` reverted:

```
FAILED TestSaveThermoLibDuplicates::test_generated_library_loads_in_rmg
  rmgpy.exceptions.DatabaseError: Adjacency list and multiplicity of R2 matches
  that of existing molecule R1 in thermo library. Please correct your library.
FAILED TestSaveThermoLibDuplicates::test_identical_species_written_once
  AssertionError: 'label = "R2"' unexpectedly found in ...
```

With only `save_arkane_thermo.py` reverted:

```
FAILED TestSaveArkaneThermoOutputPyFallback::test_both_duplicates_recover_their_own_thermo
  AssertionError: False is not true : thermo.yaml must be recovered from output.py
  when the library is absent
FAILED TestSaveArkaneThermoOutputPyFallback::test_recovered_values_match_reference_thermochemistry
  AssertionError: False is not true : thermo.yaml must be recovered from output.py
  when the library is absent
```

`test_library_path_takes_precedence` and `test_distinct_species_all_written` pass **both** before and after — they guard the unchanged paths, so they correctly do not flip.

The end-to-end test runs the **real script as a subprocess** against a fixture directory containing `output.py` and no `RMG_libraries/`, which is exactly the state Arkane leaves behind. Both A+A duplicates come out with their own recovered thermo.

## Thermochemistry the tests assert

Reference values, not whatever the code emits. The recovered product is H₂O, against JANAF:

| Quantity | Reference | Tolerance |
|---|---|---|
| ΔfH°(298.15 K) | −241.8 kJ/mol | 3 kJ/mol |
| S°(298.15 K) | 188.8 J/(mol·K) | 2 J/(mol·K) |
| H(2000 K) − H(298.15 K) | 72.79 kJ/mol | 3 kJ/mol |
| S°(2000 K) | 264.77 J/(mol·K) | 3 J/(mol·K) |

The 2000 K checks evaluate the **serialized `nasa_high` coefficients**, which is what ARC consumers use out to 3000 K — a reload that corrupted the high-temperature polynomial while leaving 298 K intact cannot pass. No assertion is a byte-exact float.

**Why OH's S° is not asserted.** The fixture's OH recovers S°(298.15 K) ≈ 178.2 against an accepted 183.708 J/(mol·K). The gap is *not* a missing R·ln 2: Arkane has no electronic-mode class, degeneracy enters through `Conformer.spin_multiplicity`, and the value already contains a spin/optical term of exactly 5.7632 = R·ln 2. What is missing is OH's **²Π orbital (Λ = ±1) degeneracy and its spin–orbit structure**: splitting 139.21 cm⁻¹ (θ = 200.29 K), q_el = 2 + 2·exp(−θ/T) = 3.0216 at 298.15 K, S_el = R[ln q + T·dln q/dT] = 11.0825 J/(mol·K); minus the 5.7632 already present leaves 5.3193, giving 183.51 against 183.708. (Adding R·ln 2 gives 183.95 — that near-agreement is coincidental.) The independent check is the heat capacity: fixture Cp₂₉₈(OH) = 29.098 vs JANAF 29.886, a deficit of 0.788 against a computed Schottky Cp_el(298) of 0.840. A temperature-independent degeneracy contributes **exactly zero** to Cp, so the deficit can only be the two-level spin–orbit term. What *is* asserted for OH is the A+A invariant: two blocks for the same species must reconstruct to identical thermo.

## Notes for review

- One commit, six files, no file touched twice.
- Two claims from the first revision of this PR description were **wrong and are withdrawn**: the reloaded `thermo.yaml` is *not* byte-compatible with the library path (`data` differs — RMG's `save_entry` routes the NASA `comment` to `shortDesc` and drops it from the repr, and `spc.thermo.data` is interpolated verbatim into ARC's library; benign today because Arkane leaves `NASA.comment` empty, but the claim was false); and `eval(..., {'__builtins__': {}})` was described as a sandbox — it is trivially escapable via `().__class__` and no longer claimed to be one. The input is a file Arkane just wrote in ARC's own run directory.
- The duplicate check skips the entry in the writer rather than mutating `include_in_thermo_lib` on the species. Mutating would serialize into `restart.yml` and permanently exclude the species on a later restart even once the duplicate is gone.
- `computed_reaction` TCKDB payloads were never affected — they carry no NASA.

### DIFF
--- a/arc/plotter.py
+++ b/arc/plotter.py
@@ -32,8 +32,9 @@ from arc.common import (NUMBER_BY_SYMBOL,
                         sort_two_lists_by_the_first,
                         )
 from arc.checks.common import TS_IRC_FAILED_MARKER, get_ts_validation_comment
-from arc.exceptions import InputError
+from arc.exceptions import InputError, InvalidAdjacencyListError
 from arc.level import Level
+from arc.molecule.molecule import Molecule
 from arc.parser.parser import parse_trajectory
 from arc.species.converter import (check_xyz_dict,
                                    get_xyz_radius,
@@ -766,6 +767,40 @@ def save_irc_traj_animation(irc_f_path, irc_r_path, out_path):
             f.write(' Normal termination of Gaussian 16\n')
 
 
+def _get_thermo_lib_duplicate(adjlist: str,
+                              written: list,
+                              ) -> str | None:
+    """
+    Return the label of an already written thermo library entry that RMG would reject
+    ``adjlist`` as a duplicate of, or ``None`` if there is none.
+
+    Applies the same test as ``rmgpy.data.thermo.ThermoLibrary.load_entry``: an isomorphic
+    molecule with an equal multiplicity. An adjacency list that cannot be parsed is reported
+    as not a duplicate, leaving the entry to be written as it was before this check existed.
+
+    Args:
+        adjlist (str): The adjacency list of the candidate entry.
+        written (list): Entries are (label, adjacency list) tuples already written.
+
+    Returns:
+        str | None: The label of the matching entry, or ``None``.
+    """
+    try:
+        mol = Molecule().from_adjacency_list(adjlist)
+    except (ValueError, InvalidAdjacencyListError):
+        logger.warning(f'Could not parse an adjacency list while checking the thermo library for '
+                       f'duplicates:\n{adjlist}')
+        return None
+    for label, written_adjlist in written:
+        try:
+            written_mol = Molecule().from_adjacency_list(written_adjlist)
+        except (ValueError, InvalidAdjacencyListError):
+            continue
+        if mol.multiplicity == written_mol.multiplicity and mol.is_isomorphic(written_mol):
+            return label
+    return None
+
+
 def save_thermo_lib(species_list: list,
                     path: str,
                     name: str,
@@ -773,6 +808,14 @@ def save_thermo_lib(species_list: list,
                     ) -> None:
     """
     Save an RMG thermo library of all species.
+
+    A species is written only if it has thermo data and ``include_in_thermo_lib`` is ``True``.
+    A species whose adjacency list and multiplicity match an already written entry is skipped
+    with a warning, because ``rmgpy.data.thermo.ThermoLibrary.load_entry`` raises
+    ``DatabaseError`` on such a pair and the whole library would then fail to load. This applies
+    to a reaction whose two reactants are the same species, which reaches this function as two
+    separately labeled entries. Skipping affects the library file only; every species keeps its
+    own computed thermo everywhere else it is reported.
 
     Args:
         species_list (list): Entries are ARCSpecies objects to be saved in the library.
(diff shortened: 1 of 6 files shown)


calvinp0:
Superseded by #1021, which carries these commits unchanged alongside the rest of the A + A handling. The three layers — per-occurrence participants in a well, the dedup-aware normal mode displacement check, and thermo recovery when Arkane declines to write its library — are one story, and the thermo recovery in particular only reads correctly with the context of why R1 == R2 reaches Arkane at all.

PR #967, opened 2026-08-12
Use the frequency log's geometry for the TS normal mode displacement …

calvinp0:
**Base of a 3-PR stack — review this one first.** Reading order: **#967 → #968 → #970**. All three change the head of `analyze_ts_normal_mode_displacement()`; they are split by *cause*, not by file.

## What breaks

- The TS normal mode displacement (NMD) check accepted or rejected a transition state using a geometry that was not in the same coordinate frame as the normal modes it was compared against.
- The geometry came from the optimization job's `final_xyz`; the modes came from the **frequency** job's log. Gaussian reports modes in that job's own orientation, which is not generally the orientation of the opt geometry.
- Measured over 996 real benchmark frequency jobs: 82% agree to within 1°, but **17.6% differ by ~180°**. These are exact rigid rotations (fit RMSD ~1e-16), not numerical noise.
- Adding a mode vector expressed in one basis to a coordinate expressed in another corrupts the cross term of every displaced bond length, so the rotation-invariance of a bond length does **not** rescue the result.

## What the fix does

- `get_ts_xyz_in_normal_mode_frame()` takes the TS geometry from the frequency job's output file — the same file the displacements are parsed from — so geometry and modes share a frame by construction.
- For Gaussian that resolves to the last `Standard orientation:` block, falling back to `Input orientation:`. Both are right: the fallback only applies when Gaussian suppressed the standard orientation (e.g. under `nosymm`, which ARC's troubleshooting can add), and in that case the frequency analysis is done in the input orientation too.
- Falls back to the TS species geometry when the freq log yields no geometry, or when the parsed geometry's element symbol sequence differs from the species'. Both fallbacks log a warning that the frame may not match.
- The fallback is one-directional. When the **species** has no geometry — `ARCSpecies.get_xyz()` returns `None` for a TS with no `final_xyz`/`initial_xyz`/conformer and no TS guess carrying an `initial_xyz` — the parsed geometry is used rather than discarded. Returning `None` there would hand the caller nothing to measure: it immediately does `len(ts_xyz['symbols'])`.
- `is_ts_atom_order_consistent_with_reactants()` makes the analysis return `None` when the TS atom order differs from the concatenated reactant order. Nothing checked this before: the forming/breaking bond indices are indices into the *reactant* order, so on a mismatch the check silently scored the wrong atoms and still returned a confident verdict.

## Why this shape

- A frame mismatch or an atom-order mismatch is a **missing precondition**, not evidence against the TS — hence `None` (analysis not performed) rather than `False` (TS rejected).
- ARC already pairs a geometry with modes from the same log elsewhere: `arc/job/trsh.py::trsh_negative_freq` parses both from one `log_file` before perturbing along the mode. The NMD check was the outlier, not the precedent.

## How it was verified

- Independently confirmed by the Eckart condition. Modes expressed in their own frame satisfy `Σ mₐ(rₐ − R_com) × dₐ = 0`, and that quantity is frame-dependent, so it is a direct test of whether the pairing is consistent: **73%** of affected jobs exceed 3σ when evaluated in the opt frame, versus **0 of 111** in the freq frame.
- 9 new tests: frames agree, a deliberately rotated geometry, an element symbol mismatch, a log with no geometry, a log that fails to parse, a species with no geometry at all, the atom-order guard (unit and end-to-end), and a scheduler-level test that an unknown (`None`) NMD verdict does not trigger a TS switch.
- `arc/checks/` passes in full — 60 tests.

## Open note for the reviewer

- The line just above the new guard, `if n_ts != n_expected: return False`, is the same class of missing precondition but still returns `False`. It predates this PR and is left alone here; worth deciding whether it should become `None` too.
- `analyze_ts_normal_mode_displacement()` still does not tolerate a `None` geometry, which after this change can only arise when the freq log **and** the species both yield nothing. That `len(ts_xyz['symbols'])` on a possibly-`None` value is unchanged from `main` and is left as-is rather than widened here.

## What I searched for

- Whether ARC already had a helper that resolves a geometry into the frame of a job's parsed modes — searched by behaviour (geometry / frame / orientation reconciliation) rather than by an intended name. `trsh_negative_freq` does it inline for a different purpose; there was no reusable helper to import.
- Confirmed the geometry parse goes through the existing `arc/parser` entry point rather than a new reader, and that the symbol-sequence comparison had no existing equivalent.

### DIFF
--- a/arc/checks/nmd.py
+++ b/arc/checks/nmd.py
@@ -33,7 +33,8 @@ def analyze_ts_normal_mode_displacement(reaction: ARCReaction,
     """
     Analyze the normal mode displacement by identifying bonds that break and form
     and comparing them to the expected given reaction.
-    Note that the TS geometry must be in the standard orientation for the normal mode displacement to be relevant.
+    The TS geometry is taken from the frequency job's output file so that it is in the same coordinate frame
+    as the normal mode displacements parsed from that file.
 
     Args:
         reaction (ARCReaction): The reaction for which the TS is checked.
@@ -58,11 +59,18 @@ def analyze_ts_normal_mode_displacement(reaction: ARCReaction,
                 not in reaction.ts_species.ts_checks['warnings']:
             reaction.ts_species.ts_checks['warnings'] += 'Atom map is None; skipped the TS normal mode displacement check; '
         return None
-    ts_xyz = reaction.ts_species.get_xyz()
+    ts_xyz = get_ts_xyz_in_normal_mode_frame(reaction=reaction, job=job)
     n_ts = len(ts_xyz['symbols'])
     n_expected = sum(spc.number_of_atoms for spc in reaction.r_species)
     if n_ts != n_expected:
         return False
+    if not is_ts_atom_order_consistent_with_reactants(reaction=reaction, ts_xyz=ts_xyz):
+        logger.warning(f'The geometry of TS {reaction.ts_species.label} lists its atoms as '
+                       f'{tuple(ts_xyz["symbols"])}, while the reactants of reaction {reaction.label} are ordered '
+                       f'{tuple(reaction.get_reactants_xyz(return_format="dict")["symbols"])}. The forming and '
+                       f'breaking bond indices refer to the reactant order, so they do not describe the intended '
+                       f'atoms of this TS. Skipping the normal mode displacement analysis.')
+        return None
     try:
         freqs, normal_mode_disp = parser.parse_normal_mode_displacement(log_file_path=job.local_path_to_output_file)
     except NotImplementedError:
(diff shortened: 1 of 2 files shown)


calvinp0:
Subsumed by #970. This branch's commit 4644b85b is a strict ancestor of #970 (verified with `git merge-base --is-ancestor`), so reviewing #970 already reviews this — the three NMD PRs were a linear stack in content while all three declared main as their base, which the PR list does not show.

#970 has been restructured so no file is touched by more than one commit, and the frequency-log-geometry rationale from this PR is preserved in full in its `arc/checks/nmd.py` commit: the frame mismatch between `final_xyz` and the modes, the per-ESS fixture verification, why `final_xyz` is not refreshed instead, and the atom-order precondition.

PR #978, opened 2026-08-13
Make the RMG family set configurable via settings, resolved in get_all_families

calvinp0:
Base: `main`. This branch absorbs **#979** (deterministic ordering + authoritative pinning) and **#982** (chemical gates on the family choice) — both were flattened into this PR and are now closed. Review this PR alone; nothing else is outstanding.

Three commits, no file touched by more than one of them. Together they make the RMG family set configurable, make family determination reproducible, and make the family that wins a chemical choice rather than an alphabetical one.

## 1. The family set is configurable, and `'all'` reaches the whole database

- `settings['rmg_family_set']` (new, ships as `'default'`) had no equivalent before: `'default'` was the signature default of `get_all_families()` and of every entry point above it, so ARC could not be asked to consider anything else short of naming a set at every call site. That excludes families ARC's own TS adapters declare support for.
- The value is resolved **inside `get_all_families()`** — the single sink every path funnels through, which already owned an in-body fallback — not threaded through signatures. Threading leaves the sink's literal `'default'` in place, so every caller that does not forward the value keeps returning the curated set; `check_family_name()` on `main` was exactly such a bare call site. Measured on a deployed installation whose `~/.arc/settings.py` asked for `'all'`: `get_all_families('all')` → 100 families, bare `get_all_families()` → 55.
- `get_product_dicts()`, `determine_family()`, `get_reaction_family_products()` and `check_family_name()` propagate `None` to mean "not specified". `determine_family()`'s shortcut to the cached `product_dicts` property keys off `rmg_family_set is None` rather than `== 'default'`, so an explicitly requested set is honoured even when it matches the configured one.
- `'all'` previously unioned only the sets named in RMG's `recommended.py`; families such as `Intra_RH_Add_Exocyclic` and `Intra_RH_Add_Endocyclic` appear in none of them despite shipping in the database. `get_all_families()` now also unions the families that exist as RMG database **directories**, via the new `get_rmg_family_directories()`, counting a directory as a family only when it holds a `groups.py`. The union is de-duplicated, which also removes the 24 duplicate labels the recommended sets alone produced (measured locally: `'default'` → 54, `'all'` → 99, of which 11 are directory-only).

**Ordering is a contract, not an accident.** Directory-only families are positioned last, so one wins only where no recommended family matched — i.e. only where the alternative was `family=None`. Every candidate must still reproduce the products the reaction asserts, so such a family cannot introduce a transformation, only propose a mechanism for one already stated. Stated in the `get_all_families()` docstring and pinned by `test_directory_only_families_are_ordered_last` and, end to end, by `test_widening_keeps_the_recommended_family_when_both_match` on `[OH] + [O]O <=> OO + [O]`, where `H_Abstraction` (recommended) and `Substitution_O` (directory-only) each match in isolation.

**This redefines `'all'` for a stock install**, because the Linear TS adapter asks for `'all'` unconditionally whenever the configured set yields no product dicts. Measured on 1,4-cyclohexadiene ⇌ benzene + H₂ at the shipped `'default'`: `origin/main` gets 0 wider-scan product dicts and `family=None`; this branch gets 16 and `H2_Loss`. Pinned by `test_wider_family_set_scan_used_by_the_linear_ts_adapter`.

`get_reaction_family_products()`'s docstring claimed `'all'` excludes surface families outright; two further statements repeated it. `'all'` skips family *sets* whose label contains `'surface'` and surface family *directories*, but a non-surface-labelled set can still list one, as `electrochem` does for `Surface_Proton_Electron_Reduction_*`. The three statements now say what the code does, and a test pins it.

## 2. Determinism and authoritative pinning

| symptom | cause |
|---|---|
| the family ARC picks depends on `PYTHONHASHSEED` | candidates came from the **set literals** in `recommended.py` (`ast.literal_eval`) and from `os.listdir()` — neither iteration order is specified |
| a pinned family could be paired with another family's atom label map | setting `family` recorded the label but left `product_dicts` holding matches from every family that matched |
| `family_own_reverse` was effectively always `False` | `__init__` set it to `False` and the property only recomputed on `None`, so the recompute branch was dead |

Measured: `[CH]1C=Cc2ccccc21 <=> [CH]1C=CC2C3=C1C=CC32` reports `Intra_Diels_alder_monocyclic` at seeds 0/1/13 and `Intra_R_Add_Exocyclic` at 2/5; `[CH2]C(C=C)OS >> [O]C(C=C)CS` flips between `intra_OH_migration` and `intra_substitutionS_isomerization`. With `intra_OH_migration` pinned and `PYTHONHASHSEED=2`, ARC reported `C0-O4` as a breaking bond — there is no `C0-O4` bond in the reactant. About **10%** of the 500-reaction benchmark selection matches more than one family.

- `recommended.py` is read with `ast.parse`, so each family set keeps its labels in written order (`get_rmg_recommended_family_sets()` now returns `dict[str, tuple[str, ...]]`). A set whose elements are not all plain string literals is sorted instead.
- `get_all_families()` sorts by label **within each tier** — recommended, then database directories, then ARC's — and de-duplicates keeping the first occurrence. Sorting within a tier rather than relying on `recommended.py`'s textual layout keeps ARC's answer independent of a cosmetic upstream reordering; sorting *across* tiers would discard the ordering contract above. Both properties hold together: `get_all_families()` is sha256-identical across `PYTHONHASHSEED` 0/1/2/3/7/13, and no directory-only family precedes a recommended one. Mutation-checked — reversing the tiers, sorting across them, or de-duplicating through a `set` each fails an ordering test; leaving a tier unsorted fails the determinism test.
- New `ARCReaction.restrict_product_dicts_to_family()` restricts `product_dicts` to a **single** family — the pinned one, or the family of the first entry when none is pinned — so a recipe is only ever combined with labels its own family generated. A pinned family matching none of the generated dicts raises `ReactionError` rather than silently deriving the TS from a foreign family or from nothing. When nothing is pinned and several match, a warning names them all.
- Assigning `family` **restricts** already-generated dicts instead of discarding them, so the Linear TS adapter's wider-family-set retry (which sets `product_dicts` then `family`) keeps what the retry recovered.
- Assigning a family invalidates `family_own_reverse`, which the new `get_family_own_reverse()` derives on demand from `ReactionFamily(label=family).own_reverse` — using the `ReactionFamily` `reaction.py` already imports at module level, so **no import line changes** and no new `py/unsafe-cyclic-import` alert is raised (those are per imported *name*). `from_dict` no longer forces `False` when the key is absent. Verified: `own_reverse` matches the `ownReverse` declaration read from `groups.py` for every family in `get_all_families()`, 0 mismatches.
- An `rmg_family_set` list mixing set names with family labels reached de-duplication as a nested list, which is unhashable there; it is now flattened.

## 3. Which family wins: direction and radical gates

New `prioritize_family_product_dicts()`, applied inside `get_reaction_family_products()` **before** the label-sort tie-break. Deterministic is not correct: the label order systematically favours the concerted families, because RMG names the pericyclic families early in ASCII and the radical-addition families late.

1. **Direction gate** — if any match was found forward, drop the reverse-discovered ones. A reverse match's `r_label_map` addresses the *flipped* reaction.
2. **Radical gate** — if the reactants carry unpaired electrons, keep only matches whose recipe gains or loses a radical, provided that leaves at least one and drops at least one.
3. **Tie-break** — fewest bonds formed/broken, then fewest changed, then the deterministic label order. Both counts are needed: `formed/broken` alone leaves `H_Abstraction` (2, 0) tied with families differing only in `changed`.

Both gates only choose among families that already matched, so they cannot invent a mechanism. The direction gate is first because it is a **validity** filter, not a preference — and because it needs no recipe, only `discovered_in_reverse`. That ordering matters operationally too: a family whose `groups.py` cannot be parsed now costs only the radical gate and the bond-change ordering, instead of silently disabling the direction gate as well, and that degradation is reported at `warning` rather than `debug`. Reading recipes after the gate also skips the recipes of dropped matches.

**Evidence.** The benchmark records which family each reaction was generated from, over 281 ambiguous reactions:

| | agrees with the recorded family |
|---|---|
| before | 122 / 281 |
| after | **227 / 281** |

Of the 130 reactions whose family changes, **110 move toward the recorded family, 5 away**. The 5 are `Ketoenol` → `intra_H_migration` and are correct: `Ketoenol/groups.py` contains no `u1` atom anywhere in its group tree — a closed-shell tautomerisation template — while all 5 reactants are delocalised doublets with spin density on the H-acceptor carbon. 4 of the 5 produce byte-identical formed and broken bonds either way. The 5th, 2-hydroxyallyl radical (`[CH2]C(=C)O`), is symmetry-degenerate and the gate measurably wins: `Ketoenol`'s atom map scrambles the four CH₂ hydrogens across both allyl termini, giving formed 5 / broken 5, which no single imaginary mode can satisfy, against 1 / 1 for `intra_H_migration`. The recorded labels are internally inconsistent for this pair anyway (5 `Ketoenol` / 3 `intra_H_migration` across 8 ambiguities, 6 of which give identical bonds).

**Contract change worth flagging.** `get_reaction_family_products()` previously returned every match; it now returns a filtered and ordered list, and the docstring says so. Measured on `C=C[CH]CCC + CC=CCCC >> C=CC(CCC)C(C)[CH]CCC` under `'all'`: 4 product dicts before (1 forward `R_Addition_MultipleBond` + 3 reverse `Retroene`), 1 after. `R_Addition_MultipleBond`/`Retroene` is the single largest ambiguous combination in the benchmark pool (118 of 281), so the direction gate is the highest-volume change here.

Downstream: `Disproportionation → H_Abstraction` moves 4 reactions from **no TS adapters at all** to heuristics/autotst/crest; `Intra_2+2_cycloaddition_Cd → Intra_R_Add_*` gains kinbot. Genuinely ambiguous pairs are left alone (`Intra_R_Add_Endo/Exocyclic`, `Cl_/H_Abstraction`, `H_Abstraction`/`Substitution_O`). **Known limitation, deliberate:** the radical gate tests `multiplicity > 1`, so a singlet biradical does not trigger it.

Chemistry review on the gates: **passed, ship as written**, with two non-blocking follow-ups neither introduced here — `get_number_of_atoms_in_reaction_zone` drops 4→3 under `intra_H_migration`, and `intra_H_migration`'s empty `changed` list will interact with the NMD family-recipe work when the two meet.

## 4. Two guards the above made reachable

- **`linear.py` stops asking for reverse-discovered matches.** `interpolate_addition()`'s wider family-set rescue passed `discover_own_reverse_rxns_in_reverse=True`, but the adapter branches on `discovered_in_reverse` without ever translating the index space. Measured on CH₄ + OH: every reverse-discovered `BREAK_BOND` pair is absent from the reactant graph and present in the product graph (they are the H₂O O–H bonds), while every consumer — `get_expected_changing_bonds()` here, and `map_rxn()`, which cuts the *reactants* with the map — reads it as reactant indices. `arc/mapping/` contains no reference to `discovered_in_reverse` at all. The kwarg is removed and the misleading docstring at `interpolate_addition()` corrected. `_strategy_ring_scission` is unaffected: the gate drops reverse matches only when a forward match exists, and a reverse-only reaction keeps all of them.
- **`make_bond_changes()` is guarded twice.** `KeyError` is added to `map_rxn()`'s except clause around it (which already caught four other types) — `r_label_dict[action[1]]` raises it when a recipe names a label the map lacks. Restricting product dicts to one family narrows this without closing it, because one family can match with different label sets (`intra_H_migration` on `[CH2]CCC` yields both a four-label and a three-label map). Separately, its charge-separating branch subtracted two radical electrons from an atom that may hold one and a lone pair from an atom that may hold none, yielding `u-1`/`c-1` atoms and, once such an atom reaches RDKit, `OverflowError: can't convert negative value to unsigned int`; that branch now declines the change. **This is a guard, not a reproduced fix** — in all 40 donor/acceptor combinations tried, the following `mol.update()` raised `AtomTypeError` and the molecule was restored from its copy, so the invalid state never survived the function. The guard is a no-op wherever the result is rejected.

## Notes for callers, and scope

- **A bare `get_product_dicts()` or `get_all_families()` is now sensitive to process-global settings state.** That is the fix, but it is a footgun for tests and scripts. Every test added here either patches the setting or passes `rmg_family_set=` explicitly.
- `get_all_families()` at the shipped `'default'` is byte-identical to `main`; `test_rmg_family_set_setting_ships_as_default` asserts the shipped constant. What is *not* unchanged is `'all'`, per §1.
- The two surface filters remain inconsistent (recommended path filters on *set label*, directory path on *directory name*) — pre-existing, not changed here. Also pre-existing and not changed: the `linear.py` wider-scan block restores `_family`/`_product_dicts`/`_atom_map` in a `finally` only after the whole weights loop, while `TSGuess(family=rxn.family, ...)` is written during it.
- No allowlist is applied to the directory union. RMG's curation inverts against TS-search utility (`Intra_RH_Add_Endocyclic`/`Exocyclic` and `Intra_R_Add_ExoTetCyclic` have 0 training reactions; `Substitution_O` has 136 and `SubstitutionS` 148), and a second hand-maintained list would drift from `ts_adapters_by_rmg_family`, which ARC already maintains. A hand-picked list would also have excluded `lone_electron_pair_bond`, a family ARC declares support for and ships a passing test for (`linear_test.py::test_interpolate_lone_electron_pair_bond`).

## Reuse check

Searched for an existing settings-resolution helper by behaviour ("argument-or-fall-back-to-`settings[key]`"), by name across `def resolve*`/`get_setting*`, and for the raw `or settings[...]` idiom at call sites — nothing of the kind exists, so the fallback stays inline in `get_all_families()`, where one already was. Nothing walked the RMG database's `kinetics/families` (`get_all_families` scanned `ARC_FAMILIES_PATH` for ARC's own families only), so `get_rmg_family_directories()` is new and sits beside `get_rmg_recommended_family_sets()`. `get_family_own_reverse()` follows the `ReactionFamily(label=...)` call form already used at two pre-existing sites in `reaction.py` and two in `arc/mapping/engine.py` rather than introducing a fifth; `ReactionFamily.__new__` interns per `(label, consider_arc_families)`, so nothing is re-read from disk. For the direction gate, `arc/mapping/` was searched for any existing reverse-aware label-map translation — there is none, which is the finding that decided it; no new helper was added there.

## Checks

- Full suite at parity with `origin/main`: the same 5 pre-existing `arc/job/adapters/torch_ani_test.py` failures on both, 0 added, 0 removed.
- Verified to fail against `origin/main`: `test_wider_family_set_scan_used_by_the_linear_ts_adapter`, `test_widening_keeps_the_recommended_family_when_both_match`, `test_bare_calls_honour_the_rmg_family_set_setting`, `test_determine_family_reaches_a_directory_only_family`.
- 16 tests in `TestFamilyChoiceGates`, each mutation-proved — removing either gate or the count tie-break fails a distinct test. `test_a_radical_cyclization_is_not_labeled_as_a_cycloaddition` exercises the real path end to end on pentadienyl cyclisation, where `Intra_2+2_cycloaddition_Cd` sorts ahead of `Intra_R_Add_Exocyclic` and the gate still resolves to the radical addition.
- Under `-n 6` the branch additionally trips `scheduler_test.py::test_initialize_output_dict`; that test is order-dependent and fails identically on `origin/main` in isolation — a pre-existing xdist worker-distribution artefact.

Supersedes `fix_family_set` (`916d9943`), which carries the same feature without the sink fix.

---

## Also folds in `fix_product_isomorphism_tautomers`

`check_product_isomorphism`'s InChI fallback compared standard InChI plus multiplicity — and standard InChI carries a mobile-H layer that deliberately collapses tautomers onto one string. So ARC accepted a family-generated product that is a **different molecule** from the one the user specified (`NC=O` accepted as `N=CO`). Observed on benchmark reaction 95, where the bad dict became `product_dicts[0]` and drove a TS guess, an atom map and an NMD check for a reaction that is not in the input.

The fallback now additionally requires `Molecule.is_isomorphic(save_order=True, strict=False)` — ARC's existing electron-agnostic comparison, which ignores bond orders, charges and radicals while keeping explicit hydrogens as graph vertices. Lewis-structure pairs like `O=C=C(O)C=O` / `O=C[C-](O)C#[O+]` still match; tautomers do not.

Because that check is a *necessary* condition it runs first, so InChIs are generated only for pairs it already admits — 30 generations and 2 failures per family resolution drop to 0 and 0 — and the new `_get_inchi` converts a deep copy and memoizes per pair, so one InChI-hostile molecule no longer rejects every candidate. Seven regression tests.

Folded here rather than shipped separately because it touches the same two files as this PR's first commit; apart, whichever landed second would have conflicted.

### DIFF
--- a/arc/family/family.py
+++ b/arc/family/family.py
@@ -56,6 +56,8 @@ REACTANTS_PATTERN = re.compile(r'reactants\s*=\s*\[(.*?)\]', re.DOTALL)
 PRODUCTS_PATTERN = re.compile(r'products\s*=\s*\[(.*?)\]', re.DOTALL)
 ACTIONS_PATTERN = re.compile(r'actions\s*=\s*\[(.*?)\]', re.DOTALL)
 
+RADICAL_RECIPE_ACTIONS = ('GAIN_RADICAL', 'LOSE_RADICAL')
+
 
 def get_reaction_family(label: str, consider_arc_families: bool = True) -> 'ReactionFamily':
     """
@@ -517,7 +519,13 @@ def get_reaction_family_products(rxn: ARCReaction,
                                                           RMG-database/input/kinetics/families/recommended.py.
                                                           Can be a name of a defined set, or a list
                                                           of explicit family labels to consider.
-                                                          Note that surface families are excluded if 'all' is used.
+                                                          Note that 'all' skips family sets whose label contains
+                                                          'surface' and surface family directories, but still
+                                                          includes an individual surface family that a set with a
+                                                          non-surface label lists (e.g. the electrochem set's
+                                                          Surface_Proton_Electron_Reduction_* families).
+                                                          ``None`` (the default) means ``settings['rmg_family_set']``,
+                                                          read on every call.
         consider_rmg_families (bool, optional): Whether to consider RMG's families.
         consider_arc_families (bool, optional): Whether to consider ARC's custom families.
         discover_own_reverse_rxns_in_reverse (bool, optional): Whether to discover reactions belonging to a family
@@ -526,10 +534,17 @@ def get_reaction_family_products(rxn: ARCReaction,
                                                                ``True`` will cause the function to discover reactions
                                                                in both directions, ``False`` will cause the function
                                                                to discover reactions only in the forward direction.
+                                                               Reverse-discovered matches that this flag admits are
+                                                               still dropped by prioritize_family_product_dicts()
+                                                               whenever a forward match exists, so the flag only
+                                                               widens the result for a reaction that no family
+                                                               matches in the forward direction.
 
     Returns:
-        list[dict]: The list of product dictionaries with the reaction family label.
-                    Keys are: 'family', 'group_labels', 'products', 'own_reverse', 'discovered_in_reverse', 'actions'.
+        list[dict]: The list of product dictionaries with the reaction family label, ordered and filtered
+                    by prioritize_family_product_dicts(), so that the first entry is the family match to use.
+                    Keys are: 'family', 'group_labels', 'products', 'r_label_map', 'p_label_map',
+                    'own_reverse', 'discovered_in_reverse'.
     """
     family_labels = get_all_families(rmg_family_set=rmg_family_set,
                                      consider_rmg_families=consider_rmg_families,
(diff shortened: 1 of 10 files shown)


kfir4444 (dismissed):
Great job @calvinp0! LGTM!

kfir4444 (approved):
Still LGTM! @calvinp0

PR #979, opened 2026-08-13
Make reaction family determination deterministic and honour a pinned family

calvinp0:
Stacked on #978 (makes `settings['rmg_family_set']` reachable and widens `'all'`). Followed by #982, which gates *which* family wins. Read #978 first — this PR reworks the ordering that #978's docstring establishes.

## What breaks today

- **The family ARC picks depends on `PYTHONHASHSEED`.** `get_all_families()` built its candidate list from the **set literals** in RMG-database's `recommended.py` (via `ast.literal_eval`) and from `os.listdir()`. Neither iteration order is specified.
  - `[CH]1C=Cc2ccccc21 <=> [CH]1C=CC2C3=C1C=CC32` (benchmark `reaction_018`) reports `Intra_Diels_alder_monocyclic` at seeds 0/1/13 and `Intra_R_Add_Exocyclic` at 2/5.
  - `[CH2]C(C=C)OS >> [O]C(C=C)CS` flips between `intra_OH_migration` and `intra_substitutionS_isomerization`.
  - About **10%** of the 500-reaction benchmark selection matches more than one family, so this is not a corner case.
- **Pinning a family was not authoritative.** Setting `family` recorded the label but left `product_dicts` holding matches from *every* family that matched, in that same unstable order, so a family recipe could be paired with another family's atom label map. With `intra_OH_migration` pinned and `PYTHONHASHSEED=2`, ARC reported `C0-O4` as a breaking bond — there is no `C0-O4` bond in the reactant.
- **`family_own_reverse` was effectively always `False`.** `__init__` set it to `False` and the property only recomputed on `None`, so the recompute branch was dead.

## What this changes

**Determinism**

- `recommended.py` is read with `ast.parse` instead of `ast.literal_eval`, so each family set keeps its labels in the order they are written rather than in a hash-seeded set order. A set whose elements are not all plain string literals is sorted instead, so the result never depends on the hash seed either way.
- `get_all_families()` sorts by label **within each tier** and de-duplicates keeping the first occurrence.
- `get_rmg_recommended_family_sets()` now returns `dict[str, tuple[str, ...]]` rather than an unordered set. Its only non-test caller is `get_all_families()`.

**Pinning**

- `product_dicts` is restricted to a **single** family — the pinned one, or the family of the first entry when none is pinned — so a recipe is only ever combined with labels its own family generated.
- A pinned family that matches none of the generated dicts raises `ReactionError` rather than silently deriving the TS from a foreign family or from nothing.
- Assigning `family` **restricts** product dicts that were already generated instead of discarding them, so the Linear TS adapter's wider-family-set retry — which sets `product_dicts` and then `family` — keeps what that retry recovered.
- A warning names every family a reaction matched and which one was used.

**Adjacent correctness**

- Assigning a family derives `family_own_reverse` from that family's `ownReverse` declaration instead of leaving it at `False`. `from_dict` no longer forces `False` when the key is absent, so a restart file without the key now derives the value rather than pinning it wrong.
- `make_bond_changes()`'s charge-separating branch declines a change that would produce a negative electron count (the donor must have a lone pair to give and the acceptor the two radical electrons the new bond consumes); the affected cut is left unchanged and the reason is logged.
- `KeyError` is added to `map_rxn()`'s except clause around `make_bond_changes()`, which already caught four other types — `r_label_dict[action[1]]` raises it when the recipe names a label the map does not carry.

## Why the ordering is *tiered* and not a flat sort

- **#978 imposes an ordering** on `get_all_families()`: recommended-set families first, RMG database directories appended last, so a family reachable through a recommended set always outranks one reachable only as a directory.
- **A flat `sorted()` over the whole list erases that tier.** Measured on this branch's head against `RMG-database` at `/home/calvin/code/RMG-database`, under `'all'`:

  | family | flat-sort index | tiered index |
  | --- | --- | --- |
  | `Intra_RH_Add_Endocyclic` (directory-only) | 50 | 86 |
  | `Intra_R_Add_Exocyclic` (recommended) | 55 | 51 |
  | `Substitution_O` (directory-only) | 77 | 94 |

  ASCII `H` < `_`, so a flat sort promotes directory-only families above recommended ones. The recommended tier ends at index 84 and the directory tier starts at 85.
- Merged independently, the two orderings would silently fight and whichever landed second would quietly undo the other. Stacked, they are reconciled by sorting **within** each tier:

  `sorted(recommended families) + sorted(database-directory families) + sorted(ARC families)`, de-duplicated keeping the first occurrence, so a family present in both tiers stays in the recommended tier.
- That is hash-seed-independent (every tier is sorted) *and* preserves "recommended beats directory-only".
- **Both properties are load-bearing**, confirmed by mutating the implementation: reversing the tiers, sorting across the tiers, or de-duplicating through a `set` each still makes #978's ordering tests fail (`test_directory_only_families_are_ordered_last`, `test_widening_keeps_the_recommended_family_when_both_match`); leaving a tier unsorted makes the determinism test fail while the ordering tests pass.

## Determinism evidence

md5 of `get_all_families()` at `PYTHONHASHSEED` 0, 1, 2, 3, 7, 13:

| | `'default'` | `'all'` | `reaction_018` family |
|---|---|---|---|
| base (#978) | 5 distinct lists across the seeds | 5 distinct lists across the seeds | flips `Intra_Diels_alder_monocyclic` ↔ `Intra_R_Add_Exocyclic` |
| this branch | 1 list, byte-identical across all 6 | 1 list, byte-identical across all 6 | `Intra_Diels_alder_monocyclic` at every seed |

Re-measured on the current head `9c3ca6aa` (rebased onto `origin/main` at `44a6b112`): `'default'` and `'all'` are each byte-identical across all six seeds. sha256 of `json.dumps([get_all_families('default'), get_all_families('all'), get_all_families(consider_arc_families=False), get_all_families(consider_rmg_families=False)])` is `40b41500bc38902029a43e4126574e70c3b637e29bad48e9aa711571a82c3b9e` at every one of `PYTHONHASHSEED` 0/1/2/3/7/13.

## Module-level import, and the CodeQL position

`get_family_own_reverse()` reads the family's `ownReverse` declaration through `ReactionFamily(label=family).own_reverse`, using the `ReactionFamily` that `reaction.py` **already** imports at module level on `main`. **This PR changes no import line at all.**

- **Reuse.** Searched for an existing accessor before adding one: `reaction.py` already constructs `ReactionFamily(label=...)` directly at two pre-existing sites (`get_expected_changing_bonds`, `get_number_of_atoms_in_reaction_zone`), and `arc/mapping/engine.py` at two more. This follows that established call form rather than introducing a fourth one.
- **The constructor *is* the memoisation.** `ReactionFamily.__new__` interns instances per `(label, consider_arc_families)` — pre-existing on `main` — so `ReactionFamily(label=...)` never re-reads `groups.py`. Measured: `ReactionFamily(label='H_Abstraction') is get_reaction_family(label='H_Abstraction')` → `True`, the same object. Cold construction 16.1 ms; warm 0.294 µs via the constructor, 0.076 µs via `get_reaction_family` — a 0.22 µs difference.
- **Nothing is re-read from disk.** By the time `get_family_own_reverse()` can run, family determination has already interned 54 `ReactionFamily` objects including the chosen one, so the cold 16.1 ms is paid once per process either way. `family_own_reverse` is additionally cached on the reaction (`_family_own_reverse`), so `get_family_own_reverse()` runs about once per reaction, at 0.310 µs.
- **CodeQL.** `reaction.py:7` already carries three `py/unsafe-cyclic-import` alerts on `main` (#1443/#1444/#1445) — one **per imported name**, all dismissed as TYPE_CHECKING false positives, which is accurate: `arc/family/family.py` imports `ARCReaction` only under `if TYPE_CHECKING:`. Because the alerts are per-name, adding a fourth name to that line would have raised a fourth, new, undismissed alert. Using the already-imported `ReactionFamily` raises none. No new import edge, no new alert.
- **Correctness of the value.** `own_reverse` matches the `ownReverse` declaration read straight from `groups.py` for **every** family in `get_all_families()` — 0 mismatches. `H_Abstraction` → `True`, `intra_OH_migration` → `False`. A family whose groups file is missing still raises `FileNotFoundError` on the first *and* every subsequent call, so the `except FileNotFoundError` fallback behaves exactly as before.

## Test parity

Re-run on the current head `9c3ca6aa`, against `origin/main` at `44a6b112`:

| | result |
|---|---|
| this branch (`9c3ca6aa`) | 5 failed, 2746 passed, 36 skipped |
| `origin/main` (`44a6b112`) | 5 failed, 2730 passed, 36 skipped |

The same 5 failures on both — all `arc/job/adapters/torch_ani_test.py`, an environment issue unrelated to this change. **No added and no removed failures.** (The branch passes 16 more tests than `origin/main` because #978 and this PR add that many; the branch is now level with `44a6b112`, 0 commits behind.)

Targeted: `arc/reaction/reaction_test.py`, `arc/family/family_test.py`, `arc/mapping/engine_test.py`, `arc/mapping/driver_test.py` — 195 passed.

## On the tie-break

- This branch makes the family order **reproducible**; it does not claim the order is chemically right.
- A label sort is **not chemically neutral**: RMG's naming puts concerted/pericyclic families early in the alphabet and radical-addition families late, so a radical reaction that also matches a concerted template loses to it by default.
- That is addressed in **#982**, stacked directly on this PR, which gates the choice on reaction direction and on whether the recipe accounts for the radical before the label sort is consulted.
- The two changes are deliberately separate. Deterministic-but-ungated is a strict improvement over an order that varies with `PYTHONHASHSEED`, and it is reviewable and revertable on its own terms; #982 needs this PR, but this PR does not need #982.
- A reaction that needs a specific family can pin it, which this branch also makes authoritative.


calvinp0:
Flattened into #978. `feature_configurable_rmg_family_set` now carries this stack's full content, rebased onto main (0 behind) and restructured so no file is touched by more than one commit.

The stack was costing more than it bought: three PRs, three commits, all touching `arc/family/family.py` and `arc/reaction/reaction.py`, so none could be reviewed independently anyway. Content identity was verified across the rebase — the changed-line sets of the old tip and the new head are byte-identical in both directions, so nothing was dropped. `restrict_product_dicts_to_family()`, the pinned-family honouring and `get_family_own_reverse()` are all in #978's second commit.

PR #981, opened 2026-08-13
Stop YAML from coercing a bare SMILES like NO into a boolean

alongd:
## Motivation

PyYAML implements YAML 1.1, in which `yes`, `no`, `on`, and `off` (in any case) are booleans just like `true` and `false`. So an ARC input file containing

```yaml
species:
  - label: hydroxylamine
    smiles: NO
```

hands ARC the Python value `False`, never the string `NO`. `Molecule(smiles=False)` does not raise — it silently returns an *empty* molecule. `get_xyz()` then sees `self.mol is not None` and calls `get_cheap_conformer()`, which calls `is_diatomic()`, which found `len(self.mol.atoms) == 0` and fell through to `self.get_xyz()` with the default `generate=True`. Infinite recursion, ending in a bare `RecursionError` that names no species and points at nothing the user wrote.

The same coercion has already corrupted ARC's own data. `data/elements.yml` maps `102: No` — nobelium. On `main` today:

```python
>>> SYMBOL_BY_NUMBER[102]
False
>>> NUMBER_BY_SYMBOL['No']
KeyError: 'No'
```

`MASS_BY_SYMBOL` likewise carries a `False` key instead of `'No'`.

## What this does

1. `ARCYAMLLoader` in `arc/common.py` narrows the implicit boolean resolver to the YAML 1.2 spellings `true`/`false`. Every other alias loads as the string the user wrote. Both YAML entry points (`read_yaml_file` and `from_yaml`) use it, so files and strings behave identically. This fixes the nobelium entry with no edit to `elements.yml`.
2. Belt and suspenders, so bad input fails loudly rather than mysteriously: `check_smiles()` raises a `SpeciesError` naming the species when a SMILES is not a string (it subsumes the previous list-only check in `from_dict`, and now runs on the `__init__` path too), and `is_diatomic()` consults `mol_list` and calls `get_xyz(generate=False)` — mirroring `is_monoatomic()` — so it can never re-enter conformer generation.

Writing is untouched. `save_yaml_file` / `to_yaml` already emit `true`/`false`, so every ARC-written `restart.yml` / `output.yml` round-trips unchanged.

## Testing

`test_read_yaml_file_does_not_coerce_yaml_1_1_booleans` (round-trips `NO`, `ON`, `yes`, `Off` as strings while `true`/`False`/`TRUE` stay bools, via both entry points), `test_read_element_dicts_nobelium`, `test_non_str_smiles_raises_named_error`, `test_is_diatomic_does_not_generate_conformers`. All four were written failing first.

End-to-end, under `sys.setrecursionlimit(300)`:

```
parsed: {'project': 'yaml_bool_repro', 'species': [{'label': 'hydroxylamine', 'smiles': 'NO'}]}
label: hydroxylamine | mol: NO | formula: H3NO | mult: 1
is_diatomic: False
```

Full local suite: 2742 passed, 7 skipped. The single failure, `functional/restart_test.py::TestRestart::test_globalize_paths`, is pre-existing and environmental — it asserts the substring `ARC/arc/testing/...` against an absolute path, so it fails in any worktree directory not named exactly `ARC`. Verified by stashing this diff and re-running it on a clean tree.

## Notes for reviewers

- One behavior change worth naming: an existing input file written as `keep_files: yes` now yields the string `'yes'` rather than `True`. Still truthy at essentially every call site, but not identical. Nothing in this repo relies on `yes`/`no` as booleans — the docs consistently use `true`/`false` — so the exposure is limited to user files in the wild.
- Not addressed here: the standalone job-adapter scripts under `arc/job/adapters/scripts/` load YAML themselves with `FullLoader`. They run detached on remote servers and cannot import `arc.common`, so covering them means duplicating the loader in seven files. Happy to open a follow-up issue.

### DIFF
--- a/arc/common.py
+++ b/arc/common.py
@@ -353,6 +353,26 @@ def get_git_branch(path: str | None = None) -> str:
         return ''
 
 
+class ARCYAMLLoader(yaml.FullLoader):
+    """A YAML loader that does not coerce YAML 1.1 boolean aliases into booleans.
+
+    PyYAML implements YAML 1.1, in which ``yes``, ``no``, ``on``, and ``off`` (in any case) are
+    booleans just like ``true`` and ``false``. That silently turns a bare SMILES such as ``NO``
+    (hydroxylamine) or an element symbol such as ``No`` (nobelium) into ``False`` before ARC ever
+    sees it. Here only the YAML 1.2 spellings ``true`` and ``false`` resolve as booleans, and
+    every other alias loads as the string the user wrote.
+    """
+
+
+ARCYAMLLoader.yaml_implicit_resolvers = {
+    first_char: [(tag, regexp) for tag, regexp in resolvers if tag != 'tag:yaml.org,2002:bool']
+    for first_char, resolvers in yaml.FullLoader.yaml_implicit_resolvers.items()
+}
+ARCYAMLLoader.add_implicit_resolver('tag:yaml.org,2002:bool',
+                                    re.compile(r'^(?:true|True|TRUE|false|False|FALSE)$'),
+                                    list('tTfF'))
+
+
 def read_yaml_file(path: str,
                    project_directory: str | None = None,
                    ) -> dict | list:
@@ -374,7 +394,7 @@ def read_yaml_file(path: str,
     if not os.path.isfile(path):
         raise InputError(f'Could not find the YAML file {path}')
     with open(path, 'r') as f:
-        content = yaml.load(stream=f, Loader=yaml.FullLoader)
+        content = yaml.load(stream=f, Loader=ARCYAMLLoader)
     return content
 
 
@@ -405,7 +425,7 @@ def from_yaml(yaml_str: str) -> dict | list:
     Returns: dict | list
         The respective Python object.
     """
-    return yaml.load(stream=yaml_str, Loader=yaml.FullLoader)
+    return yaml.load(stream=yaml_str, Loader=ARCYAMLLoader)
 
 
 def to_yaml(py_content: list | dict) -> str:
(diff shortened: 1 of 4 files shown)


arc/common.py
```diff
@@ -353,6 +353,26 @@ def get_git_branch(path: str | None = None) -> str:
...
+    sees it. Here only the YAML 1.2 spellings ``true`` and ``false`` resolve as booleans, and
+    every other alias loads as the string the user wrote.
+    """
+
+
+ARCYAMLLoader.yaml_implicit_resolvers = {
+    first_char: [(tag, regexp) for tag, regexp in resolvers if tag != 'tag:yaml.org,2002:bool']
+    for first_char, resolvers in yaml.FullLoader.yaml_implicit_resolvers.items()
+}
+ARCYAMLLoader.add_implicit_resolver('tag:yaml.org,2002:bool',
```
calvinp0, line 371: To make aware, this now applies to the whole file - the strop treating no as false, just give me the word.

And that python treats any non-empty word as true. So the word 'no' counts as true.

So, if by chance someone does

`compute_thermo: yes` then `"yes"` is true. but if `compute_thermo: no`, then `"no"` is also true.

I get that this probably impacts nobody since no one should be doing this and it is not documented in our docs.

alongd: I agree. I still think that the issue of "I wrote `NO` for SMILES and got an error" is more profound than "I wrote `no` for rotors and got rotors computed". We can over-engineer this and consider nuances, or evaluate as `bool` if it's not in the SMILES entry. I think it's OK as is since all of our examples are `true` and `false`. If it becomes an issue we'll address

calvinp0 (approved):

PR #985, opened 2026-08-15
Make queue-executed ASE (UMA/fairchem) jobs runnable on a cluster

alongd:
## What

Make queue-executed `ASEAdapter` jobs actually run on a remote cluster. The ASE adapter had only ever been exercised **incore**; its queue path was latent and broke on the first real submission. This lets ASE-backed calculators (e.g. UMA / fairchem MLIPs) run hindered-rotor and other scans as ordinary PBS/Slurm jobs.

## Why

Running a real ARC job that delegates its hindered-rotor scans to an ASE calculator on a cluster queue crashed the moment the first scan job was dispatched:

```
arc.exceptions.InputError: Cannot upload a non-existing file.
Check why file in path .../directed_scan_a2781/submit.sh is missing.
```

Two independent defects in the queue path, both masked because ASE had only ever run incore:

1. **Files were never written.** `JobAdapter.execute()` calls `upload_files()` *before* `execute_queue()`, and `_initialize_adapter()` calls `set_files()` while the job is constructed. The Gaussian/Orca/xTB adapters write `submit.sh` / input inside `set_files()`; `ASEAdapter.set_files()` only appended their names to `files_to_upload` and never wrote them, so the calcs dir was empty when `ssh.upload_file()` ran its `os.path.isfile` pre-flight.
2. **Submission was dead code.** `ASEAdapter.execute_queue()` guarded on `self.server_adapter`, an attribute nothing in ARC sets (`hasattr` → `False`), so even with the files present the job would never submit — the write fix alone just moves the crash one line down to an `AttributeError`.

## Changes (confined to the ASE adapter)

- `set_files()` now writes `submit.sh` and `input.yml` for non-incore jobs, mirroring the other adapters. **Incore is unchanged** — it still writes its input in `execute_incore()`.
- `execute_queue()` now delegates to the shared `legacy_queue_execution()` that every other adapter uses (which also records `job_status` / `job_id`), replacing the bespoke never-reached block.
- `write_submit_script()`'s non-queue branch falls back to `local_path` when `remote_path` is `None` (a server-less job), fixing a latent crash surfaced by writing at construction time.
- Adds `test_set_files_writes_the_files_of_a_queue_job` (red against the pre-fix adapter).

## Verification

- New/updated unit tests: `pytest arc/job/adapters/ase_test.py -q -n0` → **10 passed** (the file requires serial `-n0`, a pre-existing xdist `setUpClass` race unrelated to this change).
- Incore path unregressed: full `execute()` → `execute_incore()` → `ase_script.py` → `parse_results()` reproduces the prior geometry/energy to 5 decimals.
- **Live cluster run:** an ARC job (DFT opt/freq/sp on the queue, hindered-rotor scans via a UMA/ASE calculator) now runs end-to-end — the DFT stage completes and the rotor jobs are written with their `submit.sh` + `input.yml` and submitted to the queue, where the pre-fix run crashed.

The diff is limited to `arc/job/adapters/ase_adapter.py`, `arc/job/adapters/ase_test.py`, and a 1-indexed→0-indexed constraint fix in `arc/job/adapters/scripts/ase_script.py` (ASE `FixInternals` is 0-indexed; ARC constraints are 1-indexed). No other adapter's submit path is touched.


alongd:
Added a third commit (`dc23057`): fix a stale directed-scan parser call site — `check_directed_scan_job` called `parser.parse_e_elect(path=...)` but the refactored API is `log_file_path=` (as every other call site uses). Without it, ARC crashes assembling V(φ) the moment a directed-scan job's energy is parsed. Verified by parsing a real UMA rotor output (−500058.16 kJ/mol). This is needed for ASE/UMA rotor scans to run end-to-end.

arc/job/adapters/ase_adapter.py
```diff
@@ -277,13 +298,114 @@ def set_input_file_memory(self) -> None:
...
+            directives += [f'#PBS -l {select}', '#PBS -o out.txt', '#PBS -e err.txt']
+        else:
+            directives = [f'#SBATCH -J {self.job_server_name}', '#SBATCH -N 1',
+                          f'#SBATCH -n {self.cpu_cores}', f'#SBATCH --mem-per-cpu={memory}',
+                          '#SBATCH -o out.txt', '#SBATCH -e err.txt']
+            if config['queue']:
+                directives.insert(1, f"#SBATCH -p {config['queue']}")
+            if config['gpu_resource']:
+                directives.append(f"#SBATCH --gres={config['gpu_resource']}")
+        return '\n'.join(['#!/bin/bash -l', ''] + directives + [
```
alongd: this script might be cluster-software-specific. best to put it in settings/submit.py like all other submit scripts

arc/job/adapters/ase_adapter.py
```diff
@@ -277,14 +303,119 @@ def set_input_file_memory(self) -> None:
...
+        - ``python``: the python executable **on the server**. ``self.python_executable`` is
+          resolved against the ARC host's conda envs and generally does not exist on a remote server.
+        - ``queue``: the queue to submit to, if not already set on the job or in the server settings.
+
+        Returns:
+            dict: The resolved submit configuration.
+        """
+        block = (self.args or dict()).get('block', dict()) or dict()
+        default_queue, _ = next(iter(servers.get(self.server, dict()).get('queues', dict()).items()), (None, None))
+        return {'queue': self.queue or block.get('queue') or default_queue,
```
calvinp0, line 357: I would like also an update in the docs so a user is aware of being able to write in the input.yml, as an example:

```yaml
level:
  method: uma
  args:
    block:
      python: /home/User/.conda/envs/UMA_env/bin/python
      env_setup: module load CUDA/12.1; conda activate UMA_env
      queue: GPU_Long
```

One concern is the lower case situation. The lowercasing hits all four keys, not just env_setup. I checked: module load CUDA/12.1; conda activate UMA_env  $\rightarrow$  module load cuda/12.1; conda activate uma_env, /home/User/.../UMA_env/bin/python $\rightarrow$ /home/user/.../uma_env/bin/python, GPU_Long $\rightarrow$ gpu_long.

So I think we need to find a way around this? I am not sure how though, maybe something with the settings.py?

calvinp0 (changes requested):

arc/job/adapters/ase_adapter.py
alongd: Both addressed in `35e609158`:

- **Case corruption:** you were right — `Level.lower()` lowercased *every* block value, mangling paths, module names, conda envs, and queue names. Fixed at its source in `arc/level.py`: added a `CASE_SENSITIVE_BLOCK_ARGS = ('python', 'env_setup', 'queue', 'gpu_resource')` tuple, and `lower()` now keeps the original case for those four `block` keys while everything else (QC keywords, other block args) is lowercased as before. Covered by `test_lower_preserves_case_sensitive_block_args`.
- **Docs:** added a "Running ASE / UMA jobs on a cluster" section to `docs/source/advanced.rst` documenting the `block` execution knobs (`python`, `env_setup`, `queue`, `gpu_resource`) with your `input.yml` example, and noting these four keep their case.

arc/imports.py
```diff
@@ -134,6 +135,10 @@
         from submit import pipe_submit as local_pipe_submit
     except ImportError as e:
         _report_unusable_overlay(local_arc_submit_path, 'submit', e, 'pipe_submit')
+    try:
+        from submit import ase_submit as local_ase_submit
+    except ImportError:
```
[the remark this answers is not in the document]
alongd: Fixed in `454352dff`. The `ase_submit` overlay was the only one of the four `submit.py` overlays swallowing `ImportError` with a bare `pass`; it now calls `_report_unusable_overlay(...)` like its siblings (`incore_commands`, `pipe_submit`, `submit_scripts`). So a `~/.arc/submit.py` that fails to import is reported to the user, while one that merely predates the `ase_submit` key stays quiet (handled inside `_report_unusable_overlay`).

arc/job/adapters/ase_adapter.py
alongd: Quick follow-up on your "maybe something with settings.py?" — that is essentially the fix: `level.py` now holds an allow-list constant `CASE_SENSITIVE_BLOCK_ARGS = ('python', 'env_setup', 'queue', 'gpu_resource')`, and `Level.lower()` skips lowercasing those four `block` keys while everything else lowercases as before. So your example now survives verbatim — `CUDA/12.1`, `UMA_env`, `GPU_Long`, and the full python path all keep their case.

alongd:
Rebased onto current `main` (was BEHIND) and re-verified. Everything raised in review is addressed at the tip; posting the mapping so a re-review does not have to reconstruct it.

**@calvinp0 — both of your points:**

- **Case corruption.** You were right that the lowercasing hit all four `block` keys, not just `env_setup`. Fixed at source in `arc/level.py`: `CASE_SENSITIVE_BLOCK_ARGS = ('python', 'env_setup', 'queue', 'gpu_resource')`, and `Level.lower()` now preserves those four while lowercasing everything else as before. Your example survives verbatim — `CUDA/12.1`, `UMA_env`, `GPU_Long` and the full interpreter path all keep their case. Covered by `test_lower_preserves_case_sensitive_block_args`.
- **Docs.** Added *"Running ASE / UMA jobs on a cluster"* to `docs/source/advanced.rst` (§ at line 388), built around the `block` example you wrote.

**Copilot's four:**

- *Slurm would upload `submit.sh` while ARC submits `submit.sl`* → `determine_submit_filename()` keys off `submit_filenames[cluster_soft]` for queue jobs and keeps the plain name for incore.
- *Resolved queue never added to `attempted_queues`, so `trsh_job_queue()` can retry the same queue* → recorded in `write_submit_script()` before returning, as `JobAdapter` does.
- *Template always formatted with `remote_path`, so a `local` queue job `cd`s into the wrong directory* → `pwd` is `local_path` when the server is `None` or `local`.
- *PBS template omitted walltime* → `#PBS -l walltime={t_max}` is in the template, with the value formatted through `t_max_format` per cluster software.

**CodeQL's two:** implicit string concatenation in the submit list, and the bare `except: pass` in `arc/imports.py` — the `ase_submit` overlay was the only one of the four swallowing `ImportError` silently and now calls `_report_unusable_overlay()` like its siblings.

**Tests:** `arc/job/adapters/ase_test.py` 18 passed, `arc/level_test.py` 13 passed. Three `level_test` failures reproduce identically on unmodified `main` in this environment and are unrelated to this branch.

One note for whoever reviews: a separate defect in this same area is up as #1054 (`rotor_top()` inherits ASE's 0.3 Å neighbour-list `skin`, fuses 1-3 pairs into a spurious ring, and rejects every rotor of a branched molecule). It is deliberately **not** folded in here — different file, different cause, and this PR is large enough. Both are needed before an ASE rotor scan actually completes on a cluster; verified end to end on zeus today with the two together.

arc/job/adapters/ase_adapter.py
```diff
@@ -308,14 +334,119 @@ def set_input_file_memory(self) -> None:
...
+        path that only exists on the ARC host. See ``determine_submit_config()`` for the knobs.
         """
-        remote_script_path = os.path.join(self.remote_path, 'ase_script.py')
-        command = f"{self.python_executable} {remote_script_path} --yml_path {self.remote_path}"
-        content = f"#!/bin/bash\n\n{command}\n"
-        with open(os.path.join(self.local_path, 'submit.sh'), 'w') as f:
+        config = self.determine_submit_config()
+        cluster_soft = servers.get(self.server, dict()).get('cluster_soft', '').lower() \
+            if self.server is not None else ''
+        queue_job = self.execution_type != 'incore' and cluster_soft in ('pbs', 'slurm')
```
calvinp0: should we also deal with htcondor and oge?

arc/job/adapters/ase_test.py
```diff
@@ -123,6 +124,71 @@ def test_set_files(self):
...
+            self.assertTrue(os.path.isfile(file_dict['local']), msg=f"missing {file_dict['file_name']}")
+        with open(submit_path, 'r') as f:
+            content = f.read()
+        self.assertIn('#PBS -q test_q', content)
+        self.assertIn('conda activate uma_env', content)
+        self.assertIn('/remote/python', content)
+
+    def test_set_files_does_not_write_for_an_incore_job(self):
+        """Test that an incore job writes no submit script (it writes its input when it executes)"""
+        self.assertFalse(os.path.isfile(os.path.join(self.job_1.local_path, 'submit.sh')))
```
calvinp0: the assertion can't fail; setUpClass reassigns local_path after construction, so it checks a directory set_files() never wrote to.

arc/job/adapters/ase_adapter.py
```diff
@@ -308,14 +334,119 @@ def set_input_file_memory(self) -> None:
...
+        ``submit_job()`` invokes (``submit_filenames``, e.g. ``submit.sl`` for Slurm), or the
+        submission fails because the file it names is not on disk. Everything else uses the plain
+        ``submit.sh`` the bare script is written to.
+
+        Returns:
+            str: The submit-script filename.
+        """
+        cluster_soft = servers.get(self.server, dict()).get('cluster_soft', '') if self.server is not None else ''
+        if self.execution_type != 'incore' and cluster_soft.lower() in ('pbs', 'slurm'):
+            return submit_filenames[cluster_soft]
```
calvinp0: we do a lower case guard but then the lookup isn't lowercase

calvinp0 (commented):
just 3 minor comments

alongd:
Worked every open finding on this PR. Pushed two commits (fast-forward, no rebase, no force-push): `fd8e795c0` and `5647e0f8e`.

Short version: **one of the nine was still a live defect** (finding 1). Five had already been fixed on the branch but carried no test, so they now have one. Two were answered. One was already resolved.

---

### 1. `ase_adapter.py:377` — "we do a lower case guard but then the lookup isn't lowercase" (@calvinp0)

**Real, and it raised.** Reproduced against the branch tip: a server whose `cluster_soft` is anything but the exact settings spelling passes `cluster_soft.lower() in ('pbs', 'slurm')` and then dies inside the constructor —

```
cluster_soft='PBS'      adapter->'submit.sh'   submit_job()->'submit.sh'
cluster_soft='Slurm'    adapter->'submit.sl'   submit_job()->'submit.sl'
cluster_soft='slurm'    RAISED KeyError: 'slurm'
cluster_soft='pbs'      RAISED KeyError: 'pbs'
cluster_soft='SLURM'    RAISED KeyError: 'SLURM'
```

**Which way I resolved it, and why: the lookup is the correct half; the guard was wrong.** Lowercasing the lookup would break *every* server, because `submit_filenames` has no lowercase keys at all — it is `{'OGE', 'Slurm', 'PBS', 'HTCondor'}`. More importantly, this adapter is not free to pick its own convention: the filename it writes has to be the one the submission actually invokes, and `local.submit_job()` (`arc/job/local.py:246`) and `ssh.submit_job()` (`arc/job/ssh.py:475`) both resolve that as `submit_filenames[cluster_soft]`, verbatim. A case-insensitive resolution here would cheerfully name `submit.sl` on a server where `submit_command['slurm']` then KeyErrors anyway — a file named for a submission that cannot happen.

So the guard is now the lookup's own precondition, hoisted into a named constant so the two cannot drift apart again:

```python
# Cluster schedulers this adapter carries an ``ase_submit`` template for, spelled exactly as the
# keys of ``submit_filenames``/``submit_command`` in arc/settings/settings.py. ...
QUEUE_CLUSTER_SOFT = ('PBS', 'Slurm')
```

`write_submit_script()` uses the same constant, so the filename decision and the template decision can no longer disagree. The lower-cased key survives in exactly one place — `ase_submit`, which is ARC's own dict and *is* keyed in lower case, like `pipe_submit`. That split is now stated in a comment at the call site rather than being something you have to infer.

Worth noting this exact guard/lookup mismatch also exists upstream of the adapter — `ssh.py:442` validates `cluster_soft.lower() not in [...]` and `ssh.py:475` looks up verbatim — so a mis-cased `cluster_soft` was never going to run. It now fails where it is diagnosable instead of mid-construction. Not fixing that one here; it is not this branch's.

Covered by `test_submit_filename_is_the_one_the_submission_path_invokes` and `test_submit_filename_guard_admits_only_what_the_lookup_can_serve`.

### 2. `ase_test.py:162` — "the assertion can't fail" (@calvinp0)

**Real.** `setUpClass` repoints `job_1.local_path` at a flat scratch directory after construction, so `assertFalse(os.path.isfile(join(job_1.local_path, 'submit.sh')))` checked a directory `set_files()` had never written to.

I watched it fail. One-line deliberate break in `set_files()` — `if self.execution_type != 'incore':` → `if True:`, so an incore job *does* write a submit script — then the flagged assertion in isolation:

```
constructed local_path (what set_files() wrote to): .../test_ASEAdapter/test_1/calcs/Species/H2O/sp_a1
reassigned  local_path (what the assertion checks): .../test_ASEAdapter/test_1
submit.sh actually on disk in constructed path     : True

ORIGINAL assertion  assertFalse(isfile(join(job_1.local_path,'submit.sh')))  -> isfile=False -> PASSES (cannot fail)
REPAIRED assertion  assertEqual(submit.* in constructed path, [])           -> ['submit.sh'] -> FAILS
```

and through pytest, the repaired test against the same broken code:

```
>       self.assertEqual([entry for entry in os.listdir(self.job_1_constructed_local_path)
                          if entry.startswith('submit.')], list())
E       AssertionError: Lists differ: ['submit.sh'] != []
```

The break is reverted; it is not in either commit. `setUpClass` now keeps the constructed paths (`job_1_constructed_local_path`), and the assertion checks for *any* `submit.*`, so it does not go blind again if the scheduler-specific name changes.

### 3. `ase_adapter.py:429` — "should we also deal with htcondor and oge?" (@calvinp0)

**Not a defect — and my answer is "not on this branch", with reasons.**

What the adapter supports today: PBS and Slurm, the two keys in `ase_submit`. Anything else reaching `get_queue_submit_script()` raises `NotImplementedError` naming the available templates.

What ARC's other adapters do: nothing scheduler-generic at all. Gaussian, Orca, QChem, Molpro, TeraChem, CFour, xTB and the TS adapters all read `submit_scripts[server][software]` — a **per-server, per-software** dict a user fills in for their own machine. By that standard `ase_submit` already covers more schedulers than any ESS adapter in ARC. The only real precedent for a server-independent scheduler-keyed dict is `pipe_submit` (`slurm`, `pbs`, `sge`, `htcondor`, `local`), and it is instructive about why the remaining two are not mechanical:

- **HTCondor is not a shell script.** `submit_filenames['HTCondor']` is `submit.sub`, and `pipe_submit['htcondor']` is a condor submit *description* file (`executable = …`, `queue N`), not bash. Everything the ASE template does lives in shell: `cd "{pwd}"`, `JOB_DIR="$(pwd)"`, the `OMP_NUM_THREADS`/`MKL`/`OPENBLAS` pinning, `{env_setup}` (`module load CUDA/12.1; conda activate uma_env`), and `touch initial_time` / `touch final_time` — which `JobAdapter.determine_run_time()` reads back off the server (`arc/job/adapter.py:609-627`) to time the job. Under HTCondor all of that has to move into a separate wrapper that `executable =` points at. That means a second file in `set_files()`, a second name out of `determine_submit_filename()`, and `request_gpus` in place of the `gpu_select`/`--gres` mapping. That is an adapter design change, not a template.
- **OGE/SGE is a naming question I should not settle unilaterally.** `submit_filenames`/`submit_command`/`check_status_command` spell it `OGE`; `pipe_submit` spells it `sge` and `pipe_run.py:261` carries an explicit bridge, `template_key = 'sge' if self.cluster_software == 'oge' else self.cluster_software`. `submit_filenames` has no `SGE` key at all, so an SGE server already KeyErrors in `submit_job()` before ASE is reached. Picking a spelling for `ase_submit` is a settings-wide decision.
- **Neither can be verified from here.** There is no HTCondor or OGE server in `settings.py` or in the fixtures, and I am not submitting to a cluster on this branch. A template that has never run under its scheduler is worse than an explicit refusal.

What I did do, since a silent wrong answer was the actual hazard: a queue job on a scheduler with no `ase_submit` template used to write a bare directive-less script and say nothing. It now warns, and `determine_submit_filename()` and `write_submit_script()` agree on that path. Test: `test_write_submit_script_warns_on_a_scheduler_with_no_template`.

If you want HTCondor, the work is the two-file wrapper design above; say the word and it is a separate PR with a `request_gpus` mapping and a `submit.sub` + wrapper pair.

### 4. `ase_adapter.py:284` / `:414` — "Slurm jobs never upload their submit script"

**Was real; already fixed on this branch, and now has a test.** Verified rather than assumed — on the current tip a Slurm queue job gives

```
files_to_upload: ['submit.sl', 'input.yml', 'ase_script.py']
submit_filenames['Slurm'] = submit.sl -> present: True
on disk: ['input.yml', 'submit.sl']
```

`determine_submit_filename()` is what fixed it. It had no test, which is how finding 1 hid inside it; `test_submit_filename_is_the_one_the_submission_path_invokes` now pins the filename, its presence on disk, and its presence in `files_to_upload`, for PBS and Slurm.

### 5. `ase_adapter.py:354` — "local queue submissions cd into the wrong directory"

**Was real; already fixed (`pwd = self.local_path if … 'local' else self.remote_path`), now has a test.** Verified on a `local` server with a queue execution type:

```
local_path : .../calcs/Species/H2O/opt_a1
remote_path: runs/ARC_Projects/f5/H2O/opt_a1
script cd  : ['cd ".../calcs/Species/H2O/opt_a1"']
```

`test_write_submit_script_enters_the_local_path_of_a_local_server` asserts the `cd` target *and* that the remote path appears nowhere in the script.

### 6. `ase_adapter.py:361` — "the resolved queue is never recorded in `attempted_queues`"

**Was real; already fixed, now has a test.** Both routes verified: the server-default queue and `args['block']['queue']` each land in `attempted_queues` and in the `#PBS -q` directive, so `trsh_job_queue()` can move a failed submission on. `test_write_submit_script_records_the_resolved_queue`.

### 7. `submit.py:165` — "the PBS template omits walltime entirely"

**Was real; already fixed, now has a test.** `#PBS -l walltime={t_max}` is in the template and `t_max` is formatted per `t_max_format`. `test_write_submit_script_requests_the_job_walltime` checks both schedulers, including that Slurm gets its days format:

```
max_job_time=12.0, PBS   -> #PBS -l walltime=12:00:00
max_job_time=48.0, Slurm -> #SBATCH -t 2-0:00:00
```

### 8. CodeQL `ase_adapter.py:366` — implicit string concatenation in a list

**Already resolved, twice over.** Alert 1953 was dismissed as *won't fix* by @alongd on 2026-08-15, and the construct no longer exists — moving the template into `settings/submit.py` removed the list it lived in. Confirmed by walking the module's AST for adjacent string constants inside any list/tuple/set literal: `implicit concat inside list/tuple/set literals: NONE`. Nothing to change.

### 9. `ase_adapter.py:357` — "best to put it in `settings/submit.py` like all other submit scripts" (@alongd)

**Confirmed done from the code, not from the thread.** `arc/settings/submit.py:134` defines `ase_submit = {'slurm': …, 'pbs': …}` as server-independent templates; `arc/imports.py:15` imports it and lines 138-151 give it the same `~/.arc/submit.py` overlay machinery as `incore_commands`/`pipe_submit`/`submit_scripts`, reporting an unusable overlay through `_report_unusable_overlay` rather than a bare `pass` (which was the separate CodeQL finding on `imports.py:140`). `get_queue_submit_script()` is down to computing placeholder values and `.format()`-ing the template. The two companion halves are there too: `CASE_SENSITIVE_BLOCK_ARGS` at `arc/level.py:20`, and the "Running ASE / UMA jobs on a cluster" section at `docs/source/advanced.rst:388`.

---

### Tests

`conda activate arc_env`, `pytest -n0`, counts before → after:

| module | before | after |
|---|---|---|
| `arc/job/adapters/ase_test.py` | 21 | **27** (+7 subtests) |
| `arc/level_test.py` | 17 | 17 |
| `arc/job/adapter_test.py` | 42 | 42 |
| `arc/imports_test.py` | 11 | 11 |
| **total** | **91** | **97** |

Wider sweep over `arc/job/` + `arc/settings/` + `arc/level_test.py` + `arc/imports_test.py`: **1515 passed, 5 skipped, 2 failed**. Both failures are `pyscf_test.py::test_run_freq_*`, from PySCF not being installed in this environment; I re-ran them on the unmodified tip and they fail identically there, so they are not from this work.

### One thing left for you

The branch is `BEHIND` `main` (`mergeStateStatus: BEHIND`) and needs a rebase before it can merge. I deliberately did not do it — rebasing and force-pushing a branch that is under review is yours to call, not mine.

PR #990, opened 2026-08-16
Tests: stop xdist workers from deleting each other's ARC project directories

calvinp0:
`pyproject.toml` runs the suite with `-n auto --dist=worksteal` and CI uses `-n 6`. worksteal distributes *individual tests*, so one class's tests land on several worker processes and **every one of those workers runs that class's `setUpClass` and `tearDownClass`.** Any class with a fixed project directory plus a fixture that deletes it is therefore exposed: a worker starting or finishing its slice rmtrees a tree another worker is still testing inside.

It does not take two modules sharing a directory name — one class is enough. That is why the fix is broader than the collisions that are visible by grep.

**Reproduced on `main`**, not inferred: `FileNotFoundError: .../Projects/unit_test_specific_job/arc.log`. `TestARC` rmtrees six project directories in `setUpClass` *and* `tearDownClass`, and `ARC.__init__` creates the directory and then opens `arc.log` inside it, so another worker's fixture lands in that window.

### Three commits, no file touched by more than one

- **`b45ce92a`** adds `get_test_project_name` / `get_test_project_directory` to `arc/common.py` — the module that already owns `ARC_PATH` and `ARC_TESTING_PATH` — and applies them in `ts_test`, `nmd_test`, `plotter_test`, `scheduler_test`, `species_test` and `functional/restart_test`, whose private copy of the same logic is folded into the helpers rather than left as a second implementation.
- **`46d1effe`** covers `main_test` and `scale_test`. `ARC` derives its project directory from the project *name* when none is given, so the names are scoped rather than the paths.
- **`d643fb21`** covers the adapter modules, whose scratch directories live under `arc/testing/` rather than `ARC_PATH/Projects`. `common_test.py` and `gaussian_test.py` had additionally both chosen `test_GaussianAdapter`, which worker scoping alone cannot separate, so `common_test.py`'s two directories take a `_common` suffix.

### Measurements

`taskset -c 0-3 python -m pytest <target> -n 6 --dist worksteal`; per-commit numbers are in the commit messages.

| target | before | after |
|---|---|---|
| all 14 affected modules | red 4/4 runs (333–335 of 339 passing) | **339 passed, 5/5 runs** |
| six adapter modules | red 6/6 runs (58–71 passed, with errors) | **74 passed, 6/6 runs** |
| `main_test.py` | red 2/14 runs | **11 passed, 6/6 runs** |

Serial results are identical before and after (339 passed), so this changes nothing about what the tests assert.

### Reuse

`arc/common.py` is the module that owns the path constants, and after this change the repository has **exactly one production reader of `PYTEST_XDIST_WORKER`** — in `arc/common.py`. The adapter modules compose `ARC_TESTING_PATH` with `get_test_project_name` rather than re-reading the environment variable or adding a wrapper around a one-line `os.path.join`. Searched by behaviour ("scope a test directory per worker", "build a test project path") and for the raw operation (`PYTEST_XDIST_WORKER`, `ARC_PATH, 'Projects'`, `ARC_TESTING_PATH, 'test_`) before writing anything.

### Deliberately not changed

- `scheduler_test`'s remaining fixed directories are per-test and cleaned by `addCleanup` in the same worker and the same test that created them; every name was checked and none is shared between two tests.
- `common_test`'s two remaining sites are the helpers' own tests.
- `restart_test`'s remaining sites already take their name from `get_test_project_name`; the `'Projects'` literal there is just the join.
- `main_test.py`'s `assertIn('arc_project_for_testing_delete_after_usage', arc2.project_directory)` is a substring assertion on a project name, not a directory, and still holds — the worker suffix is appended, not substituted.

Unscoped `'Projects'` sites in test modules: **46 → 19**, with all 19 in the categories above.

### Follow-up, not in this PR

Roughly ten further adapter test modules — `xtb_test`, `obabel_test`, `mockter_test`, `cfour_test`, `ase_test`, `uma_test`, `pyscf_test`, `kinbot_test`, `xtbgsm_test` — carry the identical fixed-scratch-directory race and are untouched here to keep this PR reviewable.

Separately, `d643fb21` also removes a dead `arc0 = ` binding inside a `with self.assertRaises(InputError):` block in `main_test.py`. It is pre-existing on `main`; this PR only changed the `project=` argument on that line, which was enough for CodeQL to re-present the statement as new (alert 1956, `py/unused-local-variable`). The object never exists — the constructor raises — so the binding is removed rather than suppressed.

### DIFF
--- a/arc/checks/nmd_test.py
+++ b/arc/checks/nmd_test.py
@@ -12,7 +12,7 @@ import shutil
 import numpy as np
 
 import arc.checks.nmd as nmd
-from arc.common import ARC_PATH, ARC_TESTING_PATH, almost_equal_coords
+from arc.common import ARC_PATH, ARC_TESTING_PATH, almost_equal_coords, get_test_project_directory
 from arc.job.factory import job_factory
 from arc.level import Level
 from arc.molecule import Molecule
@@ -32,12 +32,13 @@ class TestNMD(unittest.TestCase):
         A method that is run before all unit tests in this class.
         """
         cls.maxDiff = None
+        cls.project_directory = get_test_project_directory('tmp_nmd_project')
         cls.generic_job = job_factory(job_adapter='gaussian',
                                       species=[ARCSpecies(label='SPC', smiles='C')],
                                       job_type='composite',
                                       level=Level(method='CBS-QB3'),
                                       project='test_project',
-                                      project_directory=os.path.join(ARC_PATH, 'Projects', 'tmp_nmd_project'),
+                                      project_directory=cls.project_directory,
                                       )
         cls.xyz_1 = {'symbols': ('C', 'N', 'H', 'H', 'H', 'H'),
                      'isotopes': (13, 14, 1, 1, 1, 1),
@@ -941,10 +942,7 @@ class TestNMD(unittest.TestCase):
         A function that is run ONCE after all unit tests in this class.
         Delete all project directories created during these unit tests
         """
-        projects = ['tmp_nmd_project']
-        for project in projects:
-            project_directory = os.path.join(ARC_PATH, 'Projects', project)
-            shutil.rmtree(project_directory, ignore_errors=True)
+        shutil.rmtree(cls.project_directory, ignore_errors=True)
         file_paths = [os.path.join(ARC_PATH, 'arc', 'checks', 'nul'), os.path.join(ARC_PATH, 'arc', 'checks', 'run.out')]
         for file_path in file_paths:
             if os.path.isfile(file_path):
--- a/arc/checks/ts_test.py
+++ b/arc/checks/ts_test.py
@@ -12,7 +12,7 @@ import shutil
 import numpy as np
 
 import arc.checks.ts as ts
-from arc.common import ARC_PATH, ARC_TESTING_PATH, almost_equal_lists
+from arc.common import ARC_PATH, ARC_TESTING_PATH, almost_equal_lists, get_test_project_directory
 from arc.job.factory import job_factory
 from arc.level import Level
 from arc.parser.parser import parse_normal_mode_displacement, parse_geometry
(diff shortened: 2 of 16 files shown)


calvinp0:
Superseded by #1008, which now carries this work.

Same problem, different mechanism. This PR added `get_test_project_name()` / `get_test_project_directory()` to `arc/common.py`, returning `ARC_PATH/Projects/<base>-<xdist worker id>`. #1008 uses `tempfile.mkdtemp()` + `addCleanup`/`addClassCleanup` instead, because a per-worker suffix does not fix the case every measured failure actually came from: `--dist worksteal` can split a single TestCase class across workers, so each worker independently runs `setUpClass` **and** `tearDownClass`. A class does not own its directory merely because only that class uses it. `mkdtemp` is unique per `setUpClass` invocation and cannot leave debris in the repo.

The helper itself was dropped rather than reshaped: `addCleanup`/`addClassCleanup` are TestCase members, so a mkdtemp-returning helper cannot register its own cleanup — it would alias a one-line stdlib call while leaving the boilerplate in place, and put test-only code in a production module. #1001 and #1009 also both already use the raw idiom independently, so a helper would have created a second mechanism the moment either landed.

#1008 now covers 26 files across 7 commits, including everything unique to this PR except four files that belong elsewhere: `gaussian_test.py` and `adapters/common_test.py` are converted on #1001 with a byte-identical idiom, and `scheduler_test.py` is restructured on #1009 (which also fixes 12 tests that mutate shared scheduler objects — a second root cause this PR did not address). Two extras came across beyond this PR's scope: `species_test`'s dead `gcn_tst` cleanup entry is removed, and `main_test`'s previously-uncleaned `adaptive_levels_test`/`test_0..3` projects now get scratch directories.

Measured on #1008: `arc/checks/nmd_test.py` and `ts_test.py` fail 3 of 3 solo runs under `-n 4 --dist worksteal` on main and pass 3 of 3 there; `main_test` + `species_test` + `scale_test` run together failed 4 of 5 on main leaving 5 stray project directories, and pass 5 of 5 with none. Full suite identical: 5 failed / 2803 passed both sides, the same 5 known torch_ani environmental failures.

The CodeQL alert raised here (unused `arc0` at `arc/main_test.py:513`) is carried over — that file is in #1008.

PR #994, opened 2026-08-17
Fix rmg_env_command dropping PYTHONPATH needed for source-tree RMG-Py

alongd:
## Failure mode

`rmg_env_command()`'s `RMG_PYTHON` branch (direct-interpreter invocation, used
on conda/mambaforge installs where micromamba's `conda` shim is broken)
unsets `_ARC_ENV_ACTIVATION_VARS` before invoking `rmg_env`'s interpreter,
to scrub ARC's env leakage (`BABEL_LIBDIR`, `LD_LIBRARY_PATH`, `CONDA_*`,
etc. — otherwise bound to `arc_env`'s tree and causing ABI-mismatch
crashes). `PYTHONPATH` is included in that unset list.

That's correct when RMG-Py/Arkane are pip-installed into `rmg_env`. But on
a host where RMG-Py/Arkane are used from a **source checkout reachable
only via `PYTHONPATH`**, unsetting it and stopping there leaves the child
interpreter unable to import `rmgpy`/`arkane` at all. Every call through
this branch dies with:

```
/home/alon/anaconda3/envs/rmg_env/bin/python: No module named arkane
```

Real-run evidence:
`/home/alon/runs/t3-pes-CH2O2/iteration_1/PDep_SA/network1_1/MSC/stderr.log`
contains exactly that line. Consequence: T3's master-equation sensitivity
job can never run on such a host, so no P-dep network can ever qualify for
QM refinement — the entire PDep→QM feature is silently dead, surfaced only
as the bland "No PDep networks qualified for QM refinement".

## Why unsetting PYTHONPATH is still right

Leaving ARC's own `PYTHONPATH` bound in the child is exactly the kind of
env leakage this branch exists to scrub (see the module docstring / the
`_ARC_ENV_ACTIVATION_VARS` comment) — a stale entry (e.g. an old checkout
on the caller's `PYTHONPATH`) could shadow `rmg_env`'s own site-packages
or ARC's own modules could bleed into the child. The unset is correct; the
bug is stopping there instead of restoring the *target* env's own path.

## The fix

After the unset, re-export `PYTHONPATH` from ARC's own `RMG_PATH` setting
(already resolved in `arc/settings/settings.py`, alongside `RMG_PYTHON`),
shell-quoted like the neighbouring lines, and only when `RMG_PATH` is
truthy (never emit an empty `export PYTHONPATH=`, which would be worse
than leaving it unset):

```bash
unset CONDA_PREFIX ... PYTHONPATH PYTHONHOME
export PATH=/home/alon/anaconda3/envs/rmg_env/bin:"$PATH"
export PYTHONPATH=/home/alon/Code/RMG-Py-t3pes   # <-- restored
/home/alon/anaconda3/envs/rmg_env/bin/python -c "import arkane; print(arkane.__file__)"
# -> arkane OK
```

Verified manually on the affected host before implementing.

## The other two branches

- `MAMBA_EXE` branch and the launcher-hunt (`bash -l`) branch never unset
  `PYTHONPATH` in the first place — they route through a launcher's `run`
  (which re-fires the target env's own activation hooks) or a login shell
  that sources the user's profile. Neither has this gap, so neither was
  changed.

## Test plan

- [x] Baseline: `python -m pytest arc/job/env_run_test.py` — 36 passed (unmodified branch)
- [x] Post-fix: same command — 38 passed (36 + 2 new regression tests)
- [x] New tests assert `PYTHONPATH` is re-exported from `RMG_PATH` (after
      the `unset`) in the `RMG_PYTHON` branch, and that nothing is emitted
      when `RMG_PATH` is falsy. Settings are patched via `patch.dict`, so
      the tests don't depend on this machine's real paths.
- [x] `ruff check arc/job/env_run.py arc/job/env_run_test.py` — clean

### DIFF
--- a/arc/job/env_run.py
+++ b/arc/job/env_run.py
@@ -168,11 +168,16 @@ def run_in_conda_env(
 # target env, but invoking rmg_env's interpreter *directly* does not: they stay
 # pointed at arc_env's tree and make the child resolve shared-library plugins
 # against the wrong env, which is what silently kills Arkane's OpenBabel
-# import. Only the direct-interpreter branch below has to scrub them.
-# PYTHONPATH is included because the RMG helper scripts in arc/scripts/ import
-# only rmgpy/arkane/rdkit plus a sibling ``common`` module (found via the
-# script's own directory), so nothing there needs ARC on the path -- while a
-# stale entry would shadow rmg_env's site-packages.
+# import. Only the direct-interpreter branch below has to scrub this whole set;
+# the launcher branches leave the ABI-sensitive vars to ``run``'s re-activation.
+# PYTHONPATH is the exception: a launcher's ``run`` re-fires the target env's
+# activation hooks but does NOT manage PYTHONPATH, so ARC's own PYTHONPATH would
+# leak into the child on *every* branch and shadow rmg_env's site-packages.
+# ``_pythonpath_lines`` therefore normalizes it uniformly across all three
+# branches. It is kept in this tuple too so the direct-interpreter branch scrubs
+# it as part of the single ``unset``; the RMG helper scripts in arc/scripts/
+# import only rmgpy/arkane/rdkit plus a sibling ``common`` module (found via the
+# script's own directory), so nothing there needs ARC on the path.
 _ARC_ENV_ACTIVATION_VARS = (
     'CONDA_PREFIX', 'CONDA_PREFIX_1', 'CONDA_DEFAULT_ENV', 'CONDA_PROMPT_MODIFIER',
     'CONDA_SHLVL', 'LD_LIBRARY_PATH', 'BABEL_LIBDIR', 'BABEL_DATADIR',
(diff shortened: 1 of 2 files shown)


calvinp0:
One question about scope. The failure mode you describe - a source-tree RMG-Py/Arkane reachable only via PYTHONPATH- applies to all three branches of rmg_env_command, but only the RMG_PYTHON one is fixed here. The MAMBA_EXE and launcher-hunt branches don't unset _ARC_ENV_ACTIVATION_VARS at all, so they inherit ARC's PYTHONPATH, which would contain the RMG-Py checkout only by coincidence. So a host with the same source-tree layout but a working micromamba shim still gets No module named arkane - and that's the branch CI runs on.

That also leaves the three branches with three different PYTHONPATH behaviours after this lands: inherited, RMG_PATH exactly, inherited-plus-profile. The inheritance in branches 1 and 3 is the same env leakage the docstring calls out as causing ABI-mismatch crashes, just unscrubbed

alongd:
Thanks, @calvinp0 , I agree. pushed a fix

calvinp0 (approved):

PR #995, opened 2026-08-17
fix: guard against a None species.thermo in parse_arkane_thermo_output

alongd:
## Crash

Real T3 campaign run (`~/runs/t3-pes-CH2O2/`) died after all QM jobs converged, during thermo
reporting:

```
AttributeError: 'NoneType' object has no attribute 'H298' and no __dict__ for setting new attributes
  File "arc/statmech/arkane.py", line 509, in parse_arkane_thermo_output
    spc.thermo.H298 = content[lbl]['H298']
```

## Root cause

`ARCSpecies.__init__` always sets `self.thermo = ThermoData()` as a default. Somewhere between
species construction and this loop, a species can end up with `spc.thermo is None` again — I could
not pin the exact assignment site that clears it in the live run, but the codebase itself already
treats this as an expected, guarded-against state in three independent places:

- `ArkaneAdapter.set_reaction_dh_rxn` (arc/statmech/arkane.py)
- `processor.process_arc_project` (arc/processor.py:198)
- `output.get_species_output_dict` (arc/output.py:510)

`parse_arkane_thermo_output`'s result-assignment loop was the one place that assumed `spc.thermo`
can never be `None` and crashed instead of guarding — classifying this as case (a): a species
legitimately reaching this code with no thermo container, not a `valid_labels` filtering bug (I
found no evidence `valid_labels` includes species it should exclude).

## Fix

Restore the `ThermoData()` default (matching the exact construction used in `ARCSpecies.__init__`)
immediately before populating it with the real, successfully-computed Arkane results, instead of
crashing and discarding those results:

```python
if spc.thermo is None:
    spc.thermo = ThermoData()
spc.thermo.H298 = content[lbl]['H298']
...
```

## Sibling kinetics path

Checked `parse_arkane_kinetics_output`/`parse_reaction_kinetics` for the same unguarded-attribute
pattern: none found. `parse_reaction_kinetics` always does one wholesale `reaction.kinetics =
kinetics` dict assignment rather than mutating attributes on an existing kinetics sub-object, so
there's no equivalent `None`-target crash risk there.

## Containment at the process_arc_project/compute_thermo boundary

Deliberately did **not** add a broader try/except there. The crash's root data problem is now fixed
at its source (the offending species gets a real `ThermoData` container populated with its actual
results, not skipped). A blanket guard at the reporting boundary would risk swallowing genuinely
new data-integrity bugs in future runs rather than surfacing them — better handled as its own
decision if a distinct need for it shows up.

## Testing

Added `test_parse_arkane_thermo_output_recovers_missing_thermo_container` to
`arc/statmech/arkane_test.py`, covering a species with `.thermo is None` reaching the loop and a
well-formed sibling species processed normally in the same call.

- Baseline (`git stash`, unmodified branch): `44 passed, 2 failed`. The 2 failures
  (`test_generate_arkane_input`, `test_run_statmech_using_molecular_properties`) are pre-existing,
  caused by a missing `arkane` module in the `rmg_env` conda environment ARC shells out to — unrelated
  to this change.
- Post-fix: `45 passed, 2 failed` (same 2 pre-existing failures, +1 new passing test).
- Verified the new test is a genuine regression test: reverted only `arc/statmech/arkane.py` to
  HEAD and re-ran it in isolation — it failed with the exact same traceback signature as the real
  crash above, then passed again after reapplying the fix.

### DIFF
--- a/arc/statmech/arkane.py
+++ b/arc/statmech/arkane.py
@@ -21,6 +21,8 @@ from arc.job.local import execute_command
 from arc.statmech.adapter import StatmechAdapter
 from arc.statmech.factory import register_statmech_adapter
 
+from arc.species.species import ThermoData
+
 if TYPE_CHECKING:
     from arc.level import Level
     from arc.reaction import ARCReaction
@@ -511,6 +513,24 @@ class ArkaneAdapter(StatmechAdapter, ABC):
                 logger.info(header)
                 for lbl in valid_labels:
                     spc = self.species_dict[lbl]
+                    if spc.thermo is None:
+                        # ``ARCSpecies.__init__`` sets ``self.thermo = ThermoData()`` unconditionally,
+                        # so no species ARC builds itself can reach this loop with ``None``. A caller's
+                        # subclass can, though, by re-assigning the attribute after ``super().__init__()``
+                        # -- which is exactly how T3's ``T3Species`` produced the crash this guards
+                        # (fixed at the source in T3 PR #187). Other call sites across the codebase
+                        # (e.g., ``ArkaneAdapter.set_reaction_dh_rxn``, ``processor.process_arc_project``,
+                        # ``output.get_species_output_dict``) already tolerate ``None`` here.
+                        #
+                        # Supply the container rather than crash: the Arkane results being assigned are
+                        # real and already computed, and losing them means discarding converged QM jobs
+                        # at the reporting stage. Warn rather than repair silently, so a caller
+                        # re-introducing this stays visible instead of being absorbed.
+                        logger.warning(f'Species {lbl} reached the Arkane thermo assignment loop with '
+                                       f'.thermo set to None, instead of the ThermoData() instance '
+                                       f'ARCSpecies.__init__() assigns. Supplying an empty ThermoData() '
+                                       f'container so the computed results are not lost.')
+                        spc.thermo = ThermoData()
                     spc.thermo.H298 = content[lbl]['H298']
                     spc.thermo.S298 = content[lbl]['S298']
                     spc.thermo.data = content[lbl]['data']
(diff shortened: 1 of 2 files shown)


alongd:
**Root cause found; PR updated.** Two corrections to what this PR originally claimed.

**1. The "restore the `__init__` default" framing was wrong.** `git log -S "self.thermo = ThermoData()" -- arc/species/species.py` returns exactly one commit (`b2be26dc8`, "Implement ThermoData and TransportData classes in species"). The line has never been removed. Nothing was restored, because nothing was dropped.

**2. The `None` does not originate in ARC at all.** I could not find the producer when opening this PR, and said so. It is now located, and the reason the search failed is that it is on the other side of the boundary:

- `ARCSpecies.__init__` sets `self.thermo = ThermoData()` unconditionally (`arc/species/species.py:362`);
- `from_dict` never touches the attribute, and it is not serialised into `restart.yml`;
- no ARC production code assigns `None` to it — `grep -rn "thermo\s*=\s*None" arc/ --include=*.py` matches only test files.

The producer is a **caller's subclass**. T3's `T3Species(ARCSpecies)` ran `self.thermo = thermo` after `super().__init__()`, with `thermo` defaulting to `None`, so *every* species T3 handed ARC arrived with the invariant already broken. Fixed at the source in ReactionMechanismGenerator/T3#187.

**This PR is still worth merging** as defence in depth, and I've kept it. ARC should not discard converged QM results because a caller violated an invariant ARC never advertised — and this loop was the odd one out, since `set_reaction_dh_rxn`, `process_arc_project` and `get_species_output_dict` all already tolerate a `None` thermo.

**Changes since the first push:**

- commit message rewritten to say what actually happens, and to name the real producer;
- the guard now emits a `logger.warning` instead of repairing silently. A silent guard means the next caller to re-introduce this gets absorbed with no trace, which is how the state stayed unexplained this long.

Tests unchanged: 45 passed, 2 pre-existing failures (`test_generate_arkane_input`, `test_run_statmech_using_molecular_properties`) that are the known `rmg_env` missing-`arkane` environment issue, identical to the baseline on `main`.

calvinp0 (approved):

PR #1000, opened 2026-08-19
Delete ESS checkfiles on the server too, not just locally

alongd:
ARC uploads a Gaussian checkpoint as a wavefunction guess for each job and downloads it again afterwards. When a run terminates it deletes those files — but only on the local machine:

- `delete_check_files(project_directory)` walks `<project_directory>/calcs`;
- `main.py` calls it unconditionally at the end of a run, since `keep_checks` defaults to `False`;
- no code path anywhere deletes anything on the remote server.

So the documented policy — *"they usually take up lots of space and are not needed after ARC terminates"* — is half-implemented, and every checkpoint ARC has ever uploaded is still sitting in the remote project directory.

**Measured impact.** On a quota-managed cluster, **82.3% of a fully characterised species' 61.4 MiB footprint is exactly this class of file** — the next largest class is 5.6× smaller — and 6.35 GiB had silently accumulated against a 300 GB quota. Nothing downstream reads the remote copy: thermochemistry is parsed from the primary output log, and the *local* copy is the one reused as a guess.

**What this does.** Adds `SSHClient.delete_check_files` and a `delete_remote_check_files` helper, and routes the existing teardown through a new `ARC.clean_check_files`. Design constraints kept deliberately tight:

- `keep_checks=True` keeps them everywhere, local and remote — the flag's meaning is unchanged;
- deletion is scoped to `*.chk` **under the run's own remote project directory**, using the same path construction `job/adapter.py` already uses, and no-ops if that directory does not exist;
- errors are logged and never raised. This runs after the science is done; an unreachable server at teardown is an inconvenience, not a reason to lose a run;
- servers named `local` are skipped; nothing about uploads, the wavefunction-guess mechanism, or `keep_checks`'s default changes.

**Tests** cover all three: deleted with `keep_checks=False` (asserting only `.chk`, and only under this project's directory), kept with `keep_checks=True`, and a run completing normally when the server call raises. Verified as a regression test rather than a tautology — removing just the remote call while leaving the helper in place makes `test_check_files_are_deleted_locally_and_remotely` fail.

### DIFF
--- a/arc/job/ssh.py
+++ b/arc/job/ssh.py
@@ -9,6 +9,7 @@ Todo:
 import datetime
 import logging
 import os
+import shlex
 import time
 from typing import Any
 from collections.abc import Callable
@@ -59,16 +60,24 @@ class SSHClient(object):
 
     Args:
         server (str): The server name as specified in ARCs's settings file under ``servers`` as a key.
+        connection_attempts (int, optional): The number of times to try connecting to the server,
+                                             waiting a minute between attempts. The default keeps trying
+                                             for 24 hours, which is appropriate while jobs are running.
+                                             Pass a low number where blocking is worse than giving up.
 
     Attributes:
         server (str): The server name as specified in ARCs's settings file under ``servers`` as a key.
         address (str): The server's address.
         un (str): The username to use on the server.
         key (str): A path to a file containing the RSA SSH private key to the server.
+        connection_attempts (int): The number of times to try connecting to the server.
         _ssh (paramiko.SSHClient): A high-level representation of a session with an SSH server.
         _sftp (paramiko.sftp_client.SFTPClient): SFTP client used to perform remote file operations.
     """
-    def __init__(self, server: str = '') -> None:
+    def __init__(self,
+                 server: str = '',
+                 connection_attempts: int = 1440,  # Continue trying for 24 hrs (24 hr * 60 min/hr).
+                 ) -> None:
         if server == '':
             raise ValueError('A server name must be specified')
         if server not in servers.keys():
@@ -77,6 +86,7 @@ class SSHClient(object):
         self.address = servers[server]['address']
         self.un = servers[server]['un']
         self.key = servers[server]['key']
+        self.connection_attempts = connection_attempts
         self._sftp = None
         self._ssh = None
         logging.getLogger("paramiko").setLevel(logging.WARNING)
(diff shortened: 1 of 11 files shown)


arc/job/ssh.py
```diff
@@ -537,6 +553,49 @@ def _create_dir(self, remote_path: str) -> None:
...
+    Args:
+        server (str): The server name as specified in ARC's settings file under ``servers`` as a key.
+        project (str): The ARC project name.
+
+    Returns: str
+        The remote path of the project directory.
+    """
+    path = servers[server].get('path', '').lower()
+    path = os.path.join(path, servers[server]['un']) if path else ''
+    return os.path.join(path, 'runs', 'ARC_Projects', project)
```
alongd: not alwyas running under 'runs', 'ARC_Projects'

arc/job/ssh.py
```diff
@@ -489,6 +489,22 @@ def change_mode(self,
         command = f'chmod{recursive} {mode} {file_name}'
         self._send_command_to_server(command, remote_path)
 
+    def delete_check_files(self, remote_path: str) -> None:
```
alongd: what's the difference between this and arc.common.delete_check_files()?

arc/job/ssh.py
```diff
@@ -489,6 +499,23 @@ def change_mode(self,
...
+        Pass ``True`` to the ``keep_checks`` flag in ARC to avoid deleting check files.
+        The local counterpart of this method is ``arc.common.delete_check_files()``.
+
+        Args:
+            remote_path (str): The remote directory path under which checkfiles will be deleted.
+        """
+        if not remote_path or not self._check_dir_exists(remote_path):
+            return
+        command = f'find {shlex.quote(remote_path)} -type f -name "*.chk" -delete'
+        _, stderr = self._send_command_to_server(command)
```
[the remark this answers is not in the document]
alongd: Good catch — the separate `_check_dir_exists()` call is gone. `delete_remote_check_files()` now sends one command that both tests for the directory and deletes, with `shlex.quote()` applied to the path in both positions:

```python
quoted_path = shlex.quote(remote_path)
command = f'[ -d {quoted_path} ] && find {quoted_path} -type f -name "*.chk" -delete'
```

The regression test no longer has to mock `_check_dir_exists` away, so the quoting is now exercised end to end against a real path containing `$(touch injected.txt)`.

arc/job/ssh.py
alongd: Agreed — `get_remote_project_path()` was deleted. The remote project path is no longer re-derived: `JobAdapter` records the `remote_project_path` it actually built (`arc/job/adapter.py`), and `Scheduler` collects one per server in `remote_project_paths`. Whatever layout a job really used is the one that gets cleaned.

arc/job/ssh.py
alongd: Same job, other side of the wire — the local one deletes `*.chk` under the project directory, this one does it under the project directory on a server. The method is now named `delete_remote_check_files()` and both docstrings name each other as counterparts.

PR #1001, opened 2026-08-19
Make remote submission over SSH work, from the host and from the container

calvinp0:
Makes ARC's remote-cluster path work over SSH, and then makes it work from inside the Docker image.

## SSH

A server's configured `key` was handed to `load_system_host_keys()` rather than to `connect()`, so it was **never offered as an authentication identity**. It is now passed as `key_filename`, and made optional so a forwarded ssh-agent or paramiko's default key paths can authenticate where no local key file exists.

The same path stops silently trusting unknown hosts — `AutoAddPolicy` becomes `WarningPolicy`, with an opt-in per-server `strict_host_key_checking` selecting `RejectPolicy` — and narrows a bare `except` around the connect retry that had been swallowing Ctrl-C and discarding the real authentication error.

A process-global connection pool replaces one-Transport-per-job, so a TS search no longer opens ~100 connections to a single cluster; teardown is wired into ARC's exit. Pipe mode now refuses a remote server, and says why, instead of deadlocking on a submission that can never land. Orca NEB resolves its reactant/product geometry paths on the executing machine.

**Reuse check:** searched `main` for existing pooling or connection reuse (`pool`, `reuse.*connect`, `persistent.*connect`, `ControlMaster` across `arc/job/ssh.py` and `arc/job/adapter.py`) — none exists, and `SSHClient()` is constructed in exactly one place (`arc/job/ssh.py:371`). `arc/job/ssh_pool.py` is genuinely new rather than a third copy.

## Docker

Adds agent-socket forwarding and read-only key / `~/.arc` mounts; fixes a PUID/PGID remap that aborted the container for most Linux desktop users; stops a `chmod o+rw` that was widening the user's live host agent socket; and adds a pre-flight that fails fast when a mounted settings overlay is unimportable, instead of letting ARC run silently against its dummy `server1.host.edu` fixtures.

## Structure

Six commits, **no file touched by more than one** — the two stories partition cleanly by path (SSH owns `arc/**` + `ARC.py`; Docker owns `Dockerfile`, `docker-compose.yml`, `dockerfiles/**`, `docs/**`), with only documentation cross-references coupling them.

## Verification

Rebased onto `main` (`d033cbbf`) from 107 behind, replayed with zero conflicts. The rebase and the subsequent 13→6 restructure were both checked by **content identity**, not just by tests: the changed-line multiset of `old-base..old-head` versus `new-base..new-head` is identical in both directions (2453 lines each), and the restructure left the tree bit-for-bit unchanged. Tests cannot prove a rebase didn't drop a hunk; this can.

`249 passed` across `ssh_test`, `ssh_pool_test`, `pipe/`, `adapter_test`, `settings`. The broader `arc/job/` sweep shows the same 5 pre-existing `torch_ani_test` environmental failures before and after — zero new.

### DIFF
--- a/arc/job/adapters/cfour.py
+++ b/arc/job/adapters/cfour.py
@@ -23,7 +23,6 @@ from arc.job.adapters.common import (_initialize_adapter,
                                      )
 from arc.job.factory import register_job_adapter
 from arc.job.local import execute_command, submit_job
-from arc.job.ssh import SSHClient
 from arc.level import Level
 from arc.species.converter import zmat_from_xyz, zmat_to_str
 
@@ -302,7 +301,7 @@ class CFourAdapter(JobAdapter):
         """
         self._log_job_execution()
         if self.server != 'local':
-            with SSHClient(self.server) as ssh:
+            with self._open_or_borrow_ssh() as ssh:
                 self.job_status[0], self.job_id = ssh.submit_job(remote_path=self.remote_path)
         else:
             self.job_status[0], self.job_id = submit_job(path=self.local_path)
--- a/arc/job/adapters/cfour_test.py
+++ b/arc/job/adapters/cfour_test.py
@@ -9,6 +9,8 @@ import math
 import os
 import shutil
 import unittest
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
 
 from arc.common import ARC_TESTING_PATH
 from arc.job.adapters.cfour import CFourAdapter
@@ -145,6 +147,55 @@ R8=0.9724
         Delete all project directories created during these unit tests
         """
         shutil.rmtree(os.path.join(ARC_TESTING_PATH, 'test_CFourAdapter'), ignore_errors=True)
+
+
+class TestCFourQueueSubmissionSharesAConnection(unittest.TestCase):
+    """CFour overrides execute_queue(), so it needs the shared borrow path of its own accord."""
+
+    @staticmethod
+    def _adapter(server):
+        """
+        A stand-in carrying only what ``execute_queue()`` reads, with the borrow path mocked.
+
+        A real adapter is not built here because constructing one runs the whole adapter
+        initialization for a test about which connection the submission uses. Naming the
+        attributes explicitly also states what ``execute_queue()`` is allowed to touch, so a
+        method that starts reading something else fails here rather than passing on a mock
+        that answers to anything.
+        """
+        return SimpleNamespace(server=server,
+                               remote_path='/remote/job',
+                               local_path='/local/job',
+                               job_status=['initializing', {'status': 'initializing'}],
+                               job_id=0,
+                               _log_job_execution=lambda: None,
+                               _open_or_borrow_ssh=MagicMock())
+
+    def test_a_remote_submission_borrows_its_connection(self):
+        """Submitting used to open a second connection right after the upload used one."""
+        adapter = self._adapter('zeus')
+        ssh = MagicMock()
+        ssh.submit_job.return_value = ('running', 4242)
+        adapter._open_or_borrow_ssh.return_value.__enter__.return_value = ssh
+        CFourAdapter.execute_queue(adapter)
+        adapter._open_or_borrow_ssh.assert_called_once()
+        ssh.submit_job.assert_called_once_with(remote_path='/remote/job')
+        self.assertEqual((adapter.job_status[0], adapter.job_id), ('running', 4242))
+
+    def test_the_borrowed_connection_is_released(self):
+        adapter = self._adapter('zeus')
+        adapter._open_or_borrow_ssh.return_value.__enter__.return_value.submit_job.return_value = \
+            ('running', 1)
+        CFourAdapter.execute_queue(adapter)
+        adapter._open_or_borrow_ssh.return_value.__exit__.assert_called_once()
+
+    def test_a_local_submission_opens_no_connection(self):
+        """A local queue job is submitted with a local command and must not touch SSH."""
+        adapter = self._adapter('local')
+        with patch('arc.job.adapters.cfour.submit_job', return_value=('running', 9)) as submitted:
+            CFourAdapter.execute_queue(adapter)
+        adapter._open_or_borrow_ssh.assert_not_called()
+        submitted.assert_called_once_with(path='/local/job')
 
 
 if __name__ == '__main__':
--- a/arc/job/adapters/common.py
+++ b/arc/job/adapters/common.py
@@ -258,13 +258,9 @@ def _initialize_adapter(obj: JobAdapter,
 
     obj.args = set_job_args(args=obj.args, level=obj.level, job_name=obj.job_name)
     if obj.execution_type != 'incore' and obj.job_adapter in obj.ess_settings.keys() and obj.server is None:
-        if 'server' in obj.args['trsh']:
-            obj.server = obj.args['trsh']['server']
-        elif obj.job_adapter in obj.ess_settings.keys():
-            if isinstance(obj.ess_settings[obj.job_adapter], list):
-                obj.server = obj.ess_settings[obj.job_adapter][0]
-            else:
-                obj.server = obj.ess_settings[obj.job_adapter]
+        obj.server = resolve_job_server(ess_settings=obj.ess_settings,
+                                        job_adapter=obj.job_adapter,
+                                        args=obj.args)
 
     obj.set_file_paths()
     obj.set_cpu_and_mem()
@@ -279,6 +275,40 @@ def _initialize_adapter(obj: JobAdapter,
 
     obj.set_files()
     check_argument_consistency(obj)
+
+
+def resolve_job_server(ess_settings: dict,
+                       job_adapter: str,
+                       args: dict | None = None,
+                       ) -> str | None:
+    """
+    Return the server that a ``job_adapter`` job will be submitted to.
+
+    A troubleshooting override in ``args['trsh']['server']`` wins, since it is set precisely to
+    move a job off the server that failed it, and it is honoured even when it is empty so that
+    the caller sees the same server the job itself would be given. Otherwise the server is the
+    first one the ESS settings name for the adapter, which is the entry ARC submits to; a bare
+    string there is read as a single server.
+
+    Args:
+        ess_settings (dict): The ESS settings, mapping an adapter to the server or the list of
+                             servers it is available on.
+        job_adapter (str): The job adapter to resolve a server for.
+        args (dict, optional): The job's arguments, whose ``'trsh'`` entry may carry a
+                               ``'server'`` override.
+
+    Returns: str | None
+        The server name, or ``None`` when neither an override nor the ESS settings name one.
+    """
+    trsh_args = (args or dict()).get('trsh') or dict()
+    if isinstance(trsh_args, dict) and 'server' in trsh_args:
+        return trsh_args['server']
+    servers_for_adapter = (ess_settings or dict()).get(job_adapter)
+    if isinstance(servers_for_adapter, str):
+        return servers_for_adapter or None
+    if isinstance(servers_for_adapter, (list, tuple)) and len(servers_for_adapter):
+        return servers_for_adapter[0]
+    return None
 
 
 def is_restricted(obj: JobAdapter) -> bool | list[bool]:
--- a/arc/job/ssh.py
+++ b/arc/job/ssh.py
@@ -6,11 +6,14 @@ Todo:
     * delete scratch files of a failed job: ssh nodeXX; rm scratch/dhdhdhd/job_number
 """
 
+import base64
 import datetime
+import hashlib
 import logging
 import os
 import shlex
 import time
+from collections import Counter
 from typing import Any
 from collections.abc import Callable
 
@@ -27,6 +30,116 @@ check_status_command, delete_command, list_available_nodes_command, servers, sub
     settings['check_status_command'], settings['delete_command'], settings['list_available_nodes_command'], \
     settings['servers'], settings['submit_command'], settings['submit_filenames']
 
+KNOWN_HOSTS_PATH = '~/.ssh/known_hosts'
+PLACEHOLDER_ADDRESS_SUFFIX = '.host.edu'
+PLACEHOLDER_USERNAME = '<username>'
+
+
+class UnknownHostKeyError(ServerError, paramiko.SSHException):
+    """
+    Raised when a server's host key is absent from ``known_hosts`` and the server sets
+    ``strict_host_key_checking``.
+
+    An ARC :class:`~arc.exceptions.ServerError`, so ARC's server error handling covers it, and a
+    ``paramiko.SSHException``, which is what a missing host key policy is expected to raise. Being
+    a distinct type, a refused host key is told apart from a transport failure without matching on
+    paramiko's message text.
+    """
+
+
+def get_host_key_fingerprint(key: paramiko.PKey) -> str:
+    """
+    Return the OpenSSH-style SHA256 fingerprint of a host key.
+
+    Args:
+        key (paramiko.PKey): The host key to fingerprint.
+
+    Returns: str
+        The fingerprint, formatted as ``SHA256:<unpadded base64>``, as reported by
+        ``ssh-keygen -lf`` and by OpenSSH when it prompts about an unknown host.
+    """
+    digest = hashlib.sha256(key.asbytes()).digest()
+    return f'SHA256:{base64.b64encode(digest).decode().rstrip("=")}'
+
+
+class LogAndAcceptHostKeyPolicy(paramiko.MissingHostKeyPolicy):
+    """
+    A missing host key policy that reports the unknown key through ARC's logger, then accepts it.
+
+    paramiko's ``WarningPolicy`` emits through ``warnings.warn()``, which ARC's
+    :func:`arc.common.initialize_log` filters out for the ``paramiko`` module, so nothing of it
+    reaches the log file or the terminal. This policy logs the host and the key's fingerprint at
+    the warning level, and connects.
+    """
+
+    def missing_host_key(self,
+                         client: paramiko.SSHClient,
+                         hostname: str,
+                         key: paramiko.PKey,
+                         ) -> None:
+        """
+        Log the unknown host key and accept it.
+
+        Args:
+            client (paramiko.SSHClient): The client the key was presented to.
+            hostname (str): The address of the server that presented the key.
+            key (paramiko.PKey): The host key that is not in ``known_hosts``.
+        """
+        logger.warning(f'Connecting to {hostname} with an unverified host key: '
+                       f'{key.get_name()} {get_host_key_fingerprint(key)} is not in '
+                       f'{KNOWN_HOSTS_PATH}, so a first-ever connection cannot be told apart '
+                       f'from an interception. Verify the fingerprint against a trusted source '
+                       f'and add the key with:\n'
+                       f'    ssh-keyscan -H {hostname} >> {KNOWN_HOSTS_PATH}')
+
+
+class RejectUnknownHostKeyPolicy(paramiko.RejectPolicy):
+    """
+    A missing host key policy that refuses the connection, raising :class:`UnknownHostKeyError`.
+    """
+
+    def missing_host_key(self,
+                         client: paramiko.SSHClient,
+                         hostname: str,
+                         key: paramiko.PKey,
+                         ) -> None:
+        """
+        Refuse the unknown host key.
+
+        Args:
+            client (paramiko.SSHClient): The client the key was presented to.
+            hostname (str): The address of the server that presented the key.
+            key (paramiko.PKey): The host key that is not in ``known_hosts``.
+
+        Raises:
+            UnknownHostKeyError: Always.
+        """
+        raise UnknownHostKeyError(f'The host key of {hostname} '
+                                  f'({key.get_name()} {get_host_key_fingerprint(key)}) is not in '
+                                  f'{KNOWN_HOSTS_PATH}, and this server sets '
+                                  f'strict_host_key_checking. Verify the fingerprint against a '
+                                  f'trusted source and add the key with:\n'
+                                  f'    ssh-keyscan -H {hostname} >> {KNOWN_HOSTS_PATH}')
+
+
+class HostKeyMismatchError(ServerError, paramiko.SSHException):
+    """
+    Raised when the host key a server presents contradicts the one stored in ``known_hosts``.
+
+    Told apart from :class:`UnknownHostKeyError`, which is the absence of a stored key, because
+    the two mean different things: an absent key is a host never connected to before, while a
+    contradicted key is either a re-keyed server or an interception, and only the second of
+    those is a security event. Being an ARC :class:`~arc.exceptions.ServerError` keeps it inside
+    ARC's server error handling, and being a ``paramiko.SSHException`` keeps it catchable
+    alongside the exception it is raised from.
+    """
+
+
+PERMANENT_CONNECTION_ERRORS = (paramiko.AuthenticationException,
+                               paramiko.BadHostKeyException,
+                               UnknownHostKeyError,
+                               )
+
 
 def check_connections(function: Callable[..., Any]) -> Callable[..., Any]:
     """
@@ -36,11 +149,16 @@ def check_connections(function: Callable[..., Any]) -> Callable[..., Any]:
     to make sure your connection still alive. If connection is bad, this
     decorator will reconnect the SSH channel, to avoid connection related
     error when executing the method.
+
+    ``connect()`` assigns ``self._sftp`` and ``self._ssh`` itself and returns nothing, so its
+    result is not unpacked into them. Unpacking it raised ``TypeError: cannot unpack
+    non-iterable NoneType object`` for any client that had not connected yet, which is the one
+    case this branch exists to serve.
     """
     def decorator(*args, **kwargs) -> Any:
         self = args[0]
         if self._ssh is None:  # not sure if some status may cause False
-            self._sftp, self._ssh = self.connect()
+            self.connect()
         # test connection, reference:
         # https://stackoverflow.com/questions/
         # 20147902/how-to-know-if-a-paramiko-ssh-channel-is-disconnected
@@ -64,19 +182,30 @@ class SSHClient(object):
                                              waiting a minute between attempts. The default keeps trying
                                              for 24 hours, which is appropriate while jobs are running.
                                              Pass a low number where blocking is worse than giving up.
+                                             A permanent failure raises on the first attempt whatever
+                                             this is set to, see :meth:`connect`.
 
     Attributes:
         server (str): The server name as specified in ARCs's settings file under ``servers`` as a key.
         address (str): The server's address.
         un (str): The username to use on the server.
-        key (str): A path to a file containing the RSA SSH private key to the server.
+        key (str | None): A path to a file containing the SSH private key to the server.
+                          Optional: when it is not set (or set to an empty string), no explicit
+                          identity is offered and paramiko falls back to a running ssh-agent and
+                          then to the default key paths (``~/.ssh/id_rsa``, ``~/.ssh/id_ecdsa``,
+                          ``~/.ssh/id_ed25519``), which is how agent-based setups authenticate.
         connection_attempts (int): The number of times to try connecting to the server.
         _ssh (paramiko.SSHClient): A high-level representation of a session with an SSH server.
         _sftp (paramiko.sftp_client.SFTPClient): SFTP client used to perform remote file operations.
+        _keepalive_interval (int | None): The keepalive interval this client was asked for, which
+                                          :meth:`connect` re-applies to every transport it opens.
+                                          ``None`` until :func:`arc.job.ssh_pool.set_keepalive`
+                                          sets it, since a client that is not pooled is not held
+                                          open long enough to be dropped while idle.
     """
     def __init__(self,
                  server: str = '',
-                 connection_attempts: int = 1440,  # Continue trying for 24 hrs (24 hr * 60 min/hr).
+                 connection_attempts: int = 1440,
                  ) -> None:
         if server == '':
             raise ValueError('A server name must be specified')
@@ -85,10 +214,11 @@ class SSHClient(object):
         self.server = server
         self.address = servers[server]['address']
         self.un = servers[server]['un']
-        self.key = servers[server]['key']
+        self.key = servers[server].get('key') or None
         self.connection_attempts = connection_attempts
         self._sftp = None
         self._ssh = None
+        self._keepalive_interval = None
         logging.getLogger("paramiko").setLevel(logging.WARNING)
 
     def __enter__(self) -> SSHClient:
@@ -122,7 +252,7 @@ class SSHClient(object):
             # and even yield different behaviors.
             # Make sure to change directory back after the command is executed
             if self._check_dir_exists(remote_path):
-                command = f'cd "{remote_path}"; {command}; cd '
+                command = f'cd -- {shlex.quote(remote_path)}; {command}; cd '
             else:
                 raise InputError(
                     f'Cannot execute command at given remote_path({remote_path})')
@@ -139,6 +269,7 @@ class SSHClient(object):
         stderr = stderr.readlines()
         return stdout, stderr
 
+    @check_connections
     def upload_file(self,
                     remote_file_path: str,
                     local_file_path: str = '',
@@ -179,12 +310,23 @@ class SSHClient(object):
             logger.debug(f'Could not upload file {local_file_path} to {self.server}!')
             raise ServerError(f'Could not write file {remote_file_path} on {self.server}. ')
 
+    @check_connections
     def download_file(self,
                       remote_file_path: str,
                       local_file_path: str,
                       ) -> None:
         """
         Download a file from the server.
+
+        The existence of the remote file is checked up to three times, one second apart, since
+        scheduler epilogues may flush a job's stdout and stderr to the work directory a second or
+        two after the job has left the queue.
+
+        If the remote file does not exist, the local path is emptied instead of being downloaded
+        to: it is created if it is absent, and truncated if a file is already there. A file left
+        at that path by an earlier job therefore never survives to be read as this job's output
+        (see :meth:`arc.job.adapter.JobAdapter._get_additional_job_info`), and the miss is
+        reported at the warning level.
 
         Args:
             remote_file_path (str): The remote path to be downloaded from.
@@ -193,19 +335,37 @@ class SSHClient(object):
         Raises:
             ServerError: If the file cannot be downloaded with maximum times to try
         """
-        if not self._check_file_exists(remote_file_path):
-            # Check if a file exists
-            # This doesn't have a real impact now to avoid screwing up ESS trsh
-            # but introduce an opportunity for better troubleshooting.
-            # The current behavior is that if the remote path does not exist
-            # an empty file will be created at the local path
-            logger.debug(f'{remote_file_path} does not exist on {self.server}.')
+        for attempt in range(3):
+            if self._check_file_exists(remote_file_path):
+                break
+            if attempt < 2:
+                time.sleep(1.0)
+        else:
+            logger.warning(f'{remote_file_path} does not exist on {self.server}. '
+                           f'Emptied {local_file_path} instead of downloading it.')
+            self._empty_local_file(local_file_path)
+            return
         try:
             self._sftp.get(remotepath=remote_file_path,
                            localpath=local_file_path)
         except IOError:
             logger.warning(f'Got an IOError when trying to download file '
                            f'{remote_file_path} from {self.server}')
+
+    @staticmethod
+    def _empty_local_file(local_file_path: str) -> None:
+        """
+        Create ``local_file_path`` if it does not exist, and truncate it if it does.
+
+        Args:
+            local_file_path (str): The local path to empty.
+        """
+        try:
+            with open(local_file_path, 'wb'):
+                pass
+        except OSError as e:
+            logger.warning(f'Could not empty the local file {local_file_path}. '
+                           f'Got {type(e).__name__}: {e}')
 
     @check_connections
     def read_remote_file(self, remote_file_path: str) -> list:
@@ -289,7 +449,7 @@ class SSHClient(object):
         cluster_soft = servers[self.server]['cluster_soft'].lower()
         for i, status_line in enumerate(stdout):
             if i > i_dict[cluster_soft]:
-                job_id = status_line.split(split_by_dict[cluster_soft])[0]
+                job_id = status_line.lstrip().split(split_by_dict[cluster_soft])[0]
                 job_id = job_id.split('.')[0] if '.' in job_id else job_id
                 running_job_ids.append(job_id)
         return running_job_ids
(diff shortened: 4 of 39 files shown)


arc/job/ssh.py
```diff
@@ -489,6 +540,19 @@ def change_mode(self,
...
+        Remove a directory on the server.
+
+        Args:
+            remote_path (str): The path to the directory to remove on the remote server.
+        """
+        command = f'rm -r "{remote_path}"'
+        _, stderr = self._send_command_to_server(command)
+        if stderr:
+            raise ServerError(
+                f'Cannot remove dir for the given path ({remote_path}).\nGot: {stderr}')
```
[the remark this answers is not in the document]
calvinp0: adopted. also found seven other sites that required this change.

alongd:
Reviewed this at `fa404f125` against `d033cbbf` with an agent fleet (8 passes, incl. mutation testing and a cross-model pass). Everything below was executed unless marked otherwise — repros in a read-only worktree, no jobs submitted, no connections opened.

**First, the credit:** this PR genuinely fixes host-key handling. Base fed the *private key* path to `load_system_host_keys()`, which yields an empty host-key store under `AutoAddPolicy` — ARC had zero host verification, ever. That's a real fix and it was invisible until you touched it. The pooling is correctly built, and 4 of the 5 behaviors I mutation-tested are properly guarded (dropping `key_filename` → 4 failures; swapping policies → 2; disabling dead-connection eviction → 2; inverting the pipe refusal → 9). The findings below are about the edges, not the direction.

---

## The one thing worth fixing first

**`orca_neb.py:229-231` writes a relative path into a template Orca can't resolve.** This is the PR's only chemistry-affecting hunk, and it has zero test coverage.

`adapter.py:442` builds `remote_path` from `servers[server].get('path', '')` — when `'path'` is absent the join collapses to a **relative** path. `'path'` is absent from all three shipped remote server examples in `settings.py` and is documented only under `local`:

```
server1    remote_path='runs/ARC_Projects/proj/spc/job1'                     isabs=False
with_path  remote_path='/storage/group_name/u/runs/ARC_Projects/proj/spc/job1' isabs=True
```

The shipped Orca Slurm template then runs the job somewhere else entirely:

```bash
WorkDir=/state/partition1/user/{un}/$SLURM_JOB_NAME-$SLURM_JOB_ID
cd $WorkDir
cp $SubmitDir/input.in .          # only input.in is copied
```

So `NEB_END_XYZFILE "runs/ARC_Projects/.../product.xyz"` resolves from `/state/partition1/...`, Orca aborts, every NEB TS guess for that reaction fails, and ARC falls through to other TS methods saying nothing about why. Secondary: `.lower()` on line 441 silently rewrites `'path': '/Storage/Users'` → `/storage/users`.

*Suggested fix:* put bare filenames in the NEB template and add `reactant.xyz`/`product.xyz` to the `cp $SubmitDir/...` line. The files are already uploaded into the remote path, so nothing needs rendering.

**And it's unguarded** — I reverted `abs_path` to always use `self.local_path` (pre-PR behavior) and re-ran: **2 passed, 0 failed.** `orca_neb_test.py`'s only `server=` is `'local'`. Worth adding a non-`local` server case asserting `abs_path == self.remote_path`, written against the *corrected* shape.

---

## One root cause behind four findings

**`connect()`'s `while times_tried < 1440` loop catches `Exception` and sleeps 60s**, turning every permanent, deterministic failure into a 24-hour silent stall. Four separate symptoms are this one defect:

1. **`strict_host_key_checking: True` doesn't refuse — it stalls for 24h.** `RejectPolicy` raises a plain `SSHException`, which `_connect` catches and retries, then the outer loop swallows:
   ```
   SSHException:            outer loop result=exhausted, _connect calls=5
   AuthenticationException: outer loop result=exhausted, _connect calls=5
   ```
   Both the `_connect` docstring ("refusing any host that is not already in known_hosts") and `docs/source/remote_submission.rst` ("The connection is refused") describe behavior the code doesn't have.
2. **Wrong key or username → 2880 auth attempts over 24h** against a login node. `AuthenticationException` is an `SSHException` subclass. Not a regression (the base bare `except` had this too), but narrowing the except was the moment to fix it.
3. **A host-key *mismatch* is newly reachable.** Base never loaded real `known_hosts`, so `BadHostKeyException` was structurally impossible; now it lands in the same loop. Note `get_servers_missing_host_keys` (`ssh.py:642`) tests `lookup(address) is None` — it sees *missing* keys and cannot see a *mismatch*, so the startup check gives false comfort for exactly the case that now hard-fails. Concretely: a login alias round-robining to `login1`/`login2` with distinct keys, or any rekeyed cluster. *(read, not executed — 8/10)*
4. **A missing `key` path also stalls 24h.** The shipped `'key': 'path_to_rsa_key'` literal triggers it, and `arc_preflight.check_key` classifies it non-fatal. *Not a regression* — base `ssh.py:373` raised `IOError` on the same input into the same loop. But `ssh.py:387`'s new docstring claims passing the key as `key_filename` "avoids paramiko raising on a configured key path that does not exist on this machine" — it doesn't; only an *unset* key does.

*One fix covers all four:* classify the exception once — re-raise `AuthenticationException`, `BadHostKeyException`, RejectPolicy rejections, and missing-key `OSError`s; retry only transport/banner failures.

---

## Two more with executed repros

**`download_file` silently reversed its contract (`ssh.py:198-206`).** Base always called `sftp.get`, and paramiko 4.0.0 opens `localpath` `"wb"` *before* fetching — so a missing remote file left a zero-byte local file. The base comment said so explicitly ("if the remote path does not exist an empty file will be created at the local path… to avoid screwing up ESS trsh"). The new retry loop `return`s before that:

```
BASE  d033cbbf: local err.txt after a MISSING remote file -> b''                        => cleared
HEAD  fa404f125: local err.txt after a MISSING remote file -> b'slurmstepd: error: *** JOB 12345 CANCELLED DUE TO TIME LIMIT ***'
                                                                                       => STALE CONTENT SURVIVES
```

`adapter.py:892-908` rebuilds `additional_job_info` from whatever `out.txt`/`err.txt` are on disk, and `determine_job_status` scans that string for server-error keywords. A latched stale `err.txt` can mark a healthy job time-limit-errored, or mask a new failure with an old one. This gets sharper once `remove_remote_files` is wired up. Either restore truncate-on-miss, or delete the local target before returning early.

**ARC suppresses its own host-key warning (`ssh.py:428` + `common.py:217` + `main.py:317`).** `WarningPolicy.missing_host_key()` emits via `warnings.warn()`, not `logging`. `initialize_log()` installs `warnings.filterwarnings(action='ignore', module='.*paramiko.*')` two lines before `check_servers_known_hosts()`, long before any connection:

```
--- BEFORE ARC's filter ---   warnings captured: 1 ['Unknown ssh-rsa host key for cluster.exa']
--- AFTER  ARC's filter ---   warnings captured: 0 []
```

On an unknown host ARC connects and prints nothing — not to stdout, not to `arc.log`. The `_connect` docstring's justification ("logs the unknown key… the warning repeats on every connection") is false on every clause. A small `MissingHostKeyPolicy` subclass calling `logger.warning()` with the fingerprint would land the user-visible half of the improvement.

---

<details>
<summary><b>Ten more findings (click to expand)</b> — pool coverage, dead fallback, silent-import fallbacks, pipe guard, style</summary>

- **The pool misses the hot path its docstring claims (`adapter.py:268`, executed).** The actual per-cycle poll is `Scheduler.get_server_job_ids`, still `with SSHClient(server) as ssh:` — full TCP+auth per remote server per label, every 30s tick, plus `:4386`/`:4391` for the `max_simultaneous_jobs` gate. `git diff -- arc/scheduler.py` is empty. Measured: `unpooled SSHClient() constructions across 5 ticks: 300` vs `pooled: 2`. Also unpooled: `trsh.py:1483,1488,1502,1524` (`:1488` never closes), `adapters/cfour.py:305` (overrides `execute_queue`, bypassing the borrow wrapper), `ssh.py:744`, `main.py:832`.
- **`_open_or_borrow_ssh`'s "degrade gracefully" fallback is unreachable dead code (`adapter.py:283-295`, executed).** `borrow` is a `@contextmanager`; calling it runs no body. The factory executes at `__enter__`, inside the *caller's* `with`, outside this `try`. Behavior is unchanged from pre-PR so nothing newly crashes — the problem is the comment asserts a safety net that doesn't exist. *Three independent passes found this separately, each with its own repro.*
- **Idle-drop exposure (`ssh_pool.py:95-110`).** `_is_alive` returns `transport.is_active()` — a local flag, no I/O — and `_default_factory` never calls `set_keepalive`. Pre-PR a connection lived for one operation; now one lives for the whole multi-day run. A half-open connection behind a NAT still reports active, gets handed out, and `check_connections`' `exec_command('ls')` blocks until TCP retransmit timeout — which, single-threaded, stalls all jobs on all servers. Also: `upload_file`/`download_file` are the only `SSHClient` methods without `@check_connections`. Suggest `transport.set_keepalive(30)` and adding `client._sftp.get_channel().active` to `_is_alive`. *(needs a live cluster to observe — 7/10)*
- **Pool teardown is CLI-only (`ARC.py:64-70`).** `reset_default_pool()` fires only in `main()`. Every in-process consumer — T3 driving ARC across RMG iterations, notebooks, `functional/functional_test.py:106,123`, `functional/restart_test.py:54,151,178,201` — leaks the pool across successive `ARC` objects. `check_servers_known_hosts()` went into `ARC.__init__`, which suggests the object is the right seam: `try/finally` in `ARC.execute()`.
- **The dummy-fixture fallback is closed on one path only.** `arc/imports.py:69-71` (untouched) still does `try: import settings / except ImportError: pass`, silently leaving ARC on `server1.host.edu`. The new preflight guards only inside the container, only on the `arc` subcommand branch — the interactive shell and `*)` pass-through bypass it, as does every host-side run. Compounding: `check_servers_known_hosts` skips addresses ending in `PLACEHOLDER_ADDRESS_SUFFIX = '.host.edu'`, so the new startup check is silent on precisely the placeholder case. Raising from `arc/imports.py` when `~/.arc/settings.py` exists but fails to import would make the container guard redundant rather than load-bearing.
- **The preflight checks `settings.py` but not `submit.py` (`arc_preflight.py:135-157`).** `arc/imports.py:86-97` swallows that `ImportError` the same way, falling back to the repo's dummy submit templates. This is a *worse* silent failure than the dummy-server case: jobs run and produce numbers, with the wrong partition/modules/scratch dir. Same for `inputs.py`. The compose file's own comment makes the argument ("a remote run needs this mount as much as it needs the SSH material").
- **The pipe's remote-server guard fails open (`pipe_coordinator.py:106-116`).** The guard resolves `first_server` with `next((s for s in server_list if s in servers_dict), None)`, skipping entries absent from `settings['servers']`; real job resolution takes `ess_settings[adapter][0]` unconditionally. With `ess_settings['gaussian'] = ['zeus', 'local']` and `'zeus'` missing from `servers`, the guard resolves `'local'` and permits the pipe while every real job resolves `'zeus'` — the exact deadlock the guard exists to prevent. It also returns without refusing when `first_server is None`, is case-sensitive where `check_ess_settings` is case-insensitive, and ignores the `args['trsh']['server']` override that `adapters/common.py` honours. Sharing one resolution function between guard and `adapters/common.py` would fix all four.
- **`remove_remote_files` / `remove_dir` added with no callers (`adapter.py:512-520`, `ssh.py:544-555`).** Only callers are tests. An unwired remote `rm -rf` tends to get connected later without re-deriving its interaction with the `download_file` change above. Wire it or drop it from this PR.
- **Docker security posture notes** *(reasoning only — see coverage caveat below)*: `docker-compose.yml:77`'s `${ARC_KNOWN_HOSTS:-/dev/null}` means the default container run has an empty host-key store — the `/dev/null` default is well-reasoned (Docker would otherwise create a root-owned dir at a missing source), the comment just doesn't state the resulting posture. The forwarded agent (`:20,37,61`) can sign anything with every loaded identity against any host for the container's lifetime; worth documenting the blast radius and suggesting `ssh-add -c`. `${ARC_IMAGE:-laxzal/arc:latest}` is a mutable tag in a personal namespace — consider a digest pin. `entrywrapper.sh:230`'s `chown -R mambauser:mambauser /work` contradicts the rule the PR's own new `is_bind_mount()` helper states 190 lines above (pre-existing, but this PR is what makes `/work` default to `${ARC_WORKDIR:-.}`). `entrywrapper.sh:147,177`: `id_has_running_process` runs as PID 1 before anything else starts, so `may_share_id` approves every non-system collision — harmless today, not the guard it reads as. On `entrywrapper.sh:64`: the socket-widening `chmod` is gated behind `ARC_WIDEN_AGENT_SOCKET=1` and loudly warned, which is the right call; one accuracy nit is that the exposure is bounded by the agent socket's 0700 parent dir.
- **Repo style (`CLAUDE.md`).** `adapter.py:245,276,284` — function-local imports; no circularity exists (`python -c 'import arc.job.ssh_pool, arc.job.adapter'` succeeds). `adapter.py:261` — no return annotation, no `Returns:` section. `adapter.py:316,449` — no return annotation. `adapter.py:477` — `_upload_with_ssh(self, ssh)` unannotated.

</details>

---

<details>
<summary><b>Checked and found sound</b> — so nobody re-treads these</summary>

- `Dockerfile:203`'s `chown` after `USER mambauser` does **not** break the build — a same-owner `chown` by the owning non-root user is permitted (probed, exit 0).
- The narrowed `except` leaks no new exceptions. `SSHException` covers `AuthenticationException`/`BadHostKeyException`/`PasswordRequiredException`; `OSError` covers `gaierror`/`NoValidConnectionsError`/`FileNotFoundError`. Only `KeyboardInterrupt`/`SystemExit` newly escape — the intended improvement.
- A `key:` pointing at `known_hosts` still authenticates: paramiko's `_auth` catches per key class, then falls through to the agent and `look_for_keys`. Same identity, still works. (Worth a release note anyway — this PR removes the `'key': '.../known_hosts'` line from the commented `pharos` example at base `settings.py:31`, so that population exists.)
- Pool races aren't reachable today. `grep -rn "threading\|multiprocessing\|concurrent.futures\|asyncio\|Thread("` over non-test `arc/` returns only `ssh_pool.py`'s own docstring; pipe mode parallelises via `subprocess.Popen` in separate processes. Design gaps for a future concurrent scheduler, not present defects.
- Credentials can't cross between pooled servers — the pool keys on server name, and the only runtime write into `servers` anywhere in `arc/` is `servers[server]['memory'] = 64` at `trsh.py:927`.
- Stale pooled connections are reaped (paramiko sets `_transport = None` on close, so `_is_alive` is False). No channel exhaustion from `check_connections`' unclosed `exec_command('ls')` — `Channel._handle_close` removes it from `Transport._channels`.
- Loading `known_hosts` doesn't break key-type negotiation — paramiko `client.py:229-232` only *reorders* `sec_opts.key_types`.
- `orca_neb`'s `abs_path` can't be `None` (`adapters/common.py:268` calls `set_file_paths()` before `:278` calls `set_files()`).
- `borrow()` lacking `try/finally` around its `yield` is correct — the pool deliberately retains the client, and `ssh_pool_test.py:122-126` covers propagation.
- No silent fallback to local execution: `should_use_pipe` → `False` propagates to `PipePlanner._try_pipe_job` → empty set → `legacy_queue_execution`, which for `server != 'local'` still goes over SSH.
- Restart/resume unaffected; `restart.yml` semantics unchanged. No second transport exists (no subprocess `ssh`/`scp`/`rsync`, no bare `Transport`, no `ControlMaster` in the submit templates) — the "no existing pooling" half of the reuse check stands.
- The rebase is clean: `d033cbbf` *is* the tip of `main`. The three nearest upstream commits touching these files were checked anyway — no overlapping hunk.

</details>

---

**Where this review is weak, so you can weight it accordingly:** the Docker half has **zero executed coverage**. `test_docker_smoke.py` gives 18 failed / 2 passed / 6 skipped on a bare host (`micromamba: command not found`), so `entrywrapper.sh` (+232 lines, runs as entrypoint with the mounted keys and forwarded agent) and `arc_preflight.py` (181 new lines) were reasoned about, not run — settling those needs a real `docker build` plus a container run. The idle-drop finding needs a live cluster. Also unreconciled: the PR reports 249 tests passed, I measured 275 passed + 15 subtests in the same stated scope — probably a collection-scope difference, but worth a glance. Separately and **not caused by this PR**: `adapter_test.py` is order-dependent under `-n auto` (2 failures, reproduced identically on `d033cbbf`), so parallel CI runs of that file are unreliable.

Happy to open a PR against your branch for the NEB path + its test if that's useful.

calvinp0:
> Reviewed this at `fa404f125` against `d033cbbf` with an agent fleet (8 passes, incl. mutation testing and a cross-model pass). Everything below was executed unless marked otherwise — repros in a read-only worktree, no jobs submitted, no connections opened.
> 
> **First, the credit:** this PR genuinely fixes host-key handling. Base fed the _private key_ path to `load_system_host_keys()`, which yields an empty host-key store under `AutoAddPolicy` — ARC had zero host verification, ever. That's a real fix and it was invisible until you touched it. The pooling is correctly built, and 4 of the 5 behaviors I mutation-tested are properly guarded (dropping `key_filename` → 4 failures; swapping policies → 2; disabling dead-connection eviction → 2; inverting the pipe refusal → 9). The findings below are about the edges, not the direction.
> ## The one thing worth fixing first
> 
> **`orca_neb.py:229-231` writes a relative path into a template Orca can't resolve.** This is the PR's only chemistry-affecting hunk, and it has zero test coverage.
> 
> `adapter.py:442` builds `remote_path` from `servers[server].get('path', '')` — when `'path'` is absent the join collapses to a **relative** path. `'path'` is absent from all three shipped remote server examples in `settings.py` and is documented only under `local`:
> 
> ```
> server1    remote_path='runs/ARC_Projects/proj/spc/job1'                     isabs=False
> with_path  remote_path='/storage/group_name/u/runs/ARC_Projects/proj/spc/job1' isabs=True
> ```
> 
> The shipped Orca Slurm template then runs the job somewhere else entirely:
> 
> ```shell
> WorkDir=/state/partition1/user/{un}/$SLURM_JOB_NAME-$SLURM_JOB_ID
> cd $WorkDir
> cp $SubmitDir/input.in .          # only input.in is copied
> ```
> 
> So `NEB_END_XYZFILE "runs/ARC_Projects/.../product.xyz"` resolves from `/state/partition1/...`, Orca aborts, every NEB TS guess for that reaction fails, and ARC falls through to other TS methods saying nothing about why. Secondary: `.lower()` on line 441 silently rewrites `'path': '/Storage/Users'` → `/storage/users`.
> 
> _Suggested fix:_ put bare filenames in the NEB template and add `reactant.xyz`/`product.xyz` to the `cp $SubmitDir/...` line. The files are already uploaded into the remote path, so nothing needs rendering.
> 
> **And it's unguarded** — I reverted `abs_path` to always use `self.local_path` (pre-PR behavior) and re-ran: **2 passed, 0 failed.** `orca_neb_test.py`'s only `server=` is `'local'`. Worth adding a non-`local` server case asserting `abs_path == self.remote_path`, written against the _corrected_ shape.
> ## One root cause behind four findings
> 
> **`connect()`'s `while times_tried < 1440` loop catches `Exception` and sleeps 60s**, turning every permanent, deterministic failure into a 24-hour silent stall. Four separate symptoms are this one defect:
> 
>     1. **`strict_host_key_checking: True` doesn't refuse — it stalls for 24h.** `RejectPolicy` raises a plain `SSHException`, which `_connect` catches and retries, then the outer loop swallows:
>        ```
>        SSHException:            outer loop result=exhausted, _connect calls=5
>        AuthenticationException: outer loop result=exhausted, _connect calls=5
>        ```
>        
>        
>            
>              
>            
>        
>              
>            
>        
>            
>          
>        Both the `_connect` docstring ("refusing any host that is not already in known_hosts") and `docs/source/remote_submission.rst` ("The connection is refused") describe behavior the code doesn't have.
> 
>     2. **Wrong key or username → 2880 auth attempts over 24h** against a login node. `AuthenticationException` is an `SSHException` subclass. Not a regression (the base bare `except` had this too), but narrowing the except was the moment to fix it.
> 
>     3. **A host-key _mismatch_ is newly reachable.** Base never loaded real `known_hosts`, so `BadHostKeyException` was structurally impossible; now it lands in the same loop. Note `get_servers_missing_host_keys` (`ssh.py:642`) tests `lookup(address) is None` — it sees _missing_ keys and cannot see a _mismatch_, so the startup check gives false comfort for exactly the case that now hard-fails. Concretely: a login alias round-robining to `login1`/`login2` with distinct keys, or any rekeyed cluster. _(read, not executed — 8/10)_
> 
>     4. **A missing `key` path also stalls 24h.** The shipped `'key': 'path_to_rsa_key'` literal triggers it, and `arc_preflight.check_key` classifies it non-fatal. _Not a regression_ — base `ssh.py:373` raised `IOError` on the same input into the same loop. But `ssh.py:387`'s new docstring claims passing the key as `key_filename` "avoids paramiko raising on a configured key path that does not exist on this machine" — it doesn't; only an _unset_ key does.
> 
> 
> _One fix covers all four:_ classify the exception once — re-raise `AuthenticationException`, `BadHostKeyException`, RejectPolicy rejections, and missing-key `OSError`s; retry only transport/banner failures.
> ## Two more with executed repros
> 
> **`download_file` silently reversed its contract (`ssh.py:198-206`).** Base always called `sftp.get`, and paramiko 4.0.0 opens `localpath` `"wb"` _before_ fetching — so a missing remote file left a zero-byte local file. The base comment said so explicitly ("if the remote path does not exist an empty file will be created at the local path… to avoid screwing up ESS trsh"). The new retry loop `return`s before that:
> 
> ```
> BASE  d033cbbf: local err.txt after a MISSING remote file -> b''                        => cleared
> HEAD  fa404f125: local err.txt after a MISSING remote file -> b'slurmstepd: error: *** JOB 12345 CANCELLED DUE TO TIME LIMIT ***'
>                                                                                        => STALE CONTENT SURVIVES
> ```
> 
> `adapter.py:892-908` rebuilds `additional_job_info` from whatever `out.txt`/`err.txt` are on disk, and `determine_job_status` scans that string for server-error keywords. A latched stale `err.txt` can mark a healthy job time-limit-errored, or mask a new failure with an old one. This gets sharper once `remove_remote_files` is wired up. Either restore truncate-on-miss, or delete the local target before returning early.
> 
> **ARC suppresses its own host-key warning (`ssh.py:428` + `common.py:217` + `main.py:317`).** `WarningPolicy.missing_host_key()` emits via `warnings.warn()`, not `logging`. `initialize_log()` installs `warnings.filterwarnings(action='ignore', module='.*paramiko.*')` two lines before `check_servers_known_hosts()`, long before any connection:
> 
> ```
> --- BEFORE ARC's filter ---   warnings captured: 1 ['Unknown ssh-rsa host key for cluster.exa']
> --- AFTER  ARC's filter ---   warnings captured: 0 []
> ```
> 
> On an unknown host ARC connects and prints nothing — not to stdout, not to `arc.log`. The `_connect` docstring's justification ("logs the unknown key… the warning repeats on every connection") is false on every clause. A small `MissingHostKeyPolicy` subclass calling `logger.warning()` with the fingerprint would land the user-visible half of the improvement.
> Ten more findings (click to expand) — pool coverage, dead fallback, silent-import fallbacks, pipe guard, style
> 
> 
> Checked and found sound — so nobody re-treads these
> 
> 
> **Where this review is weak, so you can weight it accordingly:** the Docker half has **zero executed coverage**. `test_docker_smoke.py` gives 18 failed / 2 passed / 6 skipped on a bare host (`micromamba: command not found`), so `entrywrapper.sh` (+232 lines, runs as entrypoint with the mounted keys and forwarded agent) and `arc_preflight.py` (181 new lines) were reasoned about, not run — settling those needs a real `docker build` plus a container run. The idle-drop finding needs a live cluster. Also unreconciled: the PR reports 249 tests passed, I measured 275 passed + 15 subtests in the same stated scope — probably a collection-scope difference, but worth a glance. Separately and **not caused by this PR**: `adapter_test.py` is order-dependent under `-n auto` (2 failures, reproduced identically on `d033cbbf`), so parallel CI runs of that file are unreliable.
> 
> Happy to open a PR against your branch for the NEB path + its test if that's useful.

`download_file` no longer leaves a stale local file behind: a missing remote file now empties the local path, exactly reproducing what main did (paramiko's get opens the local path `"wb"` before discovering the miss), and the miss is logged at warning level.

Orca NEB refuses a non-absolute remote path rather than emitting a deck Orca will fail on.

Permanent auth/host-key failures are classified and raised on the first attempt instead of feeding the 24-hour retry.

The per-connection host-key report now goes through ARC's logger, since initialize_log filters paramiko's warnings.warn. The pool fallback is rebuilt on `ExitStack` so it covers the lease and not the caller's body.

Regarding the `.lower()` comment - it is gone Ithis moves files for anyone whose configured `path` has uppercase) and a server with no `path` is reported once per run. I deliberately did not invent an absolute default :`~` is quoted by `shlex.quote` and is not expanded by  SFTP, so it would create a directory literall named `~`, and resolving the real home needs a connection ARC does not have at job-construction time.

Also, in the docs ` remote_submission.rst:140-144` states the 24-hour consequence verbatim. And the `_connect` dosctring clause read as wrong is about `key=None`. `adapter.py:363-365` is identical on `main`, so the defect is pre-exisiting. Fixed here now.

alongd:
Re-verified all 18 findings against the current head `db2a20a40` (the branch was rebased since I reviewed `fa404f125`, so nothing was taken on trust). **16 of 18 are fixed, and each fix is guarded by a test that fails when I revert it** — I mutation-checked the NEB path, the `PERMANENT_CONNECTION_ERRORS` classification, `download_file`'s truncate-on-miss, the host-key policy, the `ExitStack` fallback, and the `execute()` teardown. The pooling now covers the hot path: 5 ticks × 10 labels × 2 remote servers = 100 borrows → **2** connections, with exactly +1 reopen on a dead transport. Thank you — the `resolve_job_server()` extraction and the `arc/imports.py` overlay reporting are both better than what I suggested.

One thing has to change before this merges, and it's a consequence of the NEB fix rather than something the review anticipated.

---

### `orca_neb`'s new `ValueError` aborts the run, where the old bug lost one TS guess

`_get_abs_path()` (`orca_neb.py:259-277`) raises inside the **constructor**, and nothing on the path catches it:

```
_get_abs_path()  →  raises ValueError
  ← set_files()                     (orca_neb.py:279)
  ← _initialize_adapter             (adapters/common.py:275)
  ← job_factory
  ← Scheduler.run_job               (scheduler.py:1053)
  ← Scheduler.spawn_ts_jobs         (scheduler.py:1873)  ← bare call, no try
```

Executed against this head:

```
CONSTRUCTOR RAISED: Cannot write an Orca NEB input file for server "no_path": the remote path is
'runs/ARC_Projects/test_orca_neb/i-C3H7__=__n-C3H7_TS/tsg_a1', which is not absolut...
```

`orca_neb` is in the default `ts_adapters` (`settings.py:119`) and in `ts_adapters_by_rmg_family` for many families, and `'path'` is absent from all three shipped remote server examples — it is documented only under `local` and appears commented-out in the configs I can see. So the common configuration goes from "one TS guess writes a deck Orca fails on" to "the whole ARC run dies at TS spawn." That inverts the severity of the thing the fix was for.

PR #996 doesn't cover this either — it is not in `official/main`, and it catches `DependencyError` only.

Either shape works, your call: raise something `spawn_ts_jobs` already tolerates, or widen the `except` at the spawn site so the refusal costs one TS method. A test asserting the run survives a `path`-less remote server would pin it.

The refusal itself is the right design — relying on `$SubmitDir` being visible from the compute node is true on any shared-home Slurm cluster, and it avoids touching `submit.py`. It just needs to fail soft.

---

### Two one-liners worth taking while you're in these files

**A missing `key` path still stalls 24 h.** paramiko's `_auth` catches only `SSHException` around `_key_from_filepath`, so a non-existent `key_filename` raises `FileNotFoundError` → an `OSError` → not in `PERMANENT_CONNECTION_ERRORS`. Executed at this head with `connection_attempts=5`:

```
1. _key_from_filepath -> FileNotFoundError ... is SSHException? False | is OSError? True
2. connect() with 5 attempts: _connect calls=5 sleeps=4  => RETRIED (not classified permanent)
3. client.key = 'path_to_rsa_key'
```

At the default 1440 that's the original 24-hour stall, and the trigger still ships — `settings.py:55,62,70` all carry the `'key': 'path_to_rsa_key'` literal, and `arc_preflight.py:73-92` classifies it non-fatal (and only inside the container). The docstring half *is* correctly fixed at `ssh.py:583-587`. Suggest classifying a missing-key `OSError` as permanent.

**The keepalive is lost on the first reconnect.** `_default_factory` sets `transport.set_keepalive(30)` (`ssh_pool.py:60-87`), but `SSHClient.connect()` (`ssh.py:503-548`) does not, and `check_connections` reconnects the *same* client object in place. Executed: after the factory sets the keepalive on transport 0, a half-open socket drives `check_connections` → `connect()` → transport 1, on which `set_keepalive` is **never called**. The pooled client silently loses idle-drop protection for the rest of the run — which is exactly the multi-day scenario pooling introduces. One line at the end of `connect()`; nothing tests it today.

---

<details>
<summary><b>Verified fixed</b> — so nobody re-treads these</summary>

- **NEB path + `.lower()` + coverage.** `.lower()` gone from `adapter.py:439`, path used verbatim; `test_a_configured_path_is_used_verbatim` guards it. Reverting `_get_abs_path` to `self.local_path` → 2 failed; dropping only the `isabs` guard → 1 failed.
- **connect() classification (3 of 4).** `PERMANENT_CONNECTION_ERRORS` (`ssh.py:138-141`) + raise on first attempt (`:532-536`, `:638-639`). `RejectUnknownHostKeyPolicy` raising its own `UnknownHostKeyError` is what makes the strict-mode refusal distinguishable — I confirmed paramiko `client.py:456-458` calls the policy unguarded, so it really escapes. `_host_key_mismatch_error` names both fingerprints; `get_servers_with_conflicting_host_keys` adds the offline-checkable half and the old function now states its own limitation. Mutating the tuple to `(RuntimeError,)` → 4 failures.
- **`download_file`.** A pre-seeded stale `err.txt` comes back `b''` with `sftp.get` never called. 7 tests at `ssh_test.py:774-853`; no-op'ing `_empty_local_file` fails 2.
- **Host-key warning.** With `warnings.filterwarnings(action='ignore', module='.*paramiko.*')` installed exactly as `common.py:217` does: 0 warnings captured, 1 `arc` WARNING record with hostname and `SHA256:` fingerprint. Reverting to `warnings.warn` fails 3.
- **Pool hot path.** `scheduler.py:3562`, `trsh.py:1484/1489/1503/1525` (the `:1488` leak is now a `with`), `cfour.py:304`, `main.py:854` all pooled. Remaining non-test `SSHClient(`: 4, of which 2 are the pool itself and `delete_check_files_on_servers` is deliberately unpooled with a rationale.
- **`ExitStack` fallback is now reachable.** Both arms executed: a factory raising `ServerError`, and a pool whose `borrow` raises `RuntimeError` → one-shot client, closed on exit. A `ValueError` from the body still propagates untouched. Stripping it fails 5.
- **Teardown.** `main.py:561-576` `try/finally`, `ARC.py` no longer duplicates the seam; 3 tests cover finish/raise/interrupt.
- **Overlay reporting.** `_report_unusable_overlay` on settings (`:115`), submit (`:132/136/140`), inputs (`:156`), deduped and queued so it survives `initialize_log`. Executed with three broken overlays: all three warn, all three reach `arc.log`.
- **Pipe guard, all four sub-defects.** Executed against `resolve_job_server()`: the `['zeus','local']`-with-`zeus`-absent fail-open case now refuses; `None` refuses; casefolding matches `check_ess_settings`; the `trsh` override is honoured both directions.
- **`remove_dir` quoting** (Copilot's thread). `rm -rf -- {shlex.quote(...)}`; a path containing `' ; rm -rf ~` round-trips as one token. Eight quoted sites in `ssh.py`; `ssh_test.py:532-592` asserts both.

Suites `ssh_test`, `ssh_pool_test`, `adapter_test`, `imports_test`, `main_test`, `trsh_test`, `scheduler_test`, `cfour_test`, `pipe_coordinator_test`, `common_test`, `orca_neb_test`: 268 passed, stable across 3 runs.

</details>

<details>
<summary><b>Not fixed, and fine by me</b></summary>

- **All five Docker posture notes.** `docker-compose.yml`, `entrywrapper.sh` and `arc_preflight.py` are byte-identical to what I reviewed. These were reasoning-only findings with no executed coverage on my side, and they're documentation/hardening rather than defects — I'm not holding the PR for them, but the `${ARC_KNOWN_HOSTS:-/dev/null}` default posture and the forwarded-agent blast radius are each worth one sentence in `docker.rst` eventually.
- **`remove_remote_files` / `remove_dir` still have no callers.** The docstrings now say so outright, which removes the trap for the next reader. Staged rather than forgotten — good enough.
- **Style residue in `ssh_pool.py`** — `__init__`/`_close_quietly` without docstrings, missing `Args:`/`Returns:` at `:109,143,175,183,199`, double-quoted strings and `%s` logging where ARC uses f-strings, and multi-line docstrings opening on the `"""` line. The `adapter.py` sites you touched are all clean now, and `borrow_ssh_client` is in ARC style — the file is just internally inconsistent. Cosmetic; fold it in whenever.

</details>

<details>
<summary><b>Drive-by, pre-existing on <code>main</code></b></summary>

`trsh.py` has `logger.denug(...)` in the unknown-cluster-software branch of `trsh_job_on_server` (blames to `7ef79c8bc`) — an `AttributeError` waiting on any non-OGE/Slurm server. Sitting in a file this PR edits; free to fix here or separately.

</details>

---

Fix the NEB abort and I'll approve. CI is green on all nine checks and the branch is mergeable — this is the only thing standing between it and `main`.

calvinp0:
**NEB abort** — fixed, though not the way you suggested. Both options fail silently mid-run: the user loses a TS method and only finds out from a log line, and widening the `except` at the spawn site would swallow unrelated constructor failures too. So it's caught at startup instead. `check_ess_settings` already validates adapter and server names and raises `SettingsError` from `main.py:398` before anything spawns, so the check went there — a server missing `path` gets reported once, naming the server and the key. `_get_abs_path()` keeps its refusal as a backstop, now raising `SettingsError` rather than `ValueError` so the two read as one settings problem.

One thing that isn't visible in the diff and shaped the design: `OrcaNEBAdapter` subclasses `OrcaAdapter`, and `orca.py:205` sets `job_adapter = 'orca'` before `_initialize_adapter` at `:213`, which is where `set_files()` calls `_get_abs_path()`. A NEB job therefore resolves its server from `ess_settings['orca']`, so a check keyed on `'orca_neb'` would never have matched a real config.

**The root cause was the templates.** No remote server block in the docs defines `path` — nine examples, zero occurrences. `server1/2/3` in `settings.py` and all nine docs examples now set it. That fixes new configurations; the validation only covers ones already written, since `~/.arc/settings.py` replaces the whole `servers` dict and never picks up template changes.

**Where I might be wrong on this.** If you think fixing the templates is enough, the validation is arguably insurance for a population neither of us can measure — I don't know how many existing overlays lack `path`, and mine doesn't. I kept it because a stale overlay still kills the run otherwise, which is the regression you flagged, but it's ~20 lines and one call site and I'd drop it if you disagree. Same question on the backstop: I made it a hard raise, since validation covers every server in `ess_settings['orca']` and the ordinary path can't reach it. What's left is a `trsh` server override, direct in-process adapter use, and `scale.py:103` calling `check_ess_settings` with one argument. If you think one of those should degrade instead, say which.

**Missing key** — a `FileNotFoundError` naming the configured key file is now classified permanent. Other `OSError`s stay retryable; `ConnectionRefusedError` and a `FileNotFoundError` for a different file are both tested at 5 attempts / 4 sleeps.

**Keepalive** — the client records the interval it was given and re-applies it to each new transport, so `check_connections` reconnecting in place no longer drops it. I avoided importing `ssh_pool` from `ssh`; `ssh_pool.set_keepalive` stays the single definition.

**Both pre-existing ones are fixed too.** `logger.denug` → `debug`. And in the same function, the `break` was sitting at the `for` level rather than inside the `if`, so the loop always exited on the first line and the `for...else` never ran — a submit file with no node directive got uploaded unchanged, and one with a directive further down never had it replaced. Two tests now pin both branches. `orca_neb_test.py` also moved off `ARC_TESTING_PATH` onto `tempfile.mkdtemp()`.

**Two corrections.** `spawn_ts_jobs` doesn't tolerate anything — there's no `try` on `spawn_ts_jobs` → `run_job` → `job_factory` → `_initialize_adapter`, and `scheduler.py`'s eight `except` clauses are all elsewhere, so option one would have needed a new exception type and a new catch. And `test_a_configured_path_is_used_verbatim` doesn't exist; the guard is `test_task_1b_a_remote_server_resolves_the_endpoint_files_on_the_server`.

350 passed. Each fix reverted individually with the guarding test confirmed failing. Rebased onto current `main`.

arc/job/adapters/cfour.py
```diff
@@ -302,7 +301,7 @@ def execute_queue(self):
         """
         self._log_job_execution()
         if self.server != 'local':
-            with SSHClient(self.server) as ssh:
+            with self._open_or_borrow_ssh() as ssh:
```
alongd, line 304: I couldn't find where `_open_or_borrow_ssh()` is defined. could you explain?

arc/job/adapters/cfour_test.py
```diff
@@ -9,6 +9,8 @@
 import os
 import shutil
 import unittest
+from types import SimpleNamespace
```
alongd, line 12: can you explain what `SimpleNamespace`'s function here is?

arc/job/adapters/common.py
```diff
@@ -281,6 +277,40 @@ def _initialize_adapter(obj: JobAdapter,
     check_argument_consistency(obj)
 
 
+def resolve_job_server(ess_settings: dict,
+                       job_adapter: str,
+                       args: dict | None = None,
+                       ) -> str | None:
+    """
+    Determine the server a job for ``job_adapter`` will be sent to.
```
alongd: is this properly phrased? am I missreading? `Determine the server a job for ``job_adapter`` will be sent to.` maybe phrase better?

arc/job/ssh.py
```diff
@@ -27,6 +30,116 @@
     settings['check_status_command'], settings['delete_command'], settings['list_available_nodes_command'], \
     settings['servers'], settings['submit_command'], settings['submit_filenames']
 
+KNOWN_HOSTS_PATH = '~/.ssh/known_hosts'
+PLACEHOLDER_ADDRESS_SUFFIX = '.host.edu'
+PLACEHOLDER_USERNAME = '<username>'
+
+
+class UnknownHostKeyError(ServerError, paramiko.SSHException):
```
alongd, line 38: I suggest moving all of these to `/arc/exceptions.py`

arc/job/ssh_pool.py
```diff
@@ -0,0 +1,289 @@
...
+
+from arc.common import get_logger
+from arc.job.ssh import SSHClient
+
+logger = get_logger()
+
+
+SSHClientFactory = Callable[[str], SSHClient]
+
+KEEPALIVE_INTERVAL = 30
```
alongd: is this is `s`? add a comment

alongd (commented):
Thanks! please see some minor comments

arc/job/adapters/cfour.py
calvinp0: It's inherited rather than defined here — `JobAdapter._open_or_borrow_ssh()` at `arc/job/adapter.py:264`. It returns the pooled client for the server when there is one and falls back to a one-shot `SSHClient` otherwise, so the caller gets a context manager either way without needing to know which it got. `CFourAdapter` reaches for it explicitly because it overrides `execute_queue()` instead of using the base implementation, which is why it was the one adapter still opening its own connection.

arc/job/adapters/cfour_test.py
calvinp0: It stands in for the adapter, carrying only the five attributes `execute_queue()` actually reads, with `_open_or_borrow_ssh` mocked. Building a real `CFourAdapter` would run the full adapter initialization — species, level, project directories — for a test about which connection the submission uses.

Listing the attributes also pins what `execute_queue()` is allowed to touch: if it starts reading something else the test fails, where a `MagicMock` would silently answer to anything. I've expanded the docstring to say that.

arc/job/adapters/common.py
calvinp0: You're reading it right, it was clumsy. Now reads: `Return the server that a ``job_adapter`` job will be submitted to.`

arc/job/ssh.py
calvinp0: I'd rather keep it here, and the reason is import weight. `arc/exceptions.py` currently imports nothing at all, and 52 modules import from it. `UnknownHostKeyError` subclasses `paramiko.SSHException` deliberately — that's what lets a refused host key be told apart from a transport failure without matching on paramiko's message text — so moving it would pull paramiko into all 52. It appears in exactly two files today, both under `arc/job/`.

If you'd still like it central, the alternative is dropping the paramiko base and keeping it a plain `ServerError`, but then strict-mode refusal is only distinguishable by message text. Happy to do that if you prefer it.

arc/job/ssh_pool.py
calvinp0: Seconds, and it is used — `_default_factory` passes it at `:62`, and it's the default for `set_keepalive(client, interval)` at `:66`, which reaches `transport.set_keepalive(interval)` at `:98`. `SSHClient` now records the interval it was given so `connect()` can re-apply it after a reconnect, which is the other consumer.

Renamed to `KEEPALIVE_INTERVAL_SECONDS` so the unit is visible at each use site, with a docstring under it noting that long-lived pooled connections are otherwise dropped by an idle timeout on the server or a firewall between it and the client.

alongd (approved):
Thanks!

PR #1008, opened 2026-08-22
Give each test class its own scratch directory instead of writing into arc/testing

calvinp0:
Test classes were writing scratch into `ARC_TESTING_PATH`, which holds read-only fixture **inputs**. Nineteen files did it, and eight of them `rmtree`'d a subdirectory of it in a teardown — so one module's cleanup deleted a directory another module was still writing into.

The result is a suite that is broadly unstable under xdist. Measured on `main`, each module run **solo** under `-n 4 --dist worksteal`, three runs each:

| module | `main` | this branch |
|---|---|---|
| `orca_test` | 2F / 2F / 2F | 3 × 20 passed |
| `xtb_test` | 6E / 1F+4E / 6E | 3 × 8 passed |
| `adapter_test` | 1F+2E / 2F / 2F | 3 × 16 passed |
| `molpro_test` | 4E / fail / 3E | 3 × 5 passed |
| `obabel_test` | 1F+1E / fail / 2F | 3 × 4 passed |
| `mockter_test` | 4E / 1F+2E / 2E | 3 × 6 passed |
| `cfour_test` | fail / 1F+1E / 1F | 3 × 4 passed |
| `arkane_test` | 1F / pass / pass | 3 × 46 passed |

Paired (7 job-adapter modules together, 5 runs): `5 failed, 58 passed` × 5 → **`63 passed` × 5**.

**Solo failure is the important column.** `--dist worksteal` can split a single `TestCase` class across workers, so each worker independently runs `setUpClass` *and* `tearDownClass` — a class does not own its directory merely because only that class uses it. A per-worker path suffix does not fix this; `tempfile.mkdtemp()` does, because it is unique per `setUpClass` invocation.

## What changed

Every converted class allocates scratch with `tempfile.mkdtemp()` in `setUpClass`/`setUp` and releases it via `addClassCleanup`/`addCleanup`. All `tearDownClass` rmtrees of `ARC_TESTING_PATH` subdirectories are gone. **Read-only fixture reads are untouched** — those are shared safely, because nobody writes to them.

Eight of the nineteen were not findable by grepping for `rmtree(os.path.join(ARC_TESTING_PATH,…))` because they build the path through a variable: `pyscf_test`, `ase_test`, `uma_test`, `ts/orca_neb_test`, `ts/kinbot_test`, `ts/gcn_test`, `processor_test`, `common_test`, `reaction_test`.

## Cases needing more than a mechanical rewrite

- **`common_test.test_globalize_paths`** asserted on a literal repo path. The fixture is now `copytree`'d into a tmpdir and both assertions rebuilt from `project_directory` — same semantics, and it allowed deleting `_clean_globalized_restart_artifact()`, a self-healing helper that existed only because the test wrote `restart_paths_globalized.yml` back into the repo on every run.
- **`reaction_test.test_load_ts_xyz_user_guess_from_files`** passed a fixture directory as an ARC `project_directory`, so a Scheduler run created `arc.log`, `output/` and `log_and_restart_archive/` inside `arc/testing/reactions/methanoate_hydrolysis`. Its `ts_xyz_guess` entries resolve relative to `project_directory`, so the fixture is copied and the run happens in the copy.
- **`kinbot_test`** had a hand-rolled `_remove_test_dir` written specifically to survive xdist; a per-class tmp root makes it dead code.

## Not touched

`arc/job/adapters/gaussian_test.py` and `arc/job/adapters/common_test.py` are #1001's files and are converted there. They remain the one genuinely cross-module shared directory on `main`, and Group C measurements confirm that pair still races here — unchanged, as expected.

## Verification

Full suite, serial: `main` **5 failed, 2803 passed, 36 skipped** (386 s) → branch **5 failed, 2803 passed, 36 skipped** (413 s). Identical counts; the 5 are the known `torch_ani_test` environmental failures. `git status --porcelain` clean after both runs — no `arc/testing/` debris left behind.

Mutation-verified: reverting `orca_test` to `main`'s version brings the failures straight back (`1F+15P+4E`, `2F+18P`, `1F+19P`), restoring it gives `20 passed` × 3; same for `xtb_test`. Restored files md5-matched their snapshots.

Four commits, no file in more than one. Two pre-existing `py/import-and-import-from` violations fixed in files already being touched (`gcn_test`, `kinbot_test`).

### DIFF
--- a/arc/checks/nmd_test.py
+++ b/arc/checks/nmd_test.py
@@ -8,6 +8,7 @@ This module contains unit tests for the arc.checks.nmd module
 import unittest
 import os
 import shutil
+import tempfile
 
 import numpy as np
 
@@ -32,12 +33,14 @@ class TestNMD(unittest.TestCase):
         A method that is run before all unit tests in this class.
         """
         cls.maxDiff = None
+        cls.scratch_dir = tempfile.mkdtemp(prefix='arc_test_nmd_')
+        cls.addClassCleanup(shutil.rmtree, cls.scratch_dir, ignore_errors=True)
         cls.generic_job = job_factory(job_adapter='gaussian',
                                       species=[ARCSpecies(label='SPC', smiles='C')],
                                       job_type='composite',
                                       level=Level(method='CBS-QB3'),
                                       project='test_project',
-                                      project_directory=os.path.join(ARC_PATH, 'Projects', 'tmp_nmd_project'),
+                                      project_directory=os.path.join(cls.scratch_dir, 'tmp_nmd_project'),
                                       )
         cls.xyz_1 = {'symbols': ('C', 'N', 'H', 'H', 'H', 'H'),
                      'isotopes': (13, 14, 1, 1, 1, 1),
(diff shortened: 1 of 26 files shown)


calvinp0:
Superseded by #1028, which rebases this onto current `main` and combines both isolation branches into one. Measured there: random-order failure counts across eight seeds drop from 8-12 to 5-6, against a floor of 5 environmental `torch_ani_test` failures.

Closing in favour of #1028 rather than rebasing this one separately, since the two overlap in intent and `main` has moved past both.

PR #1009, opened 2026-08-22
Give each TestScheduler test its own scheduler and project directory

calvinp0:
`TestScheduler` does not pass reliably on `main`. Measured over 12 runs of `pytest arc/scheduler_test.py::TestScheduler -n 4 --dist worksteal`: **0 clean, 12 failing**. The whole file failed 5 of 5.

The failures rotate, which is why this reads as noise rather than a bug: `test_check_rxn_e0_by_spc` (7 of 12), `test_initialize_output_dict` (2), `test_conformers` (1), plus `ERROR` cascades of up to 11 tests at once.

## The cleanest demonstration

```
$ pytest arc/scheduler_test.py::TestScheduler::test_initialize_output_dict     # on main
1 failed        AssertionError: False is not true
```

That test only passes when something else runs first. Its opening line asserted that the shared `output` dict already contained information — true only because an earlier test had dirtied it, and testing nothing about the method under test. It now writes that state itself.

## Two causes, not one

**1. A shared project directory deleted by `tearDownClass`.** Every fixture used `ARC_PATH/Projects/arc_project_for_testing_delete_after_usage3`, and `tearDownClass` `rmtree`'d it. `--dist worksteal` can split a single `TestCase` class across workers, so each worker independently runs `setUpClass` *and* `tearDownClass` — worker A's teardown deletes the tree worker B is still writing into. This is the whole `ERROR` cascade, and it explains `test_check_rxn_e0_by_spc`: PIL writing `ts_guesses.png` into a directory that had just vanished.

**2. Tests mutating shared class state.** Twelve of them, including `test_non_rotor` permanently flipping `sched1.job_types['rotors']`, `test_add_label_to_unique_species_labels` appending to a list after asserting its exact initial value, and `test_troubleshoot_ess_under_max_attempts` reading what `test_troubleshoot_ess_max_attempts` wrote.

## The change

`setUpClass` → `setUp`, plus a `make_project_directory()` helper giving each test a `tempfile.mkdtemp` under `ARC_PATH/Projects` released by `addCleanup`. `tearDownClass` deleted.

Construction cost was measured before choosing: `setUpClass` takes **62 ms**, so per-test construction is cheap and the simple answer was the right one — no need to split read-only from mutating tests.

## Verification

| | |
|---|---|
| `TestScheduler`, `-n 4 --dist worksteal` | **22/22 pass** (was 0/12) |
| solo `-n 4` | 6/6 |
| whole file, `-n 4 --dist worksteal` | 5/5 |
| serial test count | 45 before, 45 after |
| serial runtime | 4.98 s → 6.70 s (+35 %) |
| `-n 4` runtime | ~4.0 s → ~3.9 s (unchanged) |

Mutation-verified: restoring the shared fixed-name project directory and its `tearDownClass` brings the failures back in 3 of 10 runs. Re-sharing the scheduler object alone did not reproduce within 10 runs, because xdist dispatches in collection order so the polluting test usually still precedes its victim on the same worker — the order dependency is instead proven directly by the solo-run result above.

Also fixed: `test_save_e_elect` cleaned up with a trailing `rmtree` that leaked its directory on any failure; now `addCleanup`.

One commit, one file.

## Relationship to other PRs

Same defect class as #1008, in a different file — #1008 converts 19 test modules writing scratch into `ARC_TESTING_PATH`. This one is `ARC_PATH/Projects` and is independent of it. Fixing this also unblocks CI on #1001, which has been failing on `test_initialize_output_dict` by luck of the draw rather than on anything in its own diff.

### DIFF
--- a/arc/scheduler_test.py
+++ b/arc/scheduler_test.py
@@ -9,6 +9,7 @@ import unittest
 from unittest.mock import MagicMock, patch
 import os
 import shutil
+import tempfile
 from types import SimpleNamespace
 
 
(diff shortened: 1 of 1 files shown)


calvinp0:
Superseded by #1028, which rebases this onto current `main` and combines both isolation branches into one. Measured there: random-order failure counts across eight seeds drop from 8-12 to 5-6, against a floor of 5 environmental `torch_ani_test` failures.

Closing in favour of #1028 rather than rebasing this one separately, since the two overlap in intent and `main` has moved past both.

PR #1015, opened 2026-08-22
Restore live conformer jobs by name on restart, not the fossil 'conformer{i}'

alongd:
### Motivation

Restarting a project that had **conformer jobs in flight** crashes with an unhelpful `KeyError: ''`:

```
  File "arc/scheduler.py", line 3508, in get_completed_incore_jobs
    job = self.job_dict[label][job_type][job_name]
KeyError: ''
```

The cause is a naming mismatch between the two places `running_jobs` is populated.

The **live** path (`run_job`) appends `f'{job_type}_{conformer}'`, e.g. `conf_opt_0`:

```python
self.running_jobs[label].append(f'{job_type}_{conformer}')
```

The **restart** path (`restore_running_jobs`) instead appends a fossil `conformer{i}` name that no
consumer of `running_jobs` accepts any more:

```python
elif 'conformer' in job_description:
    self.running_jobs[spc_label].append(f'conformer{job_description["conformer"]}')
```

`get_i_from_job_name` only recognises the `CONFORMER_JOB_TYPES` prefixes (`conf_opt`, `conf_sp`)
and `tsg`, so it returns `None` for `conformer0`. The caller then falls through to the
generic-job-name branch, which derives a job type by stripping the last `_`-delimited field —
`'conformer0'.split('_')[:-1]` is `[]`, so `job_type` becomes `''` and the lookup raises
`KeyError: ''`.

The effect is that a restart cannot reconnect to running conformer jobs. It is not specific to any
adapter or server; any project restarted while conformer jobs are queued or running hits it.

### Changes

- `arc/scheduler.py` — `restore_running_jobs` now emits the same `'{job_type}_{conformer}'` name
  the live path uses, so the restored entry is one `get_i_from_job_name` can parse.
- `arc/checks/common.py` — corrected the `get_i_from_job_name` docstring, which cited
  `'conformer12'` as an example input. That form has not been accepted since `CONFORMER_JOB_TYPES`
  was introduced, and the stale example is what makes the defect above read as intentional.
- `arc/scheduler_test.py` — regression test `test_restore_running_jobs_conformer_reconnects`.

### Verification

The test is red/green against this change, and the red reproduces the production error exactly:

- with `arc/scheduler.py` reverted to `main`: `FAILED ... KeyError: ''` at `arc/scheduler.py:3508`
- with the fix: `1 passed`

`arc/scheduler_test.py` is otherwise unaffected (44 passed alongside it locally; the remaining
local errors are a pre-existing environment issue, present identically on `main`).

### Context

Found while diagnosing a ~4 hour outage of a long-running rate campaign, where the orchestrator
died on restart and the traceback gave no indication that conformer jobs were the trigger.

### DIFF
--- a/arc/scheduler.py
+++ b/arc/scheduler.py
@@ -17,7 +17,7 @@ from typing import TYPE_CHECKING
 
 import arc.parser.parser as parser
 from arc import plotter
-from arc.checks.common import get_i_from_job_name, is_conformer_job, sum_time_delta
+from arc.checks.common import get_conformer_job_name, get_i_from_job_name, is_conformer_job, sum_time_delta
 from arc.checks.ts import check_imaginary_frequencies, check_ts, check_irc_species_and_rxn
 from arc.common import (extremum_list,
                         get_angle_in_180_range,
@@ -1098,7 +1098,7 @@ class Scheduler(object):
         elif conformer is not None:
             # Running a conformer DFT job. Append differently to job_dict.
             self.running_jobs[label] = list() if label not in self.running_jobs else self.running_jobs[label]
-            self.running_jobs[label].append(f'{job_type}_{conformer}')  # mark as a running job
+            self.running_jobs[label].append(get_conformer_job_name(job_type, conformer))  # mark as a running job
             if 'conf_opt' not in self.job_dict[label]:
                 self.job_dict[label]['conf_opt'] = dict()
             if 'conf_sp' not in self.job_dict[label] and job_type == 'conf_sp':
@@ -4131,7 +4131,10 @@ class Scheduler(object):
                             and ('tsg' not in job_description or job_description['tsg'] is None):
                         self.running_jobs[spc_label].append(job_description['job_name'])
                     elif 'conformer' in job_description:
-                        self.running_jobs[spc_label].append(f'conformer{job_description["conformer"]}')
+                        # Emit the same name the live path uses (e.g. 'conf_opt_0'),
+                        # not the fossil 'conformer{i}' that no consumer of running_jobs accepts.
+                        self.running_jobs[spc_label].append(get_conformer_job_name(job_description['job_type'],
+                                                                                   job_description['conformer']))
                     elif 'tsg' in job_description:
                         self.running_jobs[spc_label].append(f'tsg{job_description["tsg"]}')
                     for species in self.species_list:
@@ -4162,9 +4165,16 @@ class Scheduler(object):
                             and ('tsg' not in job_description or job_description['tsg'] is None):
                         self.job_dict[spc_label][job_description['job_type']][job_description['job_name']] = job
                     elif 'conformer' in job_description and job_description['conformer'] is not None:
+                        # File the job under its actual job_type ('conf_opt' or 'conf_sp'), the same
+                        # key the live path uses (see run_job) and the same key get_completed_incore_jobs
+                        # reads back -- filing a conf_sp job under 'conf_opt' would crash the first sweep
+                        # with KeyError: 'conf_sp'.
+                        conf_job_type = job_description['job_type']
                         if 'conf_opt' not in self.job_dict[spc_label].keys():
                             self.job_dict[spc_label]['conf_opt'] = dict()
-                        self.job_dict[spc_label]['conf_opt'][int(job_description['conformer'])] = job
+                        if conf_job_type == 'conf_sp' and 'conf_sp' not in self.job_dict[spc_label].keys():
+                            self.job_dict[spc_label]['conf_sp'] = dict()
+                        self.job_dict[spc_label][conf_job_type][int(job_description['conformer'])] = job
                         # don't generate additional conformers for this species
                         self.dont_gen_confs.append(spc_label)
                     elif 'tsg' in job_description and job_description['tsg'] is not None:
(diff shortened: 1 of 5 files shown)


arc/scheduler.py
```diff
@@ -4060,7 +4060,9 @@ def restore_running_jobs(self):
                             and ('tsg' not in job_description or job_description['tsg'] is None):
                         self.running_jobs[spc_label].append(job_description['job_name'])
                     elif 'conformer' in job_description:
-                        self.running_jobs[spc_label].append(f'conformer{job_description["conformer"]}')
+                        # Emit the same '{job_type}_{conformer}' name the live path uses (e.g. 'conf_opt_0'),
+                        # not the fossil 'conformer{i}' that no consumer of running_jobs accepts.
+                        self.running_jobs[spc_label].append(f'{job_description["job_type"]}_{job_description["conformer"]}')
```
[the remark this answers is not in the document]
alongd: Fixed in a692ce47f. restore_running_jobs now files each conformer job under its actual job_type — a conf_sp job goes into job_dict[label]["conf_sp"], mirroring the live path (run_job, scheduler.py:1098-1103) and what get_completed_incore_jobs reads back. Previously conf_sp was hardcoded into "conf_opt", so a conf_sp restart would have hit KeyError: "conf_sp".

arc/scheduler_test.py
```diff
@@ -240,6 +240,63 @@ def test_conformers(self):
...
+
+        # The first scheduling sweep after a restart reproduces the production crash on the unfixed
+        # code: get_i_from_job_name('conformer0') is None, the fallback derives an empty job-type
+        # from the underscore-less name, and self.job_dict[label][''] raises KeyError: ''.
+        sched.get_completed_incore_jobs()
+        self.assertEqual(sched.completed_incore_jobs, list())
+
+        # And the restored name is the live '{job_type}_{i}' format, not the fossil 'conformer{i}'.
+        self.assertEqual(sched.running_jobs[label], ['conf_opt_0'])
+
```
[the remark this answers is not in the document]
alongd: Done in a692ce47f — the regression test now carries both a live conf_opt and a live conf_sp job through the real restore + sweep path, asserting each is filed under its own job_type and that the sweep does not crash.

arc/scheduler.py
```diff
@@ -4060,7 +4060,9 @@ def restore_running_jobs(self):
                             and ('tsg' not in job_description or job_description['tsg'] is None):
                         self.running_jobs[spc_label].append(job_description['job_name'])
                     elif 'conformer' in job_description:
```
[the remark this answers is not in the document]
alongd: Not a live case, so leaving the branch as-is. Job.as_dict (arc/job/adapter.py:633) only writes the "conformer" key when it is not None, so a restart file never carries "conformer": null — no version of ARC has written one. And the first branch here already handles conformer is None explicitly (`conformer not in job_description or job_description["conformer"] is None`), so the {job_type}_None path is unreachable with real payloads.

arc/scheduler.py
LeenFahoum: This `f'{job_type}_{conformer}'` naming is now used in three places (`adapter.py:519`, `scheduler.py:1098`, and here), and parsed in `arc/checks/common.py`.
It might be worth moving this into a shared helper to keep the format consistent

LeenFahoum (commented):
Looks good overall! I left one small suggestion

arc/scheduler.py
alongd: Good call — done in a4cdeba68. Added `get_conformer_job_name(job_type, i)` to `arc/checks/common.py`, next to the parsers that consume the format, and routed all three construction sites through it (`adapter.py`, `run_job`, `restore_running_jobs`). `arc/checks/common_test.py` now round-trips it against `get_i_from_job_name` and `is_conformer_job` for both conformer job types, so the writer and the parser can no longer drift apart silently.

Branch rebased onto current main and force-pushed.

arc/scheduler.py
LeenFahoum: Not touched by this PR, but I noticed this while reviewing: this `elif` chain doesn’t use the current `job_type`, so restoring both `conf_opt` and `conf_sp` for the same species can create an empty `tsg` entry even though no TS-guess job exists. This is also the scenario exercised by `test_restore_running_jobs_conformer_reconnects`.

It doesn’t break anything currently, but the extra `tsg` key is misleading and is inconsistent with the invariant this PR is establishing, where the container key should match the actual `job_type`.
I think this block could also be keyed directly off `job_type`, similar to the fix below

LeenFahoum (dismissed):
Looks good overall. I left one non-blocking comment about the restart bookkeeping logic

arc/scheduler.py
alongd: Confirmed and fixed in 2494ccb38 — thanks, this reproduces exactly as you describe.

The pre-creation block decided the container by an `elif` chain that never looked at `job_type`: the first conformer job of a species takes the `conf_opt` branch, so the second one falls through to the `tsg` branch and creates an empty `tsg` entry. A `tsg` job restored on its own hit the mirror-image case and created a phantom `conf_opt`.

Rather than re-key that block off `job_type`, I dropped it: each of the three filing branches below already creates exactly the container it writes into, so the pre-creation was redundant as well as wrong. The container key is now the job's own `job_type` on every path.

`test_restore_running_jobs_conformer_reconnects` now asserts `set(job_dict[label]) == {"conf_opt", "conf_sp"}` — red with `'tsg'` present before the fix, green after. Full unit suite: 3600 passed (7 pre-existing local failures, identical on the unmodified branch — stale Cython `.so` and a missing PySCF in this worktree).

Pushed as a follow-up commit rather than squashed into the first, since a local integration branch has already merged this branch's current tip.

LeenFahoum (approved):

PR #1018, opened 2026-08-22
Add a one-process relaxed rotor scan to the ASE adapter

alongd:
## What

Implements `job_type='scan'` for `ASEAdapter`: a complete 1D relaxed torsional scan run **in a single process**, sweeping one torsion in fixed increments and relaxing all other degrees of freedom at each point. Previously `ASEAdapter` had no `scan` branch at all — a `scan` job fell through to the geometry-save `else` in `ase_script.py`, so `output.yml` carried no `energies`/`angles` and ARC's YAML scan parser returned `(None, None)` silently (a quiet empty potential, not an error).

## Why one process per rotor, not one job per point

A rotor scan can be decomposed into one constrained optimization per dihedral point (as `directed_scan` does), but for a machine-learned potential (e.g. UMA) each such job re-loads the model checkpoint to do a few seconds of work — for a large corpus that is hundreds of thousands of tiny jobs dominated by fixed startup. An ESS engine like Gaussian sweeps the whole rotor internally via its `scan` keyword; this gives the ASE adapter the same capability, so a full rotor is one in-process task instead of ~46 jobs. The model load is then paid once per worker and amortized across every rotor it handles.

The scan loop lives in `ase_script.py` (which runs in the calculator's own environment); the adapter only writes the input (`torsions`, `scan_res`) and parses the output. `output.yml` carries `energies` (Hartree) and `angles` (degrees), read back through the existing `yaml` parser adapter and `rotors_dict['scan_path']` — no change to the statmech/Arkane side.

## Design choices

- **Bidirectional relaxed walk.** The torsion is swept 0→360 and 360→0 from the same starting geometry, keeping the lower energy at each grid point. A single directional walk is path-dependent: if a coupled coordinate flips partway round, that walk ends in a different conformer and its curve is not a torsional potential. Taking the pointwise minimum of both walks recovers the lowest torsional path.
- **Geometry propagation is hysteretic** — each point starts from the previous relaxed geometry (which is exactly why both directions are walked).
- **Per-point constraint.** After rotating the increment, the point's actual dihedral is measured and pinned with `FixInternals`; convergence is judged on the residual force **with the constraint still applied** (reading `opt.converged()` after clearing it re-exposes the free torsional force and falsely reports non-convergence at relaxed points).
- **Non-converged points are kept, never dropped** — a hole in the periodic grid would corrupt the downstream Fourier fit. `fmax_worst` in the output records the largest residual.
- **Rotating group** is derived by graph traversal (break the pivot bond, flood-fill from the second pivot), raising on a ring pivot.

## Validation

Validated on **ethane** against an already-validated standalone MLIP scan (UMA `uma-s-1p2`), same species / torsion / starting geometry:

| | this PR (incore ASE `scan`) | reference |
|---|---|---|
| barrier | 10.6879 kJ/mol | 10.6880 kJ/mol |
| points × resolution | 46 × 8° | 46 × 8° |
| max abs deviation | — | **0.028 kJ/mol** (mean 0.012) |

The two V(φ) profiles are the same curve; the residual is at the floating-point/optimizer-noise level (the reference ran on GPU, this on CPU).

## Tests

Adds hermetic unit tests using ASE's EMT calculator (no MLIP needed in CI): rotating-group detection incl. the ring-pivot error, a full 46-point relaxed scan (grid, endpoint duplication, finiteness), output round-tripping through ARC's YAML scan parser, and the single-torsion guard.

## Scope

`scan` only; additive and gated on `job_type=='scan'`. `directed_scan` and existing job types are untouched. Wiring `rotor_scan_1d` through the pipe is a follow-up.

### DIFF
--- a/arc/job/adapters/ase_adapter.py
+++ b/arc/job/adapters/ase_adapter.py
@@ -7,7 +7,7 @@ import os
 import subprocess
 from typing import TYPE_CHECKING, List, Optional, Tuple, Union
 
-from arc.common import get_logger, read_yaml_file, save_yaml_file
+from arc.common import get_logger, read_yaml_file, save_yaml_file, torsions_to_scans
 from arc.job.adapter import JobAdapter
 from arc.job.adapters.common import _initialize_adapter
 from arc.job.factory import register_job_adapter
@@ -188,6 +188,28 @@ class ASEAdapter(JobAdapter):
             
         return ARC_PYTHON or 'python'
 
+    def set_scan_torsions(self) -> None:
+        """
+        Resolve the torsion(s) a 'scan' job should sweep, and store them in ``self.torsions``.
+
+        A scan job may be given either an explicit ``torsions`` list (the pipe path) or a species
+        ``rotors_dict`` together with a ``rotor_index`` (the scheduler path, which leaves
+        ``torsions`` None). This resolves whichever was supplied, mirroring the other ESS adapters,
+        so that ``self.torsions`` is always populated for both the input file and the scheduler,
+        which reads ``job.torsions[0]`` when the job returns.
+
+        Raises:
+            ValueError: If neither a rotor nor a torsion was supplied.
+        """
+        if self.torsions is not None and len(self.torsions):
+            return
+        if self.rotor_index is not None and self.species[0].rotors_dict:
+            scans = self.species[0].rotors_dict[self.rotor_index]['scan']
+            scans = [scans] if not isinstance(scans[0], list) else scans
+            self.torsions = torsions_to_scans(scans, direction=-1)
+            return
+        raise ValueError(f'Could not determine scan parameters for scan job {self.job_name}')
+
     def write_input_file(self) -> None:
         """
         Write the input file for ase_script.py.
@@ -202,6 +224,15 @@ class ASEAdapter(JobAdapter):
             'irc_direction': self.irc_direction,
             'settings': self.determine_settings(),
         }
+        if self.job_type == 'scan':
+            # A 'scan' job sweeps a whole rotor in one process: pass the torsion(s) and the
+            # scan resolution so ase_script.py can run the relaxed torsional scan itself.
+            # The scheduler dispatches a rotor with rotors_dict + rotor_index and leaves torsions
+            # None, so resolve the scan the same way Gaussian/QChem/TeraChem/Psi4 do and back-fill
+            # self.torsions, which the scheduler later reads directly (as job.torsions[0]).
+            self.set_scan_torsions()
+            input_dict['torsions'] = self.torsions
+            input_dict['scan_res'] = self.scan_res
         save_yaml_file(os.path.join(self.local_path, 'input.yml'), input_dict)
 
     def warn_if_unreliable_uma_sp(self) -> bool:
@@ -301,6 +332,8 @@ class ASEAdapter(JobAdapter):
             self.normal_modes = results.get('modes')
             self.reduced_masses = results.get('reduced_masses')
             self.force_constants = results.get('force_constants')
+            for warning in results.get('warnings') or list():
+                logger.warning(f"ASE job warning: {warning}")
             if 'error' in results:
                 logger.error(f"ASE job error: {results['error']}")
 
--- a/arc/job/adapters/ase_test.py
+++ b/arc/job/adapters/ase_test.py
@@ -12,10 +12,32 @@ import unittest
 from unittest.mock import patch
 import numpy as np
 
+from ase import Atoms
+from ase.calculators.emt import EMT
+
 from arc.common import ARC_TESTING_PATH, read_yaml_file, save_yaml_file
 from arc.job.adapters.ase_adapter import ASEAdapter
+from arc.parser.parser import parse_1d_scan_coords, parse_1d_scan_energies
 from arc.species.species import ARCSpecies
-from arc.job.adapters.scripts.ase_script import to_kJmol, numpy_vibrational_analysis, is_linear
+from arc.job.adapters.scripts.ase_script import (is_linear,
+                                                 merge_scan_branches,
+                                                 numpy_vibrational_analysis,
+                                                 relaxed_torsion_scan,
+                                                 rotor_top,
+                                                 run_torsion_scan,
+                                                 scan_convergence_warning,
+                                                 to_kJmol)
+
+ETHANE_XYZ = {'symbols': ('C', 'C', 'H', 'H', 'H', 'H', 'H', 'H'),
+              'isotopes': (12, 12, 1, 1, 1, 1, 1, 1),
+              'coords': ((0.0, 0.0, 0.761395),
+                         (0.0, 0.0, -0.761395),
+                         (0.0, 1.017111, 1.157241),
+                         (0.880844, -0.508556, 1.157241),
+                         (-0.880844, -0.508556, 1.157241),
+                         (0.880844, 0.508556, -1.157241),
+                         (-0.880844, 0.508556, -1.157241),
+                         (0.0, -1.017111, -1.157241))}
 
 
 class TestASEAdapter(unittest.TestCase):
@@ -30,8 +52,7 @@ class TestASEAdapter(unittest.TestCase):
         """
         cls.maxDiff = None
         cls.project_directory = os.path.join(ARC_TESTING_PATH, 'test_ASEAdapter')
-        if not os.path.exists(cls.project_directory):
-            os.makedirs(cls.project_directory)
+        os.makedirs(cls.project_directory, exist_ok=True)  # parallel workers race here
 
         xyz = {'symbols': ('O', 'H', 'H'),
                'isotopes': (16, 1, 1),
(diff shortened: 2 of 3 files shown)


arc/job/adapters/ase_adapter.py
```diff
@@ -202,6 +202,11 @@ def write_input_file(self) -> None:
             'irc_direction': self.irc_direction,
             'settings': self.determine_settings(),
         }
+        if self.job_type == 'scan':
+            # A 'scan' job sweeps a whole rotor in one process: pass the torsion(s) and the
+            # scan resolution so ase_script.py can run the relaxed torsional scan itself.
+            input_dict['torsions'] = self.torsions
```
Ophir-Weisz, line 234: blocking: This writes self.torsions straight through, but check_argument_consistency
explicitly allows torsions to be None when rotors_dict + rotor_index are set — which is
the scheduler's normal path (scheduler.py:3254 reads job.rotor_index). Gaussian, QChem,
TeraChem and Psi4 all resolve scans from rotors_dict[rotor_index]['scan'] first and
back-fill self.torsions via torsions_to_scans(scans, direction=-1). Without that here,
a scheduler-dispatched rotor writes `torsions: null` and the job fails at runtime.
Could we reuse the same resolution the other adapters use?

arc/job/adapters/scripts/ase_script.py
```diff
@@ -130,6 +131,205 @@ def apply_constraints(atoms: Atoms, constraints_data: list):
...
+        pivot_2 (int): The 0-indexed atom on the rotating side of the rotation axis.
+
+    Raises:
+        ValueError: If the pivots are not bonded, or if the pivot bond is part of a ring (a 1D
+                    rotor is then ill-defined).
+
+    Returns:
+        list: The sorted 0-indexed atoms of the rotating group.
+    """
+    nl = build_neighbor_list(atoms, natural_cutoffs(atoms, mult=1.2),
```
Ophir-Weisz: question: natural_cutoffs(..., mult=1.2) is a fixed bonding threshold. The adapter
passes is_ts — for a TS with a stretched forming/breaking bond, would the neighbor list
still find the pivot bond, or would this raise "not bonded"?

arc/job/adapters/scripts/ase_script.py
```diff
@@ -130,6 +131,205 @@ def apply_constraints(atoms: Atoms, constraints_data: list):
...
+    if step_deg <= 0:
+        raise ValueError(f'The dihedral increment must be positive, but got step_deg={step_deg}.')
+    i, j, k, l = torsion
+    if top is None:
+        top = rotor_top(atoms, j, k)
+
+    e_f, r_f = _scan_walk(atoms, torsion, step_deg, nsteps, top, fmax, steps, +1, optimizer)
+    e_b, r_b = _scan_walk(atoms, torsion, step_deg, nsteps, top, fmax, steps, -1, optimizer)
+
+    npts = nsteps  # unique grid points; the walk's index nsteps maps back onto index 0
```
Ophir-Weisz: nit: The n % npts folding and the duplicated endpoint assume nsteps * step_deg == 360.
ARC guarantees that via divmod(360, scan_res)[1] in check_argument_consistency, but this
function is public API running in a separate environment (uma_env), where that guard
doesn't apply. An assert here would make the assumption explicit.

arc/job/adapters/scripts/ase_script.py
```diff
@@ -130,6 +131,205 @@ def apply_constraints(atoms: Atoms, constraints_data: list):
...
+    for n in range(nsteps + 1):
+        p = n % npts  # forward:  phi = +n*step
+        grid_f[p] = min(grid_f[p], e_f[n])
+        q = (-n) % npts  # backward: phi = -n*step
+        grid_b[q] = min(grid_b[q], e_b[n])
+    merged = np.minimum(grid_f, grid_b)
+    energies = merged.tolist()
+    energies.append(energies[0])  # duplicate endpoint: 46-point protocol
+    angles = [n * step_deg for n in range(nsteps + 1)]
+    return {
```
Ophir-Weisz, line 344: question: The output is energies + angles + diagnostics, with no geometries — but
_scan_walk has the relaxed Atoms at line 220 and drops it on the next rotate_dihedral.

ARC wants them in two places: trsh.scan_quality_check pulls conf_xyzs[min_index] to
emit a 'change conformer' action whenever the scan finds a point >= 0.5 kJ/mol below
the start, and scheduler.py:3272 passes the trajectory in to check a TS survived the
rotation. Both get None here. (Not a regression — Psi4/CFOUR/Molpro/QChem return None
there too. Difference is that here the data exists and is thrown away.)

ase_script.py already serialises a geometry list for irc_traj at line 598.

Deliberate for a first iteration? Fine as a follow-up issue if so.

arc/job/adapters/scripts/ase_script.py
```diff
@@ -345,6 +545,13 @@ def save_current_geometry(out_dict, atoms_obj, input_xyz):
     if job_type == 'sp':
         output['sp'] = to_kJmol(atoms.get_potential_energy())
 
+    if job_type == 'scan':
+        try:
+            output.update(run_torsion_scan(atoms, input_dict, settings))
```
Ophir-Weisz, line 658: question: converged, fmax_worst and branch_gap_max are computed but nothing downstream
reads them — a scan where every point hit the step limit still reports as a clean
success. Should a badly non-converged scan (or a large branch_gap_max) set
success: False, or at least log a warning?

arc/job/adapters/ase_test.py
```diff
@@ -204,6 +224,96 @@ def test_numpy_vibrational_analysis(self):
...
+        # A pivot bond inside a ring is ill-defined for a 1D rotor.
+        cyclopropane = Atoms(symbols=('C', 'C', 'C'),
+                             positions=((0.0, 0.87, 0.0), (0.75, -0.43, 0.0), (-0.75, -0.43, 0.0)))
+        with self.assertRaises(ValueError):
+            rotor_top(cyclopropane, 0, 1)
+        # Pivots that are not bonded cannot define a rotating group.
+        with self.assertRaises(ValueError):
+            rotor_top(atoms, 2, 5)
+
+    def test_relaxed_torsion_scan(self):
```
Ophir-Weisz, line 272: nit: The bidirectional merge is the core idea of this PR, and no test exercises it —
ethane is too symmetric to show hysteresis under EMT. Mocking _scan_walk to return two
synthetic branches would pin both the pointwise minimum and the index folding.

Ophir-Weisz (commented):
Nice speedup — running the whole rotor in one process makes a lot of sense for ML
potentials, and I like that non-converged points are kept so the grid stays periodic.
I traced the ARC-side plumbing and the units and angle convention line up with what
the YAML parser and the scheduler expect.

One thing I'd like to understand before approving: the adapter passes self.torsions
through directly, and it looks like that can be None on the scheduler's rotor_index
path. Details in the inline comment. The rest are questions and nits.

(First review I'm doing here, so please tell me if I've misread something.)

arc/job/adapters/ase_adapter.py
alongd: Confirmed and fixed. Verified the failure mode first: `scheduler.py:3263` dereferences `job.torsions[0]` after the job returns, so the scheduler path (rotors_dict + rotor_index, torsions None) would both write `torsions: null` into input.yml and crash on return.

Added `ASEAdapter.set_scan_torsions()`, which mirrors the Gaussian/QChem/TeraChem/Psi4 resolution (`rotors_dict[rotor_index]['scan']` -> `torsions_to_scans(scans, direction=-1)`) and back-fills `self.torsions`. `write_input_file` calls it before writing.

Checked end to end on ethane dispatched the scheduler way: `torsions` None -> `[[2, 0, 1, 5]]`, matching `rotors_dict[0]['scan'] == [3, 1, 2, 6]`. Covered by `test_set_scan_torsions_from_rotors_dict`.

arc/job/adapters/scripts/ase_script.py
alongd: Good catch, and the probe turned up more than the question asked.

You are right that a stretched bond falls outside the default cutoff. I first tried escalating the multiplier globally (1.2 -> 1.5 -> 1.8) and the new test failed with *the ring error*: at a 2.66 Angstrom C-C, loosening the whole graph lets the two carbons' hydrogens reach each other across the widened gap, so the molecule fuses into a spurious ring. Escalating globally trades one wrong answer for another.

So only the pivot pair now gets the looser allowance (`pivot_mult=1.8` against covalent radii); the rest of the graph stays at `mult=1.2`. A genuinely non-bonded pair still raises, now with the measured distance in the message.

Also worth recording: ASE's neighbor list is more permissive than a raw radii sum (it adds a skin), so at mult=1.2 a C-C bond counts as bonded out to about 2.4 Angstrom — a good deal of TS stretch was already tolerated. `test_rotor_top` now pins both ends: a 2.66 Angstrom pivot resolves correctly, and `pivot_mult=1.0` rejects it.

arc/job/adapters/scripts/ase_script.py
alongd: Agreed, and made explicit. `relaxed_torsion_scan` now raises if `abs(nsteps * step_deg - 360) > 1e-6`, and `run_torsion_scan` raises if `scan_res` does not divide 360 evenly — the same `divmod(360, scan_res)[1]` test `check_argument_consistency` uses, so the two agree.

Raised rather than asserted, since asserts vanish under `python -O` and this guards a silent grid corruption. Both covered.

arc/job/adapters/scripts/ase_script.py
alongd: Not deliberate — and cheaper to fix than to defer, so it is in this PR. The YAML parser already has the read side: `arc/parser/adapters/yaml.py:127` `parse_1d_scan_coords` reads a `scan_coords` key. Nothing needed to be added on ARC's side at all; the script was simply not writing it.

`_scan_walk` now keeps each relaxed geometry, the merge carries the winning branch's geometry per grid point (not just its energy), and `run_torsion_scan` shapes them as ARC xyz dicts using the input symbols/isotopes. `test_scan_coords_are_parseable` round-trips them back through `parse_1d_scan_coords`.

So `trsh.scan_quality_check` can now emit its 'change conformer' action and `scheduler.py:3272` gets a real trajectory for the TS check, instead of both getting None.

arc/job/adapters/scripts/ase_script.py
alongd: Fixed, warn-not-fail. A single stubborn point still carries a usable potential, and failing the whole rotor would have ARC resubmit it indefinitely — but you are right that silent 'clean success' was wrong.

Added `scan_convergence_warning()`, which fires on either symptom: points that never reached `fmax`, or a forward/backward `branch_gap_max` above 0.05 eV (the two walks having relaxed into genuinely different conformers, which is the more insidious one). It writes into a new `warnings` key that `parse_results` logs via `logger.warning`; the adapter previously only surfaced `error`. Covered by `test_scan_convergence_warning`.

arc/job/adapters/ase_test.py
alongd: Agreed it was the untested core. Covered it without a mock, since ARC's CLAUDE.md asks for real objects: the fold-and-take-minimum step is now a pure function, `merge_scan_branches(e_f, e_b, nsteps, coords_f, coords_b)`, so it can be fed hand-built asymmetric branches directly.

`test_merge_scan_branches` pins exactly what you named — the pointwise minimum where the forward walk is trapped high, the `(-n) % nsteps` index folding (asserting the backward grid is `[0, 3, 2, 5]` from `[0, 5, 2, 3]`), the endpoint at index `nsteps` folding back onto grid point 0 without losing a lower value, and that the geometries follow whichever branch won.

Extracting it also made the `scan_coords` work above straightforward, since branch selection now happens in one place.

Ophir-Weisz (approved):
All six addressed, thanks — and the pivot-only cutoff and the scan_coords work both
went further than I asked.

Re-checked the new commit rather than the replies: set_scan_torsions matches the
Gaussian/QChem resolution and is covered with a real determine_rotors() species;
scan_coords comes back at 46 points, same length as energies, which is what
trsh.scan_quality_check needs for conf_xyzs[min_index]; and test_merge_scan_branches
now fails if the (-n) % nsteps folding is inverted, which was the gap I was worried
about. Good call raising instead of asserting on the 360 guard.

Two non-blocking observations, take or leave: set_scan_torsions prefers self.torsions
where Gaussian prefers rotors_dict (harmless unless they ever disagree), and the
isotopes fallback in run_torsion_scan yields (None, ...) if input_dict['xyz'] is
missing — unreachable through the adapter, since it always writes xyz.

Approving. codecov/project is red on a -0.03% project delta against a base that's a
commit behind; codecov/patch is green and all modified lines are covered.

PR #1019, opened 2026-08-23
Add a rotor_scan_resolution input key

alongd:
## Motivation

ARC had no user-facing way to state the resolution of its 1D rotor scans. The value came from
`rotor_scan_resolution` in the settings of *whatever process happened to be running*. This bit us
for real: **a machine-wide settings override silently put every rotor scan on a workstation below
the resolution where the downstream Fourier fit is valid, for a week, with no crash and no
warning** — a silent wrong answer. This PR makes a run able to state what it needs, so resolution is
a property of the run and not of the host that launched it.

## What changed

Adds a user-facing input key **`rotor_scan_resolution`** (degrees; points per rotor = 360 / value):

```yaml
project: my_project
level_of_theory: wb97xd/def2tzvp
rotor_scan_resolution: 4.0
species:
  - label: ethane
    smiles: CC
```

Route (input → job), through the **existing** per-job override channel rather than a new one:

```
ARC.__init__(rotor_scan_resolution=...)  ->  validated, stored, round-tripped via as_dict()
ARC.execute()  ->  Scheduler(rotor_scan_resolution=...)
run_job()      ->  set_scan_resolution(args, job_type)  ->  args['trsh']['scan_res'] = value
job_factory -> JobAdapter -> set_job_attributes  ->  scan_res used by the ESS scan
```

- Only `job_type == 'scan'` (the continuous 1D ESS scans) is affected. Brute-force ND directed
  scans (`job_type='directed_scan'`, which use `dihedral_increment`) and the pipe path are untouched.
- An explicit troubleshooting `scan_res` is never overridden — troubleshooting still wins.

## Guard: refuse a too-coarse value

A resolution coarser than 20° (fewer than 18 points per rotor) is **refused** with `InputError` at
input-parse time, not merely warned. Below 18 points RMG-Py's Fourier fitter never runs and
`get_potential()` reads an uninitialised C double — a coarse value is not a preference but a silent
wrong answer, which is exactly the incident above. `20.0°` (18 points) is the coarsest still accepted.

## Backwards compatibility

Absent the key, behaviour is **byte-identical** to before: `rotor_scan_resolution` defaults to
`None`, `as_dict()` omits it (so `restart.yml` is unchanged), and `set_scan_resolution` leaves the
job args untouched so the settings default (`8.0`) still reaches the job.

## Tests

Four new tests (parse/store/round-trip; absent-key no-op; the guard; and a real scan `JobAdapter`
built via `job_factory` receiving the value and falling back to the settings default when unset).

- `arc/main_test.py arc/scheduler_test.py`: 63 passed.
- `arc/job/adapters/common_test.py arc/job/adapters/gaussian_test.py`: 31 passed.

Run under `arc_env` (Python 3.14) with `HOME` pointed at an empty dir (ARC's suite reads
`~/.arc/settings.py`) and `-n0 -o addopts=""`.

### DIFF
--- a/arc/main.py
+++ b/arc/main.py
@@ -253,6 +253,7 @@ class ARC(object):
                  project: str | None = None,
                  project_directory: str | None = None,
                  reactions: list[ARCReaction] | None = None,
+                 rotor_scan_resolution: float | None = None,
                  running_jobs: dict | None = None,
                  scan_level: str | dict | Level | None = None,
                  sp_level: str | dict | Level | None = None,
@@ -311,6 +312,7 @@ class ARC(object):
         self.job_types = job_types or default_job_types
         self.job_types = initialize_job_types(job_types, specific_job_type=self.specific_job_type)
         self.bath_gas = bath_gas
+        self.rotor_scan_resolution = check_rotor_scan_resolution(rotor_scan_resolution)
         self.n_confs = n_confs
         self.e_confs = e_confs
         self.adaptive_levels = process_adaptive_levels(adaptive_levels)
@@ -503,6 +505,8 @@ class ARC(object):
             restart_dict['reactions'] = [rxn.as_dict() for rxn in self.reactions]
         if self.running_jobs:
             restart_dict['running_jobs'] = self.running_jobs
+        if self.rotor_scan_resolution is not None:
+            restart_dict['rotor_scan_resolution'] = self.rotor_scan_resolution
         if self.scan_level is not None and len(self.scan_level.method) \
                 and str(self.scan_level).split()[0] != default_levels_of_theory['scan']:
             restart_dict['scan_level'] = self.scan_level.as_dict() \
@@ -600,6 +604,7 @@ class ARC(object):
                                    dont_gen_confs=self.dont_gen_confs,
                                    trsh_ess_jobs=self.trsh_ess_jobs,
                                    trsh_rotors=self.trsh_rotors,
+                                   rotor_scan_resolution=self.rotor_scan_resolution,
                                    fine_only=self.fine_only,
                                    ts_adapters=self.ts_adapters,
                                    report_e_elect=self.report_e_elect,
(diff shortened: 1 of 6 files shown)


arc/main.py
```diff
@@ -1271,6 +1276,30 @@ def standardize_output_paths(self):
...
+    Returns: float | None
+        The validated resolution, or ``None`` if none was provided.
+    """
+    if rotor_scan_resolution is None:
+        return None
+    if rotor_scan_resolution <= 0 or rotor_scan_resolution > 20:
+        raise InputError(f'The rotor_scan_resolution must be a positive value no coarser than 20 degrees '
+                         f'(coarser resolutions give fewer than 18 points per rotor, below which the '
+                         f'Fourier fit is invalid and returns a silent wrong answer). Got: {rotor_scan_resolution}')
+    return rotor_scan_resolution
```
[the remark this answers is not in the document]
alongd: Fixed in f6ceb9166 — check_rotor_scan_resolution() now raises a consistent InputError for non-numeric (and bool) inputs before the numeric comparison, so a quoted YAML value no longer produces a raw TypeError. Added a test asserting the guard fires on a str.

arc/main.py
```diff
@@ -1271,6 +1276,33 @@ def standardize_output_paths(self):
             self.output = dict()
 
 
+def check_rotor_scan_resolution(rotor_scan_resolution: float | None) -> float | None:
```
Lilachn91, line 1279: Might be worth adding a guard against poor host-level settings as well? i.e., if the user defines in `~/.arc/settings.py` a `rotor_scan_resolution` higher than 20.0.
Also, divisibility is not checked until job-build time (in `common.py` when reaching `if obj.job_type == 'scan' and divmod(360, obj.scan_res)[1]:`). If this function already guards against unacceptable values at input-parse time, maybe consider moving the divisibility test in here too?

arc/main_test.py
```diff
@@ -206,6 +206,31 @@ def test_from_dict_specific_job(self):
...
+        self.assertIsNone(arc1.rotor_scan_resolution)
+        self.assertNotIn('rotor_scan_resolution', arc1.as_dict())
+
+    def test_rotor_scan_resolution_guard(self):
+        """Test that a rotor scan resolution coarser than 20 degrees is refused."""
+        with self.assertRaises(InputError):
+            ARC(project='arc_test_coarse_scan_res', rotor_scan_resolution=30.0)
+        with self.assertRaises(InputError):
+            ARC(project='arc_test_nonpositive_scan_res', rotor_scan_resolution=0.0)
+        # A non-numeric type (e.g. a quoted YAML value) is refused with a consistent InputError
```
Lilachn91, line 226: check_rotor_scan_resolution() explicitly excludes bool (isinstance(rotor_scan_resolution, bool)) - since bool is a subclass of int, without that line True/False would silently pass as 1/0 degrees. Nothing tests this defensive branch, so if it ever got removed or broken by a refactor, nothing would catch it.

Lilachn91 (commented):
Thanks, see a few comments.
Also, a general question regarding RMG-Py's Fourier fitter - if a user runs Arkane manually, and gives it a scan job with resolution higher than 20.0, would he not run into the same issue of getting a silent wrong answer?

arc/main.py
alongd: Both done in 149c79609:

- Host-level guard: `check_argument_consistency` in `arc/job/adapters/common.py` is the single chokepoint where the run-level key, the `~/.arc/settings.py` default, and trsh all converge on `obj.scan_res`. Extended its scan-job check from divisibility-only to `obj.scan_res <= 0 or obj.scan_res > 20 or divmod(360, obj.scan_res)[1]`, so a coarse or non-positive `rotor_scan_resolution` in host settings is now rejected at job build, with a message pointing at both the input file and `~/.arc/settings.py`.
- Divisibility at parse time: added the `divmod(360, rotor_scan_resolution)[1]` check to `check_rotor_scan_resolution` so the input-key path fails fast with a clear `InputError`. I kept the `common.py` divisibility guard as the catch-all for the settings/trsh paths (which don't pass through `check_rotor_scan_resolution`).

arc/main_test.py
alongd: Good catch — added coverage in 149c79609. `test_rotor_scan_resolution_guard` now asserts that both `True` and `False` raise `InputError` (pinning the bool-exclusion branch), plus a case for an indivisible value (7.0) covering the new divisibility check.

Lilachn91 (approved):
LGTM, thanks!

alongd:
A user using Arkane manually with a low resolution scan (e.g., 60 deg) will probably have a bad quality fit. This is indeed an Arkane issue (not an ARC one). But it's OK - Arkane is the bare computation, ARC as the wrapper has more protections for a quality result. If we want, we could PR a fix in Arkane too, but I think it's OK as is.

PR #1021, opened 2026-08-23
Support reactions whose participants repeat (A + A), end to end

calvinp0:
A reaction whose two reactants are the same species — `OH + OH <=> H2O + O` — is mishandled at three
layers. One commit per layer.

# 1. A species participating more than once in a well

`ARCReaction` stored each participant once, so a species appearing twice in a well contributed its
atoms once. The occurrence count is now read through `get_species_count(species=spc, well=...)` and the
participants are expanded per occurrence where the geometry needs them, rather than by holding duplicate
objects: the reactant and product well geometries place one copy per occurrence, and the translation
offsets are computed over the expanded list.

# 2. The normal mode displacement check under dedup

`analyze_ts_normal_mode_displacement` mapped bond changes against the deduplicated participant list, so
for an A + A reaction the expected bonds did not line up with the atoms the TS actually holds. The check
is dedup-aware, and a TS it cannot map is reported as uncheckable rather than discarded — an unmappable
TS is not evidence of a bad TS.

# 3. Thermo recovery when Arkane refuses to write its library

For a reaction whose two reactants are the same species, Arkane's `save_thermo_lib` raises
`DatabaseError` on the `R1 == R2` duplicate at the very end of its run — after `output.py` is written but
before `RMG_libraries/thermo.py` is. `save_arkane_thermo.py` guarded on that missing library file and
returned early, so `thermo.yaml` was never written and `parse_arkane_thermo_output` left `spc.thermo`
unset: null thermo in `output.yml` and an empty project thermo library for every species in the run,
despite Arkane having computed full NASA polynomials.

When the library file is absent, each species' thermo is reconstructed from `output.py`, which Arkane has
already written by the time `save_thermo_lib` runs, producing `thermo.yaml` exactly as the library path
would. Blocks are located with `ast.parse` and evaluated with the same rmgpy classes Arkane uses to read
these files back, so a parenthesis inside a string literal cannot drop a species. A species whose thermo
cannot be evaluated is reported on stderr and skipped rather than aborting the remaining ones. The path
where the library file exists is unchanged.

With thermo recovered, both A + A duplicates carry data, and `plotter.save_thermo_lib` emitted two
entries whose adjacency list and multiplicity match — which `rmgpy`'s `ThermoLibrary.load_entry` rejects,
taking the library from empty to unloadable. The same isomorphic-plus-equal-multiplicity test RMG applies
now omits the later entry with a warning. This affects the library file only; every species keeps its own
computed thermo elsewhere.

The species list handed to the thermo library is filtered by the same criterion one layer earlier, in
`processor.py`. The two checks sit at different points — one filters the list, the other filters entries
as they are written — and share a predicate that a later change could collapse into one helper.

# Tests

Per-occurrence well geometry and atom counts; the dedup-aware NMD check including the uncheckable-TS
path; end-to-end thermo recovery through the real script with no library present; both duplicates
recovering their own thermo; JANAF checks on H2O at 298.15 K and 2000 K; parser robustness; and an RMG
load of the generated thermo library.

### DIFF
--- a/arc/checks/nmd.py
+++ b/arc/checks/nmd.py
@@ -34,6 +34,9 @@ def analyze_ts_normal_mode_displacement(reaction: ARCReaction,
     Analyze the normal mode displacement by identifying bonds that break and form
     and comparing them to the expected given reaction.
     Note that the TS geometry must be in the standard orientation for the normal mode displacement to be relevant.
+    The forming, breaking and changed bonds of the reaction are indexed into the concatenated reactant atoms,
+    where a species participating more than once (e.g. ``A + A``, for which ``r_species`` holds a single entry)
+    contributes its atoms once per occurrence. The TS geometry must span that same set of atoms.
 
     Args:
         reaction (ARCReaction): The reaction for which the TS is checked.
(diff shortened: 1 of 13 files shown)


kfir4444 (approved):
lgtm!

PR #1024, opened 2026-08-24
Return every atom map per reaction and cluster them into distinct reaction channels

kfir4444:
Stacked on #1023 — please review that first; this branch's base will collapse once it lands.

## Problem

`map_reaction` returns **one** atom map per reaction. Several places in the pipeline compute more than one
and discard the rest:

* `map_two_species` scores every superimposable backbone candidate, then keeps the lowest-RMSD one and
  breaks ties by taking the *last* tied candidate — an arbitrary choice.
* `map_rxn` returns on the first template `product_dict` that succeeds, and only advances to the next one
  when a stage fails.

So when a reaction genuinely has more than one distinct channel, ARC silently picks one of them by a
tie-break, and there is no way to ask how many equivalent paths a channel has.

## What this adds

`arc/mapping/cluster.py`: enumerate every valid atom map, then group them into equivalence classes. Each
class is one chemically distinct reaction channel (one TS to search for); its `degeneracy` is that channel's
reaction path degeneracy.

Exposed as `ARCReaction.atom_map_clusters` — lazily computed, cached, never triggered by `atom_map`, and not
persisted by `as_dict`. `atom_map` itself is untouched, so every existing consumer is unaffected.

### The equivalence relation

Two maps are equivalent iff `sigma_2 = beta . sigma_1 . alpha` for some `alpha` in `Aut(R)`, `beta` in
`Aut(P)` — the double cosets. That reduces to a single group. Writing `C(sigma)` for the set of bonds of `R`
that break, form, or change order, and `R'` for `R` with those changes applied, any valid `sigma` is an
isomorphism `R' -> P`. So equal changed-bond sets differ by an element of `Aut(P)`, and:

```
sigma_1 ~ sigma_2   <=>   C(sigma_1) and C(sigma_2) lie in the same Aut(R) orbit
```

`Aut(P)` never appears. Automorphisms are enumerated in-module by colour-pruned backtracking over the
*core skeleton* (hydrogens folded onto their parent), which avoids the factorial blow-up — isobutane has
1296 full-graph automorphisms and 6 core ones — and needs no new dependency.

### Reaction path degeneracy

Degeneracy is **not** a count of what the enumeration found; that depends on how many template matches RMG
happens to generate, and under-reports. It is the orbit size of the reaction center under the full
automorphism group, which factorises:

```
degeneracy = |orbit of the collapsed center under Aut_core(R)|
             * product over core atoms p of  n_p! / (n_p - k_p)!
```

Verified against known values:

| reaction | degeneracy |
|---|---|
| CH4 + OH -> CH3 + H2O | 4 |
| C2H6 + OH -> C2H5 + H2O | 6 |
| C3H8 + NH2 -> n-C3H7 + NH3 | 6 |
| C3H8 + NH2 -> i-C3H7 + NH3 | 2 |
| C2H5Cl -> C2H4 + HCl | 3 |
| cyclohexane + OH -> C6H11 + H2O | 12 |

Cyclohexane exercises both halves: 6 equivalent carbons times 2 hydrogens each.

### Validity filtering

`changed_bonds` computes a center for *any* element-preserving bijection, so scrambled maps would otherwise
open bogus clusters — for the Diels-Alder of butadiene with ethene the correct map changes 2 bonds while
scrambled ones change 10 to 14, and each opened its own cluster. Maps are validated against the family
recipe (`ARCReaction.get_expected_changing_bonds` over each `product_dict`'s `r_label_map`), falling back to
a minimal-center heuristic where the recipe is unreadable. On the corpus the two agree.

## Test

`arc/mapping/cluster_test.py`, 60 tests: graph construction, automorphism counts against known group orders
(benzene 12, naphthalene 4, neopentane 24, 2x benzene 288), Kekule bond-order handling, changed-bond sets
and their `Aut(P)` invariance, canonical keys, the degeneracy factorisation, both validity filters, and an
end-to-end CH4 + OH run.

Expected values are derived independently — group orders from graph theory, degeneracies from known
reaction path degeneracies — not read off the implementation. Each behaviour was mutation-checked: reverting
the code under test makes the corresponding test fail.

```
arc/mapping/ arc/reaction/ arc/checks/nmd_test.py  ->  217 passed
```

## Corpus effect

`testing/errs_no_li.yml`, 452 reactions, 14 workers, 200 s each:

```
ok       330 (73.0%)      1 cluster:  329  (99.7%)
no_maps  120 (26.5%)      2 clusters:   1  (0.3%)
```

**One cluster is overwhelmingly the norm.** With reactant *and* product species both pinned, different
reacting sites generally give different product graphs, so naming the products pins the channel. More than
one cluster requires two `Aut(R)`-inequivalent centers yielding isomorphic products.

The single multi-cluster case is `C10H9-3 <=> C10H9-4` (cyclopentadienyl-cyclopentadiene, `|Aut(R)| = 1`):
an H shift within one ring, or one that swaps the rings' roles, both give an isomorphic product. Both
centers are predicted by the family recipe, so neither is an enumeration artifact — though whether the
second corresponds to a physically sensible (large-ring) TS is a question for a chemist. `map_reaction`
resolves that ambiguity today by an arbitrary tie-break.

So with both wells specified the feature mostly confirms degeneracy; at 1-in-330 a multi-cluster result is
signal worth investigating rather than noise. Its sharper use is enumerating channels when products are
*not* fixed in advance.

## Cost

Enumeration is substantially more expensive than return-on-first-success — it sweeps every `product_dict`,
both directions, and every backbone candidate per fragment. That is why `atom_map_clusters` is opt-in and
never triggered by `atom_map`. Caps exist at three levels (`MAX_AUTOMORPHISMS`, `MAX_ENUMERATED_MAPS`,
`map_rxn_all`'s `max_maps`) and all log when hit; a truncated automorphism group is reported through
`MapCluster.truncated`, since it would over-split rather than merely slow things down.

## Note on scope

One commit, `fix: create job directories without a check-then-create race`, is outside this PR's subject. `_initialize_adapter` and the job adapter both did `if not isdir(p): makedirs(p)`, which raises `FileExistsError` when two jobs are set up concurrently. It is a pre-existing bug on main, but this PR's added tests make it fire reliably in the parallel suite, so it is fixed here rather than left to redden CI. Happy to split it out if preferred.

### DIFF
--- a/arc/mapping/cluster.py
+++ b/arc/mapping/cluster.py
@@ -48,6 +48,7 @@ happened to produce. It is the orbit size of the reaction center under the *full
 core group extended by permutations of the hydrogens on each core atom - see :func:`center_degeneracy`.
 """
 
+import itertools
 import operator
 from dataclasses import dataclass, field
 
@@ -445,6 +446,32 @@ def _endpoint(graph: ComplexGraph, index: int) -> tuple:
     return CORE, index
 
 
+def hydrogen_stabilizer(center: frozenset, hydrogens: list[int]) -> int:
+    """
+    How many permutations of ``hydrogens`` leave the changed-bond set unchanged.
+
+    The changed-bond set is a *set*, so two hydrogens on the same parent that play the same role in it are
+    interchangeable: swapping them reproduces the same center rather than describing a second reaction path.
+    :func:`center_degeneracy` divides the ordered count by this to recover the orbit of the set.
+
+    Args:
+        center (frozenset): The changed-bond set, not hydrogen-collapsed.
+        hydrogens (list[int]): The named hydrogens of one parent core atom.
+
+    Returns:
+        int: The size of the stabilizer, at least 1.
+    """
+    hydrogens = sorted(hydrogens)
+    stabilizer = 0
+    for permutation in itertools.permutations(hydrogens):
+        relabel = dict(zip(hydrogens, permutation))
+        image = frozenset(tuple(sorted((relabel.get(i, i), relabel.get(j, j)))) + (before, after)
+                          for i, j, before, after in center)
+        if image == center:
+            stabilizer += 1
+    return stabilizer or 1
+
+
 def center_degeneracy(center: frozenset,
                       graph: ComplexGraph,
                       automorphisms: list[dict[int, int]],
@@ -464,8 +491,16 @@ def center_degeneracy(center: frozenset,
                      * product over core atoms p of  n_p! / (n_p - k_p)!
 
     where ``n_p`` is how many hydrogens hang off core atom ``p`` and ``k_p`` how many distinct ones the
-    center actually names. The falling factorial counts the ordered choices of which hydrogens play the
-    named roles. For CH4 + OH the core group is trivial and one of methane's four hydrogens is named, giving
+    center actually names. The falling factorial counts the *ordered* choices of which hydrogens play the
+    named roles, so it is divided by :func:`hydrogen_stabilizer` to recover the orbit of the center as a
+    set - see there for why.
+
+    The result is the orbit size and is deliberately *not* halved when the two reactants are the same
+    species. RMG halves it in that case (``reduce_same_reactant_degeneracy``, after Bishop and Laidler
+    1965) so a rate expression does not double count indistinguishable colliding partners:
+    ``C2H5 + C2H5 <=> C4H10`` is 1 here and 0.5 in RMG. Which convention is correct depends on how the
+    number is consumed, and a caller folding it into a rate coefficient for identical reactants has to
+    apply that factor itself. For CH4 + OH the core group is trivial and one of methane's four hydrogens is named, giving
     4; for C2H6 + OH the core group swaps the two carbons and one of three hydrogens is named, giving 6.
 
     Args:
@@ -477,7 +512,11 @@ def center_degeneracy(center: frozenset,
         int: The reaction path degeneracy.
     """
     if not center:
-        return 0
+        # The orbit of the empty set is itself, so the answer is one path, never zero. An empty center
+        # means this map changes no bonds *under the current bond-order setting* - families such as
+        # 1,2-Birad_to_alkene only change orders, so ignore_bond_orders=True erases their whole center.
+        # cluster_atom_maps() detects that case and retries with orders honored.
+        return 1
     collapsed = collapse_hydrogens(center, graph)
     images = set()
     for alpha in automorphisms:
@@ -502,9 +541,16 @@ def center_degeneracy(center: frozenset,
 
     hydrogen_factor = 1
     for parent, hydrogens in named.items():
-        available = hydrogen_counts.get(parent, 0)
+        # named[parent] is a subset of the hydrogens on that parent, so available >= len(hydrogens) and
+        # each factor below is >= 1; no clamping is needed.
+        available = hydrogen_counts[parent]
         for offset in range(len(hydrogens)):
-            hydrogen_factor *= max(available - offset, 1)
+            hydrogen_factor *= available - offset
+        # The falling factorial counts *ordered* assignments of hydrogens to the named roles, but the
+        # center is a set. Interchangeable hydrogens on one parent therefore have to be divided out, or a
+        # single path is counted once per ordering: both C-H bonds of CH2O breaking is one path to H2, not
+        # two. Dividing by the stabilizer turns the ordered count into the orbit of the set.
+        hydrogen_factor //= hydrogen_stabilizer(center, list(hydrogens))
     return core_orbit * hydrogen_factor
 
 
--- a/arc/mapping/cluster.py
+++ b/arc/mapping/cluster.py
@@ -48,6 +48,7 @@ happened to produce. It is the orbit size of the reaction center under the *full
 core group extended by permutations of the hydrogens on each core atom - see :func:`center_degeneracy`.
 """
 
+import operator
 from dataclasses import dataclass, field
 
 from arc.common import logger
@@ -621,6 +622,35 @@ def enumerate_atom_maps(rxn,
     return maps
 
 
+def is_permutation_map(atom_map, n_reactant_atoms: int, n_product_atoms: int) -> bool:
+    """
+    Whether ``atom_map`` is a genuine bijection of reactant indices onto product indices.
+
+    :func:`changed_bonds` inverts the map, so anything short of a bijection - a duplicated or missing
+    product index, an out-of-range entry, a non-integer entry - raises a ``KeyError`` there and aborts the
+    entire clustering run rather than just discarding the offending map. Callers screen maps with this
+    first, in the same spirit as ``driver.check_atom_map_and_return``.
+
+    Args:
+        atom_map: The candidate atom map, ``atom_map[reactant_index] == product_index``.
+        n_reactant_atoms (int): Number of atoms in the reactant complex.
+        n_product_atoms (int): Number of atoms in the product complex.
+
+    Returns:
+        bool: ``True`` iff the map is a bijection of ``range(n_reactant_atoms)`` onto
+              ``range(n_product_atoms)``.
+    """
+    if not isinstance(atom_map, (list, tuple)) or len(atom_map) != n_reactant_atoms:
+        return False
+    try:
+        # operator.index() accepts numpy integers, which the mapping pipeline can produce, while still
+        # rejecting floats, None and strings.
+        indices = [operator.index(value) for value in atom_map]
+    except TypeError:
+        return False
+    return len(set(indices)) == n_product_atoms and all(0 <= index < n_product_atoms for index in indices)
+
+
 def cluster_atom_maps(atom_maps: list[list[int]],
                       rxn,
                       ignore_bond_orders: bool = True,
@@ -656,9 +686,10 @@ def cluster_atom_maps(atom_maps: list[list[int]],
 
     scored: list[tuple[list[int], frozenset]] = list()
     for atom_map in atom_maps:
-        if len(atom_map) != r_graph.n_atoms:
-            logger.warning(f'Skipping an atom map of length {len(atom_map)} for {rxn}, '
-                           f'expected {r_graph.n_atoms}.')
+        if not is_permutation_map(atom_map, r_graph.n_atoms, p_graph.n_atoms):
+            logger.warning(f'Skipping an atom map for {rxn} that is not a bijection of the '
+                           f'{r_graph.n_atoms} reactant atoms onto the {p_graph.n_atoms} product '
+                           f'atoms: {atom_map}.')
             continue
         scored.append((atom_map, changed_bonds(r_graph, p_graph, atom_map,
                                                ignore_bond_orders=ignore_bond_orders)))
--- a/arc/mapping/cluster.py
+++ b/arc/mapping/cluster.py
@@ -1,0 +1,782 @@
+"""
+Enumeration and equivalence-clustering of atom maps.
+
+The standard mapping entry point, ``arc.mapping.driver.map_reaction``, returns a single atom map per
+reaction. This module enumerates *many* valid atom maps and groups them into equivalence classes, where
+each class corresponds to one chemically distinct reaction channel (one transition state to search for)
+and the size of a class is that channel's reaction path degeneracy.
+
+Theory
+------
+An atom map is a bijection ``sigma: atoms(R) -> atoms(P)`` from the reactant complex to the product
+complex, stored as a list where ``atom_map[reactant_index] == product_index``.
+
+If ``sigma`` is valid then so is ``beta . sigma . alpha`` for any ``alpha`` in ``Aut(R)`` and ``beta`` in
+``Aut(P)``, since composing with an automorphism only relabels symmetry-equivalent atoms. The equivalence
+classes are therefore the double cosets ``Aut(P) . sigma . Aut(R)``.
+
+That definition can be reduced to something much cheaper. Let ``C(sigma)`` be the *changed-bond set*: the
+bonds of ``R`` that are broken, formed, or change order under ``sigma``. Write ``R'`` for ``R`` with those
+changes applied. Any valid ``sigma`` is a graph isomorphism ``R' -> P``, so:
+
+- If ``C(sigma_1) == C(sigma_2)`` then ``sigma_2 . sigma_1^-1`` maps ``P`` to ``P`` preserving all bonds,
+  i.e. it lies in ``Aut(P)``, hence ``sigma_1 ~ sigma_2``.
+- If ``C(sigma_2) == alpha(C(sigma_1))`` for some ``alpha`` in ``Aut(R)``, then ``sigma_1 . alpha^-1`` has
+  changed-bond set ``C(sigma_2)``, so by the previous point ``sigma_2 = beta . sigma_1 . alpha^-1``.
+
+Therefore::
+
+    sigma_1 ~ sigma_2   <=>   C(sigma_1) and C(sigma_2) are in the same Aut(R) orbit
+
+Only ``Aut(R)`` is ever needed; ``Aut(P)`` does not appear. Clustering reduces to canonicalizing an edge
+subset of ``R`` under a single group action.
+
+Hydrogens
+---------
+Automorphism groups are computed on the *core skeleton* rather than the full molecular graph, because the
+full group is dominated by intra-XHn hydrogen permutations that provably cannot change a cluster (the
+hydrogens on a methyl group are one orbit of ``Aut(R)`` by construction). For isobutane the full group has
+order 1296 while the core skeleton group has order 6.
+
+A "core" atom is any non-hydrogen atom, plus any hydrogen that has no unique non-hydrogen neighbor (a free
+H radical, H2, ...). Remaining hydrogens are represented in the changed-bond set by their parent core atom,
+which is well defined precisely because hydrogens on a common parent are interchangeable. This keeps
+hydrogen-transfer reactions - where the migrating atom *is* a hydrogen - fully representable.
+
+See ``docs/atom_mapping_clustering_design.md`` for the full design, and ``docs/atom_mapping_summary.md``
+for a description of the underlying single-map pipeline.
+"""
+
+from dataclasses import dataclass, field
+
+from arc.common import logger
+from arc.mapping.driver import MAX_PDI, map_rxn_all
+from arc.mapping.engine import flip_map
+from arc.species import ARCSpecies
+
+# An upper bound on the number of automorphisms enumerated for one complex. Hitting this bound means the
+# clustering is computed under a truncated group and may split classes that are genuinely equivalent, so it
+# is reported rather than applied silently.
+MAX_AUTOMORPHISMS = 100_000
+
+# An upper bound on the number of maps enumerated for one reaction.
+MAX_ENUMERATED_MAPS = 5_000
+
+# Endpoint descriptor tags used in a changed-bond set.
+CORE = 'A'
+HYDROGEN = 'H'
+
+
+@dataclass
+class ComplexGraph:
+    """
+    A flat graph view of a reactant or product complex, indexed by *running* atom indices across the whole
+    complex - the same indexing convention used by ``atom_map``.
+
+    Attributes:
+        n_atoms (int): Total number of atoms in the complex.
+        symbols (list[str]): Element symbol per atom.
+        bonds (dict[frozenset[int], float]): Bond orders keyed by the unordered pair of atom indices.
+        core (list[int]): Indices of core (skeleton) atoms.
+        parent (dict[int, int]): Maps a non-core hydrogen index to its parent core atom index.
+        adj (dict[int, dict[int, float]]): Core-core adjacency with bond orders.
+        invariant (dict[int, tuple]): Per-core-atom initial colour used to seed the automorphism search.
+    """
+    n_atoms: int
+    symbols: list[str]
+    bonds: dict[frozenset, float]
+    core: list[int]
+    parent: dict[int, int]
+    adj: dict[int, dict[int, float]]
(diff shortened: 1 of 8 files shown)


kfir4444:
## A worked example with two real channels

`CC[CH]C <=> CCC[CH2]` (sec-butyl to n-butyl, `intra_H_migration`) is a clean case where the single-map
answer is a coin flip between two genuinely different transition states.

The reactant is `C0-C1-C2(.)-C3`. A hydrogen leaves a terminal methyl and lands on the radical carbon C2,
and either methyl works — both give n-butyl, since n-butyl is the same molecule numbered from either end:

```
C3 -> C2    1,2 shift, three-membered TS
C0 -> C2    1,3 shift, four-membered TS
```

These are not symmetry-equivalent. `|Aut_core(R)| = 1` — sec-butyl has no automorphism exchanging its ends —
so no element of `Aut(R)` carries one center onto the other, and they must not merge into a single channel of
degeneracy 6. The clustering reports two channels of degeneracy 3 each, one per methyl:

```
maps enumerated : 8
clusters        : 2
   degeneracy 3   ('C0', 'H@C0', 1, 0), ('C2', 'H@C0', 0, 1)      1,3 shift
   degeneracy 3   ('C2', 'H@C3', 0, 1), ('C3', 'H@C3', 1, 0)      1,2 shift
```

All six paths are predicted by the family recipe, so the absolute reference of §"Validity filtering" did the
filtering here and neither channel is an enumeration artifact.

`rxn.atom_map` returns a representative of the 1,3-shift channel. The 1,2-shift channel is never reported.

**Both transition states have been located and optimized independently**, at b2plyp/def2tzvp, by supplying
the two atom maps to ARC by hand — declaring the product species twice and writing out each map, which is
precisely the manual workaround this PR removes the need for. The two barriers differ substantially, so the
channels are kinetically distinct rather than a bookkeeping artifact.

Feeding those two hand-written maps to `cluster_atom_maps` puts them in two different clusters, and the
automatic enumeration reproduces exactly the same two channel keys with no maps supplied. That equivalence is
pinned by `test_sec_butyl_to_n_butyl_has_two_channels`.

arc/mapping/engine.py
```diff
@@ -79,6 +79,151 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
...
+            tied_indices = [i for i in tied_indices
+                            if displacements[i] - lowest_displacement <= RMSD_TIE_TOLERANCE]
+        chosen_candidate_index = max(tied_indices)
+        atom_map = atom_maps[chosen_candidate_index]
+        if map_type == 'list':
+            atom_map = [v for k, v in sorted(atom_map.items(), key=lambda item: item[0])]
+
+    if inc_vals is not None:
+        atom_map = [value + inc_vals for value in atom_map]
+    return atom_map
```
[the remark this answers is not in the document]
kfir4444: Confirmed and fixed in b588947c.

Iterating a dict yields its keys, so `[value + inc_vals for value in atom_map]` returned a *list* of shifted reactant indices instead of a dict with shifted product indices — both the wrong type and the wrong content, exactly as described. `map_two_species` now branches on `isinstance(atom_map, dict)` and increments the values.

Two notes on scope. This bug predates the branch: the line traces back to `846b8c6d` ("Added engine to mapping module") and is present on `main` today, so this PR only surfaced it rather than introducing it. That is why it is committed as a standalone `fix:` and not as a fixup of any commit here — it is arguably separable from this PR entirely. Also, `map_two_species_all`, the enumerating variant this PR adds, already handled the dict case correctly; only the older single-map function was wrong.

Regression test `test_inc_vals_dict` in `arc/mapping/engine_test.py` asserts the domain is untouched and that every value moved by exactly `inc_vals`. It was mutation-checked: reverting the fix makes it fail and leaves the rest of the suite green.

arc/reaction/reaction.py
```diff
@@ -169,6 +171,36 @@ def atom_map(self, value):
...
+
+        Not persisted by ``as_dict``, unlike ``atom_map`` - it is a derived quantity that can always be
+        recomputed from the species.
+        """
+        if self._atom_map_clusters is None \
+                and all(species.get_xyz(generate=False) is not None for species in self.r_species + self.p_species):
+            self._atom_map_clusters = map_reaction_clusters(rxn=self, backend='ARC')
+        if not self._atom_map_clusters:
+            logger.error(f"The requested ARC reaction {self} could not be atom mapped into channels.")
+        return self._atom_map_clusters
```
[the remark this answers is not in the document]
kfir4444: Right on both counts, and fixed in f32b9d93.

Reverting the fix quantifies it: the error was logged **twice** when no mapping had even been attempted (missing geometries, two accesses), and **three times instead of once** across three accesses after an empty result. So both failure modes you name were real — indiscriminate spam, and "not yet computable" being indistinguishable from a genuine mapping failure.

The property now separates the three cases:

- **Geometries missing** — nothing was attempted, so it logs at `debug`, returns `None`, and caches nothing. Not reported as a failure, and it stays retryable once geometries arrive.
- **Attempted and empty** — a single `logger.error`, immediately after the attempt that failed. The empty result is cached.
- **Cached** — returned as-is, silently, with no recomputation.

Two regression tests in `arc/mapping/cluster_test.py`, both using `mock`:

- `test_atom_map_clusters_does_not_log_an_error_without_geometries` asserts `map_reaction_clusters` is never called, no error is logged across repeated accesses, and nothing is cached.
- `test_atom_map_clusters_logs_a_failure_once_and_caches_it` accesses the property three times against a stubbed empty result and asserts exactly one computation and exactly one error.

Both mutation-checked — reverting the fix fails them with the 2x and 3x counts above.

arc/mapping/cluster.py
```diff
@@ -0,0 +1,852 @@
...
+    orbits = core_orbits(r_graph, automorphisms)
+
+    scored: list[tuple[list[int], frozenset]] = list()
+    for atom_map in atom_maps:
+        if len(atom_map) != r_graph.n_atoms:
+            logger.warning(f'Skipping an atom map of length {len(atom_map)} for {rxn}, '
+                           f'expected {r_graph.n_atoms}.')
+            continue
+        scored.append((atom_map, changed_bonds(r_graph, p_graph, atom_map,
+                                               ignore_bond_orders=ignore_bond_orders)))
```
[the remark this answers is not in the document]
kfir4444: Confirmed and fixed in fdbafb1d — and the blast radius is worse than a single bad map.

`changed_bonds` builds `inverse_map` by inverting the atom map, so a duplicated product index leaves that dict missing a key and the subsequent `inverse_map[i]` lookup raises. Reproduced by reverting the guard and feeding a same-length non-permutation map to the CH4 + OH case: `KeyError: 6`. Because the exception propagates out of the loop, it aborts the **entire** clustering run rather than discarding just the offending map — one malformed entry takes down every channel for the reaction.

Added `is_permutation_map(atom_map, n_reactant_atoms, n_product_atoms)`, which replaces the length-only check in `cluster_atom_maps`; maps that fail it are skipped with a warning and clustering proceeds on the rest. It validates length, integrality, range and distinctness, and checks the map is *onto* the product complex, since that is the condition inverting it actually requires. Entries are coerced with `operator.index()` so numpy integers — which the mapping pipeline does produce — still pass, while floats, `None` and strings are rejected.

Tests in `arc/mapping/cluster_test.py`:

- `TestPermutationValidation`, 7 cases: duplicates, wrong length, out-of-range, negative, non-integer, non-sequence, and a map onto a differently sized product.
- `test_cluster_atom_maps_skips_a_non_permutation_map` — a malformed map alongside the valid ones is dropped while the valid ones still cluster to one channel of degeneracy 4.
- `test_cluster_atom_maps_survives_only_malformed_input` — returns no channels instead of raising.

Both integration tests fail with the `KeyError` when the guard is reverted.

arc/mapping/cluster.py
```diff
@@ -0,0 +1,883 @@
...
+    hydrogen_counts = {v: 0 for v in graph.core}
+    for hydrogen, parent in graph.parent.items():
+        hydrogen_counts[parent] = hydrogen_counts.get(parent, 0) + 1
+    named: dict[int, set] = dict()
+    for i, j, _, _ in center:
+        for index in (i, j):
+            if index in graph.parent:
+                named.setdefault(graph.parent[index], set()).add(index)
+
+    hydrogen_factor = 1
```
calvinp0, line 578: the `n!/(n-k)!` factor counts ordered assignments, but the centre is set - so when two hydrogens on the same parent are interchangeable it counts each pair twice.
`CH2O <=> CO + H2` - `1,2 Insertion CO` comes out as 2; RMG gives 1 and so does a direct orbit count under the docstring definition.

arc/mapping/cluster.py
```diff
@@ -0,0 +1,883 @@
...
+        hydrogens even though the true degeneracy is 3. Compare the two to gauge enumeration coverage.
+        """
+        return len(self.centers)
+
+    def __repr__(self) -> str:
+        return f'<MapCluster degeneracy={self.degeneracy} enumerated={self.enumerated_degeneracy} ' \
+               f'maps={len(self.members)} signature={self.signature}>'
+
+
+def build_complex_graph(species_list: list[ARCSpecies]) -> ComplexGraph:
```
calvinp0, line 144: The degeneracy is computed from the one Lewis structure ARC holds, so a radical spread over equivalent sites looks less symmetric than it is. Measured through map_reaction_clusters, against RMG in the same forward direction:

| reaction                     | RMG | here |
|------------------------------|-----|------|
| allyl + CH4 → propene + CH3  | 8   | 4    |
| NO2 + CH4 → HONO + CH3       | 8   | 4    |
| c-C5H5 + CH4 → c-C5H6 + CH3  | 20  | 4    |
| benzyl + CH4 → toluene + CH3 | 4   | 4    |

`enumerated_degeneracy` matches degeneracy in all four.

Benzel agrees becuase its radical carbon is already the carbon with two hydrogens and its ring sites give a different product.

arc/reaction/reaction.py
```diff
@@ -169,6 +171,43 @@ def atom_map(self, value):
         """Allow setting the atom map"""
         self._atom_map = value
 
+    @property
```
calvinp0, line 174: for `C2H5OO <=> C2H4 + HO2` and `C2H5 + C2H5 <=> C2H6 + C2H4`, the map that `rxn.atom_map` returns is not in any cluster. Its center has 11 and 10 changed bonds against the recipes 3 and 2. Would be good to log it?

arc/mapping/cluster.py
```diff
@@ -0,0 +1,883 @@
...
+    scored: list[tuple[list[int], frozenset]] = list()
+    for atom_map in atom_maps:
+        if not is_permutation_map(atom_map, r_graph.n_atoms, p_graph.n_atoms):
+            logger.warning(f'Skipping an atom map for {rxn} that is not a bijection of the '
+                           f'{r_graph.n_atoms} reactant atoms onto the {p_graph.n_atoms} product '
+                           f'atoms: {atom_map}.')
+            continue
+        scored.append((atom_map, changed_bonds(r_graph, p_graph, atom_map,
+                                               ignore_bond_orders=ignore_bond_orders)))
+    if validate_centers and scored:
```
calvinp0, line 791: `expected_reaction_centers` returns `None` for every `R_Recombination` reaction, so only the remaining fileter is `filter_minimal_centers`.
`CH3 + CH3 <=> C2H6` reports one channel with a 13-bond centre and degeneracy 36.

arc/mapping/cluster.py
```diff
@@ -0,0 +1,883 @@
...
+
+    Args:
+        center (frozenset): The changed-bond set, in reactant indices and *not* hydrogen-collapsed.
+        graph (ComplexGraph): The reactant complex.
+        automorphisms (list[dict[int, int]]): The core-skeleton automorphism group.
+
+    Returns:
+        int: The reaction path degeneracy.
+    """
+    if not center:
```
calvinp0, line 550: `1,2-Birad_to_alkene` and `Singlet_Val6_to_triplet` only change bond orders, so with `ignore_bond_orders=True` the changed-bond set is empty and the cluster reports `degeneracy=0`. Was check on `[CH2][CH2] <=> C=C`. The empty centre also switches off both validity filters,

arc/mapping/cluster.py
```diff
@@ -0,0 +1,883 @@
...
+    for i, j, _, _ in center:
+        for index in (i, j):
+            if index in graph.parent:
+                named.setdefault(graph.parent[index], set()).add(index)
+
+    hydrogen_factor = 1
+    for parent, hydrogens in named.items():
+        available = hydrogen_counts.get(parent, 0)
+        for offset in range(len(hydrogens)):
+            hydrogen_factor *= max(available - offset, 1)
```
calvinp0: i think the clamp is unreachable?

arc/mapping/cluster.py
```diff
@@ -0,0 +1,883 @@
...
+        return HYDROGEN, graph.parent[index]
+    return CORE, index
+
+
+def center_degeneracy(center: frozenset,
+                      graph: ComplexGraph,
+                      automorphisms: list[dict[int, int]],
+                      ) -> int:
+    """
+    The exact reaction path degeneracy of a reaction center: the size of its orbit under the *full*
```
calvinp0, line 516: Not sure if this is impactful on the work but RMG values for degeneracy differ by a factor of 2 whenever the reactants are identical - `C2H5 + C2H5 <=> C4H10` is 1 here and 0.5 in RMG (RMG halves deliberately in `reduce_same_reactant_degeneracy`, Bishop & Laidler 1965)

calvinp0 (changes requested):

arc/mapping/cluster.py
kfir4444: Confirmed, and fixed in 59631c4c. You are exactly right about the cause: the falling factorial is the orbit of an *ordered* tuple of hydrogens, but the center is a set, so any permutation of hydrogens that reproduces the same set was being counted as a separate path.

Reproduced at unit level on CH2O, both C-H bonds breaking into H2:

```
reported=2   true orbit=1     <- both hydrogens in the same role
reported=2   true orbit=2     <- control, one hydrogen named
```

The fix divides the falling factorial by the stabilizer of the center under permutations of that parent's named hydrogens (`hydrogen_stabilizer`). For CH2O: `2!/0! = 2`, one role class of size 2, `2/2! = 1`. The control is untouched at 2, and a third test pins the case that must *not* collapse — two hydrogens on one carbon in different roles (one breaks, one changes order) stay two paths.

I cross-checked `center_degeneracy` against an exhaustive orbit count over the full group (core automorphisms x hydrogen bijections) rather than trusting the factorisation: CH2O 1 and 2, CH4+OH 4, C2H6+OH 6, cyclohexane+OH 12, propane 6 - all agree. End to end, `CH2O <=> CO + H2` now reports degeneracy 1, matching RMG.

arc/mapping/cluster.py
kfir4444: Confirmed, and fixed in 97109c14. Your diagnosis was right and the mechanism is precisely the one you name: `graph.invariant` seeded the colour refinement with `radical_electrons` (and lone pairs and formal charge), all of which are properties of the *one* Lewis structure ARC happens to hold. Allyl's two terminal carbons were then distinguished only by which end the stored structure puts the radical on, which halved every degeneracy computed through it.

`ignore_bond_orders` already normalized bond orders for exactly this reason - Kekule structures - but stopped at the bonds. Resonance moves the radical, the lone pairs and the formal charge too, so `effective_invariant` now drops all of them when orders are ignored, leaving element and attached-hydrogen count, which resonance preserves.

Your table, re-measured:

| reaction | RMG | before | now |
|---|---|---|---|
| allyl + CH4 | 8 | 4 | **8** |
| NO2 + CH4 | 8 | 4 | **8** |
| c-C5H5 + CH4 | 20 | 4 | **20** |
| benzyl + CH4 | 4 | 4 | **4** |

Benzyl is the important control, and your explanation of why it already agreed is the reason it is in the test: its radical carbon is genuinely unique, so it must stay 4 and not inflate. It does.

The trade-off is documented on `effective_invariant`: two fragments differing *only* in charge or radical count, with identical element, connectivity and hydrogen count, are now seen as interchangeable. That is the correct reading for resonance forms of one species, and the price paid elsewhere.

arc/reaction/reaction.py
kfir4444: Agreed, and it is now logged - `warn_if_atom_map_is_unclustered`, in 97109c14. It reports the map's center size against the channels', at warning level:

```
The atom map of <rxn> is in none of its N enumerated reaction channels: it changes 11 bonds,
against [3] for the channels. atom_map is most likely scrambled for this reaction.
```

Two deliberate constraints. It only inspects the atom map when it has *already* been computed - reading `rxn.atom_map` there would run the whole single-map pipeline as a side effect of clustering, which this module goes out of its way never to do, and there is a test pinning that it does not.

Second, and worth stating plainly: I think this is evidence about `map_reaction`, not about clustering. The channels are filtered against the family recipe, so a map outside all of them with 11 changed bonds against a recipe predicting 3 means the single-map pipeline returned a scrambled result. I hit the same thing independently on `CH3 + CH3 <=> C2H6`, where the sole enumerated map sends one methyl's hydrogens to the *other* methyl's carbon; handing the chemically correct map to the same code gives a clean one-bond center. So there is a real mapping defect upstream. I have not tried to fix `map_rxn_all` here - it is a much bigger change than this PR - but it is now visible rather than silent.

arc/mapping/cluster.py
kfir4444: Confirmed and fixed in 97109c14, and the cause was recoverable. `R_Recombination`'s recipe is `FORM_BOND(*, *)` - one label, two atoms - and `get_expected_changing_bonds` reads only the base label, so both endpoints resolved to the same index and the bond looked degenerate. But the label map does carry the second site under a suffixed key:

```
r_label_map: {'*': 0, '*_2': 4}      get_expected_changing_bonds -> ([], [(0, 0)])
```

`resolve_duplicate_label` recovers it, so the predicted center for `CH3 + CH3 <=> C2H6` is now `{(0, 4, 0, 1)}` - the forming C-C bond, correct. Only when no suffixed sibling exists is the dictionary skipped, so the existing guard against a genuinely degenerate action still holds.

The degeneracy is fixed as a result: **36 -> 1**.

The 13-bond center, though, is not a clustering artifact - it is a real defect in the enumeration, and the now-working recipe reference proves it by rejecting every enumerated map. The single map `map_rxn_all` produces sends one methyl's hydrogens to the other methyl's carbon; the correct map `[0, 2, 3, 4, 1, 5, 6, 7]` gives a one-bond center and degeneracy 1 through the same code path. Same root cause as your comment on `rxn.atom_map`.

Since the fallback to `filter_minimal_centers` cannot reject a uniformly scrambled enumeration, `MapCluster` now carries `recipe_validated`. A channel the recipe rejected is no longer presented as recipe-backed, so `False` next to an implausible center reads as "the enumeration is wrong here", not "this is a reaction channel".

arc/mapping/cluster.py
kfir4444: Confirmed on both halves, and fixed in 59631c4c and 97109c14.

The degeneracy first: an empty center returned 0, which is not a meaningful reaction path degeneracy for a reaction that exists. The orbit of the empty set under any group is the empty set, so the answer is 1, and it now returns 1.

That alone would have been papering over the real problem, which is the one you point at: with `ignore_bond_orders=True` the center of an order-only family is empty, and an empty center carries no information at all - nothing to cluster on, and, as you note, nothing for either validity filter to judge. So `cluster_atom_maps` now detects that *every* center came out empty and re-clusters honoring bond orders, which is the only setting that can describe these families.

On `[CH2][CH2] <=> C=C`:

```
center, orders ignored : []
center, orders honored : [(0, 1, 1.0, 2.0)]
result                 : 1 channel, degeneracy 1
```

The C-C bond going single to double is the reaction center, and one channel of degeneracy 1 is right - there is only one way to pair the two radical electrons. Note the recipe-based reference is unavailable with orders honored, since the recipe describes only breaking and formation, so validation falls back to the relative minimal-center filter there; that is called out in the code comment.

arc/mapping/cluster.py
kfir4444: You are right, it is unreachable, and it is gone in 59631c4c.

`named[parent]` is built only from hydrogens whose parent is that atom, so it is a subset of the hydrogens on it: `len(hydrogens) <= available`. The loop runs `offset` over `range(len(hydrogens))`, so the smallest factor is `available - (len(hydrogens) - 1) >= 1`, and `max(..., 1)` can never bind. Measured across the graphs in the test suite the minimum factor is exactly 1, i.e. the clamp sits precisely at its boundary and never below it.

Replaced with a comment stating the subset relation, since that is the fact that makes the clamp unnecessary and is worth keeping in view.

arc/mapping/cluster.py
kfir4444: Good catch, and it is a genuine convention difference rather than an error on either side, so I have documented it rather than changed the number.

What this reports is the orbit size of the reaction center under the full automorphism group of the reactant complex. RMG deliberately halves that for identical reactants in `reduce_same_reactant_degeneracy` (Bishop and Laidler 1965), so that a rate expression does not double count two indistinguishable colliding partners. Both are correct for their own purpose; they answer different questions.

Since ARC exposes this as a property of the channel rather than as a rate coefficient, halving it here would be wrong for anyone counting distinct transition states, and not halving it is wrong for anyone folding it straight into a rate. The `center_degeneracy` docstring now states which of the two it is, names `reduce_same_reactant_degeneracy` and the Bishop and Laidler reference, gives your `C2H5 + C2H5 <=> C4H10` example as 1 here against 0.5 in RMG, and says explicitly that a caller building a rate coefficient for identical reactants has to apply the factor itself.

If ARC ever consumes this in a rate expression, that is the place it has to be applied, and I would rather it be an explicit step there than a silent factor buried in the orbit count.

calvinp0 (approved):
Great work!

PR #1025, opened 2026-08-26
fix(species): coerce and validate ARCSpecies.run_time at the constructor boundary

alongd:
## What

`ARCSpecies.__init__` stored `run_time` unexamined, even though the annotation and class docstring promise a `datetime.timedelta`. A caller passing seconds as a plain number — `ARCSpecies(label='x', run_time=12.5)` — was accepted silently at construction and only failed much later at serialization, in `as_dict`, with `AttributeError: 'float' object has no attribute 'total_seconds'`. The traceback pointed at `as_dict`, not at where the mistake was made. `from_dict` already coerced on the way in; the constructor was the asymmetry.

## Change

Add a single module-level helper `process_run_time()` in `arc/species/species.py` and route both entry points (constructor at `:380` and `from_dict` at `:873`) through it:

- `None` → stays `None`
- `datetime.timedelta` → stored unchanged
- `int` / `float` → interpreted as **seconds** (the existing `total_seconds()` round-trip convention) and converted to `timedelta`
- anything else → immediate `SpeciesError` naming the parameter, the type received, and the accepted types

**Booleans raise.** `bool` is an `int` subclass, so it is checked first and rejected: a boolean `run_time` is almost certainly a caller mistake, and coercing `True`→1 s would manufacture a plausible-but-wrong duration — the exact fail-late-with-a-wrong-value class this change removes.

The public type of `run_time` is unchanged (`timedelta`); this makes the object honour it. No schema, data-file, migration, or API-signature changes.

## Tests

New `test_run_time_coercion` (round-trip through `as_dict`/`species_dict=`, `timedelta` stored unchanged, `None` stays `None`, `str` and `bool` raise `SpeciesError` at construction with the message asserted) and `test_process_run_time` (helper unit test) in `arc/species/species_test.py`.

- `arc/species/species_test.py` — 95 passed
- `arc/scheduler_test.py` — 52 passed

Mutation-checked: reverting only the constructor wiring makes `test_run_time_coercion` fail on the coercion assertion; restoring it passes.

### DIFF
--- a/arc/species/species.py
+++ b/arc/species/species.py
@@ -6,7 +6,7 @@ If the species is a transition state (TS), its ``ts_guesses`` attribute will hav
 import datetime
 import numpy as np
 import os
-from math import isclose
+from math import isclose, isfinite
 from typing import Any
 
 import arc.molecule.element as elements
@@ -126,7 +126,8 @@ class ARCSpecies(object):
         rxn_index (int, optional): The reaction index which is the respective key to the Scheduler rxn_dict.
         external_symmetry (int, optional): The external symmetry of the species (not including rotor symmetries).
         optical_isomers (int, optional): Whether (=2) or not (=1) the species has chiral center/s.
-        run_time (timedelta, optional): Overall species execution time.
+        run_time (timedelta | int | float, optional): Overall species execution time,
+                                                     an ``int``/``float`` is interpreted as seconds.
         checkfile (str, optional): The local path to the latest checkfile by Gaussian for the species.
         number_of_radicals (int, optional): The number of radicals (inputted by the user, ARC won't attempt to determine
                                             it). Defaults to None. Important, e.g., if a Species is a bi-rad singlet,
@@ -338,7 +339,7 @@ class ARCSpecies(object):
                  number_of_radicals: int | None = None,
                  optical_isomers: int | None = None,
                  preserve_param_in_scan: list | None = None,
-                 run_time: datetime.timedelta | None = None,
+                 run_time: datetime.timedelta | int | float | None = None,
                  rxn_label: str | None = None,
                  rxn_index: int | None = None,
                  smiles: str = '',
@@ -377,7 +378,7 @@ class ARCSpecies(object):
         self.active = active
         self.optical_isomers = optical_isomers
         self.charge = charge
-        self.run_time = run_time
+        self.run_time = process_run_time(run_time)
         self.checkfile = checkfile
         self.transport_data = TransportData()
         self.yml_path = yml_path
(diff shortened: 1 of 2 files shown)


arc/species/species.py
LeenFahoum, line 1: Type hint is stale - process_run_time (and the new tests) accept int/float seconds here too, not just timedelta.   I think it should be : 
`run_time: datetime.timedelta | int | float | None = None,`

arc/species/species.py
```diff
@@ -2938,6 +2938,31 @@ def __reduce__(self):
                                 self.polarizability, self.rotrelaxcollnum, self.comment))
 
 
+def process_run_time(run_time: datetime.timedelta | int | float | None,
```
LeenFahoum, line 2942: `process_run_time` currently doesn't validate that numeric inputs are finite before passing them to `datetime.timedelta()`. As a result, `run_time=float('inf') raises OverflowError`, while `run_time=float('nan')` raises `ValueError`.

Per the docstring's `Raises` section, invalid `run_time` values should consistently raise `SpeciesError`. I thinko It would be good to add an explicit finiteness check before constructing the `timedelta`:

```
if isinstance(run_time, bool) or not isinstance(run_time, (int, float)):
    raise SpeciesError(
        f'run_time must be a datetime.timedelta, an int/float number of seconds, '
        f'or None, got {type(run_time).__name__}: {run_time!r}'
    )
if not math.isfinite(run_time):
    raise SpeciesError(f'run_time must be finite, got {run_time!r}')
return datetime.timedelta(seconds=run_time)
```

LeenFahoum (changes requested):
Looks good overall, but I’m requesting changes for the two issues noted above: handling non-finite `run_time` values consistently and updating the corresponding type hint.

arc/species/species.py
alongd: Right, fixed. Signature is now `run_time: datetime.timedelta | int | float | None = None`, and the class docstring entry says an int/float is read as seconds. The attribute docstring still says `timedelta`, since that is what is actually stored after coercion.

arc/species/species.py
alongd: Good catch. `inf` raised `OverflowError` and `nan` raised `ValueError`, both escaping the documented `SpeciesError` contract. Added an `isfinite()` check (using `from math import isclose, isfinite`, which the module already imports from) that raises `SpeciesError` before constructing the `timedelta`, plus tests for `inf`/`-inf`/`nan` in both `test_process_run_time` and `test_run_time_coercion`.

LeenFahoum (approved):
Looks good now!

PR #1026, opened 2026-08-26
fix(orca): implement NMD parsing, guard trsh against None

Mia1Gohary:
Orca's parse_normal_mode_displacement() was a stub returning (None, None). trsh_negative_freq() then called len(None) on any species with a genuine imaginary frequency, crashing the whole ARC process instead of nudging the geometry and re-optimizing.

Implement real parsing of Orca's VIBRATIONAL FREQUENCIES / NORMAL MODES blocks, and add a defensive None check in trsh.py so any adapter lacking this capability degrades gracefully instead of crashing.

Validated against CR_1's r1 freq job on zeus: correctly identified and parsed the -33.01 cm^-1 imaginary mode; live run confirmed freq troubleshooting completes (freq: True) where it previously crashed.


arc/parser/adapters/orca.py
```diff
@@ -155,13 +155,77 @@ def parse_frequencies(self) -> np.ndarray | None:
...
+                - normal mode displacements, shape (num_modes, num_atoms, 3), same mode order as frequencies.
         """
-        # Not implemented for Orca.
-        return None, None
+        with open(self.log_file_path, 'r') as f:
+            lines = f.readlines()
+
+        all_freqs = list()
+        for i, line in enumerate(lines):
+            if 'VIBRATIONAL FREQUENCIES' in line:
```
calvinp0: I believe we do vibrational freq parsing in a function
https://github.com/ReactionMechanismGenerator/ARC/blob/14dca8a259e9bc10ff0b2585b4ec873851ee4888/arc/parser/adapters/orca.py#L114

arc/parser/parser_test.py
```diff
@@ -498,6 +498,31 @@ def test_get_normal_mode_displacement_returns_none_for_a_file_it_cannot_read(sel
         self.assertIn(missing_path, warning)
         self.assertIn('TS', warning)
 
+        path = os.path.join(ARC_TESTING_PATH, 'normal_mode', 'n_cetane', 'output.log')
```
calvinp0, line 505: Two small things. These are appended to test_get_normal_mode_displacement_returns_none_for_a_file_it_cannot_read, which is about a missing file. So I think they'd be better as their own method.

the assertions check that the parser read the right numbers out of the file, but not that the numbers are the right kind of vector. ARC need cartesian displacement. `get_displaced_xyzs` adds them straight to a geometry. 
If a log ever gave mass-weighted ones instead, the golden array would just hold different numbers, and the natural fix would be to regenerate it, so the test would stay green while every TS got displaced along a heavy-atom-tilted vector.
so, then maybe an extra assertion where a real vibration has no net momentum so $\sum{}{}m_{a} d_{a} \approx{0}$ holds fo cartesian displacements and fails for mass weighted ones. so on these files it measures $-1e-5$ versus 0.6 if the modes were mass weighted. so it discriminates.
The `assert_array_equal(freqs, parse_frequencies(...))` line also pins the index alignment that nmd.py and trsh.py rely on.

calvinp0 (changes requested):

calvinp0:
It's looking good @Mia1Gohary , just two comments for fixing!

arc/parser/parser_test.py
Mia1Gohary: Thanks!
-I moved the n-cetane checks into their own test_parse_orca_normal_mode_displacement method.
-I also added assert_array_equal to verify frequency/mode index alignment and a mass-weighted-sum assertion to confirm that the parsed vectors are Cartesian displacements.

arc/parser/adapters/orca.py
Mia1Gohary: Thanks! I refactored the frequency-reading logic into a shared _parse_frequencies() helper. Both parse_frequencies() and parse_normal_mode_displacement() now use it. The normal-mode parser retains zero-frequency translation/rotation modes internally so its matrix columns remain correctly aligned, while the public parse_frequencies() result still excludes them.

arc/parser/adapters/orca.py
```diff
@@ -153,15 +156,72 @@ def parse_frequencies(self) -> np.ndarray | None:
...
+            stripped = lines[i].strip()
+            if stripped and '.' not in stripped:
+                try:
+                    col_indices = [int(tok) for tok in stripped.split()]
+                except ValueError:
+                    i += 1
+                    continue
+                i += 1
+                for row in range(n_dof):
+                    vals = lines[i].split()[1:]
```
calvinp0, line 208: `lines[i]` is not bounds-checked here, so a log truncated inside the `NORMAL MODES` block raises `IndexError`. `trsh` is fine - `get_normal_mode_displacement` catches it but `nmd.py:71` only catches `NotImplementedError` so it propogates there.

calvinp0 (changes requested):

arc/parser/adapters/orca.py
Mia1Gohary: Thanks! fixed this.

calvinp0 (dismissed):
lgtm! thanks @Mia1Gohary

arc/parser/parser_test.py
```diff
@@ -480,8 +480,7 @@ def test_get_normal_mode_displacement_returns_the_parsed_displacements(self):
 
     def test_get_normal_mode_displacement_returns_none_when_no_modes_are_reported(self):
         """Test that an output file yielding no normal mode displacements returns None, not the (None, None) sentinel."""
-        for file_name in ['orca_neg_freq_ts.out', 'orca_example_freq.log', 'orca6_example.out',
```
calvinp0, line 483: Rather than remove the usage of these logs, we can now use them in tests.
- `orca_neg_freq_ts.out`: we can make this a positive assertion now
- `orca_example_freq.log `: also make this a positive assertion somewhere
- `orca6_example.out`: still returns (None, None) so it can still stay in this line

arc/job/trsh_test.py
```diff
@@ -989,8 +989,7 @@ def test_trsh_negative_freq(self):
 
     def test_trsh_negative_freq_for_an_ess_without_normal_mode_displacements(self):
         """Test that an ESS output file reporting no normal mode displacements is reported, not indexed."""
-        for file_name in ['orca_neg_freq_ts.out', 'orca_example_freq.log', 'orca6_example.out',
```
calvinp0, line 992: Rather than remove the usage of these logs, we can now use them in tests.
- `orca_neg_freq_ts.out`: we can make this a positive assertion now
- `orca_example_freq.log `: also make this a positive assertion somewhere
- `orca6_example.out`: still returns (None, None) so it can still stay in this line

arc/checks/nmd_test.py
```diff
@@ -765,8 +766,7 @@ def test_analyze_ts_normal_mode_displacement_for_hypervalence_nitrogen(self):
 
     def test_analyze_ts_normal_mode_displacement_skips_an_unsupported_ess(self):
         """Test that an ESS ARC cannot parse normal mode displacements from is skipped rather than raising."""
-        for file_name in ['orca_neg_freq_ts.out', 'orca_example_freq.log', 'CH2O_freq_molpro.out',
-                          'C2H6_freq_QChem.out', 'CH2O_freq_terachem.dat']:
```
calvinp0, line 769: right to take out the log files but that leaves `a supported ESS whose output has no normal modes` uncovered, so `orca6_example.out` could be the fix for that

calvinp0 (changes requested):

arc/checks/nmd_test.py
Mia1Gohary: Thanks, done!

calvinp0 (approved):

PR #1027, opened 2026-08-26
pipe: carry rotor scan resolution through the task payload

alongd:
## Motivation

A piped 1D rotor scan (`rotor_scan_1d`, the path MLIP/ASE rotors take) ran at **whatever
`rotor_scan_resolution` the worker node's settings happened to hold** — not the resolution the run
intended. The `TaskSpec` carried no `scan_res`, so the value simply rebound on the compute node.

This is the same silent-wrong-answer class as #1019: below 18 points per rotor RMG-Py's Fourier
fitter never runs and `get_potential()` reads an uninitialised C double. Nothing crashes; the
torsional contribution is just wrong. #1019 makes resolution a property of the run at the input
layer; this closes the remaining hole, where the value was lost crossing the pipe boundary.

## What changed

Capture the resolution **at staging time**, on the machine that knows what the run asked for, and
forward it in the worker:

- `pipe_planner.py` / `pipe_run.py` — `build_rotor_scan_1d_tasks` records `scan_res` into the
  `rotor_scan_1d` task payload. The value follows the same precedence `Scheduler.set_scan_resolution`
  applies to ESS scans: the run-level `rotor_scan_resolution` input key (#1019) wins, and the
  launching host's settings value is the fallback when it is unset.
- `pipe_worker.py` — forwards it as `args['trsh']['scan_res']`, the existing per-job override
  channel (the same one #1019 uses), so no new mechanism is introduced.

Absent a `scan_res` in the payload, behaviour is unchanged: the worker's settings default still
applies, exactly as before.

## Tests

New coverage in `pipe_run_test.py` and `pipe_worker_test.py` pinning that the resolution is
captured at staging and forwarded at execution.

- `arc/job/pipe/`: 157 passed.
- `arc/job/adapters/ase_test.py arc/job/adapters/common_test.py arc/scheduler_test.py`: 86 passed.

Run under `arc_env` with `HOME` pointed at an empty dir (ARC's suite reads `~/.arc/settings.py`),
`-n0 -o addopts=""`.

An end-to-end check against the benchmark: an ethane pipe scan matched the reference potential to
**0.0279 kJ/mol**.

## Relationship to other PRs

Independent of #1019 and mergeable on its own — they close the same gap at two different layers
(#1019 at the input/scheduler boundary for ESS `scan` jobs, this at the pipe boundary for piped
rotor tasks). Rebased onto `main` now that #1018 (the one-process ASE relaxed scan this builds on)
has merged.

### DIFF
--- a/arc/job/pipe/pipe_planner.py
+++ b/arc/job/pipe/pipe_planner.py
@@ -52,6 +52,7 @@ if TYPE_CHECKING:
 logger = get_logger()
 
 pipe_settings = settings['pipe_settings']
+rotor_scan_resolution = settings['rotor_scan_resolution']
 
 
 class PipePlanner:
@@ -308,7 +309,8 @@ class PipePlanner:
             job_type='scan',
             build_tasks_fn=lambda adapter: build_rotor_scan_1d_tasks(
                 self.sched.species_dict[label], label, rotor_indices,
-                self._level_dict(level), adapter, self._memory_mb),
+                self._level_dict(level), adapter, self._memory_mb,
+                scan_res=rotor_scan_resolution),
             log_msg=f'Routing {len(rotor_indices)} 1D rotor scans for {label} to pipe mode',
         )
         return set(rotor_indices) if submitted else set()
--- a/arc/job/pipe/pipe_run.py
+++ b/arc/job/pipe/pipe_run.py
@@ -1049,8 +1049,15 @@ def build_ts_opt_tasks(species, label: str, xyzs: list[dict],
 
 def build_rotor_scan_1d_tasks(species, label: str, rotor_indices: list[int],
                               level_dict: dict, job_adapter: str,
-                              memory_mb: int) -> list[TaskSpec]:
-    """Build TaskSpec objects for 1D rotor scan tasks."""
+                              memory_mb: int, scan_res: float | None = None) -> list[TaskSpec]:
+    """
+    Build TaskSpec objects for 1D rotor scan tasks.
+
+    The scan resolution is captured here, at staging time on the ARC host, and carried in each
+    task's payload so the worker runs the resolution ARC intended rather than re-reading whatever
+    ``rotor_scan_resolution`` the worker node's settings happen to hold. When ``scan_res`` is
+    ``None`` the payload omits it and the worker falls back to its own settings default.
+    """
     cores = default_job_settings.get('job_cpu_cores', 8)
     species_dict_payload = species.as_dict()
     tasks = []
@@ -1059,6 +1066,13 @@ def build_rotor_scan_1d_tasks(species, label: str, rotor_indices: list[int],
         torsions = rotor['torsion']
         if isinstance(torsions[0], int):
             torsions = [torsions]
+        input_payload = {
+            'species_dicts': [species_dict_payload],
+            'torsions': torsions,
+            'rotor_index': ri,
+        }
+        if scan_res is not None:
+            input_payload['scan_res'] = scan_res
         tasks.append(TaskSpec(
             task_id=f'{label}_scan_r{ri}',
             task_family='rotor_scan_1d',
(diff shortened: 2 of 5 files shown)


arc/job/pipe/pipe_planner.py
```diff
@@ -308,7 +309,8 @@ def try_pipe_rotor_scans_1d(self, label: str, rotor_indices: list[int]) -> set[i
             job_type='scan',
             build_tasks_fn=lambda adapter: build_rotor_scan_1d_tasks(
                 self.sched.species_dict[label], label, rotor_indices,
-                self._level_dict(level), adapter, self._memory_mb),
+                self._level_dict(level), adapter, self._memory_mb,
+                scan_res=rotor_scan_resolution),
```
Ophir-Weisz, line 313: blocking: This forwards the settings default, not the run-level value. #1019 (now merged)
stores the input key on Scheduler.rotor_scan_resolution and injects it via
set_scan_resolution for job_type == 'scan' — and its description says the pipe path is
untouched, so this PR is the one meant to close that.

As written, a user who sets `rotor_scan_resolution: 4.0` in their input gets 4.0 on ESS
scans and 8.0 on piped rotors: the same silent rebinding this PR is fixing, just moved
from the worker's settings to the host's.

self.sched is already in this call two lines up (line 311), so something like
`self.sched.rotor_scan_resolution or rotor_scan_resolution` would prefer the run's value
and keep the settings default as the fallback. Am I reading the precedence right?

arc/job/pipe/pipe_planner.py
```diff
@@ -52,6 +52,7 @@
 logger = get_logger()
 
 pipe_settings = settings['pipe_settings']
+rotor_scan_resolution = settings['rotor_scan_resolution']
```
Ophir-Weisz, line 55: question: #1019 refuses a resolution coarser than 20 deg with InputError at parse time,
because below 18 points the Fourier fit never runs. That guard sits on the input key
only — a machine-wide settings override (which is the incident #1019's description
opens with) reaches this line unguarded, and check_argument_consistency only tests
divmod(360, scan_res)[1], which 60.0 passes.

Not a regression, since the worker read the same settings before. But this PR is framed
as closing the same silent-wrong-answer class, so: should the 18-point guard apply here
too, or is settings.py considered trusted in a way the input file isn't?

Ophir-Weisz (commented):
Traced the worker path — job_factory gets `args` from nowhere else, so the trsh channel
is clean and doesn't clobber anything.

One thing before I approve: line 313 forwards the settings default rather than
Scheduler.rotor_scan_resolution from #1019, so a user-set resolution looks like it'd
reach ESS scans but not piped rotors. Details inline, plus a scope question on the
18-point guard.

arc/job/pipe/pipe_planner.py
alongd: Done in `63637652e` — and yes, you're reading the precedence right.

The planner now takes `self.sched.rotor_scan_resolution or rotor_scan_resolution`, so the run-level input key wins and the settings value is only the fallback — the same order `set_scan_resolution()` applies to ESS scans. `check_rotor_scan_resolution` rejects anything `<= 0`, so the `or` can't swallow a legitimate value.

One trap writing the test: `_make_mock_sched` builds a `MagicMock`, whose `rotor_scan_resolution` attribute is truthy out of the box and would have posed as a run-level value. It's now pinned to `None`, which is what a real `Scheduler` defaults to. Two new cases in `pipe_planner_test.py` cover both directions (input key set → 4.0 in the payload; unset → the settings default).

arc/job/pipe/pipe_planner.py
alongd: Good question, but the guard already reaches this path, so settings.py isn't trusted more than the input file.

#1019 widened `check_argument_consistency` past the divmod test. On `main` it reads (`arc/job/adapters/common.py:709`):

```python
if obj.job_type == 'scan' and (obj.scan_res <= 0 or obj.scan_res > 20 or divmod(360, obj.scan_res)[1]):
```

and `ase_adapter.py:94` goes through `_initialize_adapter`, which sets `obj.scan_res` from `args['trsh']['scan_res']` or the settings default (`common.py:286-292`) and then runs that check. So the piped rotor hits the same chokepoint an ESS scan does.

Adding a staging-time check in `try_pipe_rotor_scans_1d` would close that.

PR #1037, opened 2026-09-01
Fix Gaussian 1D scan angles losing direction and gaining a spurious 360

calvinp0:
Gaussian 1D scan angles lost their direction on a reversed scan and gained a spurious 360 degrees on a
short one.

## The defect

`arc/parser/adapters/gaussian.py` normalised a scan's dihedrals by subtracting the first angle,
lifting negatives by 360, then adding 360 to the last point if it looked too small:

```python
angle -= angle[0]
angle[angle < 0] += 360.0
if len(angle) > 1 and angle[-1] < 2 * (angle[1] - angle[0]):
    angle[-1] += 360.0
```

**Reversed scans.** A scan with a negative step produces decreasing dihedrals, so after the
subtraction they are negative and the second line lifts the whole interior — not just the endpoint.
The H₂O₂ series read backwards came out as `[0, 350, 340, …, 20, 10, 360]`: not monotone, and the
direction is gone. On a reversed sBuOH scan the final point reached **719.999**.

**Short scans.** With two points, `angle[0]` is 0 after the subtraction, so the guard reduces to
`angle[1] < 2 * angle[1]` and holds for every positive angle. A 10 degree step parsed as `[0, 370]`.

Three points are not unconditionally affected — uniform steps give `60 < 60`, which is false. But
`arc/testing/rotor_scans/TS_errored.out`, already in the repository, has raw steps of 30.0001 and
29.9999 and therefore parses to `[0, 30, 420]` today.

That last case points at the origin: the heuristic is a mis-port of Arkane's, which compares against
the **declared** step size from the log (`arkane/ess/gaussian.py:520`). ARC substituted the first
**observed** step, which makes it sensitive to the printed precision of the dihedral.

## What the heuristic was protecting

A forward sweep completing a full turn. Gaussian reports dihedrals in (-180, 180], so the last point
of a 360 degree scan equals the first and reads as approximately 0, which would appear to run
backwards. `sBuOH.out` (46 points) and `H2O2.out` (37 points) are exactly this case; both still end
at 360, bit for bit unchanged.

## The fix

Per-step phase unwrapping. Each raw consecutive difference is wrapped into ±180 with
`arc.common.get_angle_in_180_range(step, round_to=None)` and cumulatively summed from zero. This
reads the direction from every raw step before any transformation, rather than inferring it from one
transformed step, and it subsumes direction handling instead of branching on it.

`round_to=None` is required: the default rounds to two decimal places, and rounding per step
accumulates across a long sweep.

Values are deliberately not folded into [0, 360). A sweep may legitimately exceed 360 where that
keeps it monotone, and monotonicity is the property callers rely on.

Verified identical to the previous behaviour on all eleven well-formed Gaussian 1D fixtures; the
output differs only where the previous answer was wrong.

## Consumers

Only one consumer reads the angles: `scheduler.py:3283` passes them to
`plotter.plot_1d_rotor_scan`, whose x-axis is already labelled "Dihedral angle increment".
`output.py:654`, `species.py:3026`, `species.py:3121` (`determine_rotor_symmetry`, which walks
energies by index) and `trsh.py`'s `scan_quality_check` all take element `[0]` — energies only. No
physical quantity changes. Arkane is unaffected: ARC hands it the raw log and RMG-Py performs its own
fold. `parse_1d_scan_energies_from_specific_angle` now yields a monotone absolute series for a
reversed scan where it previously yielded a scrambled one.

## Reuse

`np.unwrap(angle, period=360.0) - angle[0]` was considered and measured. It is numerically identical
on all fifteen fixtures, so the two differ only on an exactly-180-degree step — and there it is
backwards for real Gaussian input. Gaussian records such a scan as `[118.87, -61.13, 118.87]`, steps
of -180 and +180; `np.unwrap`'s asymmetric half-period rule gives `[0, -180, 0]`, which is not
monotone, where per-step unwrapping gives `[0, -180, -360]`.

`signed_angular_diff` was the other candidate and is semantically the closer name, but it has no
`round_to` passthrough, so it would reintroduce the accumulating rounding.

## Tests

`arc/parser/` — 44 passed, serially and under `-n 6 --dist worksteal`. Consumers: 238 passed across
`trsh_test`, `output_test`, `ase_test`, `plotter_test` and `species_test`.

Validated by mutation rather than coverage — five mutants, each killed by a test: the original
condition restored, direction handling dropped, values folded into [0, 360), a uniform-step
strawman, and the non-optimised-point deletion moved before the unwrap.

The repository had no genuine negative-step or two-point Gaussian log, so both fixtures are derived
from real per-point blocks of `H2O2.out` (reversed and truncated respectively, with the ModRedundant
directive changed to match). For H₂O₂ this is faithful: with four atoms, every non-scanned degree of
freedom has a unique minimum at each dihedral, so a genuine reversed scan gives the same energies.

https://claude.ai/code/session_01MbGeU8wLpafo3YFzTky2ho

### DIFF
--- a/arc/parser/adapters/gaussian.py
+++ b/arc/parser/adapters/gaussian.py
@@ -8,7 +8,7 @@ import numpy as np
 import pandas as pd
 import re
 
-from arc.common import SYMBOL_BY_NUMBER, is_same_pivot
+from arc.common import SYMBOL_BY_NUMBER, get_angle_in_180_range, is_same_pivot
 from arc.constants import E_h_kJmol, bohr_to_angstrom
 from arc.species.converter import str_to_xyz, xyz_from_data
 from arc.parser.adapter import ESSAdapter
@@ -326,6 +326,13 @@ class GaussianParser(ESSAdapter, ABC):
         """
         Parse the 1D torsion scan energies from an ESS log file.
 
+        The reported angles are the cumulative rotation relative to the first scan point,
+        accumulated from the per-step displacements of the raw dihedral values.
+        They start at zero, retain the sign of the scan direction (negative for a scan with a
+        negative step size), and are not folded into the 0-360 degree range, so a sweep that
+        completes a full turn ends at +/-360 degrees. Consecutive scan points recorded in the log
+        are assumed to be less than 180 degrees apart.
+
         Returns: tuple[list[float] | None, list[float] | None]
             The electronic energy in kJ/mol and the dihedral scan angle in degrees.
         """
@@ -376,11 +383,8 @@ class GaussianParser(ESSAdapter, ABC):
             angle = np.array(angle, float)
             if len(angle) != len(vlist):
                 return None, None
-            angle -= angle[0]
-            angle[angle < 0] += 360.0
-            if len(angle) > 1 and angle[-1] < 2 * (angle[1] - angle[0]):
-                angle[-1] += 360.0
-            angle = angle.tolist()
+            steps = [get_angle_in_180_range(step, round_to=None) for step in np.diff(angle)]
+            angle = np.concatenate(([0.0], np.cumsum(steps))).tolist() if len(steps) else [0.0]
 
         vlist = np.array(vlist, float)
         vlist -= np.min(vlist)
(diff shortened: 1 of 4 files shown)


kfir4444 (approved):
LGTM! great fix.

PR #1040, opened 2026-09-03
Skip the TS-GCN installer in the CI install lane

calvinp0:
## Why CI is red

Every CI run since 2026-09-02 fails in `make install-ci` before a single test executes, on `main` and on every open PR. The GCN installer builds the `ts_gcn` environment from `devtools/gcn_environment.yml`, whose PyTorch Geometric wheels (`torch-scatter`, `torch-sparse`, `torch-cluster`, `torch-spline-conv`) are linked from `pytorch-geometric.com/whl/torch-1.7.1+cpu.html` to `data.pyg.org`. That host stopped resolving on 2026-09-02 (dangling CNAME to a retired CloudFront distribution; tracked upstream at pyg-team/pyg-lib#719 with no ETA). `install_all.sh` runs under `set -euo pipefail`, so the failed pip step aborts the whole install.

## What we realised along the way

TS-GCN is not tested anywhere in our CI flow. `gcn_test.py` patches `TS_GCN_PYTHON` and mocks the subprocess boundary, `settings.py` resolves `TS_GCN_PYTHON` to `None` when the environment is absent, and `gcn_available()` already handles that. No workflow step runs the `ts_gcn` interpreter. CI has been building a heavyweight environment on every run and never using it.

## Change

- `devtools/install_all.sh` gains `--no-gcn`, the same shape as `--no-goflow` and `--no-rits`.
- `make install-ci` passes it. `make install` and `make install-gcn` still build the environment for real deployments.
- The workflow's PYTHONPATH step now adds only AutoTST. Nothing clones `../TS-GCN` on the runner any more, so the old line would print a `realpath` error on every run.

Verified that no other installer still in the CI lane fetches from the dead host (`install_goflow.sh` and `install_rits.sh` reference it but were already outside `install-ci`), that `docker_build.yml` never calls `install_all.sh`, and that the associative-array `unset` behaves under `set -u`.

https://claude.ai/code/session_0155RJ4VFRKpXQPEUBU2G4ho

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,11 +101,11 @@ jobs:
             bash ARC/devtools/install_rits.sh --cpu
           fi
 
-      - name: Set TS-GCN and AutoTST in PYTHONPATH
+      - name: Set AutoTST in PYTHONPATH
         shell: micromamba-shell {0}
         working-directory: ARC
         run: |
-          echo "PYTHONPATH=$(realpath ../TS-GCN):$(realpath ../AutoTST):$PYTHONPATH" >> $GITHUB_ENV
+          echo "PYTHONPATH=$(realpath ../AutoTST):$PYTHONPATH" >> $GITHUB_ENV
 
       - name: Compile ARC molecule
         shell: micromamba-shell {0}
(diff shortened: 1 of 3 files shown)


calvinp0:
data.pyg.org is back

Issue #1032, opened 2026-08-30
Atom mapper places hydrogens on the wrong heavy atom, returning chemically invalid maps


kfir4444:
## Summary

`map_reaction` returns atom maps that place hydrogens on the wrong heavy atom. The heavy-atom
correspondence is correct, but hydrogens belonging to one heavy atom are mapped onto hydrogens of a
*different* heavy atom, so the resulting map implies many spurious bond breaks and formations.

For `CH2 + C2H4 <=> C3H6` the carbons map `3 -> 0` and `4 -> 1`, while the hydrogens of those carbons map
as though `3 -> 1` and `4 -> 0`. The two carbons are effectively swapped for hydrogens only.

Reproduced on `main` at 83173030.

## Reproducer

Self-contained — no corpus files, no template lookups, no reference map. For a cycloaddition every X–H bond
present in the reactants is present in the products, so a correct map must satisfy
`parent_P(sigma(h)) == sigma(parent_R(h))` for every hydrogen `h`. The check uses only the returned map and
the two molecular graphs.

<details>
<summary><code>arc_bug_repro.py</code></summary>

```python
#!/usr/bin/env python
"""
Self-contained reproducer: ARC's atom mapper places hydrogens on the wrong heavy atom.

For a cycloaddition every X-H bond in the reactants is present in the products, so a correct atom map
satisfies parent_P(sigma(h)) == sigma(parent_R(h)) for every hydrogen h. The maps ARC returns for the two
reactions below violate this: hydrogens of one carbon are mapped onto hydrogens of a different carbon while
the carbons themselves map elsewhere.

The check uses only the returned map and the two molecular graphs -- no reference map, no template.

    python arc_bug_repro.py
"""

import logging
import os
from contextlib import redirect_stderr, redirect_stdout

CH2 = """1 *3 C u0 p1 c0 {2,S} {3,S}
2    H u0 p0 c0 {1,S}
3    H u0 p0 c0 {1,S}"""

C2H4 = """1 *1 C u0 p0 c0 {2,D} {3,S} {6,S}
2 *2 C u0 p0 c0 {1,D} {4,S} {5,S}
3    H u0 p0 c0 {1,S}
4    H u0 p0 c0 {2,S}
5    H u0 p0 c0 {2,S}
6    H u0 p0 c0 {1,S}"""

C3H6 = """1 *3 C u0 p0 c0 {2,S} {3,S} {4,S} {5,S}
[...]

PR #1041, opened 2026-09-07
Fixes the atom mapper placing hydrogens on the wrong heavy atom (issue #1032).

kfir4444:
When a reaction involves a symmetric fragment, several atom-map candidates tie on every geometric score, so the choice was made arbitrarily and then corrected after the fact, `glue_maps` forced the family template's labeled atoms into position by transposing them, but applied that swap only to the labeled atoms themselves, stranding the hydrogens bonded to them. The result was a map where the carbons are right and their hydrogens are attached to each other's carbon, implying spurious C–H breaks and formations.

The fix moves the template constraint upstream so nothing needs patching afterwards, at the two places the arbitrary choice was being made: fragment pairing now breaks ties between isomorphic cuts in favor of the one sharing the most template tags, and backbone-candidate selection filters to candidates satisfying the labeled-atom assignment before they're scored. Both new parameters are optional and default to the previous behavior, so existing callers are unaffected.

On the issue's two reactions the misplaced-hydrogen count goes 8 → 0, and across a 10-reaction probe spanning `Diels_alder_addition, 2+2_cycloaddition, Disproportionation and H_Abstraction`, the four cases where the template override previously degraded the map drop to zero — including a degenerate abstraction that main maps as "no reaction at all."

### DIFF
--- a/arc/mapping/engine.py
+++ b/arc/mapping/engine.py
@@ -93,7 +93,8 @@ def map_two_species(spc_1: ARCSpecies | Molecule,
     atom_map = None
 
     if backend.lower() == 'arc':
-        candidates = identify_backbone_candidates(spc_1, spc_2, consider_chirality=consider_chirality)
+        candidates = identify_backbone_candidates(spc_1, spc_2, consider_chirality=consider_chirality,
+                                                  constraint=constraint)
         if not candidates:
             return None
         candidates = filter_candidates_by_constraint(candidates=candidates,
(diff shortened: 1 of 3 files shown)


kfir4444:
Linked issue #1032

arc/mapping/engine.py
```diff
@@ -287,6 +296,41 @@ def identify_backbone_candidates(spc_1: ARCSpecies,
...
+    Returns:
+        list[dict[int, int]]: The candidates satisfying the constraint, or all of them if none does.
+    """
+    if not constraint or not candidates:
+        return candidates
+    backbone_constraint = {k: v for k, v in constraint.items() if any(k in c for c in candidates)}
+    if not backbone_constraint:
+        return candidates
+    filtered = [c for c in candidates if all(c.get(k) == v for k, v in backbone_constraint.items())]
+    if not filtered:
```
calvinp0, line 329: before falling back to unconstrained candidates, could we retyr the backbone search with the template correspondences fixed? like for CH2 + allene -> methylenecyclopropane, it appears there is a valid constraint compatible permutation exists, but the current search misses it.

calvinp0 (commented):
lgtm, just one suggestion

calvinp0 (approved):

PR #1044, opened 2026-09-07
Let the Linear TS adapter generate guesses for requested atom maps

### LINKED ISSUE #1042: LinearAdapter silently ignores a pre-set atom map, returning identical TS guesses for different maps
## Describe the bug

`LinearAdapter` recomputes the atom map internally, so an atom map set on the reaction beforehand has no effect on the transition-state guesses it returns. Calling the adapter with two different pre-set maps for the same reaction yields **byte-identical** guess geometries, with nothing in the log indicating that the supplied map was not used.

This may well be by design — the adapter is built to iterate over `rxn.product_dicts` and derive its own map for each. What makes it worth reporting is that there is no signal: `execute_incore()` reports guesses as successful either way, so a caller who supplies a map has no way to discover it was ignored.

The behavior is easy to read the other way, because the adapter snapshots and restores the attribute:

```python
# arc/job/adapters/ts/linear.py, execute_incore()
atom_map_snapshot = rxn._atom_map
...
rxn._atom_map = atom_map_snapshot
```

That protects the caller's value, but the value is never consumed. The consumption point is instead:

```python
# same file
atom_map = cached_map_rxn(rxn=rxn, product_dict_index=i, map_cache=map_cache)   # -> map_rxn(...)
```

## How to reproduce

On `main` @ `a24e13b0`:

```python
import arc.job.adapters.ts.linear as linear
from arc.job.adapters.ts.linear import LinearAdapter
from arc.reaction import ARCReaction
from arc.species.species import ARCSpecies

def build():
    return ARCReaction(r_species=[ARCSpecies(label='r', smiles='CC[CH]C')],
                       p_species=[ARCSpecies(label='p', smiles='CCC[CH2]')])

geometries = []
for forced in (map_a, map_b):          # two different valid maps for this reaction
    rxn = build()
    rxn._atom_map = list(forced)
    rxn.ts_species = None
    LinearAdapter(project='p', project_directory='/tmp/p', job_type='tsg',
                  reactions=[rxn], testing=True).execute_incore()
    good = [g for g in rxn.ts_species.ts_guesses if g.success and g.initial_xyz]
    geometries.append(good[0].initial_xyz)

# every pairwise coordinate RMSD is 0.000 A
```

I ran this over several distinct maps of one `intra_H_migration` reaction: each call reported 8 successful guesses, and all of them were the same geometry.

Monkeypatching `linear.map_rxn` to return a chosen map does change the output, which confirms that is the consumption point. It is not a clean workaround though — with one of the forced maps the adapter then produced zero guesses, so the guess strategies appear to assume the map they derived themselves.

## Expected behavior

Any one of these would resolve it, in decreasing order of usefulness:

1. Honor `rxn.atom_map` when it is already set, and derive a map only when it is not.
2. Accept an explicit map (or a list of maps) as an adapter argument, so a caller can request a guess for a particular correspondence.
3. If neither is wanted, log a warning when the adapter overwrites a pre-set map, and say in the docstring that guesses are generated per product dictionary rather than per supplied map.

The general point is that for a reaction with more than one valid correspondence, there is currently no supported way to ask for a transition-state guess for a specific one, and the attempt fails quietly rather than loudly.

## Installation information

- ARC `main` @ `a24e13b0`
- Python 3.14, Linux

kfir4444:
## Summary

`LinearAdapter` derives an atom map per reaction path via `map_rxn` and never reads a map set on the reaction, so pre-setting one changed nothing and was not reported. This adds `LinearAdapter(atom_map=...)`, taking one map or several: each requested map replaces the derived one for a full pass over the weight grid, so a TS guess can be requested for a specific reactants-to-products correspondence. Requested maps are validated (permutation, atom count, element-preserving) and unusable ones are reported rather than silently dropped. A pre-set `rxn.atom_map` still does not take part in the derivation - other adapters populate it as a side effect of reading it - but the adapter now warns that it is unused and points at the new argument, and the docstrings say guesses are generated per product dictionary.

Closes #1042

### DIFF
--- a/arc/job/adapters/ts/linear.py
+++ b/arc/job/adapters/ts/linear.py
@@ -28,7 +28,10 @@ Pipeline (per reaction path / ``product_dict``)
     (e.g., which of several equivalent H's migrates, etc.). A reaction-level
     fallback map is used when the per-path call fails; if no map is obtained,
     the per-path strategy pipeline is skipped and the trivial-map fallback
-    (below) takes over.
+    (below) takes over. An atom map set on the reaction beforehand does not
+    take part in this derivation; to ask for guesses for one specific
+    correspondence, pass it (or several) to ``LinearAdapter(atom_map=...)``,
+    which runs the whole pipeline once per requested map instead of deriving one.
 
 2. **Near-attack conformation** (:func:`get_near_attack_xyz`): the reactant
     and product geometries are pre-folded into the ring-like TS shape by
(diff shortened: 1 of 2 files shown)


kfir4444:
Tried this branch (merged onto `main` @ `4d67ad4c`) because I need guesses for particular correspondences, and it does what it says. Two observations from using it, one positive and one gap.

**It works.** Requesting a map per path now returns a distinct geometry per path. Before, every request came back byte-identical:

| case | paths | guesses | identical pairs |
|---|---|---|---|
| `CC[CH]C -> CCC[CH2]` | 6 | 6 | 0 |
| `CC[CH]CC -> CCCC[CH2]` | 6 | 6 | 0 |
| `CCC[CH2] -> CC[CH]C` | 4 | 4 | 0 |
| `[CH2]CC -> C[CH]C` | 2 | 2 | 0 |
| `[CH2]C(C)CC -> CC(C)[CH]C` | 2 | **1** | 0 |

**The gap is the last row.** For `[CH2]C(C)CC -> CC(C)[CH]C` (`intra_H_migration`), one of the two valid maps yields no guess at all. The two differ only in which of the two hydrogens on one carbon migrates, so they are about as similar as two paths can be, and the one that fails is the map that is close to the identity:

```
p0  map = [2, 1, 0, 3, 4, 9, 11, 8, ...]   guess produced
p1  map = [0, 1, 2, 3, 4, 5,  6, 8, ...]   no guess
```

`interpolate_isomerization` raises for every weight on that path; the exception is caught, so the adapter reports success overall and simply returns fewer guesses than maps requested:

```
File "arc/job/adapters/ts/linear.py", line 1059, in execute_incore
    xyzs = interpolate(rxn=rxn, weight=w, existing_xyzs=all_xyzs_so_far,
                       map_cache=map_cache, forced_atom_map=forced_atom_map)
File "arc/job/adapters/ts/linear.py", line 1813, in interpolate
    return interpolate_isomerization(...)
```

Two suggestions:

1. When an explicit `atom_map` is supplied and no guess survives for it, say so distinctly — an explicitly requested map returning nothing is different from a derived map that happened not to work, and a caller iterating over maps cannot currently tell the two apart without counting.
2. Worth checking whether near-identity maps are the trigger. I saw the same map fail before this branch existed, when I forced it by monkeypatching `map_rxn`, so it looks like a property of the interpolation rather than of the new argument.

Not a blocker for the branch as far as I'm concerned — it is a clear improvement over the previous behavior.

arc/job/adapters/ts/linear.py
```diff
@@ -1868,7 +2036,8 @@ def interpolate_addition(rxn: ARCReaction,
         # Convert to uni-species ordering.
         if uni_is_product:
             try:
-                atom_map = cached_map_rxn(rxn=rxn, product_dict_index=i, map_cache=map_cache)
+                atom_map = cached_map_rxn(rxn=rxn, product_dict_index=i, map_cache=map_cache,
```
calvinp0, line 2039: the requested atom map can still be ignored in some cases. when a molecule splits into fragments, parts of the code still choose their own atom correspondence. so you ask for map a, and you may instead receive another map, reported as successful

calvinp0 (commented):
lgtm, just one comment

PR #1046, opened 2026-09-07
Report a requested atom map that yields no Linear TS guess

### LINKED ISSUE #1045: Linear adapter returns no guess for one requested atom map, and reports success anyway
## Describe the bug

Follow-up to #1042. On the branch that addresses it (`linear_adapter_explicit_atom_map`, PR #1044) an explicit `atom_map` is honored and different maps now give different geometries — but for one reaction a requested map returns **no guess at all**, and the adapter reports success regardless, simply returning fewer guesses than maps requested.

Measured on `main` @ `4d67ad4c` with that branch merged, requesting one guess per distinct correspondence:

| reaction | distinct maps | guesses returned | identical pairs |
|---|---|---|---|
| `CC[CH]C -> CCC[CH2]` | 6 | 6 | 0 |
| `CC[CH]CC -> CCCC[CH2]` | 6 | 6 | 0 |
| `CCC[CH2] -> CC[CH]C` | 4 | 4 | 0 |
| `[CH2]CC -> C[CH]C` | 2 | 2 | 0 |
| `[CH2]C(C)CC -> CC(C)[CH]C` | 2 | **1** | 0 |

The last row is the problem. The two maps differ only in which of the two hydrogens on one carbon migrates, so they are about as similar as two paths of one reaction can be, yet one produces nothing. The failing one is the map closest to the identity:

```
map A = [2, 1, 0, 3, 4, 9, 11, 8, ...]   guess produced
map B = [0, 1, 2, 3, 4, 5,  6, 8, ...]   no guess
```

`interpolate_isomerization` raises for every weight on that path. The exception is caught, so nothing surfaces to the caller:

```
File "arc/job/adapters/ts/linear.py", line 1059, in execute_incore
    xyzs = interpolate(rxn=rxn, weight=w, existing_xyzs=all_xyzs_so_far,
                       map_cache=map_cache, forced_atom_map=forced_atom_map)
File "arc/job/adapters/ts/linear.py", line 1813, in interpolate
    return interpolate_isomerization(...)
```

## How to reproduce

Request a guess per distinct atom map for `[CH2]C(C)CC -> CC(C)[CH]C` (`intra_H_migration`), enumerating the maps first and passing each one to the adapter as `atom_map`. One of the two comes back empty.

## Expected behavior

1. When a map is supplied explicitly and no guess survives for it, report that distinctly. An explicitly requested map returning nothing is a different situation from a derived map that happened not to work, and a caller iterating over maps cannot currently tell them apart without counting the results.
2. Worth checking whether near-identity maps are the trigger. The same map failed before that branch existed, when I forced it by monkeypatching `map_rxn`, so it looks like a property of the interpolation rather than of the new argument.

This is not a regression — the branch is a clear improvement on the previous behavior, where every requested map returned the same geometry. Filing it separately so it is not lost when that PR merges.

## Installation information

- ARC `main` @ `4d67ad4c` with `linear_adapter_explicit_atom_map` merged
- Python 3.14, Linux

kfir4444:
## Summary

A requested atom map for which no TS guess survives validation is now a distinct outcome instead of a smaller result set: it is appended to `LinearAdapter.atom_maps_without_guesses`, recorded on the reaction as an unsuccessful `TSGuess` (so it persists into the restart file), and logged as a warning, with the summary line reporting how many of the requested maps produced guesses.

Investigating the reported case: nothing raises - `interpolate_isomerization` returns empty, every candidate having been discarded by the H-migration validators. Near-identity maps are not the trigger; over the 7 enumerated maps of `[CH2]C(C)CC -> CC(C)[CH]C` the most identity-like map (11/16 fixed points) does produce a guess. The two maps that matter share an identical reaction center (formed `(0, 12)`, broken `(3, 12)`) and differ only in the permutation of the spectator hydrogens, yet one yields 3 guesses and the other none - the ordered-product geometry is built by reordering P through the map, so spectator labeling changes the Z-matrix blend. That sensitivity is a property of the interpolation path rather than of the explicit-map argument, and is left for a separate change; the warning notes that another map with the same reaction center may still produce a guess.

Based on #1044.

Closes #1045

### DIFF
--- a/arc/job/adapters/ts/linear.py
+++ b/arc/job/adapters/ts/linear.py
@@ -796,6 +796,9 @@ class LinearAdapter(JobAdapter):
     derivation (it is only consulted by the degraded fallback paths), and the adapter logs a
     warning when it encounters one. To request guesses for one specific reactants-to-products
     correspondence, pass it via the ``atom_map`` argument, which overrides the derived map.
+    A requested map for which no guess survives validation is reported: it is appended to
+    ``self.atom_maps_without_guesses``, recorded on the reaction as an unsuccessful ``TSGuess``,
+    and logged as a warning.
 
     Args:
         project (str): The project's name. Used for setting the remote path.
@@ -896,6 +899,8 @@ class LinearAdapter(JobAdapter):
         self.job_adapter = 'linear'
         # Explicitly requested atom map(s); ``None`` means "derive a map per reaction path".
         self.forced_atom_maps = normalize_atom_maps(atom_map)
+        # Requested atom maps for which no TS guess survived, in the order they were processed.
+        self.atom_maps_without_guesses: list[list[int]] = list()
         self.command = None
         self.execution_type = 'incore'  # incore-only (no server); ignore any execution_type the scheduler passes
 
@@ -1050,6 +1055,8 @@ class LinearAdapter(JobAdapter):
                         # the degraded fallback paths honor the requested correspondence too.
                         rxn._atom_map = list(forced_atom_map)
                     map_suffix = f', map={map_i}' if len(forced_atom_maps) > 1 else ''
+                    map_t0 = datetime.datetime.now()
+                    guesses_before = len(rxn.ts_species.ts_guesses)
                     # Per-map memo: atom maps and the wider-family-set scan
                     # outcome are weight-invariant, so compute them once per map
                     # instead of once per weight iteration.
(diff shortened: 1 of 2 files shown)


arc/job/adapters/ts/linear.py
```diff
@@ -1093,6 +1100,27 @@ def execute_incore(self):
...
 
+                    if forced_atom_map is not None and len(rxn.ts_species.ts_guesses) == guesses_before:
+                        # An explicitly requested map that yields nothing is a distinct outcome from a
+                        # derived map that happened not to work: the caller asked for this specific
+                        # correspondence and must be able to tell that it came back empty. Record it
+                        # both on the reaction (an unsuccessful TSGuess, which is persisted) and on the
+                        # adapter, so neither a human reading the log nor a caller iterating over maps
+                        # has to infer it from a guess count.
+                        self.atom_maps_without_guesses.append(list(forced_atom_map))
+                        rxn.ts_species.append_ts_guess(TSGuess(method=f'Linear (map={map_i})',
```
calvinp0, line 1111: a failed map can stop the remaining maps working. so if we save a failed entry without coords, then the next map produces a geom it tries to compare the geom with the missing coords and crashes. shuold also have a test bad map, then good map

arc/job/adapters/ts/linear.py
```diff
@@ -1093,6 +1100,27 @@ def execute_incore(self):
                                          comment=f'Linear w={w:.2f}, {xyz_i}{map_suffix}, family: {rxn.family}',
                                          )
 
+                    if forced_atom_map is not None and len(rxn.ts_species.ts_guesses) == guesses_before:
```
calvinp0, line 1103: a valid duplicate guess is called a failure. if a map produces a geom already found earlier, obv it will avoid saving it twice but it then interprets nothing new was saved as "this map failed"

arc/job/adapters/ts/linear.py
```diff
@@ -1106,10 +1134,15 @@ def execute_incore(self):
 
             if len(self.reactions) < 5:
                 successes = len([tsg for tsg in rxn.ts_species.ts_guesses if tsg.success and 'linear' in tsg.method.lower()])
+                requested = len(self.forced_atom_maps) if self.forced_atom_maps is not None else 0
```
calvinp0, line 1137: when reporting results for the second reaction, the code also counts failures from the first reaction. this can produce results such as "-1 out of 1 requested maps succeeded"

calvinp0 (commented):
thanks mate! just a few comments

PR #1053, opened 2026-09-12
Write the Molpro command that exists for each open-shell method

calvinp0:
## The defect

`arc/job/adapters/molpro.py` prefixed `u` to the method for any unrestricted species, unconditionally:

```python
if not is_restricted(self):
    input_dict['restricted'] = 'u'
```

`u<method>` is a Molpro command for some methods and not for others, so ARC could write an input Molpro has no command for.

## What changes

A per-method table decides the open-shell command, with a suffix rule for the methods Molpro does not implement open-shell at all.

| requested method | ARC wrote | ARC now writes | basis |
| --- | --- | --- | --- |
| `ccsd-f12`, `ccsd(t)-f12`, `dcsd` | `uccsd(t)-f12` | unchanged | "Open-shell unrestricted UCCSD-F12 approximations ... Restricted open-shell Hartree-Fock (RHF) orbitals are used" |
| `mp2`, `df-mp2`, `mp2-f12`, `df-mp2-f12` | `ump2-f12` | `rmp2-f12` | no `UMP2-F12` command exists; "all methods except `ump2` use spin-restricted Hartree-Fock (RHF) reference functions" |
| anything ending `-f12c` | `uccsd(t)-f12c` | `NotImplementedError` | "Currently CCSD-F12c is not available for open-shell cases" |
| `casscf` | `ucasscf` | `casscf` | "The program is invoked by the command MULTI, MCSCF, CASSCF, or CASCI"; no `UCASSCF` exists |
| anything else | `u<method>` | `u<method>` | see below |

A rewrite is logged at warning level naming both forms, since it changes the command from the one the user wrote. The substitution is the same theory: `RMP2-F12` is open-shell MP2-F12, and the mapping is density-fitting consistent on both rows.

Unrecognised methods keep the `u` prefix deliberately rather than by omission. It is correct for the whole coupled-cluster and configuration-interaction family, and defaulting to no prefix would silently select Molpro's **closed-shell** program — a worse failure than an invalid command.

The refusal keys on the `-f12c` suffix rather than two exact names, so `df-ccsd(t)-f12c` and `mp2-f12c` are covered too. It raises `NotImplementedError` from job construction, matching the existing precedent for Molpro plus IRC in `check_argument_consistency`.

The multireference path is unaffected: `mrci`/`rs2` already clears the prefix and builds its own command. A test fails if that guard is removed.

## Not a change: the `u` prefix on coupled cluster is correct

Worth stating, because it looks like the same class of problem and is not. Molpro's `HF`/`RHF` is the spin-restricted program, and ARC's template hardcodes `{hf; ... wf,spin=N,charge=C;}`, so ARC never requests UHF orbitals. The `u` selects the open-shell correlation ansatz on an RHF reference, giving `RHF-UCCSD(T)-F12`, which the manual requires for open-shell F12. ARC's own checked-in output confirms it: `arc/testing/sp/TS_x118_sp_CCSD(T).out` pairs `wf,spin=2` with `uccsd(t)-f12;` and reports `!RHF-UCCSD(T)-F12a energy`. ARC's default single point, `ccsd(t)-f12/cc-pvtz-f12`, was correct before this change and is unchanged by it.

## Tests

`arc/job/adapters/molpro_test.py` gains coverage of each class above for an open-shell species, the closed-shell counterpart in every case, the multireference path, and an unrecognised method. Assertions are on the rendered Molpro input where the method name appears.

Five of the new tests fail against the previous unconditional prefix.

- `molpro_test.py`, `level_test.py`, `common_test.py`: 88 passed
- `arc/job/` and `arc/parser/`: 1516 passed, 36 skipped (5 pre-existing `torch_ani_test.py` failures, reproduced with this change reverted)

https://claude.ai/code/session_01MndwJm4BEs4DN3vJQsEwD3

### DIFF
--- a/arc/job/adapters/molpro.py
+++ b/arc/job/adapters/molpro.py
@@ -35,6 +35,15 @@ default_job_settings, global_ess_settings, input_filenames, output_filenames, se
     settings['default_job_settings'], settings['global_ess_settings'], settings['input_filenames'], \
     settings['output_filenames'], settings['servers'], settings['submit_filenames']
 
+MOLPRO_OPEN_SHELL_METHOD_NAMES = {'mp2': 'rmp2',
+                                  'df-mp2': 'df-rmp2',
+                                  'mp2-f12': 'rmp2-f12',
+                                  'df-mp2-f12': 'df-rmp2-f12',
+                                  'casscf': 'casscf',
+                                  }
+
+MOLPRO_SUFFIX_WITHOUT_AN_OPEN_SHELL_FORM = '-f12c'
+
 input_template = """***,${label}
 memory,Total=${memory},m;
 
@@ -52,7 +61,7 @@ int;
  wf,spin=${spin},charge=${charge};
 }
 
-${restricted}${method}
+${method}
 
 ${job_type_1}
 ${job_type_2}${block}
@@ -214,7 +223,6 @@ class MolproAdapter(JobAdapter):
                     'keywords',
                     'memory',
                     'method',
-                    'restricted',
                     ]:
             input_dict[key] = ''
         input_dict['auxiliary_basis'] = self.level.auxiliary_basis or ''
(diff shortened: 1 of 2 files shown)


calvinp0:
Folded into #917, which already carries the companion Molpro version-parser fix. The commit is unchanged; the branch stays on origin for reference.

PR #1055, opened 2026-09-13
Fix bi-rad singlet composite jobs silently running a restricted reference

alongd:
## What

`is_species_restricted()` returned early for `REFERENCE_AGNOSTIC_METHOD_TYPES` (composite,
force-field, semiempirical) **before** reading `number_of_radicals`, so the bi-rad-singlet branch
below it was unreachable for exactly those method types. Move the `number_of_radicals > 1` check
ahead of that early return.

## Why

An open-shell singlet biradical is declared in ARC as an `ARCSpecies` with `multiplicity=1` and
`number_of_radicals=2` — that is what the `ARCSpecies` docstring prescribes, and it is the only way
to express the state. The existing code already knew what to do with it; its own comment says so:

```python
# don't run unrestricted for composite methods such as CBS-QB3, it'll be done automatically if the
# multiplicity is greater than one, but do specify uCBS-QB3 for example for bi-rad singlets.
```

That comment describes a case the function could never reach. Composite methods take
`method_type == 'composite'`, so `G4`/`CBS-QB3` returned `True` (restricted) at the top, and the
branch the comment documents never ran. A declared bi-rad singlet therefore went out as plain `g4`
with a **restricted** reference — no `u` prefix — and converged to the wrong electronic state
without warning. Nothing errors; the thermochemistry is simply wrong.

The asymmetry is the whole point: for a composite method an unrestricted reference is applied
automatically when multiplicity > 1, but a bi-rad singlet's multiplicity **is** 1, so the method
gets no signal and ARC has to say `uG4` explicitly.

## What changed

`arc/job/adapters/common.py` — `number_of_radicals > 1` is now evaluated first and returns
unrestricted, then the reference-agnostic early return, then the `multiplicity > 1` case. The
existing `# don't run unrestricted for composite methods...` comment is replaced by one that
explains why the order matters, so the next reader does not "simplify" it back.

Unaffected: plain closed-shell singlets (`number_of_radicals` unset or 1) and composite jobs at
multiplicity > 1 both take exactly the paths they did before.

## Testing

`arc/job/adapters/common_test.py` — 61 passed, with new cases covering a composite bi-rad singlet
(unrestricted), a composite closed-shell singlet (restricted), and a composite triplet (unchanged).

Verified end to end on the route section: a `g4` job for a species with `multiplicity: 1,
number_of_radicals: 2` now writes `ug4`.

### DIFF
--- a/arc/job/adapters/common.py
+++ b/arc/job/adapters/common.py
@@ -660,21 +660,28 @@ def is_species_restricted(obj: JobAdapter,
         bool: Whether to run as restricted (``True``) or not (``False``).
     """
 
-    if obj.level.method_type in REFERENCE_AGNOSTIC_METHOD_TYPES:
-        return True
-
     multiplicity = obj.multiplicity if species is None else species.multiplicity
     species_obj = obj.species[0] if species is None else species
     number_of_radicals = species_obj.number_of_radicals
     species_label = species_obj.label
-    if multiplicity > 1 or (number_of_radicals is not None and number_of_radicals > 1):
-        # run an unrestricted electronic structure calculation if the spin multiplicity is greater than one,
-        # or if it is one but the number of radicals is greater than one (e.g., bi-rad singlet)
-        # don't run unrestricted for composite methods such as CBS-QB3, it'll be done automatically if the
-        # multiplicity is greater than one, but do specify uCBS-QB3 for example for bi-rad singlets.
-        if number_of_radicals is not None and number_of_radicals > 1:
-            logger.info(f'Using an unrestricted method for species {species_label} which has '
-                        f'{number_of_radicals} radicals and multiplicity {multiplicity}.')
+
+    if number_of_radicals is not None and number_of_radicals > 1:
+        # A declared number_of_radicals > 1 with multiplicity 1 (e.g., a bi-rad singlet) always
+        # needs an unrestricted reference, *even for a reference-agnostic method type* such as a
+        # composite method (e.g., CBS-QB3, G4): for those, an unrestricted reference is applied
+        # automatically when the multiplicity is greater than one, but a bi-rad singlet's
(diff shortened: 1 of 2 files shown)


alongd:
Added a second, related fix to this branch (51f93ce), since it is the same failure — a job that runs the wrong electronic state and reports it as a normal result — reached by a different route.

`ARCSpecies.__init__` and `ARCSpecies.from_dict` disagree on the multiplicity of the same species:

```python
xyz = "<the optimized geometry of C[C]C#N>"
ARCSpecies(label="s", smiles="C[C]C#N", xyz=xyz).multiplicity                      # 3
ARCSpecies(species_dict={"label": "s", "smiles": "C[C]C#N", "xyz": xyz}).multiplicity  # 1
```

`__init__` reads multiplicity/charge off the molecule built from the SMILES/adjlist/InChI (species.py:507-508) and only then may overwrite `.mol` from the coordinates, guarded by `regen_mol and not keep_mol`. `from_dict` runs `mol_from_xyz()` first with no guard (1011-1014) and reads the multiplicity afterwards (1024-1025). The two differ by statement order alone. Perception picks an electronic state, and for an open-shell species it routinely picks a different one: `C[C]C#N` perceives from its own optimized coordinates as `C[C][C][N]`.

A deck travels the `from_dict` path, so this is not hypothetical — it silently ran 13 of 31 open-shell species in a controlled A/B study at the wrong multiplicity. The jobs converged and terminated normally.

The fix captures multiplicity and charge from the structure-derived molecule before the override, falling back to the perceived molecule only when no structure was given at all. An explicitly declared multiplicity still wins. `arc/species/` + `arc/main_test.py` + `arc/job/adapters/common_test.py` run serially: 446 passed, 1 failure (`main_test.py::TestARC::test_as_dict`) that is present on this branch before the change as well.

